PR #279, opened 2015-04-10
Recognize more comdirect documents

Necoro:
This (should) fix issue #277. As I did not want to have another counter to see how many words I have to jump, I relied on 'the last word on the line'. As I needed more functionality for that, I took the liberty of replacing some `while charAt incr/decr` loops by calling `indexOf/lastIndexOf` instead.

Additionally it recognizes more documents, where it does not say 'Zum Kurs von' but different things (I've seen 'Zum comdirect Preis von' and 'Zum Preis von' so far). I have achieved that by checking for 'Nennwert' instead. I hope that does not break anything.


buchen:
@simpsus hat den comdirect Import gebaut. Da ich selber keine comdirect PDF Dateien habe, kann ich das auch nicht prüfen. :cry: Wenn @simpsus nichts gegenteiliges vermeldet, nehme ich das in die nächste Version mit auf. :smile:

PR #330, opened 2015-07-07
Bookmark for finanztreff.de (similar to Finanzen.net).

sebasbaumh:
A default bookmark, which allows browsing additional details on finanztreff.de.
This website is used by some people because it could be connected to ETrade accounts, which were later on taken over by Flatex


buchen:
Danke :+1:

PR #331, opened 2015-07-08
New quote feed for finanztreff.de

sebasbaumh:
New quote feed for finanztreff.de and small changes to allow querying it without having to select an exchange in edit securities dialog.
----
Quote feed for finanztreff.de: parses the pages for single quotes into a set of quotes.

Note: the granularity of quotes varies. Currently there are intraday, interday and longer quotes with ascending intervals.
But they enable the user to at least get some quotes for funds or bonds missing in Yahoo Finance.

Build temporary security for quote preview in edit securities dialog.
Query feeds without exchange and url directly in edit securities dialog.


sebasbaumh:
Closed as it could conflict with data usage policies.

PR #526, opened 2016-06-05
Fix for RegEx in Flatex PDF extractor

sebasbaumh:
Fix for RegEx in Flatex PDF extractor resulting in an empty import if the amount stated in the PDF exceeds 1000 and gets noted as "1.000". Additionally escape dot characters correctly and be less strict about spacings.
Fix typo in messages.


buchen:
👍 Vielen Dank. Habe ich direkt mal gemergt 😄

sebasbaumh:
:thumbsup:

PR #687, opened 2017-01-03
Minor change regarding context menus to be more consistent

cmaoling:
Add Transfer to SecuritiesTable for consistency and convenience
"Referencing #658: Get SecurityTransaction Heading based on message text
rather than type of SecurityTransactionDialog.class

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountContextMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountContextMenu.java
@@ -49,13 +49,6 @@ public class AccountContextMenu
 
         manager.add(new Separator());
 
-        new OpenDialogAction(owner, Messages.AccountMenuTransfer) //
-                        .type(AccountTransferDialog.class) //
-                        .with(account) //
-                        .addTo(manager);
-
-        manager.add(new Separator());
-
         // show security related actions only if
         // (a) a portfolio exists and (b) securities exist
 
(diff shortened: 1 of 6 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountContextMenu.java
```diff
@@ -49,13 +49,6 @@ public void menuAboutToShow(IMenuManager manager, final Account account, final S
 
         manager.add(new Separator());
 
-        new OpenDialogAction(owner, Messages.AccountMenuTransfer) //
```
buchen: This change - the deletion of the the account transfer - I do not understand. I do not see it moved to another location.

buchen:
You have seen my comment about the deletion of the "Account transfer"?

Apart from that, the change is fine. I was wondering though whether "Kaufen" (verb) and "Steuern" (noun) really are consistent to begin with - but for now at least the dialogs and menu entries are consistent.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountContextMenu.java
cmaoling: My apologies, I overlooked that this was a cash transfer rather than a security transfer, so you are right, this item should remain.

cmaoling:
Valid point. You could also go the other way and change "Kaufen" and "Verkaufen" to the corresponding noun.

buchen:
You closed the issue which indicates to me you would also rather change it to "Kauf" und "Verkauf". Do you plan to update your PR? ☺️

cmaoling:
Yep, will produce a different PR. :)
Working on a different git branch to introduce dividends per stock....

PR #693, opened 2017-01-05
Various UI changes to reduce inconsistency

cmaoling:
Second (First one: #687) attempt to address the concerns highlighted in #658.
Based on comments the items are now nouns and minor alignment of English and German wording.
**Modification to be mentioned is, that DividendTransaction now extents AccountTransaction to enable the common types to be reused and used as basis in the SecurityPerformanceSnapshot view.**

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
@@ -65,7 +65,18 @@ public abstract class AbstractSecurityTransactionModel extends AbstractModel
     @Override
     public String getHeading()
     {
-        return type.toString();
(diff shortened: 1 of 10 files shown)


buchen:
Merged w/ 77a02af7565934794b3b0797b8d50712c3e9a8f2

> Modification to be mentioned is, that DividendTransaction now extents AccountTransaction to enable the common types to be reused and used as basis in the SecurityPerformanceSnapshot view.

I took the liberty to remove this refactoring. The "DividendTransaction" anyway is an oddity (why copy values instead of using the original AccountTransaction?) which was introduced some time ago and I never came around to refactoring. 😒 Because there is code that relies on the "instanceof", now the order of those instanceof statements would matter: if the code does an "instanceof AccountTransaction" first, then it suddenly would pick up a dividend where it previously did not. As the code improvement for the labels is minor, I reverted that small piece of your change.

Important is that the UI is much more consistent now. Thanks for the change. I have added you also to the list of contributors in the about dialog.

Issue #704, opened 2017-01-15
Online: Volumes not extracted from provided URL


cmaoling:
When providing an webpage for historical quote loads the volume is not extraced:
for example:
http://www.ariva.de/allianz-aktie/historische_kurse
Volumes of all zero are loaded.

PR #705, opened 2017-01-15
Rework and enhancement in HTMLTableQuoteFeed.java

cmaoling:
Make class Column stand alone rather than incorporated to be usable by other classes, address #704.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
@@ -1,0 +1,107 @@
+package name.abuchen.portfolio.online;
+
+import java.math.BigInteger;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.text.ParseException;
+import java.util.Hashtable;
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.Set;
+
+import org.jsoup.nodes.Element;
+
+import name.abuchen.portfolio.model.LatestSecurityPrice;
+import name.abuchen.portfolio.money.Values;
+
+
+public abstract class Column
+{
+    static final ThreadLocal<DecimalFormat> DECIMAL_FORMAT_GERMAN = new ThreadLocal<DecimalFormat>()
+    {
+        @Override
+        protected DecimalFormat initialValue()
+        {
+            return new DecimalFormat("#,##0.###", new DecimalFormatSymbols(Locale.GERMAN)); //$NON-NLS-1$
+        }
+    };
+
+    static final ThreadLocal<DecimalFormat> DECIMAL_FORMAT_ENGLISH = new ThreadLocal<DecimalFormat>()
+    {
+        @Override
+        protected DecimalFormat initialValue()
+        {
+            return new DecimalFormat("#,##0.###", new DecimalFormatSymbols(Locale.ENGLISH)); //$NON-NLS-1$
+        }
+    };
+
+    static final ThreadLocal<DecimalFormat> DECIMAL_FORMAT_APOSTROPHE = new ThreadLocal<DecimalFormat>()
+    {
+        @Override
+        protected DecimalFormat initialValue()
+        {
+            DecimalFormatSymbols unusualSymbols = new DecimalFormatSymbols(Locale.US);
+            unusualSymbols.setGroupingSeparator('\'');
+            return new DecimalFormat("#,##0.##", unusualSymbols); //$NON-NLS-1$
+        }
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
```diff
@@ -0,0 +1,166 @@
+package name.abuchen.portfolio.online;
```
buchen: Is there a specific reason to move "Column" to another package than the only class that is using it? I would prefer to have it next to HTMLTableQuoteFeed (at least as long it is bound very close to HTMLTableQuoteFeed).

name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jsoup.nodes.Element;
+
+import name.abuchen.portfolio.model.LatestSecurityPrice;
+import name.abuchen.portfolio.money.Values;
+
+
+public abstract class Column
```
buchen: If it is in the same package, the Column could be package local. It is not API to be used anywhere else (particularly not in the GUI parts).

name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
```diff
@@ -0,0 +1,166 @@
...
+                    match = match.replaceAll("([0-9])\\.([0-9][0-9][0-9])\\.([0-9][0-9][0-9])\\.([0-9][0-9][0-9])","$1$2$3$4");
+                    match = match.replaceAll("([0-9])\\.([0-9][0-9][0-9])\\.([0-9][0-9][0-9])","$1$2$3");
+                    match = match.replaceAll("([0-9])\\.([0-9][0-9][0-9])","$1$2");
+                }
+                value = Double.valueOf(match);
+            }
+        }
+        catch (Throwable ex)
+        {
+            System.err.println("Uncaught exception - " + ex.getMessage());
```
buchen: Writing the error to System.err will loose the message (only visible in the Dev environment). You can
* store it and return it via the "errors" Object
* or ignore it at completely.

Generally, I am not writing errors per (attempted) import line, because this will results in possibly hundreds of error messages.

Then there is the question what Exception you expect here. I guess it is the Double.valueOf so rather a NumberFormatException

name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
```diff
@@ -0,0 +1,166 @@
...
+        String text = elem.text().trim();
+        int multiplier = 1;
+        double value = -0.1;
+        try
+        {
+            Pattern numberPattern = Pattern.compile("([0-9,.]+)");
+            Matcher numberMatcher = numberPattern.matcher(text);
+            if (numberMatcher.find())
+            {
+                String match = numberMatcher.group(1);
```
buchen: This "lastDot" and "lastComma" unfortunately does not really work well.
I had similar logic for the quote initially. But what if the value is "1,045". It could be read as American format (one thousand and forty five) or as German format (one euro and 4 and half cents).

That's why "asQuote" also uses the "langaugeHint" which is basically the language given in the HTML page. I guess the volumne should also use that.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeed.java
```diff
@@ -219,6 +129,32 @@ void setValue(Element value, LatestSecurityPrice price, String languageHint) thr
...
+            {
+                if ("-".equals(value.text().trim())) //$NON-NLS-1$
+                    price.setVolume((int) LatestSecurityPrice.NOT_AVAILABLE);
+                else
+                    price.setVolume(super.asInt(value));
+            }
+            catch (Throwable ex)
+            {
+                System.err.println("Uncaught exception - " + ex.getMessage());
+                ex.printStackTrace(System.err);
```
buchen: Which exception ends up here? The "asInt" already catches Throwables. 
And: if we catch the exception, it makes sense the set the value to NOT_AVAILABLE.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/Column.java
cmaoling: No specific reason. Moving it to the name.abuchen.portfolio.online.impl

cmaoling:
Thanks for your detailed feedback and consideration

PR #733, opened 2017-04-16
updated readme

Yky:
changed maven requirement to 3.x.x as it also runs with 3.2.x

updated to latest Eclipse release (tested, works)

PDT is commonly used as an abbreviation for Eclipse PHP Development Tools, PDE is the Plug-in Development Environment required for this project. Changed to avoid confusion.


buchen:
Thanks!

PR #787, opened 2017-07-15
Issue 783 flatex pdf import revise to fin tech group bank ag

Ragas13:
Belonging to issue 783 as the header has changed from "biw AG" to "FinTech Group Bank AG".
- native converter is not changed
- test cases replaced from biw to FinTech
- 3 new test case files added, 1 new implementet


Ragas13:
Leave a comment

PR #814, opened 2017-08-28
enhanced INGDiBa Extractor to deal with joint accounts

stephanmunich:
In joint accounts, the lines starting with "Kirchensteuer" and
"Solidaritätszuschlag" appear two times and have to be summed up, the
line starting with "Kapitalertragsteuer" is replaced by the two lines
starting with "KapSt anteilig" and have to be summed up as well.

There is no PDF to test but there is a test case and the problem was
described in
https://forum.portfolio-performance.info/t/steuern-bei-pdf-import-von-dividenden-richtig-buchen/631/12


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaExtractor.java
```diff
@@ -151,38 +170,43 @@ private void addSellTransaction()
...
+        
+        addTaxSectionToBuySellEntry(type, transaction);
+        block.set(transaction);
     }
 
     @SuppressWarnings("nls")
     private void addErtragsgutschrift()
     {
-        DocumentType type = new DocumentType("Ertragsgutschrift");
+        DocumentType type = new DocumentType("Ertragsgutschrift", (context, lines) ->
```
buchen: Das scheint mir immer der gleiche Block zur Bestimmung zur Bestimmung eines Gemeinschaftskontos zu sein, oder?

Könnte man das nicht rausziehen, z.B.

```
private static final BiConsumer<Map<String, String>, String[]> detectJointAccount = (context, lines) ->
        {
            Pattern pJointAccount = Pattern.compile("KapSt anteilig 50,00 %.*");
             Boolean bJointAccount = false;
...
        });
```

buchen:
Ist drin - ich habe noch das isJointAccount raus-fefactored.

Danke!

PR #839, opened 2017-09-29
Fix date - sell date vs. valuta date at Consorsbank

Ragas13:
Belonging to https://forum.portfolio-performance.info/t/import-fehlerhaft-bei-cortal-consors/1500/4

According to forum, Consorsbank selling transaction should determine trading date, not valuta date.


buchen:
Danke!!!!

PR #840, opened 2017-09-29
Add regions according to market classification by MSCI.

sebasbaumh:
This makes it easier to classify securities as it does include meta classification like "Emerging Markets".
Furthermore the user is assisted by allowing him to choose a country, which then gets mapped to such a region.

Source: https://www.msci.com/market-classification respective https://www.justetf.com/uk/news/etf/msci-index-classification-and-how-they-divide-up-the-world.html

Note: I just included the English version as I do not see too much sense in translating specific terms like "Frontier Markets".


buchen:
Thanks! :+1:

PR #884, opened 2017-11-12
Change of securites chart - appearance of investments (scatter chart)

Ragas13:
Belonging to issue #862

This pull changing how the investments will be displayed by removing the MarkerLine and add them as scatter chart with diamonds as plot symbol. This upgrade the presentaton of investments especially at monthly saving plans.

The information of each buy / selling price is now added to the mouse tooltip.

Further improvment is to exclude the portfolie transfer in/out transactions as I guess this is just a transfer from one portfolio to each other and therefore should not be displayed as an investment at the chart.

Newly I've added a legend at the chart bottom to get a quick overview of the added informations.

Last but not least please find below the replaced investments appearance with a 2 vs, 10 years scaling,
unbenannt


Ragas13:
Leave a comment

PR #956, opened 2018-04-12
SMA correction according to Forum

Ragas13:
Please see https://forum.portfolio-performance.info/t/moving-average-im-kursdiagram-falsch/2600/

Unfortunately some test cases fails but I don't know how to solve, except to remove them. However, I do not have a possibility till mid/May to revise this PR, because I have only a smartphone at this time.
Br
Marco


buchen:
rebased and merged.

PR #996, opened 2018-06-30
Feature kraken api

fuchsst:
I'd like to add support for kraken.com:
- read ticker data via kraken api
- fetch trades (closed orders) and import them
- fetch funding (deposit/withdraw) and import them

So far i created the QuoteFeed and the properties screen. Maybe you can give me some hints, so i don't need to search the whole project:
* where do i find the model classes that represent the transactions/depots/accounts
* where do i find the Services that manage (read /store/convert) those
* where it's best to hook in to automatically fetch the transactions from Kraken


buchen:
Hi Stefan,

> * read ticker data via kraken api

Great. I believe *this* PR is about reading data via the kraken API.

I suggest to keep them separate. And if avoidable, do not merge the "master" branch into your feature branch. That makes reviewing the changes harder. If you needed, rebase your changes onto the tip of master and push into the same branch. Or just leave it as it is - it is much easier for me to merge at the end.

Just for my understanding: to use the Kraken API, no API key or similar is needed, right? I do not see any use of the data collected via the preference page (yet)?

Then the API is using the "Ticker Symbol". Does it make sense? The ticker symbol is typically used with Yahoo.

And finally: the prices have only 4 decimal places. Is that enough for crypto currencies? I am pondering how best to change it already (the "long" is anyway reaching it's limits - but changing to BigDecimal is a huge change).

Do you have a sample symbol that I could use to try out the Kraken quote importer?

My proposal is to use this PR to finish the quote provider.

> * fetch trades (closed orders) and import them
> * fetch funding (deposit/withdraw) and import them

Let's put those changes into a new PR.

> where do i find the model classes that represent the transactions/depots/accounts
> where do i find the Services that manage (read /store/convert) those

All "model" classes are in the package [name.abuchen.portfolio.model](https://github.com/buchen/portfolio/tree/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model). These constitute the ```Client``` (= one file) and only these classes are persisted in the XML files.

There are two types of transactions: ```AccountTransaction```s for ```Account``` (Konto) und ```PortfolioTransaction``` for ```Portfolio``` (Depot). Because some transactions like a "sales" includes both account and portfolio transaction, I am using ```CrossEntry``` classes to reflect that.

Maybe take a look at the [DeutscheBankPDFExtractor](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java) that is creating transactions from PDF documents. PDF examples show how to create a transaction with all data (fees, taxes, forex, ...).

Typically the approach is:
* create the transaction
* wrap it into a [Extractor.Item](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java#L43)
* show a dialog that let's the use see what is about imported

This is used for PDF and CSV imports, and you could write a similar "Handler" for Kraken.

This is the first code that automatically would import transactions. So there is no good previous sample. Even you do not want to do the dialog step (understandably), I would prefer nonetheless to go via the Extractor.Item. When inserting them into the model, there are some sanity and consistency checks. If I extend them in the future, those checks would also apply to Kraken imports.

> where it's best to hook in to automatically fetch the transactions from Kraken

This is no hook at the moment. The ```PortfolioPart``` triggers off some ```Job```s after opening the file. As a starting point, create a new job and trigger it there as well. Later we can make up our mind how to make it more generic. 

Feel free to push "intermediate" code into a (separate) PR to discuss / if you have questions.

Andreas

portfolio-product/name.abuchen.portfolio.product
```diff
@@ -58,8 +58,6 @@
       <plugin id="org.eclipse.equinox.ds" autoStart="true" startLevel="2" />
       <plugin id="org.eclipse.equinox.simpleconfigurator" autoStart="true" startLevel="1" />
       <plugin id="org.eclipse.osgi" autoStart="true" startLevel="-1" />
-      <property name="osgi.instance.area.default" value="$LOCALAPPDATA$/PortfolioPerformance/workspace" os="win32" />
```
buchen: Those configuration items are needed for windows and linux. The editor unfortunately removes them.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/KrakenPreferencePage.java
```diff
@@ -0,0 +1,52 @@
...
+        link.addSelectionListener(new SelectionAdapter()
+        {
+            @Override
+            public void widgetSelected(final SelectionEvent event)
+            {
+                DesktopAPI.browse(krakenApiSettingsUrl); // $NON-NLS-1$
+            }
+        });
+
+        addField(new StringFieldEditor(UIConstants.Preferences.KRAKEN_API_KEY, Messages.PrefKrakenAPIKey,
```
buchen: Looks like the preference keys are not needed for the QuoteProvider implementation. If so, then I would prefer to have them in the next PR.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/KrakenQuoteFeed.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    public boolean updateLatestQuotes(Security security, List<Exception> errors)
+    {
+        if (security.getTickerSymbol() == null)
+        {
+            errors.add(new IOException(MessageFormat.format(Messages.MsgMissingTickerSymbol, security.getName())));
+            return false;
+        }
+
+        List<LatestSecurityPrice> prices = getHistoricalQuotes(security, LocalDate.now(), errors);
```
buchen: The latest quotes are also fetched via the same API as the historical quotes. Which can lead to two calls to the API (one for latest and for historical) which essentially retrieve the same. Is there a rate limit on the API? If so, it might make sense to cache the results for 5 minutes and avoid the second call.

fuchsst:
Hi,

I reduced this PR to the fetch of the Kraken Feed. 
I'll create a separate one with a "search form" where applicable tradable pairs can be selected (similar to the Yahoo Search, but easier, since the API does not return much usefull data, see: https://api.kraken.com/0/public/AssetPairs).
And another one for the fetch of the private data (funding/trades) - this would be the one where the API Settings would be needed.

4 digits is probably not always enough precision, for most it will be OK, but for very small like Ripple tere will be "rounding errors".

The AssetPairs endpoint gives a list of tradable pairs. Examples could be XETCZEUR, BCHUSD...

buchen:
@fuchsst - merged. Mach einfach neue PR für die weiteren Änderungen auf. Vielen Dank!

PR #1013, opened 2018-07-29
Bug in IBFlexQueryImporter where precision of BigDecimal would cause a validation exception

mccare:
Problem: if the ibcommision is something non zero for a buysell trade (and the account information is given) the Transaction:Unit sanity check in the constructor would cause an IllegalArgumentException.

Fix: I've forced BigDecimal to use 4 digits after the decimal separator by creating the BigDecimal from the string representation instead of valueOf(double). 

Also modified the xml and the test to have a transaction with a fee.

Minor change: modified readme with my experiences from setting up portfolio in Eclipse. Very smooth setup btw, didn't run into major problems.


buchen:
It took me some time to realize why the unit tests fail on my machine despite Travis reporting okay. And the "problem" was that I was running the tests with German locale. The issue is that ```String.format``` creates a locale-dependent string that the BigDecimal constructor then cannot parse.

I propose instead of
```
fxRateToBase = new BigDecimal(String.format("%.4f", Double.parseDouble(fxRateToBaseString)));
```
to do
```
fxRateToBase = new BigDecimal(fxRateToBaseString).setScale(4, RoundingMode.HALF_DOWN);
```

I have made the changes on top and merged it.

mccare:
Your fix seems much cleaner than going through a string representation.

PR #1019, opened 2018-08-09
Feature automatic symbol calculation

mccare:
Another addition to the Interactive Broker Importer:
- to facilitate quote updates, the importer needs to translate the ticker symbol. For German and US stock there is a heuristic to guess the correct symbol

* US: the symbol will just stay the same (US == currency is USD)
* other: 
  * sometimes the ticker symbol is appended with the exchange, e.g. BMWd or EZJl, the lower case character is then stripped from the ticker symbol
  * sometimes the ticker symbol includes EUR, e.g. BMWEUR. EUR is stripped if the currency is EUR
  * if the exchange is known, the exchange shortcut will be appended (e.g. BMW.DE or BMW.F)
  
* stocks will be using the AlphaVantage updater
* options will use the yahoo updater

This can still produce invalid ticker symbols, since the ticker symbol is taken from the first occurence of the security in the XML file. If the exchange where the stock was bought is not the usual exchange for this stock, the symbol will come out wrong. Still works most of the time, though.

There is no checking if the alpha vantage key is set or not (in the preferences).


buchen:
Looks good. I will get to this on Saturday.

buchen:
I have rebased your commit and reformatted the code to comply with the code formatter. Nothing else changed, then merged. Thanks! 👍 

I noticed you als included the "merge" commit into master in the PR. It is easier for me if the PR is just a separate branch. Either you leave it as it is (most of the time I know the other changes and and it is easy for me to merge), or you rebase your PR to master and push it again.

PR #1040, opened 2018-10-13
Fix for issue #886.

christian-wehrheim:
Define new label text, externalize string, add translations in
messages_xx.properties.
Create method to update dropdown label with new text and selected year.


buchen:
There is one corner case - on the first tab only, one can select to only show one year. However, I believe this changes nonetheless improves the UI for most users. Will merge!

bildschirmfoto 2018-10-13 um 16 53 38

PR #1041, opened 2018-10-13
Fix for issue #923

christian-wehrheim:
Rearrange buttons under chart view.


christian-wehrheim:
We moved the buttons under the chart view. It fixes the issue and looks a little bit nicer (at least for us :-)).

buchen:
> We moved the buttons under the chart view.

I have two issues with this PR. First, it must update the GridLayout column configuration for all uses of the ```SecuritiesChart``` (agreed, it is stupid to not do this inside the class itself). Second, placing the buttons below the chart uses up quite a lot of precious screen real estate. Not only is there the legend, but now a list of buttons as well.

But you inspired me to look into this again. And the problem is that the RowLayout is asked to calculate the preferred size with SWT.DEFAULT. So it calculates the size if all buttons fit in one row. And that is the size it gets - even though it does not get the preferred hight. Anyway, I have added a custom layout to calculate the preferred width given the current height. 9fa652255e1785941f182dbddfee0ca25be782bd

PR #1045, opened 2018-10-20
Issue 962

tomcastro89:
- adjusted account balance chart update mechanism when transactions change
- adjusted account balance chart update when an account has no transactions

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountBalanceChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountBalanceChart.java
@@ -33,26 +33,22 @@ public class AccountBalanceChart extends TimelineChart{
         try
         {
             suspendUpdate(true);
-
+            
             for (ISeries s : getSeriesSet().getSeries())
                 getSeriesSet().deleteSeries(s.getId());
-
+            
             if (account == null)
                 return;
-
+            
             List<AccountTransaction> tx = account.getTransactions();
             
-            LocalDate now = LocalDate.now();
-            LocalDate yesterday = LocalDate.now().minusDays(1);
-            
             if (tx.isEmpty()) {
-                //add something to update the chart when switching to an account without transactions
-                addDateSeries(new LocalDate[] {yesterday,now}, new double[]{0,0}, Colors.CASH, account.getName());
                 adjustRange();
                 suspendUpdate(false);
                 return;
             }
             
+            LocalDate now = LocalDate.now();
             LocalDate start = tx.get(0).getDateTime().toLocalDate();
             LocalDate end = tx.get(tx.size() - 1).getDateTime().toLocalDate();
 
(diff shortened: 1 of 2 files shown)


tomcastro89:
#962

tomcastro89:
I am currently unsure on how to add tests.
What is your strategy regarding testing ? In this case I could not decide which kind of tests would be suitable (Integration tests for updateChart or AccountBalanceChart in general or heavy UI tests using e.g. SWTBot)

buchen:
> What is your strategy regarding testing ?

Well... my strategy is currently: have as many calculations as possible in the *name.abuchen.portfolio* plugin and do plain unit tests there. I have no UI tests at all - such as SWTBot. So far that worked quite okay because the UIs are fairly separate (and the community quickly detects bugs and I can fix them :wink:). Of course, I am always happy to learn more. Tests could go into *n.a.p.ui.tests* plugin.

I'll have now a look at your change. But it looks good also to extract the separate class.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountBalanceChart.java
```diff
@@ -0,0 +1,91 @@
...
+                getSeriesSet().deleteSeries(s.getId());
+            
+            if (account == null)
+                return;
+            
+            List<AccountTransaction> tx = account.getTransactions();
+            
+            if (tx.isEmpty()) {
+                adjustRange();
+                suspendUpdate(false);
```
buchen: Isn't the ```suspendUpdate(false)``` already part of the *finally* block? And if we need the ```adjustRange``` always, we might as well move it there as well.

I would take the liberty to quickly change that, squash your commits and merge it.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AccountListView.java
```diff
@@ -362,7 +346,31 @@ protected void createBottomTable(Composite parent)
 
         item = new CTabItem(folder, SWT.NONE);
         item.setText(Messages.TabAccountBalanceChart);
-        item.setControl(createAccountBalanceChart(folder));
+        accountBalanceChart = new AccountBalanceChart(folder);
+        folder.addSelectionListener(new SelectionListener() {
+            @Override
+            public void widgetSelected(SelectionEvent e)
+            {
+                accountBalanceChart.updateChart((Account) ((IStructuredSelection) accounts.getSelection()).getFirstElement(),exchangeRateProviderFactory);
```
buchen: Just FYI - Instead of the cast, you could also use ```accounts.getStructuredSelection()```

buchen:
Merged. Danke! Thanks!

buchen:
BTW, I hope this commit still counts towards your #Hacktoberfest. Reading the site it says "PRs can be to any public repo on GitHub, not just the ones highlighted. The PR must contain commits you made yourself." It does not make any mentioned about merging, etc. Anyway, this one should count because it fixed the issue. 😄

PR #1046, opened 2018-10-20
Issue 977

tomcastro89:
- adjusted drag and drop mechanisms to handle a list of securities instead of a single security

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/ClientEditorSidebar.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/ClientEditorSidebar.java
@@ -141,7 +141,8 @@ import name.abuchen.portfolio.ui.util.SimpleAction;
                 if (subMenu == null || action == null)
                     continue;
 
-                // cannot use the original action b/c it will not highlight the selected entry
+                // cannot use the original action b/c it will not highlight the
+                // selected entry
                 // in the sidebar
                 String text = indent > Sidebar.STEP ? "- " + action.getText() : action.getText(); //$NON-NLS-1$
                 SimpleAction menuAction = new SimpleAction(text, a -> sidebar.select(entry));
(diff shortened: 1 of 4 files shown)


tomcastro89:
#977

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dnd/SecurityDragListener.java
```diff
@@ -20,34 +24,37 @@ public SecurityDragListener(StructuredViewer viewer)
...
+        Iterator<Security> selectionIterator = selection.iterator();
+        while(selectionIterator.hasNext()) {
+            Object object = selectionIterator.next();
+            Security selectedSecurity;
+            if(object instanceof Security) {
+                selectedSecurity = (Security) object;
+                selectedSecurities.add(selectedSecurity);
+            }else if(object instanceof Adaptable) {
+                selectedSecurity = ((Adaptable) object).adapt(Security.class);
+                selectedSecurities.add(selectedSecurity);
```
buchen: I believe we need to have a check for *null* here. Previously, it there was only one adoptable, returning null was not a problem because it meant just that: no security selected. However, in the list of securities, there must not be null values.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/ClientEditorSidebar.java
```diff
@@ -213,23 +213,25 @@ public void drop(DropTargetEvent event)
...
-                        // if the security is dragged from another file, add
-                        // a deep copy to the client's securities list
-                        if (!editor.getClient().getSecurities().contains(security))
-                        {
-                            security = security.deepCopy();
-                            editor.getClient().addSecurity(security);
+                    List<Security> securities = SecurityTransfer.getTransfer().getSecurities();
+                    if (securities != null)
+                    {   
+                        for(Security security : securities) {
```
buchen: If you are using Eclipse, then the default formatter for this project should be configured by default. If you run *ctrl-f* it also should format the file. It would help me if the code is formatted with the project formatter because then later commits do not include formatting changes and are easier to read. And here the curly brace would be on the next line. Just a minor thing...

buchen:
Merged.

I did one more change to get rid of warnings: Instead of ```<Security>``` (which the iterator definitely does not return, but which the compiler cannot know) and said ```<?>``` (which not only gets rid of the compiler warning, but also reflects the state better because we don't know what is in the iterator).

```
Iterator<?> selectionIterator = selection.iterator();
```

PR #1048, opened 2018-10-23
Enhancement while importing CSV files

cmaoling:
Not sure, whether I am the only one with this issue, but I am unable to import CSV files, which have Umlaute in the filename. This pull request uses the nio library to convert the filename to UTF-8 charset before reading the stream.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
@@ -3,9 +3,15 @@ package name.abuchen.portfolio.datatransfer.csv;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.net.URI;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.FieldPosition;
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
```diff
@@ -457,45 +463,70 @@ public void setFirstLineHeader(boolean isFirstLineHeader)
...
+        byte[] ptext = inputFile.toString().getBytes(StandardCharsets.UTF_8);
+        String str = new String(ptext, StandardCharsets.ISO_8859_1);
+        Path path = Paths.get(URI.create("file://" + str)); //$NON-NLS-1$
+        try (InputStream stream = Files.newInputStream(path))
+        {
+            processStream(stream);
+        }
+        catch (IOException e)
+        {
+            if (buffer != null)
```
buchen: BTW, my IDE says here *Redundant null check: The variable buffer cannot be null at this location*. Which I agree too. I will change the code accordingly.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
```diff
@@ -457,45 +463,70 @@ public void setFirstLineHeader(boolean isFirstLineHeader)
         return columns;
     }
 
-    public void processFile() throws IOException
+    public void processStream(InputStream stream) throws IOException
```
buchen: The ```processStream``` method then should be private.

buchen:
Merged.

Very strange problem. Sounds to me like the file open dialog on Linux is broken and is not providing the right encoding.... Anyways, if it helps, the workaround is fairly small.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
cmaoling: Valid Point.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
cmaoling: Yep. This is an orphan. While working in it I wanted to get the initial exception... But it is useless now...

PR #1049, opened 2018-10-25
Enhancement more OTF edits

cmaoling:
- new function to get sibling Type in AccountTransaction
- enable Portfolio-Transfers (beside AccountTransaction-Transfers)
- introduce generic revertAccount(Transaction)Action
- introduce Convert Actions to all AccountTransactions
- enable all AccountTransactions to be changed between each other
- enable OTF-edit of transactiontype in SecurityListView


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java
```diff
@@ -918,7 +921,7 @@ public String getText(Object element)
         support.addColumn(column);
 
         column = new NoteColumn();
-        column.setEditingSupport(null);
+        new StringEditingSupport(AccountTransaction.class, "note").addListener(this).attachTo(column); //$NON-NLS-1$
```
buchen: The ```NoteColumn``` already has editing support. If then, it should be activated here, e.g. remove the "null" assignment that deletes the editing.

This code also does not work for me. And it would limit it to AccountTransactions, but there are also many PortfolioTransactions. If at all, it should go against the ```Annotated``` interface. Keep in mind, that for transactions with ```CrossEntry``` *both* notes should be updated.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/actions/ConvertToDepositRemovalAction.java
```diff
@@ -0,0 +1,51 @@
...
+    public void run()
+    {
+        AccountTransaction accountTransaction = transaction.getTransaction();
+
+        if (AccountTransaction.Type.DEPOSIT.equals(accountTransaction.getType())
+                        || AccountTransaction.Type.FEES_REFUND.equals(accountTransaction.getType())
+                        || AccountTransaction.Type.TAX_REFUND.equals(accountTransaction.getType())
+                        || AccountTransaction.Type.INTEREST.equals(accountTransaction.getType()))
+            accountTransaction.setType(AccountTransaction.Type.DEPOSIT);
+        else if (AccountTransaction.Type.REMOVAL.equals(accountTransaction.getType())
```
buchen: REMOVAL is converted to REMOVAL? That is kind of unnecessary, isn't it?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/actions/ConvertToDepositRemovalAction.java
```diff
@@ -0,0 +1,51 @@
...
+                        )
+            throw new IllegalArgumentException();
+    }
+
+    @Override
+    public void run()
+    {
+        AccountTransaction accountTransaction = transaction.getTransaction();
+
+        if (AccountTransaction.Type.DEPOSIT.equals(accountTransaction.getType())
```
buchen: DEPOSIT is converted to DEPOSIT?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/actions/ConvertToTaxAction.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void run()
+    {
+        AccountTransaction accountTransaction = transaction.getTransaction();
+
+        if (AccountTransaction.Type.DEPOSIT.equals(accountTransaction.getType())
+                        || AccountTransaction.Type.FEES_REFUND.equals(accountTransaction.getType())
+                        || AccountTransaction.Type.TAX_REFUND.equals(accountTransaction.getType())
```
buchen: TAX_REFUND to TAX_REFUND?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/actions/RevertAccountAction.java
```diff
@@ -0,0 +1,41 @@
...
+    private final Client client;
+    private final TransactionPair<AccountTransaction> transaction;
+
+    public RevertAccountAction(Client client, TransactionPair<AccountTransaction> transaction)
+    {
+        this.client = client;
+        this.transaction = transaction;
+
+        if (transaction.getTransaction().getType() != AccountTransaction.Type.DEPOSIT
+                        && transaction.getTransaction().getType() != AccountTransaction.Type.DIVIDENDS
```
buchen: At least for DIVIDENDS there is no sibling defined. Would this work?

Plus: dividends can have taxes, other transactions not necessarily.

I also have the feeling we need some more tests here. For example, I want to add taxes to interest transactions. But does that apply to interest_charge as well? And probably the other way around, right. Taxes are not removed, but actually "paid" back. For now it is okay, but in the future one always has to keep in mind to change those "revert" actions.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/InvestmentPlanTest.java
```diff
@@ -162,40 +162,40 @@ public void testGenerationOfDepositTransaction()
         assertThat(((AccountTransaction) tx.get(1)).getType(), is(AccountTransaction.Type.DEPOSIT));
     }
 
-    @Test
-    public void testNoGenerationWithStartInFuture()
```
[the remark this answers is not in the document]
buchen: BTW, I found the issue with this tests - due to the test design, it failed on some days.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/actions/RevertAccountAction.java
cmaoling: Valid point, the Units are not covered by this code at this point.
Regarding the sibling of DIVIDEND: it seems like I introduced DIVIDEND_CHARGE in my code in March 2018...

Issue #973, opened 2018-05-10
Missing dark theme support


Anth0rx:
When using a dark theme PP won't adapt to it. It would be cool to support custom colors for icons, fonts diagrams and backgrounds.

---

Current open issues:

- [x] Embedded HTML pie charts should use light/dark CSS
- [x] color of text in diagram legend does not change
- [x] charts still have white background - for example when displayed on the dashboard
- [x] monthly / yearly returns table looses the colored background when in dark mode
- [x] currency diagrams use black color for data series - not visible in dark mode
- [x] Money Trail Tooltip
- [x] Preferences in light mode have grey bars
- [x] highlight colors in tables (for example if the trade is open)
- [ ] check use of ```org.eclipse.swt.display.useSystemTheme```

If you find more issues, post screenshots to this issue.

Anth0rx:
Is there any progress on this matter @buchen?

buchen:
I know Eclipse (the framework I use) has some kind of support for theming. I am happy if somebody can show/explain/prototype how to exactly to this. At the moment, I do not have lot of time (and other feature requests) and unless there are many, many requests, I would not spend time on a dark theme at the moment.

Anth0rx:
Okay, I understand. Unfortunately I do not have enough experience with Java/Eclipse to provide support for this.

tomcastro89:
+1 for this request
The Eclipse IDE itself has a darkmode, but even the "best" there it is not perfect and needs some adjustments ( scrollbars are white etc. )

The strategies are:
1. try to add a dark theme from the eclipse marketplace (https://marketplace.eclipse.org/content/darkest-dark-theme-devstyle)
2. implement a theme ourselves (which takes some effort)

buchen:
So sieht das Dark Theme bei PP aktuell aus:

<img width="1292" alt="Bildschirmfoto 2020-09-13 um 20 27 55" src="https://user-images.githubusercontent.com/587976/93025877-fc820100-f601-11ea-8987-c1b082287c46.png">

buchen:
If you want to help styling the dark theme of PP, then this is a good introductory article:
https://www.vogella.com/tutorials/Eclipse4CSS/article.html

SWT uses CSS files to style the components, but in the end it is a custom implementation, not HTML. Therefore the behavior is not always really predictable. There are lots of custom properties starting with "swt-". With a little bit of trial and error I was able to create the result above.

sebasbaumh:
[...]

PR #1051, opened 2018-10-26
Issue 973: Add Dark Theme support

tomcastro89:
- Added Preference
- Switch WelcomePage Background Color depending on Preference

This is an example of how we could implement a dark mode.
If this is a reasonable implementation we can try to

1. find proper color codes / material designs
2. change every UI element to ask the ColorManager for its colors

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/ColorThemeManager.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/ColorThemeManager.java
@@ -1,0 +1,19 @@
+package name.abuchen.portfolio.ui;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.widgets.Display;
+
+public class ColorThemeManager
+{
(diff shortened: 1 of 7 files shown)


tomcastro89:
#973

buchen:
Hi @tomcastro89,

I am sympathetic to a dark theme support. Not so much because I personally use the dark themes (I never do), but because I would hope to get more flexibility with colors and font sizes, etc.

I am a little skeptical of such a brute force approach. It is a lot of work. My fear is this: it could spill into many UI classes, it might be hard to style SWT default widgets such as tabs, toolbars, etc. And it will never allow dynamic change of the theme.

**Did you take a look at the [Theming Support](http://www.vogella.com/tutorials/Eclipse4CSS/article.html) in Eclipse 4 RCP?** Do you have experience with this theming where you say that is more work/more impact on the code? I would assume we would anyway need to use this theming support to change all of the default elements.

With the "theming" I do not want to imply that the end-user gets arbitrary choice in styling. I don't think that makes sense. There are maybe two carefully crafted themes: a light one and a dark one.

Andreas.

buchen:
Implemented via #973

PR #1169, opened 2019-06-07
fixed time import in case time is given including seconds

rbaeumler:
if time in csf file is given in format hh:mm:ss the import currently fails because the parsing of mm:ss to integer fails.
With this change tokens after the minutes are disregarded.


buchen:
Super. Vielen Dank! Kommt mit der nächsten Version - ich denke Montag finde ich Zeit.

PR #1173, opened 2019-06-17
allow more foreign currencies in viac pdf extractor buy transaction

Mauin:
The VIAC PDF extractor currently only supports USD as foreign
currencies. In reality there can be other foreign currencies as well.

This change allows for any foreign currency in the VIAC PDF extractor
and adds a test for this case.


buchen:
Thanks!!!

PR #1174, opened 2019-06-17
add support for VIAC dividend PDFs

Mauin:
Adds support for dividend PDFs from VIAC. 

With these changes I still see one issue: CSIF Pacific ex Japan currency is generally USD. However the dividend PDF for some reason denotes the dividend in CHF. 

This leads to an error in the import. Any special handling that should be added for this case?


buchen:
Thanks!!!

PR #1179, opened 2019-07-02
Postfinance pdf extractor

steve84:
Initial set of PostFinance (swiss bank) transactions. Transfer in/out and deposit transactions are not possible at the moment (same document headers).


buchen:
Super! Vielen Dank! Thanks a lot. I plan to build and publish a version this weekend.

Issue #1180, opened 2019-07-03
Add extraETF - https://extraetf.com as default bookmark


MarkusTiede:
(imho) it would be useful having [extraETF](https://de.extraetf.com) also as a default bookmark available.

MarkusTiede:
I'll be filing a PR for this later.

PR #1181, opened 2019-07-03
Add extraETF.com linking

MarkusTiede:
Implementation and "fix" for #1180.


buchen:
Thanks. Merged.

They way it is designed at the moment means that this bookmark is only available to *new* files. I do create copy of the default list whenever a file is created. That copy can then be edited. Maybe I add a menu to add those standard bookmarks to the current file as well.

PR #1199, opened 2019-08-09
Python script to convert exported portfolio into Yahoo-Finance format.

DirkBalthasar:
Very useful for generation of mobile access to portfolio.

### DIFF
--- a/export/convertYahoo.py
+++ b/export/convertYahoo.py
@@ -1,0 +1,109 @@
+# -*- coding: utf-8 -*-
+import os
+import csv
+
+# Convert Portfolio-performance csv-export files to yahoo finance csv format.
+# Purpose: Portfolio performance is great but not mobile. The Yahoo finance export enable portfolio overview for mobile devices via Yahoo.
+# License: Eclipse Public License - v 1.0
+# Author: Dirk Balthasar, 2019
+#
+# Prerequisite:
+# 1. Create free yahoo finance account. (This is not a recommendation from my side for this product) 
+# 2. Be aware that your data will be uploaded into the cloud
+# 3. Make sure that you are fine with the yahoo terms of usage
+#
+# Usage: 
+# 0. Install python 3
+# 1. Export portfolio: File->Export
+# 2. Copy scrip convert.py to export directory
+# 3. run convert.py (script opens all csv files and creates one csv file "yahoo.csv" that can be uploaded to yahoo finance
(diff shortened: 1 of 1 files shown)


buchen:
Hi @DirkBalthasar,

I am sorry for not commenting on this PR earlier. To be honest, I was not quite sure what do to with it.

I understand the Python converts the CSV Export of PP into such a format that it can immediately imported to Yahoo Finance.

What do you suggest to do?
* Just add it to the repository for others to copy & use
* Add it to the application download
* Add a custom export functionality to PP to create this format directly? I have done something similar for "Aktienfreunde.net". But I can only use Java for this and not Python - I also do not want to embed a Python script engine

At least with respect to option 1, the PR is kind of a place where users can download the script.

What do you think?

DirkBalthasar:
Hello Andreas, 

I think we should start with the first or second option you mentioned, if the PP CSV format is not subject to changes.
The CSV is also a good way to keep export filters away from the main system and distribute the workload on other developers and languages. 
Also, tools which are not subject to PP functionality wise can easily start on the CSV interface.

Full integration into PP from my point of view seems to be more comfortable for the user, but comes with some issues in the PP development:
1. If PP would support yahoo, requests for other services will follow from the user base.
    I have a second python export script for "simplywallst" that I could share, you mentioned   "Aktienfreunde.net". Altogether this could potentially generate some work and maintenance in PP.
2. Loss of scope / Feature creep(?)
3. If PP supports export functionality, the next question could be how to automize the upload to the cloud. 
I would avoid the full integration if the user base is not heavily demanding this. But it is a strategic decision you need to take for PP. 
If you want to go into direction of full integration, I would recommend keeping the core small, i.e. use existing Java, no python binding. If you decide to go that route, I can implement / support the "yahoo" export filter, if you send me the entry point for "Aktienfreunde.net" as template. 

Best Regards
Dirk

inv-trad:
Hi Dirk, 
is this the entry point you are looking for? 

https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/AktienfreundeNetExporter.java

Best regards
Sven

DirkBalthasar:
Thanks Sven, 
this looks easy to adapt. 
Good starting point for Chrismas vacation to renovate the yahoo exporter.

Dirk

syedabdullahali (approved):

PR #1212, opened 2019-08-24
Fix: only look at rows having the same number of columns as header

KopolJunam:
In HTMLTableQuoteFeed to extract prices we only look at rows that have the same number of columns as the header column to prevent "index out of bounds" and ignore candidates that are not relevant in the first place (like some extra info lines at the beginning or end of table).


buchen:
I rebased the PR branch on top of master and merged it. Thanks! Looking forward to the VIAC stuff.

When creating a pull request, it helps if the new code is on a separate branch. Instead of merging the new changes of the master branch into your feature branch, rebase the feature branch on top of the tip of the master. This way all the unnecessary merge commits are not distracting in the pull request.

KopolJunam:
You're absolutely right about me having to learn more about rebasing/squashing commits. So far I have never worked on an OpenSource project but only on inhouse development where having the whole history is often a desired effect. In other words we (or at least I) normally always merge and never rebase. This clearly is not practical in an open source scenario for reasons that are obvious to you but were not yet so obvious to me.

buchen:
No problem. Thanks for the contribution.

I believe the difference is: you have your own fork, a separate repository, with your own "master" branch - independent of the "master" branch of this repository. In my opinion, the easiest thing is to always create a new feature branch on this master and not work with the master branch of the fork.

KopolJunam:
I think I know what I need to know now, we will see with my next pull request :-) As for my own fork: I didn't know I even have the right to create feature branch on your repo. But having my own fork is totally fine. I have a few bigger ideas I'd like to tinker around with and to have my own experimental branches is practical. Due to my personal situation I am still dreaming of choosing a currency a chart is based on. Another idea is when creating new classifications to split an equity according to number it appears in a certain depot. Something along those lines. Anyway these are things I will discuss with you and the community before starting serious work but before any such discussion I'd just like to play around a little bit in the code to get a feeling for what can and needs to be done and for that I guess my fork is handy.

PR #1222, opened 2019-09-02
Make HTMLTableQuoteFeed derivable - part 2

KopolJunam:
It took me a while to learn about git squash, rebase, cherry pick etc.. and it shows. This pull request only has one relevant commit in it.

I changed HTMLTableQuoteFeed a little more so now it is even better derivable = the new class CSQuoteFeed (no pull request yet, but it will come as soon as this PR has beeen approved) only has to implement very few methods.

My changes are:

- the user agent and ignoring the content type of the returned data will be obtained through (virtual) methods.
- The test method "doLoad" is non-static now so it can also be used by derived classes.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeed.java
@@ -388,14 +388,23 @@ public class HTMLTableQuoteFeed implements QuoteFeed
         return Collections.emptyList();
     }
 
+    protected String getUserAgent()
+    {
+        return OnlineHelper.getUserAgent();
+    }
+
+    protected boolean isIgnoreContentType()
+    {
+        return false;
+    }
+
     protected List<LatestSecurityPrice> parseFromURL(String url, List<Exception> errors)
     {
         try
         {
             String escapedUrl = new URI(url).toASCIIString();
-            return parse(escapedUrl,
-                            Jsoup.connect(escapedUrl).userAgent(OnlineHelper.getUserAgent()).timeout(30000).get(),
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeed.java
```diff
@@ -582,12 +591,12 @@ public static void main(String[] args) throws IOException
         PrintWriter writer = new PrintWriter(System.out); // NOSONAR
         for (String arg : args)
             if (arg.charAt(0) != '#')
-                doLoad(arg, writer);
+                new HTMLTableQuoteFeed().doLoad(arg, writer);
```
buchen: I am not sure if this will really help you as the static test method always instantiates HTMLTableQuoteFeed.

Why not move the loop (in the main method) into the class as well?

Then each class can have a very simple local test method which is something like
```java
public static void main(String[] args) throws IOException
{
    new HTMLTableQuoteFeed().runWith(args);
}
```

However, I can merge the change nonetheless (once I am back at my laptop) - it is just a remark.

KopolJunam: It does help me as the only static test method is "main" now. So in my new class it just looks like this:
```
       for (String arg : args)
            if (arg.charAt(0) != '#')
                new CSQuoteFeed().doLoad(arg, writer);
```
doLoad is inherited from the HTML feed. So all fine so far. Of course this test framework could be improved a little bit if you want but as far as providing an URL as parameter and dumping the prices everything works fine with base class and derived class. I think if the rest is fine with you maybe just merge it please.

PR #1223, opened 2019-09-02
Add VIAC/Credit Suisse funds quote feed

KopolJunam:
Even though a user wrote in the forum that the quotes of the Credit Suisse institutional funds can also be obtained from the ARD website I think this quote feed still makes sense as it gets the quotes from more than 10 years back whereas ARD only has them for the past month.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/security/AbstractQuoteProviderPage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/security/AbstractQuoteProviderPage.java
@@ -38,6 +38,7 @@ import name.abuchen.portfolio.model.Exchange;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.online.QuoteFeed;
 import name.abuchen.portfolio.online.impl.AlphavantageQuoteFeed;
+import name.abuchen.portfolio.online.impl.CSQuoteFeed;
 import name.abuchen.portfolio.online.impl.EurostatHICPQuoteFeed;
 import name.abuchen.portfolio.online.impl.HTMLTableQuoteFeed;
 import name.abuchen.portfolio.ui.Messages;
@@ -129,11 +130,11 @@ public abstract class AbstractQuoteProviderPage extends AbstractPage
 
     private Group grpQuoteFeed;
     private Label labelDetailData;
-    
+
     private ComboViewer comboExchange;
     private Text textFeedURL;
     private Text textTicker;
-    
+
     private PropertyChangeListener tickerSymbolPropertyChangeListener = e -> onTickerSymbolChanged();
 
     private final EditSecurityModel model;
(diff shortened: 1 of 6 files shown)


buchen:
Thanks!

Let's see how long the "trick" with the "curl" user agent works. However, with a separate quote feed, we would have options for further fine tuning.

KopolJunam:
Thank you for adding "my stuff" and for making me a "GIT master" :-) I have now established a good and useful GIT workflow for myself.

image

From the "workingbase" branch I can branch off my feature branches to work on so that the launch configuration and the ignores are they way that I need them. At the end I can just branch off a branch for the pull request into which I can cherry pick the commits and squash them to one. The master I keep in sync with upstream via rebasing.

I assume the "curl" trick is how they do their automated tests.

If you add VIAC to the release notes somehow (and I am sure you will) I will get back in touch with them. Maybe they will use this info somehow, e.g. put it on their [blog-like "Academy"](https://viac.ch/academy/). I don't have very high hopes but to let them know about this feature can't hurt..

buchen:
> If you add VIAC to the release notes somehow (and I am sure you will) 

Yes, of course. I might not get around to creating a new release this weekend... could take 14 days or something.

KopolJunam:
No problem. Is there a way to install a snapshot? I looked through a few projects whether I find something but I am not familiar with Eclipse enough. I want to use my new feature but not start PP via Eclipse IDE all the time.

buchen:
> Is there a way to install a snapshot?

If you run the Maven build, the repository is created in ```portfolio-product/target/repository```. You can use that to update your installation (change URL in settings).

Or you run the Maven build with the ```mvn clean verify -Ppackage-distro``` command, then you have the ZIP files created in ```portfolio-product/target/products```.

KopolJunam:
Thank you for the info. Somehow it doesn't work and stops with the error below. I thought I'd let you know because maybe the information could be relevant to you. Please do not spend (waste) more time on explaining me how it works unless one sentence is enough. I just start PP via Eclipse until the next release is out.

Btw, I wondered what the "Verified" next to the commits meant and now created and installed a gpg key too so I hope my next contribution will be "verified" too.

```
[ERROR] Internal error: java.lang.RuntimeException: Could not resolve target platform specification artifact name.abuchen.portfolio:portfolio-target-definition:target:0.41.2-SNAPSHOT -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: Could not resolve target platform specification artifact name.abuchen.portfolio:portfolio-target-definition:target:0.41.2-SNAPSHOT
```

buchen:
> Thank you for the info. Somehow it doesn't work and stops with the error below.

Are you using the Maven version 3.6.1? This particular version does not work - see #1168 Any other 3.x should work.

But then again, the RuntimeError confuses me. And it fails while resolving the target definition. Not clear what the problem is...

KopolJunam:
I am using maven 3.5.0. If you're interested in looking into this error further I will gladly be of assistance otherwise we can also let this rest.

PR #1247, opened 2019-10-06
Bug 1245: Update portfolio-target-definition

sennahLessert:
this commit fixes wrong dependency package version and uri


sennahLessert:
#1246

PR #1253, opened 2019-10-14
Swissquote pdf import

tzemp:
Swissquote PDF Import.

The documents are kinda similar to those of PostFinance, with minor adjustments :-)


buchen:
Super! Thanks for the contribution.

PR #1267, opened 2019-10-27
csv import: proof of concept for importing into multiple different ac…

verybadsoldier:
…counts at once by providing extra columns with account names

*This PR is not finished and just meant as a proof of concept.*

This is meant to implement what was proposed here:
https://forum.portfolio-performance.info/t/batch-import-automatisierter-import/6038

So with this PR CSV files are allowed to have 2 extra columns "account name" and "portofolio name". If a row has one or both of these values set then the row is imported into the particular account/portfolio (instead of the account/portfolio chosen in the UI combo box).

So this allows to have a CSV file which can fill multiple accounts inside PP.
image

This PR is not finished yet and needs alot more polish. I opened this PR as I am hoping to get feedback regarding the general approach. If the approach is considered feasible then I will try to complete the code in a proper way.


buchen:
Hi @verybadsoldier,

first: Deutsch oder Englisch? Die Commit Historie hätte ich gerne in Englisch, aber wir können hier auch auf Deutsch diskutieren.

second: adding the option to define account and portfolio in the import source (be it CSV, but also PDF) makes absolutely sense. Let's add that.

third: I haven't come around yet, but with #1193 I wanted to do the import on a JSON definition. That JSON definition includes account and portfolio. And then convert CSV and PDF into that JSON format - let the use change whatever is needed - and then convert to the transaction object and add to the Client file.

The JSON object also has the property "account" - see [JTransaction](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/json/JTransaction.java). That would be my idea for setting the account upon import, let the user review / change, and then add the transaction to the account.

Alas, that does not exist yet.

I quickly went over your change. Generally it looks good to me. The change extends the import mechanism as it exists today.

There is one thing I do not like. You added a property ```accountContext``` to the ```AccountTransaction```. Because that object is serialized into XML, the AccountTransaction would always carry that property. But it is needed only for the import.

My proposal would be to add those properties to the [Extractor.Item](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java#L50). It has already a generic ```Object data```  but I think we can add primaryAccount, secondaryAccount, primaryPortfolio, secondaryPortfolio (basically the names that are also available in the ReviewDialog).

The ReviewDialog then would say: if the special property on the importer is set, then import the transaction into that account/portfolio. If not, then use the default account/portfolio selected on the top of the page. Later (with another change) we could allow the user to change that line by line.

buchen:
BTW, thanks for pushing this "unfinished" change. I like it that we can discuss and agree on an approach - based on code - before you spend a long time polishing details.

buchen:
@verybadsoldier you let me know when I should have a look again

BTW, the test failure is because the AccountTransaction still has the ```accountContext``` property.

verybadsoldier:
Hi, Andreas, danke für die Antwort! Ja, gern auf deutsch. Da du offenbar mit dem Ansatz grob einverstanden bist, hab ich mal in dem Sinne weitergemacht und deine Vorschläge eingearbeitet.

>There is one thing I do not like. You added a property accountContext to the AccountTransaction. Because that object is serialized into XML, the AccountTransaction would always carry that property. But it is needed only for the import.

>My proposal would be to add those properties to the Extractor.Item. It has already a generic Object data but I think we can add primaryAccount, secondaryAccount, primaryPortfolio, secondaryPortfolio (basically the names that are also available in the ReviewDialog).

Ja, stimmt. Ich muss sagen, ich hab die Abgrenzung `AccountTransaction` vs `AccountTransactionItem` nicht so 100% durchdrungen. Da einige Klassen dann sogar schon Felder für Account und/oder Portfolio mitgebracht haben, habe ich die erstmal nachgenutzt.

Ich hab das jetzt auf jeden Fall auf `Extractor.Item` umgebaut mit der Konsequenz, dass die eben genannten vorhanden Account/Portfolio-Felder nicht mehr genutzt werden.

Zu `ReviewExtractedItemsPage`:
In dem Fenster werden ja die zu importierenden Daten als Vorschau angezeigt. Dort sollten also ebenfalls die drei neuen Spalten angezeigt werden. Nach meinem Verständnis vom Code zeigt das Fenster unabhängig  von der Import-Art (Kontoumsätze, Depots, Wertpapiere) immer die gleichen fixen Spalten, die evtl. für den aktuellen Fall nicht relevant sind. Ich hab jetzt drei Spalten "Konto", "Zielkonto" und "Depot" hinzugefügt ("Zieldepot" wird mMn nicht benötigt).
Weißt nicht, ob man das noch schöner machen könnte?

> BTW, the test failure is because the AccountTransaction still has the accountContext property.

Ah ja danke. Hab ich gefixt.

Also wenn du magst, kannst du jetzt gerne nochmal drauf gucken und Änderungswünsche äußern. Aus meiner (begrenzten) Sicht, wäre es jetzt für mich ok, aber da du einen viel besseren Überblick hast, hast du sicherlich noch potentielle Probleme im Blick, die ich nicht sehe.

Danke dir!

Ich werde dann am Ende nochmal ein Rebase spendieren und die Commits zu einem zusammen fassen, also squashen (wenn du möchtest).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractor.java
```diff
@@ -102,6 +108,12 @@ void extract(List<Item> items, String[] rawValues, Map<String, Column> field2col
 
         Unit grossAmount = extractGrossAmount(rawValues, field2column, amount);
 
+        Account account = getAccount(getClient(), rawValues, field2column);
+        Account account2nd = getAccount(getClient(), rawValues, field2column, true);
+
+        Portfolio portfolio = getPortfolio(getClient(), rawValues, field2column);
```
buchen: Konto und Portfolio werden ausgelesen, aber nie ins Extractor.Item gesetzt.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioExtractor.java
```diff
@@ -45,6 +47,8 @@
         fields.add(new DateField("date-quote", Messages.CSVColumn_DateQuote).setOptional(true)); //$NON-NLS-1$
         fields.add(new AmountField("quote", Messages.CSVColumn_Quote, "Schluss", "Schlusskurs", "Close") //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
                         .setOptional(true));
+        fields.add(new AccountNameField("account", Messages.CSVColumn_AccountName).setOptional(true)); //$NON-NLS-1$
+        fields.add(new PortfolioNameField("portfolio", Messages.CSVColumn_PortfolioName).setOptional(true)); //$NON-NLS-1$
```
buchen: Die Felder werden definiert, aber nie gelesen und nie in das Extractor.Item gesetzt.

vipuldamani:
Very Good Morning Sir,
is that solve multi pal client trade import via csv ?
Thanks In Advance.
Have a Nice Time.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractor.java
verybadsoldier: Ohh sorry, blöd von mir. Evtl. Sachen, die ich im POC angefangen hatte und dann beim Finalisieren übersehen hab :/

buchen:
> After this Need to Filter and /or search Option in all report and Performance tab.

I have noted this down. But I cannot promise anytime line. But it sure would be nice to have such a filter there. 😄

vipuldamani:
Dear Buchen Sir,
Thanks For Same.
Have A Nice Time.

REGARDS, 
VIPUL DAMANI

    On Tuesday, 12 November, 2019, 12:04:11 pm IST, Andreas Buchen <notifications@github.com> wrote:  
 
 

After this Need to Filter and /or search Option in all report and Performance tab.


I have noted this down. But I cannot promise anytime line. But it sure would be nice to have such a filter there. 😄

PR #1284, opened 2019-11-19
HTMLTableQuoteFeed & skip tables w/ empty table rows

Ragas13:
https://forum.portfolio-performance.info/t/keine-coinmarketcap-kurse-seit-dem-06-11-19/6349/

Tables are not handled, if they are duplicated. First occurance with header only and second occurance repeated with header and values.


buchen:
Drin. Danke! 👏

PR #1293, opened 2019-11-30
SecuritiesTable ChangeOnPrevious

Ragas13:
Change calculation of col. "ChangeOnPrevious" from current date logic where data provided by LatestSecurityPrice, as this record set is only updated if quotes are online updated, manual maintained quote prices bypassing this logic.

This change is use now the closest fitting date to current date and the previous date.

https://forum.portfolio-performance.info/t/kursaenderung-zum-vortag-ist-falsch/5189/22


buchen:
rebased and merged.

Ich habe den "lookup" von aktuellen und "previous" in eine Methode von ```Security``` gepackt da es sich da einfacher testen lässt.

PR #1302, opened 2019-12-16
Feature bitfinex

debegr92:
Added QuoteFeed for cryptocurrency exchange Bitfinex.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/BitfinexQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/BitfinexQuoteFeed.java
@@ -10,7 +10,6 @@ import java.util.Collections;
 import java.util.List;
 
 import org.json.simple.JSONArray;
-import org.json.simple.JSONObject;
 import org.json.simple.JSONValue;
 
 import name.abuchen.portfolio.Messages;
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/BitfinexQuoteFeed.java
```diff
@@ -0,0 +1,201 @@
...
+    @Override
+    public String getId()
+    {
+        return ID;
+    }
+
+    @Override
+    public String getName()
+    {
+        return "Bitfinex Cryptocurrency Exchange"; //$NON-NLS-1$
```
SimonScholz: Please consider to localize the name like it is done in https://github.com/buchen/portfolio/blob/004996d700cca625ed8674ca90bbe2708ffc021e/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooFinanceQuoteFeed.java#L56

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/KrakenQuoteFeed.java
```diff
@@ -35,7 +35,7 @@ public String getId()
     @Override
     public String getName()
     {
-        return "Kraken Bitcoin Exchange"; //$NON-NLS-1$
+        return "Kraken Cryptocurrency Exchange"; //$NON-NLS-1$
```
SimonScholz: Please consider to localize this as well.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/BitfinexQuoteFeed.java
```diff
@@ -0,0 +1,201 @@
...
+                    Long high = YahooHelper.asPrice(quoteEntry.get(3).toString());
+                    Long low = YahooHelper.asPrice(quoteEntry.get(4).toString());
+                    Integer volume = YahooHelper.asNumber(quoteEntry.get(5).toString());
+
+                    LatestSecurityPrice price = new LatestSecurityPrice(LocalDate.ofEpochDay(timestamp / 1000 / secondsPerDay), close);
+                    price.setHigh(high);
+                    price.setLow(low);
+                    price.setVolume(volume);
+                    price.setPreviousClose(open);
+
```
Ragas13: There is no need to store all high, low, volume information as is blow up the data storage. Only the latest is needed. ```SecurityPrice``` should be preferable.

debegr92: When there is no need for full OHLCV data, why is there an extended class for this?
If we only should use SecurityPrice class, the open, high, low price should also be removed from KrakenQuoteFeed and AlphaVantageQuoteFeed

buchen: The currently implemented logic is:
If it is the "latest price" then it contains high, low, etc.
For historical quotes, PP only stores the "daily closing price".

The rationale is: to calculate past performance, only the closing price is used. To store high, lows,volumne in XML, it would only blow up the XML. PP does not intent to be a charting tool - there are many (and better) out there.

I will fix ```KrakenQuoteFeed``` as well b/c it does it wrong.

```AlphavantageQuoteFeed``` passes the type of price class to be created to the method:
```java
T price = klass.getConstructor().newInstance();
```

debegr92: I will keep this in mind! Thank you for the explanation.

buchen: FIY - I am in the process of merging the PR.

buchen:
Rebased & merged.

Thanks! Vielen Dank!

PR #1318, opened 2019-12-28
add URLs found in notes of security to URLs context menu.

sebasbaumh:
Adds URLs from notes of a security to the context menu.
Provides an easy way to add multiple custom URLs to a security by just writing them in the notes.
Looks like that:

for that security notes:
javaw_2019-12-28_12-32-25


debegr92:
Thumbs up for this one! 👍

PR #1319, opened 2019-12-28
make chart colors more stable when new years are added to the chart.

sebasbaumh:
replaces the switch from 5 to 10 colors, which lets the colors in the chart jump by a stable color order.


buchen:
Sieht viel besser aus. Der "Sprung" in den Farben war unschön.

PR #1325, opened 2019-12-30
Improve SutorBank PDF import

tenjaa:
https://github.com/buchen/portfolio/issues/1324


buchen:
Hi @tenjaa,

looks good. Thanks for your contribution.

Usually I do not *edit* existing test files, but add new test files. My thinking is always I want to be able to import old files even if there are subtle changes to the format. In this case, it does not make a big difference.

I'll be maybe mid to end of January before I will come around to publishing a new version including your change.

Andreas.

PR #1332, opened 2020-01-05
Add justTRADE pdf support

tenjaa:
The security name takes multiple lines, therefore as a small workaround I only parse ISIN / WKN to identify the security.

The test asserts that the name is null. But when running the app and importing the pdf file, the name of the newly created security is set anyway. How?


buchen:
> The test asserts that the name is null. But when running the app and importing the pdf file, the name of the newly created security is set anyway. How?

I have to admit: I have no idea. 🤔  I just checked the code, but I do not see that the name is assigned anywhere... Anyway, PP expects the name to be available. In other PDF documents the name is also split across multiple lines. In my opinion, it makes sense to add least include the first line. I have adopted the PR accordingly: 55832eddc998d2f99e58cf73ff43bb0a9f8511ad

rebased & merged.

PR #1358, opened 2020-01-16
added support for onvista Vorabpauschale PDF

verybadsoldier:
Is it correct to use ex-date as date for the tax payment?

Not sure if "Vorabpauschale" should be translated?

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OnvistaPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OnvistaPDFExtractor.java
@@ -48,6 +48,7 @@ public class OnvistaPDFExtractor extends AbstractPDFExtractor
         addAccountStatementTransaction();
         addAccountStatementTransaction2017();
         addRegistrationFeeTransaction();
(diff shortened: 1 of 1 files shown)


buchen:
> Not sure if "Vorabpauschale" should be translated?

I fear that these parts of the code - parsing German PDF documents - is anyway only accessible to German speaking developers. So I am lenient with the translation there... it's okay.

> Is it correct to use ex-date as date for the tax payment?

If I see this correctly, you are not using the ex-date but the date, aren't you?

```java
.match("^Nominal Ex-Tag Zahltag Jahreswert Vorabpauschale pro Stück")
// STK 1.212,000 02.01.2020 02.01.2020 EUR 0,1019
.match("STK .* (?<exdate>\\d+\\.\\d+\\.\\d{4}+) (?<date>\\d+\\.\\d+\\.\\d{4}+) (\\w{3}+) .*")
.assign((t, v) -> {
})```

As it is identical here, it is hard to say what makes more sense. I agree that using "date" makes more sense. Then the transaction is at that point when you have to pay taxes.

I'll merge.

What really would be helpful is to add a test case. The Vorabpauschale has no text code which makes it very easy to break old documents once a new version comes around.

buchen:
rebased & merged.

PR #1359, opened 2020-01-16
added testcase for onvista vorabpauschale PDF import

verybadsoldier:
> What really would be helpful is to add a test case. The Vorabpauschale has no text code which makes it very easy to break old documents once a new version comes around.

Out of interest: what is a "text code"? Sounds like some ID which could  identify different version of "Vorabpauschale"-PDFs. Could you point me to an example of other PDFs that contains such a thing?


buchen:
> Out of interest: what is a "text code"?

I wanted to write "test code".... 🤣

Thanks for the test case.

PR #1384, opened 2020-02-15
Remove Umlaute in Deutsche Bank import JSON

cmaoling:
Hi Andreas,
due to the Umlaute the build fails for me.
Thanks,
Nicolas

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/deutsche-bank-purchase.json
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/deutsche-bank-purchase.json
@@ -56,7 +56,7 @@
 					"context": "UNIT",
 					"isOptional": true,
 					"pattern": [
-						"Weitere Provision der Bank bei der börslichen Orderausführung (?<currency>\\w{3}+) (?<amount>[\\d.]+,\\d+)"
+						"Weitere Provision der Bank bei der b.*rslichen Orderausf.*hrung (?<currency>\\w{3}+) (?<amount>[\\d.]+,\\d+)"
 					],
 					"attributes": {
 						"type": "FEE"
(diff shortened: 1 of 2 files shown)


buchen:
> due to the Umlaute the build fails for me.

Is it just your build? Or is it your import of Deutsche Bank documents?

The former would be something related to UTF-8 encoding of the file. The latter would be about the PDF converter.

I am asking because there are many more such files that include umlauts - ebase.json, etc.

cmaoling:
The tests for deutsche bank are failing during the build. 
I have no PDFs at hand to check whether the import at runtime works or fails.
I agree that it might be encoding related...
For my education : where could I change the encoding for one specific (or this) file to be different?

cmaoling:
The tests for deutsche bank are failing during the build. 
I have no PDFs at hand to check whether the import at runtime works or fails.
I agree that it might be encoding related...
For my education : where could I change the encoding for one specific (or this) file to be different?

buchen:
To be honest, it is not clear to me what goes wrong.

On my macOS system, the encoding is UTF-8. I just cloned the repository again into an empty directory so I am pretty sure that the file in the repository is UTF-8.

```
> file -I portfolio/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/deutsche-bank-purchase.json 
portfolio/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/deutsche-bank-purchase.json: application/json; charset=utf-8
```

The JSON definition is read with UTF-8 encoding specified:

https://github.com/buchen/portfolio/blob/0.44.1/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/JSONPDFExtractor.java#L55-L57

And the test case is read in UTF-8 as well:

https://github.com/buchen/portfolio/blob/0.44.1/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFImportAssistant.java#L124


Can you check with "files" what your system says as encoding? Are you on Linux?

cmaoling:
Yep. I am working under Linux. And it seems that the files are ASCII encoded in my repo. Not sure how this happened. When I checkout upstream again, they are UTF-8 encoded :( I cherry-picked them on my master so not sure how come and why git is not complaining that the encode differs...

buchen:
If you do a ```git reset --hard HEAD``` does it "fix" the encoding? Then maybe the git tooling mixed it up?

PR #1388, opened 2020-02-19
Fix sorting by volume of security deposits in security accounts

inv-trad:
This is fixing the problem discussed in the forum: https://forum.portfolio-performance.info/t/sortieren-nach-depotvolumen-in-depotansicht/7621/4

In addition a cosmetic change was done to the ConsorsbankPDFExtractor.


buchen:
Danke!

PR #1404, opened 2020-02-27
Asset statement view - share column margin

Pnda87:
added a space to force the column to have a margin on the right side

margin_share_column


buchen:
Vielen Dank!

PR #1416, opened 2020-03-04
Fix Consorsbank PDF import with multiple account holders

inv-trad:
When selling a security and taxes are supposed to be payed, taxes are
split when there are multiple account holders. The result was that only
half of taxes was recorded.
Added "multipleTimes" flag and extended sections parsing function.
Added test file and test case.


buchen:
Just for my understanding: can there be an arbitrary number of accounts? Or is the two accounts from the example?

(I will get to it this weekend)

inv-trad:
Well, in my case it is a joined account with my wife, and the standard cases should be individual or joined accounts of two individuals, but I guess, there could theoretically be more account holders than two, even though not very likely. "Erbengemeinschaften" maybe?! The logic used would work for an arbitrary number of account holders, thus tax lines.

buchen:
rebased to HEAD, merged. Thank you very much!

PR #1419, opened 2020-03-06
Add Exponential Moving Average (EMA) curves to charts

chapph:
Hi,

thanks for such a really great tool!

I found it interesting that you included the Simple Moving Average (SMA) but not the Exponential Moving Average (EMA). I prefer the latter (and it's easier to calculate) so I built it in.
Nothing fancy - the same as the existing implementation just duplicated.

I also saw your comment in #695 re: "Ich will kein technisches Chart Tool bauen", even so, I find it much better to use the information I have locally rather than to use external websites which change from day to day.

Cheers,

Chappy


buchen:
Rebased, reformatted, merged.

Thanks! I do not think this feature bloats the tiny charting component that PP has at the moment.

PR #1426, opened 2020-03-07
Introduce CSV and PDF importer call in context menu of AccountListView

cmaoling:
@buchen, the lambda call looked better to me. I was thinking about removing the doExecute method in addition, but left it for the moment.
FYI: I also removed the filterpath as you did in 8feeecbb505a92ff2d15cf79632973a7f2d69f37 even though I am not happy with this.


buchen:
Ist drin. Mit ein paar Anpassungen:

- dutch and spanish labels
- removed return type Status as not needed (the "dirty" event is triggered by the wizard)
- used SimpleAction + lambda for shorter code

Issue #1433, opened 2020-03-11
0.44.1: Small typo


ronnykrijt:
In the English version, under Reports -> Securities, hovering the field 'Purchase value of held shares' shows 'inteval' instead of 'interval', so it is missing an 'r'.

![image](https://user-images.githubusercontent.com/897441/76411107-64c10300-6391-11ea-89a1-92e85e1dd0e8.png)

PR #1434, opened 2020-03-11
Fix some spelling mistakes

chrisaut:
Inspired by #1433, but I noticed a few more.


buchen:
Thanks. Every detail matters...

PR #1463, opened 2020-03-26
Support latest quote using JSON feed

adrienkohlbecker:
This PR adds support for getting the latest quote of a security using a JSON feed.

The current JSON feed implementation only considers historical quotes, using `p.getDate().isBefore(LocalDate.now())`. This change takes the first price that has the current date and uses it as the latest quote.

I'm very new to java and this codebase, so let me know if there is anything I can improve!

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
@@ -62,6 +62,15 @@ public final class GenericJSONQuoteFeed implements QuoteFeed
     @Override
     public boolean updateLatestQuotes(Security security, List<Exception> errors)
     {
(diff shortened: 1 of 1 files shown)


buchen:
Thanks! Merged. I plan to publish a new version today or tomorrow

adrienkohlbecker:
Great, thanks! 

I was wondering about how to build an OSX .app bundle, I know I can launch it through eclipse but would like to have a standalone build from my fork. Is there a way to do that myself?

buchen:
> I was wondering about how to build an OSX .app bundle

If you build via Maven ```mvn clean verify -Ppackage-distro```, then an OS X App Bundle is created in the folder ```portfolio-product/target/products/..."

buchen:
Or you can point your existing installation to the update repository created in ```portfolio-product/target/repository```

adrienkohlbecker:
Thanks! Got it to work :+1:

PR #1471, opened 2020-04-02
Add support for more date formats in JSON quote download

adrienkohlbecker:
This PR adds support for two additional timestamps formats in the JSON quote downloader:
- Timestamps in milliseconds (eg. from Robeco's API here https://www.robeco.com/en/api/funds/prof-glob-en-11/LU0085136942/pricehistory.json?5dc68d) 
- Timestamps in days (eg. from Boursorama's API here https://www.boursorama.com/bourse/action/graph/ws/GetTicksEOD?symbol=AAPL&length=7300&period=0&guid=)

Since it is not possible to determine the format from a single integer without knowing more about the source, I used the arbitrary date of 2200-01-01 as boundary:
- any number greater than this date is assumed to be a unix timestamp in milliseconds
- any number lesser than the days between 1970 and 2200, is assumed to be a unix timestamp in days 

This means that:
- millisecond timestamps before 1970-03-26 00:08:38 can't be parsed with this method
- second timestamps before 1970-01-01 23:20:06 can't be parsed with this method

These seem like reasonable tradeoffs, what do you think?
Happy to make adjustments as you see fit!

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
@@ -196,7 +196,9 @@ public final class GenericJSONQuoteFeed implements QuoteFeed
                 if (object instanceof String)
                     price.setDate(YahooHelper.fromISODate((String) object));
                 else if (object instanceof Long)
-                    price.setDate(LocalDateTime.ofEpochSecond((Long) object, 0, ZoneOffset.UTC).toLocalDate());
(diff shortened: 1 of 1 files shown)


buchen:
Merged. I like the solution. The trade-offs will work.

adrienkohlbecker:
Hello @buchen! Do you have any plans for a release soon? I plan on writing a tutorial for a french investment community on how to setup Boursorama with the JSON provider

buchen:
@adrienkohlbecker - I plan to build and release an update today

adrienkohlbecker:
Perfect, thank you!

PR #1472, opened 2020-04-02
Add Boursorama quotes provider

adrienkohlbecker:
Boursorama is a major French news website on the stock market, as well as a bank. They have data on a lot of European mutual funds, that is often absent from Yahoo Finance. 

They expose a JSON endpoint able to gather historical quotes in a simple way.

This PR adds support for this endpoint directly in the UI, by extending `GenericJSONQuoteFeed`. A user then only has to add the proper symbol, and select Boursorama as a provider.

It requires https://github.com/buchen/portfolio/pull/1471 to work properly.

Let me know what you think of this approach!

### DIFF
--- a/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
+++ b/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
@@ -1,6 +1,7 @@
 name.abuchen.portfolio.online.impl.ManualQuoteFeed
 name.abuchen.portfolio.online.impl.AlphavantageQuoteFeed
 name.abuchen.portfolio.online.impl.BitfinexQuoteFeed
+name.abuchen.portfolio.online.impl.BoursoramaQuoteFeed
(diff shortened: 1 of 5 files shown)


buchen:
Hi @adrienkohlbecker,

your code itself looks good, but usually these sites have terms and conditions that forbid the automated download by computer programs. A couple weeks ago we had a similar case with the *Deutsche Börse* (German Exchange) where one contributor then actually wrote an email and got a decline. Therefore I cannot add specific quote feeds for such sites.

I went through the [fine print](https://www.boursorama.com/aide/avertissement-legal) using Google translate. I am not really sure if the translation is any good, but there are quite a lot of mentions that seem to forbid this. How do you read this?

> Il est notamment interdit d'utiliser tout moyen de récupération automatique (programme informatique) pour recueillir les données présentes sur BOURSORAMA.

adrienkohlbecker:
Ah, I was expecting this might happen, but I wasn't sure under which conditions the other providers were added. You're reading the fine print correctly!

That's ok, it was only a cosmetic/usability improvement, as you can achieve the same result manually with the JSON provider.

PR #1503, opened 2020-04-21
Mocktesting for YahooQuoteFeed and AlphavantageQuoteFeed

tzemp:
These QuoteFeeds are widely used but not really tested. I added Mockito (needed to add the net.bytebuddy and org.objenesis for some weird reasons..) and mocked the rpc call of those two methods.

This way we can at least test whether the parsing of the response is done correctly. In order to mock the  WebAccess call, i needed to factor our the WebAccess (see rcpLatestQuote()).


buchen:
squashed, rebased, merged. Danke

PR #1507, opened 2020-04-23
Fix wrong periodicity when multiple bookings

inv-trad:
When a security exists in multiple accounts and dividends are recorded separately, periodicity might be incorrect


inv-trad:
Problem was reported here: https://forum.portfolio-performance.info/t/periodizitaet-bei-dividendenzahlungen-in-mehreren-depots/8727

buchen:
squashed, rebased & merged.

Danke

PR #1521, opened 2020-05-03
Added menu item to convert transfer into deposit/removal

fvollmer:
Allow the conversion from a transfer action into a deposit and a removal transaction. Works similar to `ConvertBuySellToDeliveryAction` etc.


buchen:
Danke!

PR #1558, opened 2020-05-20
Fix NPE occuring in TradeDetailsView when input is null

SimonScholz:
Hi,

unfortunately a NPE occurs in the `TradeDetailsView `in case the input is null.
This PR fixes this, unless the input in the `TradeDetailsView `is supposed to be never null.


buchen:
Thanks!

PR #1582, opened 2020-05-31
ImportPDFHandler: sort input files by mod. time

Al2Klimov:
Rationale:
I import all docs at once per year w/o pre-creating the securities manually.
So new securities' names depend on the docs import order.
Comdirect seems to provide a security's most import-friendly name
(not messed up with spaces) in buy billings, not in the dividend ones.


buchen:
Merged. thanks 👍 

I moved the sorting to the #runImportWithFiles methods. The sorting also makes sense when the files are imported via drag & drop into the sidebar navigation

Issue #1593, opened 2020-06-12
Company Logo?


chrisaut:
How about we support adding an optional company logo to securities? It could make the interface a bit more beautiful, while also aiding in quickly finding what you are looking for in a list of securities.

How would it work? Well how about under "Additional Attributes" when creating/adding a security we add a field "Logo" (or something) where the user could paste a base64 encoded image as a string. Whenvever a list is displayed instead of the "green file icon" it would render that image.

Later we could make it easier and add an option to enter a url, which could download the image by itself and then internally b64 encode it, but that can be done as a seperate enhancement.

What do you think?

Quick mockup
![image](https://user-images.githubusercontent.com/8252355/84481772-dc574480-ac96-11ea-88c2-d7a8ba78f70b.png)

Step by step other views could then also take advantage of it where it makes sense, for example this view:
![image](https://user-images.githubusercontent.com/8252355/84482379-e9286800-ac97-11ea-9fa9-03e03ed4e722.png)

chrisaut:
Implemented in v0.47.0

PR #1594, opened 2020-06-12
Logo support

chrisaut:
Initial implementation for #1593

It adds a new AttributeType (Image)...currently this is a raw b64 encoded string. In future we should have a nice editor/etc. for it.

It special cases if Securities have such an attribute called LOGO and uses that image instead of the generic green icon.

Hopefully this can be considered, it doesn't yet work everywhere, but I thought I'd give it a shot.

image
image

image
image
image
image


chrisaut:
Now works pretty much everywhere I could find

image

chrisaut:
I also implemented a filepicker now so users can easily choose an image, and added the ability to display arbitrary many image attribute columns.

image
image
image
image

buchen:
Vielleicht kann man die Logos auch von https://clearbit.com bekommen.

buchen:
I really like the idea. This would also help to identify securities much more easily.

Some thoughts:

* Should we add a size restriction to the image? My fear is that the underlying storage format (XML) cannot cope very well with huge images and I do not want to invest time just to optimize the storage for images.
* I believe we need to cache the images created by the ImageConverter or otherwise the SWT resources are never freed. Maybe a ```LogoCache#getDefaultColumnImage(Object object)```  that a) returns the default column image if nothing else can be found, b) lookups up the logo if the Object implements Attributable and c) creates and caches the images. Then LabelProvider just can call this LogoCache class to resolve the image.
* Image resolution - see screenshot: Interestingly, the resolution of the column image next to the name is much worse than in the "logo" column. It is not really clear to my what the difference is. It looks like that the TableWidget can do a much better job in resizing the image to the target size (on my high definition display, I need 32 x 32 pixel images. On low definition displays, it is 16 x 16 - I just played around a little bit - it looks like SWT table widget auto-resizes always).
Bildschirmfoto 2020-06-19 um 08 29 44

* I don't think you need the class ```AttributableBase``` but could implement the #getImage method as default implementation in the interface. But then, I think the image creation probably would go into the LogoCache/ImageCache anyway.

* The editing dialog needs some more love - particularly when editing a security which has a logo attached: a) the width goes always to the maximum and b) the logo preview looks strange. I also think one should not be able to edit the base64 text - what do you think?

Bildschirmfoto 2020-06-19 um 09 41 07

Bildschirmfoto 2020-06-19 um 09 41 22

How to proceed? Do you have more time to go over it? Or do you want me to do some polishing?

chrisaut:
Hallo,

normalerweise hätte ich gerne auf dein Feedback reagiert, leider kann ich mich aber momentan aufgrund von plötzlichen familiären Gründen auf nix konzentrieren, obwohl ich mich gerne ablenken würde. Ich hab mir dein Feedback nur so kurz durchgelesen....einige der Dinge hab ich mir auch schon gedacht.

Auf jeden Fall wenn du möchtest dass ich weiter mache wird es wohl eine Weile dauern bis ich dazu in der Lage sein werde, ansonsten kannst du gerne in meine branch rein committen oder auch cherry picken. Ich hoffe das ist ok.

chrisaut:
I have refactored a bit to cleanup and introduced the ImageCache. Please take a look.

I'm stuck ATM as I'm not quite sure how to deal with the Image/Iconsize, SWT seems to just display them whatever size you give it, so for me when I use 32x32 the rows become way too tall, and also we need different sizes depending on where it is displayed.

Is there a way to to tell a (for example) TableLayout to be get its auto height without taking the image into account, or alternatively a way to get the expected height so I can manually resize the image?

image

chrisaut:
I have limited image size (48x48 for now) and have it automatically downsize on import if needed, reworked the editing dialog to not show the raw b64 string (both in securities editor/attribute tab page and in the celleditor for for other items (account, etc) along with some refactoring.

Resized image quality should be improved too.

There is an error building in CI (works fine locally) when I try to import "org.eclipse.swt.SWT" and use some of its defined constants:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.5.1:compile (default-compile) on project name.abuchen.portfolio: Compilation failure: Compilation failure: 
[ERROR] /home/runner/work/portfolio/portfolio/name.abuchen.portfolio/src/name/abuchen/portfolio/util/ImageUtil.java:[10] 
[ERROR] 	import org.eclipse.swt.SWT;
[ERROR] 	       ^^^^^^^^^^^^^^^^^^^
[...]
[ERROR] /home/runner/work/portfolio/portfolio/name.abuchen.portfolio/src/name/abuchen/portfolio/util/ImageUtil.java:[98] 
[ERROR] 	gc.setAntialias(SWT.ON);
[ERROR] 	                    ^^
[ERROR] Access restriction: The field 'SWT.ON' is not API (restriction on classpath entry '/home/runner/.m2/repository/p2/osgi/bundle/org.eclipse.swt.gtk.linux.x86_64/3.108.0.v20180904-1901/org.eclipse.swt.gtk.linux.x86_64-3.108.0.v20180904-1901.jar')
[...]
```

I don't know what it means, but since it's only 3 int constants I've just inlined them as a workaround and left a comment.

chrisaut:
OK I set the "logical" img/logo height back to 16x16 and added dpi awareness so that it should dynamically return higher res/larger images now depending on the display dpi. With that it looks alright on my 1080p screen again too, hopefully this also works for better displays with scaling > 100.

buchen:
Here are some observations (I am running macOS 10.15.5):

By default, the logo does not have a good resolution - much worse than above:

Bildschirmfoto 2020-07-05 um 09 15 17

What is interesting also: apparently the logo of the first line determines the row height. If the row with the logo is the first one (when initially rendering the table), then the rows are much smaller.

Bildschirmfoto 2020-07-05 um 08 48 02

Just for testing, I removed the code that is resizing the image:

If the logo is not the first row, it looks good. Much more crisp.

Bildschirmfoto 2020-07-05 um 08 45 39

If the logo is in the first row, the row height increases for all 😒 

Bildschirmfoto 2020-07-05 um 08 46 59


I Check DPI_CURRENT:
```
public static final int DPI_CURRENT = Display.getDefault().getDPI().x;
```
It has a value of 72 on macOS. And together with DPI_DEFAULT it resizes the images to 12x12

```
private static final float DPI_DEFAULT = 96.0f;
```

Where did you take the 96 from?

I also wonder if it makes sense to update to the latest SWT version and see if the HiDPI support is better. I have pushed a branch with some preliminary changes, but it will also means I cannot build 32bit versions anymore.

chrisaut:
Haha, yes I stumbled upon some stackoverflow post I think it was that also said the first image rendered basically determines the height for all the other images. It was from 2012, although I can't find it anymore now. I was able to reproduce this too.

Basically the way I understand it is we need to return different pixel sizes for the icons based on display scaling as swt will just dump render whatever size you give it. Above you said for you 32x32 works for you, while for me 32x32 results in each row being double the height (which doesn't even look all that bad, it's just that fits much less info on the screen and it's also inconsistent, eg. if the first item doesn't have a logo the default icon rendered will cause everything back to "normal" height). This leads me to assume that your device has 200% display scaling

The idea to get display DPI and adjust icon sizes idea I got from this
https://stackoverflow.com/questions/19900033/how-to-compute-swt-widget-size-according-to-windows-dpi
which also states that the default is 96 (which it is for me on win10 with a botch standard 1080p screen and scaling at 100%). The thinking was on your mac you would have a much higher dpi., and thus 16px would turn into 32px. This seems to be the wrong approach.

Another approach I played around with is to add a listener to the the MEASURE event of the table, this gives you a swt GC object which allows measuring text height:
https://stackoverflow.com/a/12861337/378666
However, this gets called so often and is completely detached to where we need this number. But the approach itself could work outside of this measure callback if we just 
```
Point size = event.gc.textExtent("foo")
```
assuming that for you size.height would result in something like 32 while for others it would result in 16 due to scaling. Problem is I don't know pretty much anything about swt (and I'm not good in Java in general).

Another piece of information I found is this stackoverflow answer:
https://stackoverflow.com/questions/47543568/how-do-you-utilize-swts-hi-dpi-support-for-icons
this talks about 
```
ImageDataProvider in order to provide image data for different zoom levels. if the application is moved to a monitor with different DPI or the zoom level is changed, the provider will be asked to return an image for the new zoom level (e.g. 150 or 200).
```

Perhaps this is the "correct" way to do this, but I'm thoroughly confused. on how this works, my eclipse doesn't even know about ImageDataProvider so perhaps this does need an swt update?

chrisaut:
I have implemented a custom ImageDataProvider now.

The idea is as follows:
 - when the user picks an image, it is size restricted to 48x48, this image is serialized and stored in the attribute data
 - when an image is requested this is deserialized, and given a logical size (16x16 currently everywhere) that can be dependent on where it's used
 - when the framework displays an image, it goes through this imagedataprovider that has a zoomlevel, the provider than scales dynamically based on physical size, logical size and zoomlevel on first use

This should give crisp image....or at least so I hope.

buchen:
> The thinking was on your mac you would have a much higher dpi., and thus 16px would turn into 32px. This seems to be the wrong approach.

Or, so I fear, SWT is giving you only the "legacy" 72 DPI even though the actual DPI of the display is 144. Which means on macOS we would have to live with 16x16 for now (instead of the 32x32).

Anyway, I have not forgotten about this change - only I did not have the time to dig deeper. I'll not get it this weekend, but I plan to create a release next weekend and include this change.

Two more thoughts:
* I am thinking of "moving" the logo property as a regular property of the java objects. For one, we would not have to pass the ClientSettings into all those methods. But more importantly, I think many users will want to use it and it is kind of strange to first have to create a "logo" attribute. I think the functionality will be loved very much. Do you think one additionally then also needs to options to create additional custom image attributes? (Like the flags)
* Not now, but I am thinking to "embed" a couple of custom logos: maybe the existing "sheet" in different colors, but also default logos for ishares, db x-strackers. Therefore I wonder if it makes sense to prefix the base64 encoding with "b64:". Then we can have something like "embedded:" or "library:" later on. What do you think?

chrisaut:
> > The thinking was on your mac you would have a much higher dpi., and thus 16px would turn into 32px. This seems to be the wrong approach.
> 
> Or, so I fear, SWT is giving you only the "legacy" 72 DPI even though the actual DPI of the display is 144. Which means on macOS we would have to live with 16x16 for now (instead of the 32x32).

Well, I've already changed how this all works. Please give it another try, I believe the ImageDataProvider is the "expected" way to do this in SWT.

> Anyway, I have not forgotten about this change - only I did not have the time to dig deeper. I'll not get it this weekend, but I plan to create a release next weekend and include this change.

No worries.

> Two more thoughts:
> 
> * I am thinking of "moving" the logo property as a regular property of the java objects. For one, we would not have to pass the ClientSettings into all those methods. But more importantly, I think many users will want to use it and it is kind of strange to first have to create a "logo" attribute. I think the functionality will be loved very much. Do you think one additionally then also needs to options to create additional custom image attributes? (Like the flags)

I see. Some thoughts of my own:
 - I think it would be nice, although not crucial, to be able to keep additional attributes possible as well. With the current approach we get this for free. While having to inject the settings is a bit annoying (I actually had a hard time understanding how attributes work from an object model at first) it's not the end of the world IMO.
 - I think I implemented the code to automatically create this logo attribute as a default attribute for new files. That leaves existing files, perhaps we could instead add code to automatically create this attribute also for these files? I'm just not sure where to plug in there.
- Just architecturally, the only thing that irks me a bit is the special casing of the name "logo", this would obviously go away having them just be object properties. But on the other hand this string appears in exactly one place, so it's fairly isolated and it allows additional image columns/attributes completely free.
 - For what I intent do use the additional columns: Mostly for the different exchanges some of my securities are traded at (ie. NYSE/XETRA), and country flags for the stocks.

> * Not now, but I am thinking to "embed" a couple of custom logos: maybe the existing "sheet" in different colors, but also default logos for ishares, db x-strackers. Therefore I wonder if it makes sense to prefix the base64 encoding with "b64:". Then we can have something like "embedded:" or "library:" later on. What do you think?

Regarding the prefix, yeah it should probably be prefixed you are right. But if we prefix I suggest to go the full route and prefix it like this: "data:image/png;base64,<DATA>"
I'm not sure the legal ramifications of including copywritten logos in the distributed application, but if possible, sure.

buchen:
Hi @chrisaut - I have rebased, squashed, and merged your change.

There are a couple changes I did in a follow-up commit. Most notably, I changed the resizing method to use ```ImageData#scaledTo(w,h)```. The most important reason was that this was the only way I was able to keep the transparency of the embedded images. The second change I did was to add the "data:image/png;base64," prefix.

I plan to create a release today. I will add some "library" images in a later release as we plan to go swimming now.

buchen:
@chrisaut I realize the ```ImageData#scaledTo(w,h)``` does not work well for actual logos. I tested it with some icons and the result looked good + the transformation kept the transparency. But if I apply it to actual logos - bigger size, more complicated drawing - the result looks crappy.

Bildschirmfoto 2020-07-20 um 07 49 34

If I revert back to your code, the image looks better. But as you can see on the Adidas logo, the background is not transparent anymore.

Bildschirmfoto 2020-07-20 um 07 51 43 and Bildschirmfoto 2020-07-20 um 07 53 21

Maybe it is a macOS problem... Does the transparent work for you?
The statement ```gc.setAlpha(image.alpha);``` does not seem to make a difference

chrisaut:
No, transparency didn't work, I noticed it too and found a few ways that are supposed to make it work, like this:
https://stackoverflow.com/questions/8270474/drawing-on-a-transparent-image-using-java-swt/15685473#15685473

I've sort of decided against bothering, the whole swt image support seems very lacking overall,, I shouldn't have to write my own image resizing method in the first place. But looping over an alpha array and dealing with different image formats etc I thought was a bit risky and I sort of hoped you would have more experience with swt.

buchen:
I found a way to keep the transparency while keeping a higher quality image.

Bildschirmfoto 2020-07-23 um 20 56 39

chrisaut:
Awesome, what was the key thing to change?

chrisaut:
```
imageData.transparentPixel = imageData.getPixel(0, 0);
```
Ich glaube nicht dass das richtig ist, ausser wenn das (Original-)Bild den Pixel links oben transparent hat, was ja nicht unbedingt der Fall sein muss, oder?

buchen:
> Ich glaube nicht dass das richtig ist, ausser wenn das (Original-)Bild den Pixel links oben transparent hat, was ja nicht unbedingt der Fall sein muss, oder?

Zu dem Zeitpunkt ist das noch ein "leeres" Bild, d.h. alle Pixel sind gleich. Im ersten Schritt erstelle ich einen transparenten Hintergrund. Erst im zweiten Schritt wird das Logo gezeichnet.

tut das denn unter Windows?

chrisaut:
Achso, dann hab ich das falsch gelesen.

```
tut das denn unter Windows?
```
Ich hab es nicht probiert, war nur neugierig wie du es gelöst hast.
(Weiss nicht weiso ich auf einmal auf Deutsch gewechselt habe, war schon spät...)

Issue #1514, opened 2020-04-25
VIAC PDF-Import - "Rückerstattung Quellensteuer"


C1618A:
Hello
Thank you very much for the fantastic software. I'm using the PDF-Import for VIAC and when it comes to dividends some PDFs work, others not.
I leave here one PDF that work and one that doesn't.
Working: 
https://www.amazon.com/clouddrive/share/sU3XvlBY4UFj18QMPmDs1d47OqnOLLDEPSJf0stMqzK
Not Working:
https://www.amazon.com/clouddrive/share/d4dwdnTrErUciVW3EcZlPjiZ7AWm2FIW7K0iDfl3B8
The difference is that one says "Rückerstattung Quellensteuer"

Mauin:
As a follow-up issue to this: The "Rückerstattung Quellensteuer" appears to always be in CHF, whereas the underlying security might have a different base currency. But a transaction of type TAX_REFUND seems to require a currency conversion in such a case.

@buchen what would be the best way to handle such a situation?

PR #1628, opened 2020-06-28
Showcase issue of VIAC dividend imports

Mauin:
Add a (currently failing) test for VIAC dividend imports.

VIAC dividends of type "Rückerstattung Quellensteuer" seem to only ever be in CHF. However the underlying securities of the tax refund might use different currencies.

In this case the import will fail due to the missing exchange rate, which is not provided in the PDF.

What would be the best way to handle such a situation?
- Pull exchange rate of the given date and calculate the forex value?
- Only add the tax refund to the account, with just a note of which security it was for?
- Alternatives?

Happy to update the PDF extractor in this PR with the preferred solution.

Related #1624 #1514


buchen:
Both options sound reasonable. The first one is probably more "correct", the latter one much easier to implement.

> Pull exchange rate of the given date and calculate the forex value?

It is tricky to get to the CurrencyConverter. You need to get a ExchangeRateProviderFactory which "lives" in the ```ClientInput```. I guess it makes sense to extend the constructor of the ```PDFImportAssistent``` and pass the factory.

> Only add the tax refund to the account, with just a note of which security it was for?

That is easier to implement: if there is a currency mismatch, do not set the security.

Mauin:
Checked a few of the securities and there were no recent dividend payments for these securities. They funds are accumulating. So I'm not sure anymore whether calculating the forex value would be more correct.

Nicest option would probably be linking the tax refund to the security, but just in the currency of the account, without any currency conversion. Just like saying "for this security (even though it's in USD) you received a tax refund of CHF 1.23.

For now, I think the best way is removing the security from the transaction and setting it as a note, I'll make a quick change for that.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AbstractPDFExtractor.java
```diff
@@ -105,8 +106,15 @@ public String getPDFAuthor()
...
-                item.getSubject().setNote(filename);
-
+            {   
+                // Don't overwrite note set by the PDF extractor
+                Annotated subject = item.getSubject();
+                String note = subject.getNote();
+                if (note == null || note.isEmpty())
+                {
+                    subject.setNote(filename);
+                }
```
Mauin, line 116: Open for suggestions

inv-trad:
As @buchen Andreas didn't merge the other pull request https://github.com/buchen/portfolio/pull/1676 last weekend, I realized that it most likely was, because this one exists as well. 

Using most parts of what was proposed here, the other PR now includes that solution as well. 

If transaction currency and security currency are idential, the security is attached, if not, the security name is put as a note and the document name appended later.

Hope, this helps to get the fix included.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AbstractPDFExtractor.java
inv-trad: @Mauin - usually the note should be null after extraction, but in cases like yours, the filename should be appended to be consistent and the note to show, which document the import came from.

see PR https://github.com/buchen/portfolio/pull/1676

buchen:
Covered by #1676

PR #1631, opened 2020-07-01
Enhance Targobank PDF import with dividend import

inv-trad:
Dividend and accompanying tax documents are merged during import.

For sell transactions now the tax document can be imported in combination with the transaction document and is merged.

https://forum.portfolio-performance.info/t/import-fuer-targobank-pdfs/5537/8


buchen:
squashed, merged. Danke!

PR #1634, opened 2020-07-03
Refactor FixedExchangeRateProviders

chrisaut:
Refactor to get rid of all the code duplication for ExchangeRateProviders with a fixed rate.
Adding these is now a one liner, and should make it easier to user define these in the future too (eg. for all the old EUR currencies).

It also provides them with a "start date" so that the Exchange Rate graph under currencies can draw a line.


buchen:
Good idea to provide at least two dates to have something visible in the chart!

PR #1637, opened 2020-07-03
Fix DKB PDF Import for Sell Transactions

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-wertpapierabrechnung-dkb/4449/20


buchen:
squashed, rebased to master, merged

PR #1650, opened 2020-07-13
Update ClientSettings.java: Add Bookmark

Daresch:
Added bookmark for shares (not for ETF, funds and so on) at comdirect.de


buchen:
Thanks for adding the bookmark. It will show up under "templates" and can be added by users to the primary bookmarks. New files will include it there by default.

PR #1672, opened 2020-07-27
Enhancement to show tooltips for diagram buttons

inv-trad:
Issue: https://forum.portfolio-performance.info/t/mouseover-funktion-fuer-diagrammbuttons/10297


inv-trad:
@buchen Andreas, one quick question. As this change touches messages file, I guess you would prefer to have the changes in all files. How should I deal with the languages I cannot support. Should I simply add the corresponding entries in English for some else to later translate them?

buchen:
> As this change touches messages file, I guess you would prefer to have the changes in all files. How should I deal with the languages I cannot support. Should I simply add the corresponding entries in English for some else to later translate them?

I also do not speak Dutch, Spanish or Portuguese. All those translations have been contributed. Some translators are still active, some not. I am still torn what is better: a (potentially) badly translated label or a handful of English labels thrown in. At the moment, I think the former is better.

For new labels, I typically use Google Translate to translate the English label into the other languages. After creating a new release, I upload the labels to [POEditor](https://poeditor.com/join/project/4lYKLpEWOY). Before creating the next release, I download the changes and merge them into the code base.

I added the missing languages.

PR #1674, opened 2020-07-28
Enhancement match investment plan transaction when importing pdfs

inv-trad:
This enhancement figures out, whether an imported transaction relates to
an existing transaction that has been generated via a savings plan (see
DetectDuplicatesAction).

Issue: https://forum.portfolio-performance.info/t/importierte-buchungen-als-sparplan-kennzeichnen/1993/6


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java
```diff
@@ -344,7 +355,14 @@ public Status apply(ImportAction action, Context context)
...
                 portfolio = context.getPortfolio();
 
-            return action.process(entry, account, portfolio);
+            Status status = action.process(entry, account, portfolio);
+
+            // check if message was set in DetectDuplicatesAction
+            if (status.getMessage() != null && status.getMessage().equals(ImportAction.IMPORT_SAVINGS_PLAN_ITEM))  
+            { 
+                this.savingsPlanItem = true;
+            }
```
buchen, line 364: The ```Extractor.Item``` has already a property ```data```. I wonder if it makes sense to change ```data``` to something like ```Map<String, Object>```  and attach the detected investment plan transaction to the item itself. Then we do not have to re-search for the transaction later.

What do you think of passing the ```item``` also to the ```ImportAction#process``` methods? The ```DetectDuplicateAction``` can attach it, and the ```InsertAction``` can replace the transaction.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/DetectDuplicatesAction.java
```diff
@@ -112,7 +161,37 @@ private boolean isPotentialDuplicate(Transaction subject, Transaction other)
...
+            return false;
+        
+        // number of shares might differ due to differing prices per share used by savings plan
+        // accept a difference of 10%
+        long shares = subject.getShares();
+        if (shares * 1.1 < other.getShares() || shares * 0.9 > other.getShares())
+            return false;
+        
+        // use the behavior that savings plans get a specific note and check for it
+        if (other.getNote() == null || !other.getNote().startsWith(Messages.InvestmentPlanAutoNoteLabel.substring(0, 8)))
```
buchen, line 191: If we change the ```ImportAction#process``` methods, then let's consider adding the ```Client``` as well and checking whether the transaction is part of an InvestmentPlan (instead of comparing the comment).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
```diff
@@ -87,6 +93,17 @@ public Status process(BuySellEntry entry, Account account, Portfolio portfolio)
...
 
+        // when importing transactions that have already been generated by a savings plan
+        // find the existing item and delete it - it will be created again below
+        if (savingsPlanItem)
+        {
+            DetectDuplicatesAction action = new DetectDuplicatesAction();
+            AccountTransaction existingTransaction = (AccountTransaction) action
+                            .findSavingsPlanTransaction(entry.getAccountTransaction(), account.getTransactions());
+
+            account.deleteTransaction(existingTransaction, client);
```
buchen, line 104: I propose not only to delete the transaction, but add the newly created transaction to the investment plan as well.

inv-trad: The investment plan doesn't offer a method to create a transaction from outside. Maybe it is best to find and adjust the existing transaction by copying over differing information like shares, date, note, etc.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -335,9 +339,33 @@ public Image getImage(ExtractedEntry element)
...
+                    {
+                        image = Images.WARNING;
+                    }
+                }
+
                 return image != null ? image.image() : null;
             }
 
+            @Override
+            public String getToolTipText(Object entry)
```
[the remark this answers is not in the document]
inv-trad: My challange here is, that I would prefer the the entry in the list not to show as strike-through. Means either using the status OK (doesn't usually show a message) or setting the isImport flag. Will have a look, how to get this resolved.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/DetectDuplicatesAction.java
inv-trad: Thanks for the feedback. I didn't want to change the methods and was looking for other options to pass the values around. I will try to figure out, what is the best option to get this done, ideally without searching the investment plan twice...

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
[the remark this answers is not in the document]
inv-trad: Changed the way `isImported` is provided in `ExtractedEntry`. This way a warning image is displayed and the line not strike-through and all warning or error messages are now displayed as tooltip.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java
inv-trad: I extended `data` the way proposed, but didn't use it. To pass the `item` from and to `process` methods would require too many changes for this special case. The attached message is now used in a bit different way to check for an item to replace an investment plan item.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/DetectDuplicatesAction.java
inv-trad: The `client` object is passed to the action during initinalization, like it is done for `InsertAction` already. Then only the investment plan transactions for the specific security are searched. Should make it more reliable and faster, too.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java
inv-trad: Just reverted the `data` object, because the JSON PDF test cases went wrong...

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
inv-trad: Added a `addTransaction()` method to `InvestmentPlan` and solved it this way.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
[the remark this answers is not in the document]
buchen: Ok, now I get the issue. I think it is okay the way it is solved now. If we have - in the future a similar need - we could add a new ```Status.Code``` besides OK, WARNING, and ERROR that has encodes the behavior "this import updates an existing transaction".

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
```diff
@@ -87,6 +98,38 @@ public Status process(BuySellEntry entry, Account account, Portfolio portfolio)
...
+            Transaction existingTransaction = null;
+            PortfolioTransaction t = entry.getPortfolioTransaction();
+
+            // search for a match in existing investment plan transactions
+            List<InvestmentPlan> plans = client.getPlans();
+            Iterator<InvestmentPlan> i = plans.stream().filter(p -> p.getSecurity().equals(t.getSecurity())).iterator();
+            while (i.hasNext())
+            {
+                List<Transaction> transactions = i.next().getTransactions();
+                existingTransaction = action.findInvestmentPlanTransaction(t, transactions);
```
buchen, line 115: Shouldn't the *first* transaction break the loop?
Let's say I have two investment plans on the same security. And the duplicate is found in the first, then the second would overwrite it with null again.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
```diff
@@ -87,6 +98,38 @@ public Status process(BuySellEntry entry, Account account, Portfolio portfolio)
...
+                existingTransaction.setDateTime(t.getDateTime());
+                existingTransaction.setNote(t.getNote());
+                existingTransaction.setShares(t.getShares());
+                existingTransaction.clearUnits();
+                t.getUnits().forEach(existingTransaction::addUnit);
+                
+                if (existingTransaction.getCrossEntry() != null)
+                    existingTransaction.getCrossEntry().getCrossTransaction(existingTransaction).setNote(t.getNote());
+            }
+            return Status.OK_STATUS;
```
buchen, line 130: Maybe the OK_STATUS should only be returned *if* a transaction if updated (i.e. inside the if statement).

If the duplicate is not found (for whatever reasons - parallel updates), then a new transaction is created (in the later code block).

What do you think?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
inv-trad: Good point. New version breaks and returns after `existingTransaction != null`.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
inv-trad: Well, that should not happen, but you never know. I adjusted it accordingly.

buchen:
okay, rebased, squashed and merged into master. 😄

PR #1676, opened 2020-07-29
Enhanced VIAC PDF import to support English documents

inv-trad:
Issue: https://github.com/buchen/portfolio/issues/1664

Including the relevant parts of Pull request: https://github.com/buchen/portfolio/pull/1628

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ViacPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ViacPDFExtractor.java
@@ -21,14 +21,17 @@ public class ViacPDFExtractor extends SwissBasedPDFExtractor
         addBankIdentifier("Terzo"); //$NON-NLS-1$
 
         addDepositTransaction();
+        addDepositTransactionEnglish();
         addBuyTransaction();
         addBuyTransactionEnglish();
         addSellTransaction();
+        addSellTransactionEnglish();
         addInterestTransaction();
         addInterestTransactionEnglish();
         addFeeTransaction();
         addFeeTransactionEnglish();
         addDividendsTransaction();
+        addDividendsTransactionEnglish();
     }
 
     @SuppressWarnings("nls")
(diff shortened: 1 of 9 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ViacPDFExtractor.java
```diff
@@ -21,11 +21,18 @@ public ViacPDFExtractor(Client client)
         addBankIdentifier("Terzo"); //$NON-NLS-1$
 
         addDepositTransaction();
+        addDepositTransactionEnglish();
```
Mauin, line 24: Does VIAC also support PDFs in other languages (e.g. French, Italian) as well? To not have to duplicate the each transaction for each language - is there an easier way to add localication to the PDF imports?

inv-trad: It might be that VIAC supports French and Italian as well. Haven't heard of it, yet. 

I am aware of an easier way to add localization to regex. First I thought of adjusting the regex like this:

`.match("(Aussch.ttung: |Dividend payment: )(?<currency>\\w{3}+) .*")`

That could be done for multiple languages, still makes it harder to read, and I chose the other option. 

I didn't came across such kind of request earlier and don't know, how many banks offer a set of different languages to their clients?!

lucacame: VIAC also supports French and Italian, I can confirm.
I think generally most Swiss banks offer documents in 3 languages (IT, FR, DE) and some even in English, so a straightforward solution for localisation might be helpful :)

buchen: I also do not have a good idea how to localize the regex expressions.

The only other case I know is Degiro extractor which supports Dutch and German. However, that was a trivial fix because the only relevant delta was the heading - see commit c5fb722cc1c8a9de2085866d3a0f93ca03170d28


We could build in some "replacement logic" and have the PDF Parser try all patterns.
Do you think this would improve readability?
```
.match("{TERM1} (?<currency>\\w{3}+) .*", new String[] {"Aussch.ttung", "Dividend payment"})
```

inv-trad: > Do you think this would improve readability?

Not really. As there are so little cases where this is required, I would assume having separate methods per language is fine.  And, structure might be different as well (not the case here).

PR #1678, opened 2020-07-29
Change format of performance percent numbers in Statement of Assets

inv-trad:
Issue: https://forum.portfolio-performance.info/t/kursaenderung-zum-vortag-in-vermoegensuebersicht/10323/9


chrisaut:
Do I understand correctly that this is percentage change, eg. -1.2%?

Could we also add a column for "Value change", eg. -45.23 EUR? I often find myself wondering how much money I lost over the last 24 hours from each security ;-)

chrisaut:
Actually, is this needed, if you go here:
image
Performance -> Kursgewinn (FIFO, aktueller Bestand) -> 1 Handelstag

That's the same, right?

PR #1680, opened 2020-07-30
Fix Interactive Brokers importing incorrect amount

mpeterson:
In exported XML's from FlexQueries the `cost` attribute is incorrect for `sell` operations.
There is however another attribute that has the correct value for both `buy` and `sell` operations.

This patch modifies the source of truth for amount using `netCash` instead.


buchen:
Thanks. Sorry that it took some time to incorporate your change - summer vacation...

PR #1683, opened 2020-08-02
Copy Account attributes to ReadOnlyAccount on instantiation

chrisaut:
This fixes an issue with Logos not showing up for accounts in Statement of Assets view when a filter is applied

@buchen I'm not sure what the purpose of the ReadOnlyAccount wrapper class is, but I don't think it hurts to copy the attributes, does it?

https://forum.portfolio-performance.info/t/logos-fuer-konten-einstellen/10318/12


buchen:
> I'm not sure what the purpose of the ReadOnlyAccount wrapper class is, but I don't think it hurts to copy the attributes, does it?

If the user applies a filter, PP is building a new "client" with accounts and securities as needed and then runs it through the same calculation code. In order to avoid that any code accidentally allows the user to add transactions to this copy, I have created a read only wrapper that is throwing exceptions.

PR #1685, opened 2020-08-12
bondora go and grow pdf extractor

jdstamp:
The P2P platform Bondora offers the managed portfolio "Go and Grow". The interests are deposited daily. Therefore I would like to include the PDF import as manual maintenance of the account is tedious.

I still have to 
- [ ] include a withdrawal transaction 
- [ ] clean up the transaction type with withdrawal
- [x] integrate the new extractor


buchen:
I have merged this extractor (only now realizing that you marked the PR as "draft"). From a coding perspective, it looks good. And it is okay to start by detecting some transactions and add others (such as withdrawal later).

PR #1689, opened 2020-08-15
Fix small typo in German localization

udopt:
Fix "Kategoire -> Kategorie" in Asset Allocation tooltip


buchen:
Thanks for the contribution. Every detail matters. 😄

PR #1691, opened 2020-08-19
CSV import / filter negative security price

Ragas13:
Avoid negative security prices according to https://forum.portfolio-performance.info/t/import-pflege-negativer-kurse-indikatoren/6176


buchen:
Hi @Ragas13 - why did you close the PR? PP cannot handle negative values and should not import them...

Ragas13:
Sorry, I've closed the wrong PR. One failed ond one works as intended. I should not create PR's via Smartphone

buchen:
Merged :-)

PR #1695, opened 2020-08-22
Update industry-gics.properties

Alain-CYS:
Updated per https://www.msci.com/documents/1296102/11185224/GICS+Methodology+2020.pdf.


buchen:
Thanks for updating the structure and English translation.

I found the German translation on the MSCI website (https://www.msci.com/gics) and updated the German labels.

There are more translations in case folks care: French, Portuguese, Spanish (to only take the languages that PP supports).

PR #1698, opened 2020-08-26
ImportAssistent Embedded Error Handling

Ragas13:
Show import errors (e.g. Exchange rate is missing) which is only visible via right mouse button:
AdditionalErrorMessage

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
@@ -107,6 +107,8 @@ public class ReviewExtractedItemsPage extends AbstractWizardPage implements Impo
     private Portfolio portfolio;
 
     private List<ExtractedEntry> allEntries = new ArrayList<>();
+    
+    private List<Exception> errors = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -706,9 +708,15 @@ private void checkEntries(List<ExtractedEntry> entries)
         for (ExtractedEntry entry : entries)
         {
             entry.clearStatus();
-            for (ImportAction action : actions)
-                entry.addStatus(entry.getItem().apply(action, this));
+            for (ImportAction action : actions) {
+                ImportAction.Status actionStatus = entry.getItem().apply(action, this);
+                entry.addStatus(actionStatus);
+                if (actionStatus.getCode() == ImportAction.Status.Code.ERROR)
+                    errors.add(new IncorrectFileException(entry.getItem().getSubject().getNote() + ": " + actionStatus.getMessage())); //$NON-NLS-1$
```
buchen, line 715: The "import actions" (which are checking the transactions) run not only after the initial import but also when the user switches the accounts. The reason is that if you switch to a USD account, the transaction now might be importable - or still not importable. Either way, the list of error messages should not be calculated dynamically.

I haved changed it now this way: save the extraction errors separately, merge the lists with the errors from checking the extracted items accounts the accounts.

buchen:
Rebased, merged.

PR #1703, opened 2020-08-27
Add Weberbank PDF Import

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-der-weberbank/10542/6


buchen:
Thanks.

The build is failing b/c the Sonar upload _only_ works if the PR is done out of my repository (but that was all I was testing...)

PR #1728, opened 2020-09-03
Fix NUP when updating investment plan transaction

inv-trad:
Issue: https://forum.portfolio-performance.info/t/java-lang-nullpointerexception-beim-pdf-import/10724/23


buchen:
I do not think that the iterator can be null (it might be empty, but not null).
I have therefore only merged the first commit.

Issue #1466, opened 2020-03-29
Issue in security performance when selling short


PhKP:
Hi

Thanks again for working on this awesome program!

I stumbled on a weird issue with a security I have set up. Please see this screenshot:
<img width="1440" alt="Skærmbillede 2020-03-29 kl  19 34 19" src="https://user-images.githubusercontent.com/11960377/77856116-3cd8f880-71f5-11ea-8234-10a8d94c2d3f.png">

The security is a CFD's that I have traded 6 times (3 positions). Two times where I entered into a long position and one time I entered into a short position.

I am not sure if short positions are unsupported in some calculations in the tool? I was actually able to sell the security even though I had 0 shares, so I figured it was okay, but then I saw "purchase value (MA)", "purchase price (MA)" and "capital gains (MA, current ...)" in the Security Performance page. Those values do not seem correct and therefore I am reporting this issue.

buchen:
PP does not support short positions (but also does not prevent you from entering one).

Do you have a good source/link explaining how to calculate a valuation, fifo purchase values, etc?

PhKP:
Well, I am not completely sure how to calculate all of these performance values, but the general concept is described here: https://www.investopedia.com/ask/answers/05/maxreturnshortsale.asp

I think it would be great to add the feature of short selling but maybe with some "N/A" in the performance calculations. The calculations can always be added in a later release.

 After the upgrade to PP Version 0.46.1 I have the following problem since the transaction is already in my xml file:

![image](https://user-images.githubusercontent.com/11960377/79258432-664b8280-7e8b-11ea-9632-1d341d5d2964.png)

pfalcon:
Initial issue with short positions is that PP is unable to calculate "Purchase Value" for them, showing as 0. The FIFO-based calculation happens in https://github.com/buchen/portfolio/blob/cbf2b5aa0264f3385ac40f8c984769e8ee5248a7/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CostCalculation.java#L90

The problem is that PP thinks that a trade starts with a "buy" transaction. Contrary, a trade starts with the *first* transaction. The algo would be something like:

1. If fifo is empty, add xact to it.
2. Otherwise, check if head of fifo is the same type of xact (buy/sell) and if so, add to it. (invariant: all xacts in fifo are always of the same type).
[...]

PR #1736, opened 2020-09-07
New Feature: support selling short

jamespan:
Selling short and covering short support for PP.

Screen Shot 2020-09-07 at 9 41 11 AM

image

image

image

image

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -940,6 +940,7 @@ public class Messages extends NLS
     public static String SecurityMenuAddPrice;
     public static String SecurityMenuBuy;
     public static String SecurityMenuConfigureOnlineUpdate;
+    public static String SecurityMenuCover;
     public static String SecurityMenuCreateManually;
     public static String SecurityMenuCreateQuotesFromTransactions;
     public static String SecurityMenuDebugGetHistoricalQuotes;
@@ -960,6 +961,7 @@ public class Messages extends NLS
     public static String SecurityMenuRemoveFromWatchlist;
     public static String SecurityMenuSearch4Securities;
     public static String SecurityMenuSell;
+    public static String SecurityMenuShort;
     public static String SecurityMenuStockSplit;
     public static String SecurityMenuAddEvent;
     public static String SecurityMenuTransfer;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/palette/TransactionElements.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/palette/TransactionElements.java
@@ -120,6 +120,8 @@ import name.abuchen.portfolio.ui.selection.SelectionService;
         for (final PortfolioTransaction.Type type : EnumSet.of( //
                         PortfolioTransaction.Type.BUY, //
                         PortfolioTransaction.Type.SELL, //
+                        PortfolioTransaction.Type.COVER, //
+                        PortfolioTransaction.Type.SHORT, //
                         PortfolioTransaction.Type.DELIVERY_INBOUND, //
                         PortfolioTransaction.Type.DELIVERY_OUTBOUND))
         {
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
@@ -604,9 +604,11 @@ public abstract class AbstractSecurityTransactionModel extends AbstractModel
         switch (type)
         {
             case BUY:
+            case COVER:
             case DELIVERY_INBOUND:
                 return Math.max(0, total - feesAndTaxes);
             case SELL:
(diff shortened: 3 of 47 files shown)


jamespan:
#1466

nmaeding:
@jamespan Does this pull request cover that the holding pie now longer reports an error, that there are negative values?
(fourth screenshot, red note in the upper right corner...)

bbatsov:
Do you really need a "Short" action, though? It seems to me it's more natural to implement this in terms of "Sell" and maybe add some checkbox that you're shorting there.  After all, that's also how shorting works in brokers.

buchen:
I like the change. I appreciate contributing this change. The support for selling short has been asked for quite some time. I glanced over the code and I think it is a good idea to add new types. Due to the size of the change, I did not find the time yet to look at the code in detail. I will plan to review and merge the code for the next week(s) - as my job permits me to spend time on PP.

chirlu:
I agree with @bbatsov – this change needlessly complicates things by duplicating functionality, while not creating any advantage. It’s based on flawed or lacking analysis of the problem at hand (“enable short selling”).

It already works perfectly fine in PP to sell securities that you don’t own, and later buy them (back) to get to zero. The backend (model) can handle that, and the user interface, too. It’s just some of the reports and widgets and diagrams that get confused when they see negative amounts. Adding transaction types won’t fix them.

Duplicate transaction types for selling and buying not only make the code more complex, they also create user-visible issues. Assume, e.g., that someone short-sells 5 shares of some company, then some time later buys 20 shares. How would you represent that when there is an artificial distinction between “buy” and “buy to cover”? Does the actual buy of 20 need to be split into a 5-shares cover buy and a 15-shares regular buy?

Or let’s assume you short-sell 3 shares in one account, then buy 3 shares in another. Would that buy be a cover? How is that to be interpreted when viewing just that second account? You could then transfer the shares from the second account to the first, bringing each account to zero. For that, wouldn’t we need another split transaction type, “inbound transfer to cover”?

jamespan:
Direction of buy and sell matters.

chirlu:
Would you mind to elaborate?

MichaelSp:
I think the idea is
 * 5 long + 10 short = 5 short?
 * 5 long + 5 short = 5 long + 10 short

Without the direction you will always end up in one case.

jamespan:
My broker is  [futubull](https://www.futunn.com/?lang=en-us). 

If I am holding a short position, I have to cover it before I can open a new long position.

I don't know whether other brokers have the same limit about the long/short position or not.

Also, PP dose not support opening a position with a SELL, so I decide to add SHORT/COVER operations to PP.

SHORT is a selling that open a position, and COVER closes a SHORT position, just like BUY opens a long position and SELL close it.

Base on the SHORT/COVER, I fix some calculation issues comes with fees and taxes.

MichaelSp:
I think there are 3 different aspects to cover: UI, Storage/Export-Formats and Performance calculation

Just checked with my broker (Interactive Brokers, with Lynx as reseller).
1. In the UI you see a negative quantity if you have a short position.
2. In the export it is marked if a trade (sell/buy) opens a new position or closes an exist.
3. For the performance: I don't know. That would be interesting, but I guess it's just treated as long.

Not sure if that works everywhere but treating short as negative quantity seems to be reasonable.

aeneasr:
Sorry for digging this one up, I think there are even more complications with these types of transactions. Usually, you have two ways of put / call options. As already stated you can buy a put, sell a put, buy an option, sell an option - and all of those you can do to either open or close a position.

This also plays into risk calculations. So let's say you sell a put on AMD for 15$. Price falls to 10$ and you have to buy 1000 stock at 10$. I would argue that the 10.000$, which you *should* have in your bank account do not fall under "risk free" as you need the money to cover the put in case things go south.

Right now, puts and calls are not reported correctly and cause the consistency check to fail. They also do not report in your performance  correctly due to the negative prefix and because the system does not understand the concept of an option! I think a couple of things we need to keep in mind. For selling PUTs:

1. You sell a PUT with strike price 15$ end expiry EOY 2022. Half a year later, you buy a PUT with the same strike price and expiry -> position is closed.
2. Options can expire (they almost never do though) -> you make a profit if you sold the PUT (you get the full premium). Here, PP would just keep the option open forever unless you mark the security as bought.
3. The PUT realizes and you now own the stock, with the strike price. Here, you would need to create a new buy for the security, and a buy for the stock.
4. You roll over to a different strike price or a later period. This typically involves sell to close and buy to open on the same underlying stock.

Things get even more complicated with calls of course, because here it is easier to sell unbacked calls, have calls backed by some or all stock, and so on.

And then we still have the inverse, so buying PUTs and CALLs. Although, I think almost no private investor really does this except for rolling positions or closing positions.

Inbetween however, the reports can not properly reflect this. I think one issue too is that options are typically with a 10x or 100x multiplicator. So selling one put means you have to buy 100 stocks.

So in conclusion, I think options are a different type of security with different characteristics and implications. Wouldn't that warrant a new datatype in PP that can reflect this, and also interact with regular security types?

jamespan:
I have built my own trade tracker based on [Obsidian.md](https://obsidian.md/) and [Dataview](https://github.com/blacksmithgu/obsidian-dataview) plugin, this PR is no longer supported

pfalcon:
@jamespan :

> I have built my own trade tracker based on [Obsidian.md](https://obsidian.md/) and [Dataview](https://github.com/blacksmithgu/obsidian-dataview) plugin, this PR is no longer supported

Very interesting, do you have more info about your solution, a blog post perhaps?

PR #1741, opened 2020-09-09
Fix bondora pdf import to support different format

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-von-bondora-go-grow/10760/6

In addition, a new test case for ebase was added, which works fine. Issue reported here: https://forum.portfolio-performance.info/t/pdf-import-fuer-ebase/7204/23


buchen:
rebased, squashed, merged

Vielen Dank wie immer 👍

PR #1744, opened 2020-09-10
Re-apply width of first column

cmaoling:
It seems like the pack() function changes the width of the first element, as soon as a column is "packed" with a non-default width.
This causes the first column to be set to the TableViewerWidth.
I have no good explaination why, but this hack fixed the issue #1723 for me.


buchen:
Interesting. I cannot reproduce this problem on macOS.

But the code is also interesting. It packs all columns, and then reverts all **but the first column** back to the previous size. But isn't it the first column that is taking up all the space? And if we revert it back to the original size, we do not have to "pack" in the first place.

For now, I have just skipped packing the columns at all.

cmaoling:
@buchen The code reverts back the width of the first column when columns beyond the first are processed....
Anyway never tried to drop the pack() function.
This might be an issue w/ Linux then? I observed this for some time there...

PR #1745, opened 2020-09-10
Warn in case the investplan has is based on a security w/o quotes

cmaoling:
proposal to address https://github.com/buchen/portfolio/issues/1660
introducing ability to have more than one warning message


buchen:
Good catch. This warning messages will help users to understand what goes wrong

PR #1747, opened 2020-09-10
Fix Comdirect PDF import to support negative transaction amount

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/45


buchen:
The UI does not allow you to create a sale transaction with a negative value. But this example, as painful as it is, shows that this is a valid use case. I have to check if it works in all corners (particularly when filtering accounts) and then we should enable this use case.

PR #1752, opened 2020-09-16
Fix of NUP in MoneyTrailToolTipSupport

inv-trad:
Issue: https://github.com/buchen/portfolio/issues/1749

I don't get the error myself, but think that fix should work for others. What do you think, Andreas?


buchen:
I also do not really understand the problem.
I guess there is a slight change of behavior with the latest SWT version. Maybe if the tooltip is an area of the table that does not contains rows/columns (=no cells). Anyway, I think your fix is a safe bet.

PR #1758, opened 2020-09-21
Fix change to previous day

inv-trad:
Issue: https://forum.portfolio-performance.info/t/kursaenderung-zum-vortag-ist-falsch/5189/38

The date used so far (`last`) is the last one coming from quote provider rather than the last one stored as historic security price. Now the comparison should be correct.

Open is still the question, whether todays (`latest`) should be added as historic price as well? I changed the if statement back to `isBefore` to not have the latest price stored as historic price. That seems to work in diagrams and tables (e.g. Kursänderung zum Vortag).

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java
@@ -385,6 +385,15 @@ public final class Security implements Attributable, InvestmentVehicle
             return false;
 
         LocalDate now = LocalDate.now();
+        LocalDate lastStoredDate = null;
+        LocalDate previousStoredDate = null;
+
+        Optional<Pair<SecurityPrice, SecurityPrice>> previous = getLatestTwoSecurityPrices();
+        if (previous.isPresent())
+        {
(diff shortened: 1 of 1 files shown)


buchen:
> The date used so far (last) is the last one coming from quote provider rather than the last one stored as historic security price. Now the comparison should be correct.

That was a bad one. Because the attribute as well as the member variable is called "prices", it got mixed up.

Nonetheless, I do not want to use the ```getLatestTwoSecurityPrices``` method because it mixes the historic prices with the latest price. Both might have different sources (and providers) configured.

> Open is still the question, whether todays (latest) should be added as historic price as well? I changed the if statement back to isBefore to not have the latest price stored as historic price. That seems to work in diagrams and tables (e.g. Kursänderung zum Vortag).

We changed that because of this behavior of the Portfolio Report provider: https://forum.portfolio-performance.info/t/kurse-werden-nicht-mehr-aktualisiert/1453/64

Therefore I propose to
a) add today's date
b) and overwrite it because it is the last date in the list

Let's see if that works better. I am planning to create a lease next weekend (10/11 Oct)

buchen:
Merged with cb983940db430e5eb7c83d32170583f67a8861fa

PR #1760, opened 2020-09-22
Earnings per year chart coloring

inv-trad:
The new coding uses the same colors for each year that are used in monthly and quarterly charts.

Issue: https://forum.portfolio-performance.info/t/farbliche-differenzierung-der-dividenden-pro-jahr/10660/2

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/earnings/EarningsPerQuarterChartTab.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/earnings/EarningsPerQuarterChartTab.java
@@ -136,10 +136,11 @@ public class EarningsPerQuarterChartTab extends AbstractChartTab
             IBarSeries barSeries = (IBarSeries) getChart().getSeriesSet().createSeries(SeriesType.BAR,
                             String.valueOf(year));
 
-            double[] series = new double[Math.min(12, model.getNoOfMonths() - index)];
-            for (int ii = 0; ii < series.length; ii++)
+            double[] series = new double[Math.min(4, (int)(model.getNoOfMonths() - index) / 3)];
+            int length = Math.min(12, model.getNoOfMonths() - index);
+            for (int ii = 0; ii < length; ii++)
             {
-                series[(int) ii / 3] = series[ii / 3] + model.getSum().getValue(index + ii) / Values.Amount.divider();
+                series[ii / 3] = series[ii / 3] + model.getSum().getValue(index + ii) / Values.Amount.divider();
             }
             barSeries.setYSeries(series);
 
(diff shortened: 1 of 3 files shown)


Ragas13:
@inv-trade
I got the error that the stacked chart could not contain negative values. For dividends it works but for expenses for example unfortunately not :thinking:

inv-trad:
@Ragas13 
Thank you for catching that. Sorry, I have to test more thoroughly before creating a pull request.... 

I commited a (temporary) change containing a check for negative numbers before enabling to stack, what makes it work for earnings as well as expenses. But, it looks a little odd for expenses, because the bars are distributed evenly an the x axis even though it is showing only one bar per year. 

grafik

On the sum page a mixture of stacked and non-stacked can be displayed, still not looking perfect. In this example 2019 and 2020 are stacked, thus the space on the x axis is split in halves.

grafik

I will try to find another solution to have a proper display even if there is a negative bar.

One option could be in a first step to display the stacking only for earnings (when all bars are positive) and show the old behaviour in case there exists a negative bar. **Not perfect, but what do you think about that**?!

buchen:
> One option could be in a first step to display the stacking only for earnings (when all bars are positive) and show the old behaviour in case there exists a negative bar. Not perfect, but what do you think about that?!

If we do it, then this is the option. I think the "not aligned" bars look really ugly.


Another alternative is, of course, to ditch SWTChart for this diagram and use d3.js and the Browser.
This is what I do for the pie charts. And d3.js is pretty flexibly so that something like you, Marco, showed is sure possible.
And I have the impression that this is fairly stable - only a few people complaining not to see the pie chart.

Basically, you would create a HTML containing the d3.js JavaScript code. And provide a Java method that creates the JSON data structure to be used (including the colors, etc). See here for the holdings chart: https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/HoldingsPieChartView.java#L143

sebasbaumh:
>Another alternative is, of course, to ditch SWTChart for this diagram and use d3.js and the Browser.
>This is what I do for the pie charts. And d3.js is pretty flexibly so that something like you, Marco, showed is sure possible.
>And I have the impression that this is fairly stable - only a few people complaining not to see the pie chart.

d3.js needs some additional styling for dark mode (see #1751), but I would also see it as the way to go.
It allows far more sophisticated diagrams and maybe there are less people complaining about missing charts with the new eclipse/SWT version.

inv-trad:
The version that works with coloring when there is no negative bar and showing the old behavior, when there is a negative bar is commited.

When checking, whether it would be an option to overwrite the `getBarColor()` method, I came across a feature that might solve the problem. The current version of `BarSeries `class contains an attribute called `barOverlay`, which is checked when calculating the width of the bars doing so only, if not set. 

https://github.com/eclipse/swtchart/blob/develop/org.eclipse.swtchart/src/org/eclipse/swtchart/internal/series/BarSeries.java

Would it be an option to switch to a newer version (0.12.0) of SWT and use this feature?

https://github.com/eclipse/swtchart/releases

buchen:
> Would it be an option to switch to a newer version (0.12.0) of SWT and use this feature?

Yes, that is an option.

Sometime back I already created the [feature_org.eclipse.swtchart](https://github.com/buchen/portfolio/tree/feature_org.eclipse.swtchart) branch containing some of the API changes necessary to move to newer SWTChart version. There might be more API changes as the SWTChart projects did quite some major changes.

* Updating the target platform to include SWTchart from Eclipse
* cherry-pick the change from the branch (the branch is quite old, I doubt that merging is helpful).

buchen:
I have merged the change regardless. I think it is nice to have the right colors. And mostly the values will be positive.

PR #1766, opened 2020-10-01
Change shares display precision

### LINKED ISSUE #1459: Increase shares display precision
Hello!

First of all thanks for this awesome tool which I've just started using.

Most of my transactions have partial shares with 5 numbers of precision, however in the UI everything is displayed with 3 numbers. That makes my transactions harder to reconcile and navigate, and my account statements less useful, as I have to hover around each number. Could this be increased? Or made configurable?

Screenshot 2020-03-22 at 12 31 23

Thanks :-)

pascalhein:
This adds a new preference page with one setting to change
the share display precision from 0 up to 6 decimal digits.
This is applied to every shares column across the
application. Fixes #1459.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -910,6 +910,7 @@ public class Messages extends NLS
     public static String PrefLabelNote;
     public static String PrefLabelProxyHost;
     public static String PrefLabelProxyPort;
+    public static String PrefLabelSharesDigits;
     public static String PrefLabelUseIndirectQuotation;
     public static String PrefMsgConfigureUpdates;
     public static String PrefMsgLanguageConfig;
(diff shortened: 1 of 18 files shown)


buchen:
Hi @pascalhein,

thank you very much for your contributions. Due to my work schedule, I will only get to them in detail on the weekend.

BTW, in other cases I have created an "addon" that is updating a static field whenever the configuration changes. I think we can consider this over extending every view with injection of the preferences, overwriting the method, etc. Yes, there is some magic how the preference are then passend on, but it would also reduce boilerplate. What do you think?

For example, here: https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/Preference2EnvAddon.java#L58-L66

pascalhein:
Hi,
thanks for your suggestion.
I moved the code as you proposed - indeed that seems like a much cleaner solution to me. Initially, I was worried that changes might not be visible (immediately) without a `table.refresh`, but it looks like this is also "magically" taken care of ;-)

Take your time in reviewing the changes, I'm looking forward to your comments!

buchen:
squashed, rebased to master, merged. Thanks!

> Initially, I was worried that changes might not be visible (immediately) without a table.refresh, but it looks like this is also "magically" taken care of ;-)

I believe the ```OwnerDrawLabelProvider``` gets a paint event whenever the screen is painted - also after closing the dialog.

PR #1768, opened 2020-10-01
Add context menus to many more tables

### LINKED ISSUE #674: Rechtsklick-/Popup-Menü in KLASSIFIZIERUNG>...>REBALANCING um KAUFEN, VERKAUFEN etc. ergänzen
Im Rechtsklick-/Popup-Menü auf ein Wertpapier in der Rebalancing-Ansicht (KLASSIFIZIERUNG>...>REBALANCING) würden uns die Einträge für KAUFEN, VERKAUFEN etc. sehr helfen (wie bspw. in der Ansicht WERTPAPIERE>ALLE WERTPAPIERE).

Momentan gibt es dort nur ENTFERNEN oder IM BROWSER ÖFFNEN.

Vielen Dank.

pascalhein:
Fixes #674, and also for other views. In the next step one
could potentially add the shortcuts (usually Edit at C-e) to
more views as well, if that is useful at all.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/TransactionContextMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/TransactionContextMenu.java
@@ -1,0 +1,204 @@
+package name.abuchen.portfolio.ui.views;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+
+import org.eclipse.jface.action.Action;
+import org.eclipse.jface.action.IMenuManager;
+import org.eclipse.jface.action.Separator;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.KeyEvent;
+
+import name.abuchen.portfolio.model.Account;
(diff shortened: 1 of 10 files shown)


buchen:
Hi @pascalhein,

I rebased and merged the addition of the security context. Thanks.

I removed, however, the addition of the transaction context menu to the ```TransactionsTab``` in the earnings view. The reason is this: if the user selects a filter (in the top right), PP might create new transactions. For example, if there is a delivery from security account to another, then PP might create a pseudo transaction if only one security account is included in the filter. Therefore the UI might or might not have the original transaction at hand when opening the edit dialog.

How to fix this? One option could be to add a field ```Map<SomeEnum, Object> data``` to the ```Transaction``` class and have the filter attach the original transaction with ```psoudoTx.setData(SOURCE, t)```. Then the transaction context menu could retrieve the original transaction with ```transaction.getData(SOURCE)```. The map would have to be transient in order to not show up in the XML. I would use a map because I had similar use cases in the past. And I propose to use an enum so that it is easier to find/add usages in the code (as opposed to using strings as keys). I would be happy to include such a change if you have the time to develop and test it.

I left the ```TransactionContextMenu``` class even though it is at the moment only used in one place. But it makes sense to refactor this code anyway.

Issue #1732, opened 2020-09-05
incorrect performance calculation on weekends


DomBok:
The calculation keeps showing the value change of friday and if there are changes of the value on weekends they are calculated with the value from friday.

Last change of value was on friday:
![1](https://user-images.githubusercontent.com/39921798/92302188-7236fe80-ef6a-11ea-8621-ddcbc0b6b8d5.jpg)

The value of saturday shows the change that happend on friday:
![2](https://user-images.githubusercontent.com/39921798/92302193-824ede00-ef6a-11ea-9af7-7560310bf0d7.jpg)

 - OS: Windows 10
 - Version: 0.48.1

PR #1772, opened 2020-10-04
PR/1732 - Bugfix of end date @ AccountSnapshot

Ragas13:
Bugfixe of end date determination, not the date before
#1732


buchen:
What a bug. Interesting that is was around for such a long time. Thanks @Ragas13  for finding and fixing!

PR #1780, opened 2020-10-10
Enhancement to security views to show change to previous day as amount

inv-trad:


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesTable.java
```diff
@@ -389,13 +390,74 @@ private void addDeltaColumn() // NOSONAR
...
+
+            return Double.compare(v1, v2);
+        }));
+        support.addColumn(column);
+    }
+
+    private void addDeltaAmountColumn() // NOSONAR
+    {
+        Column column;
+        column = new Column("6", Messages.ColumnChangeOnPreviousAmount, SWT.RIGHT, 80); //$NON-NLS-1$
```
buchen, line 409: One note: way back, the columns were just ordered. For these historical reasons, some columns have a number as an "id". It was the position of the column in the original set of columns. Today, I give them short readable names. The requirement is that the name must be unique for a set of columns in a view.

buchen:
I have changed the column id and then merged the change. Thanks also for providing the translations for the other languages.

PR #1781, opened 2020-10-11
Fix Flatex PDF import to not read dividend cancellations

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/79

Dividend cancellations "Storno Dividendengutschrift für ausländische
Wertpapiere" were detected as "Dividendengutschrift für ausländische
Wertpapiere" and dividends booked once again.


buchen:
squashed & merged

Danke!

PR #1782, opened 2020-10-17
Change Yahoo API from SPARK to CHART

Ragas13:
SPARK response removed the adjusted quote price, CHART still contain both prices.

According to https://forum.portfolio-performance.info/t/kurse-werden-nicht-mehr-aktualisiert/1453/72


buchen:
With this change, I am not able to load any quotes anymore:

```
java.io.IOException: https://query1.finance.yahoo.com/v7/finance/chart?symbols=BAS.DE&range=3mo&interval=1d
```

@Ragas13 - can you also try with the BAS.DE symbol? You get data alright?

buchen:
BTW, with the "old" code I am able to retrieve data (yeah, not I get it, I get historical data, but not adjusted close)

rbaeumler:
With "chart" the URL is not working for me in the browser: 
`https://query1.finance.yahoo.com/v7/finance/chart?symbols=BAS.DE&range=3mo&interval=1d`
With "spark" I get data:
`https://query1.finance.yahoo.com/v7/finance/spark?symbols=BAS.DE&range=3mo&interval=1d`

Is historical data usefull at all iof it is not adjusted close price? Without taking stuff like splits into account?

Ragas13:
Strange, maybe I've used a wrong hyperlink?

For DAI.DE I've found the URL https://query1.finance.yahoo.com/v7/finance/chart/DAI.DE?interval=1d&range=max and for BAS.DE https://query1.finance.yahoo.com/v7/finance/chart/BAS.DE?interval=1d&range=max

But I guess that I do have actually limited access to a ECLIPSE engine, so I could not fully debug my faulty code 😞

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooFinanceQuoteFeed.java
```diff
@@ -206,8 +206,8 @@ else if (days < 600)
         else if (days < 1500)
             range = "5y"; //$NON-NLS-1$
 
-        return new WebAccess("query1.finance.yahoo.com", "/v7/finance/spark") //
-                        .addParameter("symbols", security.getTickerSymbol()).addParameter("range", range)
+        return new WebAccess("query1.finance.yahoo.com", "/v7/finance/chart" + security.getTickerSymbol()) //
```
buchen, line 209: Thanks. It works for me if I add a dash "/" after the chart path element

buchen: And then it looks to me that the API is now using "v8" as version identifier. So I bumped that up too.

buchen:
squashed, merged, thanks!

buchen:
@rbaeumler 

> Is historical data usefull at all if it is not adjusted close price? Without taking stuff like splits into account?

The *adjusted close price* is very tricky. Whenever there is a dividend payment, it lowers the historical prices before this dividend payment. Therefore the adjusted close does not make sense for any instruments that you hold in your portfolio (as you record dividend payments separately). The adjusted close can make sense if you want to use an instrument as benchmark. But you have to be aware that you have to delete all prices with every material change (split, dividend).

rbaeumler:
@buchen 
Thanks a lot for the explanation. Looks like I need to update my settings then, as I have all my holdings configured with the adjusted close price (where it was available).

PR #1806, opened 2020-11-03
Importer Fix - comdirect

Pnda87:
fix comdirect import after commerzbank-comdirect merger

https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/53


buchen:
cherry-picked. That is a simple enhancement.

PR #1812, opened 2020-11-05
Import from Flatex documents

666wacken666:
1. Old bank statements.
2. Creation of a note from the payment reference
3. Creation of a transaction with inverse sign for reversal documents.
4. Small bug fixes.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/fintechgroupbank/FinTechGroupBankDividendeAusland5.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/fintechgroupbank/FinTechGroupBankDividendeAusland5.txt
@@ -1,0 +1,31 @@
+FinTech Group Bank AG
+Hausbroicher Str. 222
+47877 Willich
+USt-IdNr.: DE 246 786 363
+Kundenservice:
+Tel.: +49 (0)9221 - 7035898
+E-Mail: kunden@flatex.de
+FinTech Group Bank AG -  Rotfeder-Ring 7 - 60327 Frankfurt am Main
+XXXXXXXXXXXXXXXXXXX
+Storno Dividendengutschrift für ausländische Wertpapiere
+Ihre Depotnummer: XXXXXXXXXX
+Depotinhaber    : XXXXX, XXXXX
+Nr.111111111                  STARBUCKS CORP.           (US8552441094/884437)
+St.             :         105          Bruttodividende
+                                       pro Stück       :       0,2500 USD
+Extag           :  08.08.2017          Bruttodividende :        26,25 USD
+Zahlungstag     :  25.08.2017          Bemessungs-
+Valuta          :  25.08.2017          grundlage       :         X,XX EUR
+Devisenkurs     :    1,180800         *Einbeh. Steuer  :         1,11 EUR
+Quellenst.-satz :       30,00 %        Gez. Quellenst. :         7,88 USD
+                                       Endbetrag       :        14,45 EUR
+* Enthalten sind folgende Steuern (negative Werte bedeuten Steuererstattung)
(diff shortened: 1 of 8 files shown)


buchen:
Hi @666wacken666,

thank you very much your code contribution. The code as such looks good.

I have one architectural concern, however: So far, PP assumes that the transactions values are always positive. There is a transaction type FEES with a positive number, and a transaction type FEES_REFUND with a positive number. I agree one can argue of the merit of such an assumption, but that how it is today. Your code creates negative values. I can fairly easily change the one location of FEES to use FEES_REFUND.

But: there are many "reversal" dividend statements where PP has no corresponding transaction type. In your code, you generate DIVIDEND transactions with negative value for "Storno Dividendengutschrift", "Storno Ertragsmitteilung", "Storno Zinsgutschrift").

We could not introduce a new transaction type DIVIDENDS_REVERSAL.

However, I am hesitant because **what does that mean from a performance perspective?** Let's say you get the dividend payment on the first day of month, the reversal (Storno) on the 10th day of the month - do you have a positive performance impact on the first, and a negative on the 10th? Actually, no, because it is just correcting a performance that never was there in the first place. And are we sure the "false" performance is completely reversed by the second transaction? I do think that IRR and TWRROR will have artifacts of this performance.

Personally, when I get such bank documents, I *manually* delete the original dividend transaction. And the updated dividend transaction is imported with other documents. But I know, this manual...

Let's brainstorm what options we have (and maybe @Ragas13 @inv-trad have more ideas):

0) Leave as today: do not import automatically, leave it to manual work
1) Accept proposed change: have negative values and check which implicit assumption in the code might break
2) Introduce DIVIDENDS_REVERSAL transaction type and accept the (probably small) impact on dividends. And if we use same day for importing the transaction (*Valuta* or *Extag*) there probably is not performance impact
3) Upon import, try to find and correct the original transaction. That can be tricky, for example if the original transaction is deleted or if bank documents are not imported in order?

**What do you all think? Other ideas?**
I am warming to the idea that the reversal transaction maybe has limited impact on the performance calculation.


A little bit illogical, PP does know a INTREST_CHARGE transaction. That transaction type can be used if you have to pay interest. If it is used to "correct" a transaction, however, you could have the same impact on performance.  

Andreas.

666wacken666:
Hi @buchen 
I agree with you, the implemantion is not very nice. But I tested it with the real appliaction and it worked very good, with several different documents. The drawback that i noticed is, that it is not possible to manual change the transaction, because of the forbidden negative values. Due to the fact, that the Flatex reversal document contains the value date (valuta), which is the same as for the normal document, there was no influance for the performance calculation as far as i could evaluate it.
Response to your suggestions:

> Leave as today: do not import automatically, leave it to manual work

For me that is not an option, due to too many reversal documents. I normally import all documents form my broker at once without preselecting. Finding an error afterwards would be very messy.


>  Accept proposed change: have negative values and check which implicit assumption in the code might break

See explanation above and example file. [Storno.xml.txt](https://github.com/buchen/portfolio/files/5505403/Storno.xml.txt)

> Introduce DIVIDENDS_REVERSAL transaction type and accept the (probably small) impact on dividends. And if we use same day for importing the transaction (Valuta or Extag) there probably is not performance impact

That would be the prefered option for me, but you need more knowledge about the source code. So for me it was no solution.

> Upon import, try to find and correct the original transaction. That can be tricky, for example if the original transaction is deleted or if bank documents are not imported in order?

I prefer, if the transaction are identical with the bank statements, so that i don't need the bank statements any more. So for example, i would implement interest transactions with a vaule of 0,00EUR too.

inv-trad:
I came across such kind of request a couple of times and didn't implement it so far. My primary approach would have been to identify the corresponding transaction and correct/reverse it. For the situation that the transaction doesn't (yet) exist, I never really had a solution. 

From my point of view, the best solution would be to combine the two. First, try to find an existing transaction to revert/correct, if that fails, use the DIVIDENDS_REVERSAL transaction type. Does that make sense?

PR #1837, opened 2020-11-23
IB cfd import

MichaelSp:
support for CFD trades from IB via the flex.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
@@ -52,7 +52,26 @@ public class IBFlexStatementExtractor implements Extractor
     private final Client client;
     private List<Security> allSecurities;
 
-    private Map<String, String> exchanges;
+    private final Map<String, String> exchanges;
+    private static final Map<String, String> CFD_MAPPING;
+
+    static
+    {
+        Map<String, String> m = new HashMap<String, String>();
+        m.put("IBUS500", "^GSPC");
+        m.put("IBUS30", "^DJI");
+        m.put("IBUST100", "^IXIC");
+
+        m.put("IBGB100", "^FTSE");
+        m.put("IBEU50", "^STOXX50E");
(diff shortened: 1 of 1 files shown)


buchen:
To be honest, I haven't had the time to wrap my head around the short selling.
PP assumes you invest X to purchase an asset and that it can make daily valuation of that asset.

My naive thinking is:
For the transaction above, the fees (5.8) what you actually pay to purchase the asset (CFD).
The daily valuation of your position would be something like *currentPrice - tradePrice*.
PP does not support that (yet?)


How is InteractiveBrokers showing the valuation of your CFD?
How is it showing the investment.

MichaelSp:
I think "Valutation" is called "Net asset value"? (I'm not to familiar with all these financial terms)

In NAV you'll find basically these entries:
image
Stock, Cash and Dividend as absolute values
But CFD only as difference: In Profit/Loss

I' try to find out how it's rated performance wise. 

Meanwhile I found a bug in the import, which I'm gonna try to fix too:
https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java#L483-L487 
Here the `gross unit` is added based on total` value.

And I have to think about this change here: https://github.com/buchen/portfolio/commit/d57af920da794667dd63bb29a8db272a71ee7c82 I was about to use `cost` instead of `netCash`...

MichaelSp:
I think "Valutation" is called "Net asset value"? (I'm not to familiar with all these financial terms)

In NAV you'll find basically these entries:
image
Stock, Cash and Dividend as absolute values
But CFD only as difference: In Profit/Loss

I' try to find out how it's rated performance wise. 

Meanwhile I found a bug in the import, which I'm gonna try to fix too:
https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java#L483-L487 
Here the `gross unit` is added based on total` value.

And I have to think about this change here: https://github.com/buchen/portfolio/commit/d57af920da794667dd63bb29a8db272a71ee7c82 I was about to use `cost` instead of `netCash`...

PR #1838, opened 2020-11-23
add support for CTRL+A -> select all

MichaelSp:


buchen:
Hi @MichaelSp,

thanks for the contribution. I was not aware that Ctrl-A does not work in e4. I have implemented it slightly different so that it works on all tables and trees - see b5fc7e403dd06aa7688132e1d541422471bcc4d7



Andreas.

MichaelSp:
Uhh I'm impressed. That's some serious eclipse sorcery I must learn....
Thanks!

PR #1854, opened 2020-12-07
Enhancement to add monthly average to earnings heatmap

inv-trad:
https://github.com/buchen/portfolio/issues/1795


buchen:
Merged.

I did one tiny change: instead of introducing a new keyword to store the information whether to show the average, I reused the "LAYOUT" keywords. It is also used for the other heatmaps. And it allows us to have additional options in the future (besides the average).

uok:
@buchen, @inv-trad, I'm happy that my idea made it into reality :-)

2 small suggestions:
+ change position of "average" to left side of "sum"
+ make background color of "average" and "sum" a little darker for visual separation
Thanks!

PR #1857, opened 2020-12-10
Make HTMLTableQuoteFeed work for latest security price

fcaputo:
I found out that getting the latest quote from "Tabelle auf einer Webseite" doesn't work, because the website usually only contains the time and not the date.
I fixed it and added a unit test for historical and latest quote.


fcaputo:
Found an NPE, will create a new PR.

PR #1859, opened 2020-12-11
Fix wrong and inconsistent performance calculation for part of portfolio

inv-trad:
Issue: https://github.com/buchen/portfolio/issues/1817

The problem was that transfers (in and out) were not part of the lineitems used to calculate costs and capital gains in `SecurityPerformanceRecord`. Adding them to the list, cross portfolio calculations work correctly.


inv-trad:
@buchen: Andreas, I uploaded a very minimalistic example in the forum: https://forum.portfolio-performance.info/t/performance-berechnung-falsch-rendite-berechnung/4718/10

buchen:
Hi @inv-trad - thanks for fixing!! That is an important one. It is merged.

I did change your commit slightly - there was already a helper method to re-created the transfers (used from other filter implementations). And thanks also for the test cases. I added one of the test cases simply to the test project. Before applying your fix, it failed, afterwards it ran okay.

PR #1860, opened 2020-12-13
Html table quote feed for latest security price

fcaputo:
I found out that getting the latest quote from "Tabelle auf einer Webseite" doesn't work, because the website usually only contains the time and not the date.
I fixed it and added a unit test for historical and latest quote.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeedHistoricalSample.html
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeedHistoricalSample.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+        <title>Druckansicht - Titel</title>
+        <meta name="description" content="">
+
+    </head>
+    <body>
+    <div class="container">
+    <header>
+        <h1>Historische Kurse zu DNB Fund - Technology - A EUR ACC (LU0302296495)</h1>
+    </header>
+    <main role="main">
+        <table>
+            <thead>
+            <tr>
+                                    <th class="text-left" width="64px">Datum</th>
+                                    <th >Eröffnung</th>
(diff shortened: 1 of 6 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/impl/HTMLTableQuoteFeedHistoricalSample.html
```diff
@@ -0,0 +1,49 @@
...
+                                    <th >Eröffnung</th>
+                                    <th >Hoch</th>
+                                    <th >Tief</th>
+                                    <th >Schluss</th>
+                                    <th >Volumen</th>
+                            </tr>
+            </tfoot>
+            <tbody>
+                            <tr>
+                                    <td class="text-left" width="64px">03.12.2020</td>                                                                                                <td>665,89</td>                                                                            <td>675,00</td>                                                                            <td>665,89</td>                                                                            <td>675,00</td>                                                                                                <td>0</td>                                </tr>
```
buchen, line 38: It really drove me nuts to understand where the quotes are coming from... until I realized I needed to scroll.
I have now formatted the HTML page in the repository. The spaces should not make a difference 😄

name.abuchen.portfolio/src/name/abuchen/portfolio/model/SecurityPrice.java
```diff
@@ -15,11 +16,12 @@
         @Override
         public int compare(SecurityPrice p1, SecurityPrice p2)
         {
-            return p1.date.compareTo(p2.date);
+            return p1.getDateTime().compareTo(p2.getDateTime());
         }
     }
 
     private LocalDate date;
+    private LocalDateTime dateTime;
```
buchen, line 24: Added member variables to ```SecurityPrice``` is something I do not want to do.
Any change here has an impact on the XML serialization. And the ```SecurityPrice``` is the most used object in the XML.
Plus the time is not used in any of the calculations - PP does not do intra-day performance.

I have changed your implementation: the time is now held only internally in the ```HTMLTableQuoteFeed```. Before passing it to the outside, the feed keeps on the last price of a given day.

PR #1879, opened 2020-12-18
Fix dividend yield per year display when split in multiple transactions

inv-trad:
Issue: https://forum.portfolio-performance.info/t/dividendenrendite-div-jahr-bei-gleichem-wertpapier-in-mehreren-depots/11911

An example can be found in the thread in the forum.

The calculation logic works fine for line items like in the transaction tab, where the individual number of stocks and their specific cost needs to be taken into consideration.
When calculating the average dividend yield per year, this is done by summing up the individual rates of return and here the calculation needs to relate to the total costs, otherwise multiple percentages are summed up.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/DividendTransactionTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/DividendTransactionTest.java
@@ -176,7 +176,7 @@ public class DividendTransactionTest
                                                                     // 1800
         t1.setMovingAverageCost(movingAverageCost);
 
-        double result = t1.getPersonalDividendYieldMovingAverage(); // hence
+        double result = t1.getPersonalDividendYieldMovingAverage(true); // hence
                                                                     // 5.5556%
                                                                     // yield
 
@@ -190,7 +190,7 @@ public class DividendTransactionTest
                         LocalDateTime.of(2019, 01, 15, 12, 00));
         // Transaction has no movingAverageCosts
 
-        double result = t1.getPersonalDividendYieldMovingAverage();
+        double result = t1.getPersonalDividendYieldMovingAverage(true);
 
         assertEquals(0.0, result, 0.001d);
     }
@@ -209,7 +209,7 @@ public class DividendTransactionTest
                                                                     // 1800
         t1.setMovingAverageCost(movingAverageCost);
 
-        double result = t1.getPersonalDividendYieldMovingAverage(); // hence
+        double result = t1.getPersonalDividendYieldMovingAverage(true); // hence
                                                                     // 6.66666%
                                                                     // yield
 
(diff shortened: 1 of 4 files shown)


buchen:
Thanks @inv-trad for tackling this one. Oh, there is some technical debt here. The DividendCalculation relies upon the CostCalculation running before and setting the fifoCosts and averageCosts values before. That dependency is not really easy to understand (for now I do not have a better idea though).

Your change fixes it for the sample file.

But looking at the DividendCalculation#finish method, I wonder if should change that calculation. The algorithm is summing up percentages and then dividing by the number of years. First it does not remove years for which there are no payments (possibly because there were no holdings). Second it could take into account that the rate of return was only calculated for a partial position. I think I have to sleep over it. What do you think?

The relevant snippet:
```java
    public void finish(CurrencyConverter converter, List<CalculationLineItem> lineItems)
    {

        double sumRateOfReturn = 0;

        for (Payment p : payments)
        {
            // other stuff
            sumRateOfReturn += p.rateOfReturn;
        }

        int years = 0;

        // now walk through individual years
        for (int year = firstPayment.getYear(); year <= lastPayment.getYear(); year++)
        {
            years++;

            // other calculations
        }

        this.rateOfReturnPerYear = sumRateOfReturn / years;
    }
```

inv-trad:
Hi Andreas (@buchen),

I saw that logic of summing up and dividing and that is where the problem arises. If you have e.g. 100 shares of a security you paid 100 EUR for in 2 portfolios (50 and 50) and get two individual dividend payments e.g. of 10 EUR, which is a total of 20 EUR = 2%. The old calculation will calculate 10/50 = 2% twice and sums it up later. That results in 4%, where 20 EUR with a total cost of 100 should be 2%.

Because the percentages are summed up, it needs to be calculated based on the total costs, not parts of it. I tested in different variations and they all worked. well.

On individual item level (transaction tab) it is correct to base the cost on the number of shares the dividend is paid for. That is why I split that logic.

In fact I was considering rewriting it from scratch first. Or going a different approach, calculating the percentage of a rolling 12 months period. Would that make sense instead of in addition to the current solution of looking over several years?

Take care,
Sven

buchen:
Hi @inv-trad 

I have merged your change. Independent of our discussion, I think it improves the calculation.

However, I have slightly changed it. My thinking was this: Let's think of the ```CalculationLineItem.DividendPayment``` as the values **per** payment of the ```DividendCalculation``` **per** Client. The knowledge that all those yields are summed up and divided by the years should be private to the DividendCalculation class.

Then I added the sample as a test case. If we improve the calculation later, we will see if we break something.


> Or going a different approach, calculating the percentage of a rolling 12 months period. Would that make sense instead of in addition to the current solution of looking over several years?

Personally, I am not a such a dividend focused investor. I just do not know. Looking around some of the web sites, I have the impression that the dividend yield is always calculated on the calendar year. If we would adding a rolling 12 months period, one would suddenly see numbers that do not fit. On the other hand, in the security performance view, the user can choose arbitrary reporting periods. If one chooses just 3 months, the number will be off anyway (if, at least, the security has multiple dividends payments per year).

[The Nest](https://budgeting.thenest.com/calculate-5year-average-dividend-yield-28644.html) writes:
> To calculate the dividend yield over five years, first calculate the dividend yield for each of the past five years. Divide the dividends paid during each year by the stock price for that year. Be sure to include both periodic dividends and special dividends when calculating the total paid. For example, a company might pay a quarterly dividend and a special dividend at the end of the year or after selling off a portion of the company. Second, add up all of the dividend yields for the previous five years. Third, divide the result by 5 to calculate the average dividend yield.

PR #1894, opened 2020-12-20
[translation properties only] Correct IB XML import related text

damarvin:
Correct naming to "Activity Flex Query".
Add tool-tip hints for Interactive Brokers related relevant parts.
See e.g. https://minddumpdotblog.wordpress.com/portfolio-performance/


buchen:
Merged. Thanks!

I like the blog 😄 If I come around to updating the documentation, I will link to it.

PR #1895, opened 2020-12-20
Trim values during CSV import

chrisaut:
Fixes https://github.com/buchen/portfolio/issues/1880

Note: There is also the option to perform this trimming at a lower level, eg. directly in the file-handling, but this seems to be meant to return true RAW values, so I've done it in the Extractor. Also added a quick test.


buchen:
rebased, merged, thanks!

PR #1896, opened 2020-12-20
CSV import: dynamic currency

chrisaut:
If no currency column is provided, try to use the selected account currency.

Fixes https://github.com/buchen/portfolio/issues/1877

Note: it's a bit trickier than I thought as the account is not known (or at least not fixed) when the extraction happens, as the user can change the target account on the review page. So instead mark the currency as dynamic and update on the fly on the review page before checks and the import is run.


name.abuchen.portfolio/src/name/abuchen/portfolio/model/Transaction.java
```diff
@@ -144,6 +144,7 @@ public int compare(Transaction t1, Transaction t2)
 
     private LocalDateTime date;
     private String currencyCode;
+    private boolean isDynamicCurrencyCode;
```
buchen, line 147: Adding a property to any class in the "model" packages has an immediate impact on the XML model. It means the property will show up in the XML and we cannot remove it anymore.

One alternative is that we define the property as "transient". Then the field is not stored and read from the XML file.

However, in this case I propose to check the property value itself. There are two options:

* we can leave the transaction ```null``` which identifies that no transaction has been found. I believe that would pretty much reflect the result of the import: no currency found. However, as PP is also often working with ```Money``` and that class requires a currency,
* we can use a special value - I would prefer ```UNDEFINED``` over ```DYNAMIC```. Instead of checking with an additional property, we can check with ```"UNDEFINED".equals(tx.getCurrencyCode())```

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/BaseCSVExtractor.java
```diff
@@ -140,7 +140,7 @@ protected String getCurrencyCode(String name, String[] rawValues, Map<String, Co
     {
         String value = getText(name, rawValues, field2column);
         if (value == null)
-            return client.getBaseCurrency();
+            return "DYNAMIC"; //$NON-NLS-1$ if not defined in CSV, mark to be dependend on target currency dynamically
```
buchen, line 143: As I written before, I prefer ```UNDEFINED```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -746,6 +747,34 @@ private void checkEntriesAndRefresh(List<ExtractedEntry> entries)
 
     private void checkEntries(List<ExtractedEntry> entries)
     {
+        // update dynamic currency if any
+        for (ExtractedEntry entry : entries)
```
buchen, line 751: If we update the transactions at this point of time, then the currency is set to the first account selected. If the user selects a different account, then the transactions will not be updated again. So instead of the currency of the file, it now depends on the currency of the first account.

Instead, my proposal is:
* Extend the ```CheckCurrencyAction``` to accept transactions with the ```UNDEFINED``` currency **if** the transactions are okay if they would have the currency of the account.
* Extend the ```InsertAction``` to update the currency to the currency of the account if the currency code is ```UNDEFINED```

buchen (changes requested):
Hi @chrisaut,

thanks for the pull request. I think it make sense to enhance the import in such a way that the currency of the account is taken if the imported data contains no currency information.

Please have a look at my comments. I am happy to review an updated PR.

My line of argument is:
* Do not change the model classes as it has impact on the persisted XML + we can check against the value of currency anyway --> no additional attribute needed
* Move the logic of determining the currency into the CheckCurrencyAction and InsertAction --> apply not to the first account, but to the actual account chosen

chrisaut:
Thanks, will try to address over the next few days.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/BaseCSVExtractor.java
chrisaut: Done

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
chrisaut: I don't believe this is true.

checkEntries gets called when the account selection changes on the review page.
maybeUpdateDynamicCurrency gets the target currency from getAccount....which gets it from the currently selected account, not the first.

So it works, but here's the problem. If I don't store somewhere that it was "undefined" (this isDynamicCurrencyCode in the Model) then the first selection will override, and subsequent account selection changes will be stuck at the first account's currency.

I did actually go down the path of modifying the CheckCurrencyAction before, but:
 - I found there were at least 4 or 5 places that needed patching up, and it wasn't always clear
 - I was also a bit worried to give the UI check different values to when the import finally runs....it seemed cleaner to me to run through the checks with the currency set, so it can properly do it's thing.

How about I just put the is isDynamicCurrencyCode field into the reviewpage itself, it will get set on first load based on the first entries currencyCode == "UNDEFINED". Then the model doesn't have to change, the UI updates nicely with the currency, and we don't have to run preview checks differently from the final input checks?

chrisaut:
Rebased, renamed to UNDEFINED, waiting for further discussion as I'm not clear.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
buchen: > How about I just put the is isDynamicCurrencyCode field into the reviewpage itself

That is a good idea!

Today, every extracted transaction is already wrapped in to a [Extractor.Item](https://github.com/buchen/portfolio/blob/a8eabbbe3b91f448c93f07e1021a50d6965c5a37/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java#L52). This item already has a generic ```data``` field to store the JSON representation (if a JSON-based extractor is used).

What do you think about this algorithm?
* The extractor creates a valid transactions - also with a valid currency (just to make sure we do not pollute with incorrect currencies if the UI is not used)
* The extractor tells the ```Extractor.Item``` that the currency is dynamic / undefined - I think a new boolean property to the ```Extractor.Item``` does not hurt
* Your logic reads this property and adopts the currency as needed *before* passing it to CheckCurrencyAction and InsertAction

buchen: @chrisaut - just a quick check-in. Are you planning to update the change?

chrisaut: Yes, sorry for the delay. I just didn't get to much of anything over the holidays. I should have some more time next weekend to look at all my open tasks for PP, if that's ok?

PR #1897, opened 2020-12-20
Allow input of fees for dividend transactions

chrisaut:
Fixes https://github.com/buchen/portfolio/issues/1889

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
@@ -47,6 +47,7 @@ import name.abuchen.portfolio.snapshot.PortfolioSnapshot;
 import name.abuchen.portfolio.snapshot.SecurityPosition;
 import name.abuchen.portfolio.ui.Messages;
 import name.abuchen.portfolio.ui.UIConstants;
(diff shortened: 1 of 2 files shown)


Fenrihr:
i will love see this feature implemented

buchen:
Hi @chrisaut,

thanks for the pull request. I had to add a couple more changes so that the calculations (and in particular the filters) properly take into account that FEES are now available on dividend transactions as well. See b011e46e07a9df9c489b709a467803044dec6f87

I have tested a couple of scenarios. If you could take your file and update your transactions and check that the calculations are as expected, that would be very helpful as well.

Andreas.

chrisaut:
Thanks @buchen 

Anything I particular I should look for? In the PR I checked that the fees would show up properly under earnings and expenses.

I apologize, I clearly don't understand the model enough to understand all the implications. I, apparently wrongly, assumed that a a FEE transaction would automatically be handled in line with other fees. Does every "type" (eg. dividend) need to be handled separately in all calculations with regards to taxes and fees?

buchen:
> Does every "type" (eg. dividend) need to be handled separately in all calculations with regards to taxes and fees?

Not necessarily. But this was the first AccountTransaction that has a separate FEE unit. Previously, only the PortfolioTransaction had this FEE Unit. That's why more calculations needed to be adopted. I noticed when checking the Performance calculation that the fees were missing. The model has grown over the years... What I did: I searched for the usage of the DIVIDENDS enum and then check if anything additional is needed.

PR #1898, opened 2020-12-20
Fix IRR calculation for securities with portfolio transfer

inv-trad:
Issue: https://forum.portfolio-performance.info/t/izf-negativ-bei-positiver-rendite/12363/10

Issue: https://forum.portfolio-performance.info/t/zuordnung-verkauf-zu-kauf-fuer-trades-nach-umbuchung/9404/10

A file with an example for each problem is given in the forum.


buchen:
Thanks. That was a tricky one to fix.

PR #1900, opened 2020-12-21
Fix JSON quote feed to retrieve latest quote

inv-trad:
Issue: https://forum.portfolio-performance.info/t/json-kursquelle-vermischt-parameter/9639/4


buchen (approved):
Merged. Thanks 😄 

And as a side remark: oh boy, the AbstractQuoteProviderPage eventually will need some refactoring. Over the years, I have added so many if and else that I am surprised that anybody can actually understand the behavior anymore...

I slightly changed your implementation. Instead of "instanceof HistoricalQuoteProviderPage" I used the abstract method #getJSONDatePropertyName which is to be implemented by both sub-classes. I thought it keeps it in line with the #getFeedURL method which is also overwritten by sub-classes.

inv-trad:
Thanks, Andreas (@buchen). 

It's funny, because I first went a simliar path to the one you chose in the end, but didn't really manage to get it done to the end... The way you solved it is for sure the more elegant way.

PR #1909, opened 2020-12-25
Fix Fifo calculation in statement of assets for multiple portfolios

inv-trad:
Issue: https://forum.portfolio-performance.info/t/einstandskurs-und-preis-in-der-vermoegensaufstellung-bei-einem-wertpapier-in-zwei-depots-nicht-korrekt/12042

Fifo was done per security not taking into consideration buy and sell
activities per portfolio. The fix calculates the fifo prices per
portfolio and sums them up.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/SecurityPosition.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/SecurityPosition.java
@@ -11,6 +11,7 @@ import java.util.stream.Collectors;
 
 import name.abuchen.portfolio.model.Classification;
 import name.abuchen.portfolio.model.InvestmentVehicle;
+import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.PortfolioTransaction;
 import name.abuchen.portfolio.model.PortfolioTransaction.Type;
 import name.abuchen.portfolio.model.Security;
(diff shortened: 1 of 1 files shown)


buchen:
@inv-trad That is a tricky one. There are two cost calculations. For historical reasons, there also was also one in SecurityPosition. It is not easy to fix, because you need the Portfolio (e.g. the owner of the transaction), but by merging the PortfolioSnapshots, that information get's lost.

The past 2 hours I played around how to best pass the context to the SecurityPosition and allow it to reuse the CostCalculation. The longer I think about it, it might be better to remove this calculation from the SecurityPosition completely, and have the UIs use the SecurityPerformanceRecord instead. I will play around with this option now.

inv-trad:
@buchen Thanks, Andreas. I didn't see your comment until now and was trying to figure out, how to get the portfolio information passed to SecurityPosition as well... 
In case of buy and sell transactions there is the CrossEntry that contains portfolio information, but in case of a delivery that information doesn't exist and the logic breaks.

You know the project/coding and it's history/legacy much better (historisch gewachsen... 😉) and hope, you'll figure out, how to get this one solved. Let me know, if and/or how I can help.

Frohe Weihnachten!

buchen:
So here is the change. I kicked out the duplicate calculation of the FIFO values and replaced it with the one in the SecurityPerformanceRecord. While updating the test cases, I also found some more rounding errors.

I also compared the results by opening my personal file in the currently released version on this HEAD version. So far it all looks okay. If you noticed strange values and/or behavior, please let me know.

Ebenfalls frohe Weihnachten! And thank you very, very much for all the contributions. I was very busy at work and wasn't able to do much for PP...

PR #1911, opened 2020-12-26
Increase valid formatting range from 6 to 8

suriyaa:
Servus, @buchen!

I increased the formatting range in [`FormattingPreferencePage.java`](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/FormattingPreferencePage.java) from 6 to 8. This allows Portfolio Performance users to use Bitcoin and other crypto currencies to the max. allowed "Nachkommastelle" / decimal point.

Cheers,
Suriyaa

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/FormattingPreferencePage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/FormattingPreferencePage.java
@@ -19,7 +19,7 @@ public class FormattingPreferencePage extends FieldEditorPreferencePage
     {
         IntegerFieldEditor sharesPrecisionEditor = new IntegerFieldEditor(
                         Preferences.FORMAT_SHARES_DIGITS, Messages.PrefLabelSharesDigits, getFieldEditorParent(), 1);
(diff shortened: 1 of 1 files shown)


suriyaa:
Related forum posts:
- https://forum.portfolio-performance.info/t/bitcoin-nachkommastellen/1667/2
- https://forum.portfolio-performance.info/t/anzahl-der-nachkommastellen-bei-buchungen/4660/15
- https://forum.portfolio-performance.info/t/rundungsfehler-durch-begrenzung-der-nachkommastellen/777
- https://forum.portfolio-performance.info/t/p2p-kredite-bei-portfolio-performance/142/20
- https://forum.portfolio-performance.info/t/bitcoin-nachkommastellen/1667

chirlu:
You can’t enter nor store more than six decimal places, without a lot of more invasive changes all over the place. So, with your proposal, the only thing that would change is that “00” would be appended to some numbers.

buchen:
As @chirlu noted, just extending the formatting does not work. Internally, PP is calculating with - and always rounding to - 6 digits. There is more work in the backend needed.

PR #1912, opened 2020-12-27
Fix DAB/Smartbroker PDF import

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-smartbroker/7684/78

And a bit of regex clean up using \\\\. instead of just . and removing the + after \\\\w{3}

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
@@ -78,9 +78,9 @@ public class DABPDFExtractor extends AbstractPDFExtractor
                         })
 
                         .section("amount", "currency", "exchangeRate", "forex", "forexCurrency", "curr").optional() //
-                        .find(".* (Ausmachender Betrag|Kurswert) (?<forexCurrency>\\w{3}) (?<forex>[\\d\\.]+,\\d+)-")
+                        .find("^Handel.* (Ausmachender Betrag|Kurswert) (?<forexCurrency>\\w{3}) (?<forex>[\\d\\.]+,\\d+)-")
                         .find("Wert Konto-Nr. Devisenkurs Betrag zu Ihren Lasten")
-                        .match("^(\\d+\\.\\d+\\.\\d{4}) ([0-9]*) (?<curr>\\w{3})/\\w{3} (?<exchangeRate>[\\d\\.]+,\\d+) (?<currency>\\w{3}) (?<amount>[\\d\\.]+,\\d+)$")
+                        .match("^(\\d+\\.\\d+\\.\\d{4}) ([0-9]*) (?<curr>\\w{3})\\/\\w{3} (?<exchangeRate>[\\d\\.]+,\\d+) (?<currency>\\w{3}) (?<amount>[\\d\\.]+,\\d+)$")
                         .assign((t, v) -> {
 
                             Money amount = Money.of(asCurrencyCode(v.get("currency")), asAmount(v.get("amount")));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -60,47 +60,55 @@ private void addBuyTransaction()
                         .section("isin", "name", "currency") //
                         .find("Gattungsbezeichnung ISIN") //
                         .match("^(?<name>.*) (?<isin>[^ ]*)$")
-                        .match("STK [\\d.]+(,\\d+)? (?<currency>\\w{3}+) ([\\d.]+,\\d+)$")
+                        .match("STK [\\d\\.]+(,\\d+)? (?<currency>\\w{3}) ([\\d\\.]+,\\d+)$")
```
chirlu, line 63: Inside character classes, `.` is not special, so `"[\\d.]"` (as it was) already does what it should do, i.e. match any digit or the dot (but no other characters).

The `+` in `x{3}+` turns the repetition into a possessive quantifier. I don’t think this will change the meaning in any way in this particular case, but removing the `+` might slow down the matching slightly.

In short, I think you should leave those as they were.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -60,47 +60,55 @@ private void addBuyTransaction()
...
-                        .match("^STK (?<shares>[\\d.]+(,\\d+)?) (\\w{3}+) ([\\d.]+,\\d+)$")
+                        .match("^STK (?<shares>[\\d\\.]+(,\\d+)?) (\\w{3}) ([\\d\\.]+,\\d+)$")
                         .assign((t, v) -> t.setShares(asShares(v.get("shares"))))
 
                         .section("amount", "currency") //
                         .optional() //
-                        .find("Wert Konto-Nr. Betrag zu Ihren Lasten")
-                        .match("^(\\d+.\\d+.\\d{4}+) ([0-9]*) (?<currency>\\w{3}+) (?<amount>[\\d.]+,\\d+)$")
+                        .find("Wert Konto-Nr\\. Betrag zu Ihren Lasten")
+                        .match("^(\\d+\\.\\d+\\.\\d{4}) ([0-9]*) (?<currency>\\w{3}) (?<amount>[\\d\\.]+,\\d+)$")
```
chirlu, line 74: It’s different here with the first few dots, because here they are used outside of character classes. An unescaped dot there matches any character, so the old date pattern would match `27X12Y2020`, too. So, changing it here is correct (but not in the “amount” part, where it is again in a character class).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
buchen: > The ```+``` in ```x{3}+``` turns the repetition into a possessive quantifier. I don’t think this will change the meaning in any way in this particular case, but removing the ```+``` might slow down the matching slightly.

I did not know about the impact of ```+```. Interesting!
Either way, I think speed is not the problem here (if feels the conversion of the PDF to text uses all the time).
I agree on your first point.

buchen:
rebased. squashed. Merged.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
inv-trad: Thanks, Ulrich. The timing is funny though, because just today I was reading some more about the performance of regex and how to speed it up. You are right with the `+` and its impact on performance here. In future, I will try to stick to what is suggested here: https://www.loggly.com/blog/five-invaluable-techniques-to-improve-regex-performance/

buchen: Nice link. thanks.

Issue #1876, opened 2020-12-18
UI: Hide unused currencies in submenu in dropdown


chrisaut:
**Is your feature request related to a problem? Please describe.**
When you click the dropdown to switch currencies (eg. under statement of assets) it brings up a very long dropdown menu with all currencies the system knows about. Currencies that have been used are nicely grouped at the top, however since the list is so long it still looks weird. It is especially noticeable on a 4k screen where even if the app itself is not in fullscreen suddenly a menu taking up the entire height of the screen (and in fact more as you still have to scroll) shows up.

**Describe the solution you'd like**
I suggest we put the rest of the currencies into a submenu in the dropdown, this way they are still reachable if you want to know your value in Kenyan Schillings for some reason, but it doesn't clutter up the UI. I would bet the vast majority of people never need to access any currency that they don't at least have a single security in.

**Describe alternatives you've considered**
None

**Additional context**
Example from the sample kommer.xml:
![image](https://user-images.githubusercontent.com/8252355/102600843-a8664b80-411f-11eb-9850-cf4147b82a43.png)
Here CAD and EUR would show up top level, while the rest below the seperator line would go into a submenu called "Other..."

buchen:
I agree that current drop down could be improved. 😄 

Also, it would be great to have one solution for all the places where the currency is selected
* Statement of Assets for (reporting currency)
* Security Detail Dialog (currency of security prices)
* Transaction dialog (currency of dividends)

There are multiple options - from cheap to expensive.

### Sub-Menu

I like your idea about the sub-menus. But maybe not all into one sub-menu, but also grouped:

* EUR
* CAD
* xxxxxxxxxx
* A-D
  * AED (Vereinigte Arabaische...)
  * AFN
  * ...
* E-I
* J-M
* N-R
* S-V
* W-Z

### Popup Search Box

I always dreamt of a popup search box - similarly to Ctrl-G - but it shows you the most popular currencies and you can go and easily search for others. But that is a custom implementation. I think the sub-menu version already improves it by quite a lot.

Do you want to create a PR? Something re-usable in the "ui.util" package that can be used by all code locations?

chrisaut:
> ### Sub-Menu
> I like your idea about the sub-menus. But maybe not all into one sub-menu, but also grouped:
> 
> * EUR
> * CAD
> * xxxxxxxxxx
> * A-D
>   
>   * AED (Vereinigte Arabaische...)
>   * AFN
>   * ...
> * E-I
[...]

PR #1915, opened 2020-12-29
Allow user to select an explicit currency for dividends

chrisaut:
Allows the user to select a different currency for dividend payments, different from either security currency and/or account currency.

image

Example use case:
US stocks bought (and tracked) in EUR at a German exchange, but Dividend gets paid in USD.

I'm not sure if this is done yet, putting it up in the hope for a quick review if the approach is acceptable.

[Some Forum discussion](https://forum.portfolio-performance.info/t/jaehrliche-dividendensteigerung-auswerten/12589)


chirlu:
The custom drop-down menu doesn’t fit on my screen, and there is no way to scroll. If it is to stay at this place, it should be an actual drop-down (showing only the ISO codes), like when changing the currency of an account. But I would prefer a separate line with a label like “payment currency” or “dividend currency”. This would also provide enough space for a full currency drop-down, like in the dialogue for editing a security.

Another remark: There are lots of lines with dangling spaces or tabs at the end.

chrisaut:
> The custom drop-down menu doesn’t fit on my screen, and there is no way to scroll. If it is to stay at this place, it should be an actual drop-down (showing only the ISO codes), like when changing the currency of an account. 

Well it's the same dropdown as under statement of assets. I don't particularly like it either (see #1876 ) but I still think it's better then the currency selection dropdown style IMO because at least it has the option of seperators for previously used currencies. I guess we could extend the other dropdowns with that too though. If other's disagree I can change it though, I'm not stuck on it.

I assume you mean this style?
image

There is also this currency selection:
image

I kinda think we should unify currency selection though, create a reusable component that looks the same everywhere and is easier to use. Try switching from EUR to USD quickly in the dropdown under accounts for example, it's a usability nightmare. Don't think this should be part of this though.

> But I would prefer a separate line with a label like “payment currency” or “dividend currency”. This would also provide enough space for a full currency drop-down, like in the dialogue for editing a security.

I can put it on a new line, just thought it would waste space. Do others agree?

> Another remark: There are lots of lines with dangling spaces or tabs at the end.

Yeah good point, will clean it up with next commit.

buchen:
I would also propose to separate the two things:
* First: allow to record dividends in a currency other than the security currency.
* Second: have a better usable drop-down for selecting currencies

I propose to make this PR about the first topic.

Allowing other currencies makes sense. Particularly because the currency determines both: the currency of the security prices downloaded and the currency of the dividend statements. But one wants to be able to separates this: trade a security on a EUR exchange and still receive USD dividends.

The change of the UI, however, is only one part. I do not know for sure, but so far in many calculations I assume that the currency fits to the currency of the security. To merge the change, I would like to see a test case (similar to what you find in the n.a.p.tests/src/scenarios folder) that tests ClientSnapshot, ClientPerformanceSnapshot, and SecurityPerformanceSnapshot procure the results you expect.


> I can put it on a new line, just thought it would waste space. Do others agree?

I think most of the time users will not touch the currency. I kind of like it right next to the "per share" item.

chirlu:
> Well it's the same dropdown as under statement of assets.

Oh, indeed, that one is unusable in the same way. :-) (But I never use it, so I never noticed.) I‘ve just created issue #1920 for that.

> one wants to be able to separates this: trade a security on a EUR exchange and still receive USD dividends.

As an aside, I own some shares that used to pay dividends in DEM once upon a time and now pay in EUR. So that is a use case, too.

chrisaut:
> I would like to see a test case (similar to what you find in the n.a.p.tests/src/scenarios folder) that tests ClientSnapshot, ClientPerformanceSnapshot, and SecurityPerformanceSnapshot procure the results you expect.

Ok good, I will add tests and see if things break

> > I can put it on a new line, just thought it would waste space. Do others agree?
> 
> I think most of the time users will not touch the currency. I kind of like it right next to the "per share" item.

Ok, let's leave it where it is then

tomasherman:
Hey everyone, i was wondering it this PR is going to be worked on further or merged :) Maybe one thing to consider would be adding `Dividend currency` as a paramter of `Security` (with default set to currency used to denominate the currency). So that the user can set it once when creating the `security` and then the dialog pics the current currency. What do  you think?

chrisaut:
I do still want this feature, very much so in fact, nearly all my dividends come from US but paid in EUR. If I remember correctly I got a bit stuck trying to setup a test context that would be able to exercise all the new cases. Then I unfortunately got sidetracked and busy. From the manual testing I had done it did seem to work though.

So, perhaps someone can take this as a starting point, or inspiration and finish it. If not I *may* pick it up sometime over the summer, but I just can't promise when.

tomasherman:
i wanted to pick this up but having a really rough time with eclipse (intellij guy here :D) ... it keeps crashing on me, im not sure i will be able to get at it properly, it drives me crazy

bbatsov:
Is the project built using Eclipse RCP? Haven't done any Java in ages, but I remember this was the last reason I had to use Eclipse for anything. :D

PR #1917, opened 2020-12-29
Reset exchange rate when currencies coincide

chirlu:
When currencies in a transaction coincide, controls relating to currency conversion are automatically hidden. In this case, the exchange rate is necessarily 1. However, when an existing transaction was edited to make the currencies coincide, the (invisible) exchange rate remained and still influenced the transaction. It did work when a new transaction was created. The bug was introduced by commit d0fdc85c14e1b42e66bea4fc333ec35609993730.

To fix this, make sure that the exchange rate is reset to 1 even when editing an existing transaction.

A similar issue, where hiding the forex controls was caused by deselecting an optional security, was described on the forums in https://forum.portfolio-performance.info/t/11094 and already fixed in commit 702212e15998260e3e9f4810eb8f60e9ad4a39b3.


buchen:
Thanks for the fix. These details make all the difference.

Issue #1918, opened 2020-12-29
Onvista PDF Import of Vorabpauschale: Error Zero Taxes Apply


fabtagon:
**Describe the bug**
When trying to import an Onvista Vorabpauschale PDF where **all** taxes are covered by Freistellungauftrag or Verlustopf (see supplied test case), import fails as no import converter recognises the data in that file.

**To Reproduce**
Try to import such a PDF.

**Expected behavior**
Import shall succeed (no error; no taxes added - basically that's a NOP)

**Desktop (please complete the following information):**
 - OS: Linux amd64
 - Version: 0.49.4; master branch

**Problem Details:**
Onvista Vorabpauschale PDF import relies on three sections to be present in PDF:
1. "Gattungsbezeichnung ISIN"
1. "Wert Konto-Nr. Betrag zu Ihren Lasten" (optional)
1. "Nominal Ex-Tag Zahltag Jahreswert Vorabpauschale pro Stück"

In case of Freistellungauftrag/Verlustopf-covered taxes, 2. is not present. Thus, transaction amount is zero and currency is null, leading to the transaction being considered non-present. As a result, the PDF import decides that this file is no Onvista Vorabpauschale PDF. As no other import converter recognises that file either, the file is considered not supported at all by PP.

**Solution:**
Pull request along with test is en route.

fabtagon:
Works - nothing to do anymore here

PR #1919, opened 2020-12-29
Allow zero tax for Onvista Import of Vorabpauschale PDF

fabtagon:
See description in issue #1918

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OnvistaPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OnvistaPDFExtractor.java
@@ -1782,12 +1782,16 @@ public class OnvistaPDFExtractor extends AbstractPDFExtractor
                         .section("date").optional()
                         .match("^Nominal Ex-Tag Zahltag Jahreswert Vorabpauschale pro Stück")
                         // STK 1.212,000 02.01.2020 02.01.2020 EUR 0,1019
-                        .match("STK .* (?<exdate>\\d+\\.\\d+\\.\\d{4}+) (?<date>\\d+\\.\\d+\\.\\d{4}+) (\\w{3}+) .*")
+                        .match("STK .* (?<exdate>\\d+\\.\\d+\\.\\d{4}+) (?<date>\\d+\\.\\d+\\.\\d{4}+) (?<currency>\\w{3}+) .*")
                         .assign((t, v) -> {
+                            // if all taxes are covered by Freistellungauftrag/Verlustopf, section "Wert Konto-Nr. Betrag zu
+                            // Ihren Lasten" is not present -> extract currency here
+                            if (t.getCurrencyCode() == null)
+                                t.setCurrencyCode(asCurrencyCode(v.get("currency")));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OnvistaPDFExtractor.java
```diff
@@ -1782,12 +1782,16 @@ private void addTaxVorabpauschaleTransaction()
...
                         .assign((t, v) -> {
+                            // if all taxes are covered by Freistellungauftrag/Verlustopf, section "Wert Konto-Nr. Betrag zu
+                            // Ihren Lasten" is not present -> extract currency here
+                            if (t.getCurrencyCode() == null)
+                                t.setCurrencyCode(asCurrencyCode(v.get("currency")));
                             t.setDateTime(asDate(v.get("date")));
                         })
 
-                        .wrap(t -> t.getAmount() != 0 ? new TransactionItem(t) : null);
+                        .wrap(TransactionItem::new);
```
chirlu, line 1794: Won’t this create a transaction even if the amount is 0?

fabtagon: Yes. The amount of 0.00 € paid taxes does exactly represent what is in the PDF (so there's nothing to debit from the reference account). My impression is that at least one transaction (even if it is somewhat empty) is required for the importer to recognise this as a proper Onvista file. I'm open for advise on how else to achieve that ;-)

Ragas13: But what is the benefit to create a zero amount booking especially if they will never be editable via the PP interface? :thi

fabtagon: Benefit is that it fixes the bug described in #1918 (failure to import PDF file)

chirlu: > My impression is that at least one transaction (even if it is somewhat empty) is required for the importer to recognise this as a proper Onvista file. I'm open for advise on how else to achieve that ;-)

If that’s currently the case, it could be changed by modifying the importer framework.

buchen: Last year, I had a similar problem with the Consorsbank importer. For this, I created a ```Extractor.NonImportableItem```. This way, PP does not import a transactions that cannot be edited / used later on but at the same time the importer does not fail.

https://github.com/buchen/portfolio/blob/815a0d2f248f4faa1a5508775a6642c03b856ae8/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java#L543

buchen:
Thanks @fabtagon for the contribution.

I have update the code to use the NonImportableItem and merged it into master.

PR #1928, opened 2021-01-02
Use Macedonia's new official name 'North Macedonia'

fabtagon:
From 12 February 2019 on, the official name of the state of Macedonia has changed from 'The former Yugoslav Republic of Macedonia' to 'North Macedonia'; cf. treaty text at https://vlada.mk/node/15057?ln=en-gb


buchen:
Thanks!

PR #1932, opened 2021-01-03
Feature 8digits precision for shares and quotes

buchen:
This pull request increases the precision for shares and quotes to 8 digits.

The relevant change is that while the data is still kept as ```long``` this change is now using ```BigDecimal``` for the calculations, for example in ```SecurityPosition``` to calculate the market value of a position.

I also tested converting the shares to always use ```BigDecimal```. I decided against it, because from my rough performance tests it showed that it slows down the calculation further (the change in this PR adds up to 10% already). There are a lot of additions and subtractions of shares that are apparently faster if done as ```long```. Plus, most of the code is more readable when working with long.

The last commit uses ```BigDecimal#movePointLeft``` instead of dividing the values to get the correct fractional amount. In this case my performance tests are inconclusive: sometimes it is faster, sometimes the overall calculation is slower.

And finally, I tested whether it makes sense to store the quotes as fractional amounts in the XML file. By adding a lot of trailing zeros, the file size increases quite a lot (although only if the file is stored as an uncompressed XML). In my tests, reading the file was slower however if I had to convert the values into long. Therefore I left it as is for now.

To get some idea on the performance impact, I calculated these three snapshots for a 5 year period on a set of 5 XML files with a file size between 10 and 60 MBytes.
```
SecurityPerformanceSnapshot snapshot1 = SecurityPerformanceSnapshot.create(client, converter, interval);
ClientSnapshot snapshot2 = ClientSnapshot.create(client, converter, interval.getStart());
ClientPerformanceSnapshot snapshot3 = new ClientPerformanceSnapshot(client, converter, interval);
```

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/money/Values.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/money/Values.java
@@ -1,6 +1,8 @@
 package name.abuchen.portfolio.money;
 
 import java.math.BigDecimal;
+import java.math.MathContext;
+import java.math.RoundingMode;
 import java.text.DecimalFormat;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
@@ -11,6 +13,8 @@ import java.util.Locale;
 
 public abstract class Values<E>
 {
+    public static final MathContext MC = new MathContext(10, RoundingMode.HALF_UP);
+
     public static final class MoneyValues extends Values<Money>
     {
         private MoneyValues()
@@ -58,9 +62,13 @@ public abstract class Values<E>
         private static final ThreadLocal<DecimalFormat> QUOTE_FORMAT = ThreadLocal // NOSONAR
                         .withInitial(() -> new DecimalFormat(QUOTE_PATTERN));
 
+        private final BigDecimal factorToMoney;
+
         private QuoteValues()
         {
             super(QUOTE_PATTERN, 10000D, 10000);
+
+            factorToMoney = BigDecimal.valueOf(factor() / Values.Money.factor());
         }
 
         @Override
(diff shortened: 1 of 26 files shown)


buchen:
If you have comments, ideas, suggestions, I am happy to get your feedback. Also if you see impact on your files.

Please note, that if you store your file with this version, it cannot be opened with a previous version of PP. I typically work/test with a copy therefore.

name.abuchen.portfolio/src/name/abuchen/portfolio/money/Values.java
```diff
@@ -320,33 +340,44 @@ public String format(Integer amount)
...
     private final BigDecimal bdFactor;
 
-    private Values(String pattern, double divider, int factor)
+    private Values(String pattern, int factor)
     {
         this.pattern = pattern;
-        this.divider = divider;
         this.factor = factor;
+        this.divider = factor;
+        this.precision = (int) Math.log10(divider);
```
chirlu, line 353: Isn’t there (in principle) a risk that the `log10` might return something like 7.99999999, which is then rounded to 7 by the cast? Adding, e.g., 0.1 before casting might be safer.

chirlu (commented):
Personally, I am fine with the current limits, but it seems they aren’t sufficient for Bitcoin, so why not …

name.abuchen.portfolio/src/name/abuchen/portfolio/money/Values.java
buchen: Thanks. I turned it around now. I configure the precision, and then calculate the factor with ```BigInteger.TEN.pow(precision).intValue()```

PR #1938, opened 2021-01-07
Fix BuySellEntry cannot be cast to class PortfolioTransferEntry

inv-trad:
Issue: https://github.com/buchen/portfolio/issues/1931

and https://forum.portfolio-performance.info/t/fehler-buysellentry-cannot-be-cast-to-class-portfoliotransferentry/12778/9

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/filter/ClientSecurityFilter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/filter/ClientSecurityFilter.java
@@ -178,6 +178,10 @@ public class ClientSecurityFilter implements ClientFilter
     private void convertTransfer(Function<Portfolio, ReadOnlyPortfolio> getPortfolio,
                     TransactionPair<PortfolioTransaction> pair)
     {
+        // in rare cases (wrong imports) the cross entry can be of type BuySellEntry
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/filter/ClientSecurityFilter.java
```diff
@@ -178,6 +178,10 @@ private PortfolioTransaction convertToDelivery(PortfolioTransaction t, Portfolio
     private void convertTransfer(Function<Portfolio, ReadOnlyPortfolio> getPortfolio,
                     TransactionPair<PortfolioTransaction> pair)
     {
+        // in rare cases (wrong imports) the cross entry can be of type BuySellEntry
+        if (!(pair.getTransaction().getCrossEntry() instanceof PortfolioTransferEntry))
+            return;
+
```
chirlu, line 184: I’d advise against hiding bad data, which might cause other issues down the line. When people see it failing, they can fix the transaction; the issue doesn’t seem to be widespread.

But +1 for fixing the importer.

buchen: As mentioned in #1931, either let's write a ```Check``` that fixes the data or - if the user can fix it himself - then ask the user to fix it. Independently, we need to fix the importer.

PR #1953, opened 2021-01-09
Support transaction import for securities with different currencies

inv-trad:
Issue: https://forum.portfolio-performance.info/t/buchungen-importieren-fuer-doppelt-angelegtes-wertpapier-mit-unterschiedlichen-waehrungen/13301

Im Forum kam schon mehrfach die Frage auf, warum ein Import für Dokumente mit gleichen Wertpapieren aber unterschiedlichen Währungen nicht auch funktioniert. Hierzu diese Erweiterung als Vorschlag.

Und: Der hinzugefügte Test `testThatSecurityIsMatchedByISINandName` lief erst in einen Fehler, weil die Exception for der `lookupSecurityByName` geworfen wurde. Deswegen habe ich den Lookup nach vorne gezogen.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/SecurityCacheTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/SecurityCacheTest.java
@@ -24,7 +24,31 @@ public class SecurityCacheTest
         security.setName("Security Name");
         security.setIsin("DE0007164600");
         security.setWkn("716460");
+        security.setCurrencyCode("EUR");
         client.addSecurity(security);
+
+        security = new Security();
+        security.setName("Security Name USD");
+        security.setIsin("DE0007164600");
+        security.setCurrencyCode("USD");
+        client.addSecurity(security);
+
+        for (int i = 0; i < 500; i++)
+        {
+            security = new Security();
+            security.setName("Security Name" + i);
+            security.setIsin("DE00071646" + i);
+            security.setCurrencyCode("USD");
+            client.addSecurity(security);
+        }
+        for (int i = 0; i < 500; i++)
+        {
+            security = new Security();
+            security.setName("Security Name USD" + i);
+            security.setIsin("DE00071646" + i);
+            security.setCurrencyCode("USD");
+            client.addSecurity(security);
+        }
     }
 
     @Test
@@ -49,5 +73,38 @@ public class SecurityCacheTest
         SecurityCache cache = new SecurityCache(client);
         Security lookup = cache.lookup(null, null, null, "Security Name", () -> new Security());
         assertThat(client.getSecurities().get(0), is(lookup));
+    }
+
+    @Test
+    public void testThatSecurityIsMatchedByISINUnique()
+    {
+        SecurityCache cache = new SecurityCache(client);
+        try
+        {
+            cache.lookup("DE0007164600", null, null, null, null, () -> new Security());
+            assertThat(true, is(false));
+        }
+        catch (IllegalArgumentException e)
+        {
+            assertThat(true, is(true));
+        }
+    }
+
+    @Test
+    public void testThatSecurityIsMatchedByISINandName()
+    {
+        SecurityCache cache = new SecurityCache(client);
+        Security lookup = cache.lookup("DE0007164600", null, null, "Security Name", () -> new Security());
+        assertThat(client.getSecurities().get(0), is(lookup));
+        assertThat(client.getSecurities().get(1), not(is(lookup)));
+    }
+
+    @Test
+    public void testThatSecurityIsMatchedByISINandCurrency()
+    {
+        SecurityCache cache = new SecurityCache(client);
+        Security lookup = cache.lookup("DE0007164600", null, null, null, "USD", () -> new Security());
+        assertThat(client.getSecurities().get(0), not(is(lookup)));
+        assertThat(client.getSecurities().get(1), is(lookup));
     }
 }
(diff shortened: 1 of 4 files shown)


inv-trad:
Sehe gerade, dass CSVSecurityExtractorTest einen Fehler produziert. Schaue mir das gleich noch mal an.

inv-trad:
@buchen BTW: gibt es eine Möglichkeit alle Tests einfach vorab lokal in Eclipse zu starten?

inv-trad:
Eine offene Frage für mich ist, ob dann die PDF Importer noch so angepasst werden sollten, dass die Währung nach der gesucht wird, immer der Transaktionswährung entspricht. Den Umbau des Consorsbank Importers habe ich hinzugefügt.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
```diff
@@ -325,16 +325,12 @@ public void testErtragsgutschrift10() throws IOException
         assertThat(t.getSecurity().getName(), is("OMNICOM GROUP INC. Registered Shares DL -,15"));
         assertThat(t.getSecurity().getIsin(), is("US6819191064"));
         assertThat(t.getSecurity().getWkn(), is("871706"));
-        assertThat(t.getSecurity().getCurrencyCode(), is(CurrencyUnit.USD));
+        assertThat(t.getSecurity().getCurrencyCode(), is(CurrencyUnit.EUR));
```
buchen, line 328: I am not sure why we should change the currency of the created security from USD to EUR.

If you look at the extracted text, it is clear that the security is in USD:

```
OMNICOM GROUP INC. Registered Shares DL -,15 871706 US6819191064
Bestand
25 Stück
Dividende pro Stück 0,60 USD Schlusstag 17.12.2017
Brutto in USD 15,00 USD
```

buchen (commented):
From my point of view, we should change only one the two things:

Yes, it can make sense to improve the matching of securities. If multiple securities are found by the same ISIN, make a decision based on the currency.

No, I think we should not change the how the security is extracted from PDF. If the PDF indicates, it is in USD and it includes exchange rate information, we should not throw it away.


Now I am thinking how to reconcile this change with #1915 by @chrisaut. The latter is proposing that the currency of dividends can be different to the currency of the security. That would mean you can have USD dividends for a EUR security (where the security is actually the currency of the prices - in which currency the security is traded. I wonder if this change "picking the security based on a security" then still makes sense.

I wonder if we should evolve PP in this direction:
* the *currency* of the security indicates only the currency of the prices - I need this because most source of historical prices do not include currency information
* enable dividend payments in arbitrary currency
* enable purchase/sales in arbitrary currency as well (one could by at different exchange)

If we have his, then the answer to the user in the forum is:
* just have *one* security
* in one securities account you trade in USD, in the other in EUR
* upon import, the dividends are attached to this one security (of course, choose different cash accounts based on the type of the document - one bank, other bank)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
inv-trad: I rolled back the changes to the importer and the corresponding test case.

inv-trad:
Thank you for your comprehensive feedback. 

> I wonder if we should evolve PP in this direction:
>   -   the currency of the security indicates only the currency of the prices - I need this because most source of historical prices do not include currency information
>   -   enable dividend payments in arbitrary currency
>   -   enable purchase/sales in arbitrary currency as well (one could by at different exchange)

I agree to those points, because if you have e.g. Apple shares in a Germany portfolio (Depot) and bought it in EUR, you want to see the prices in EUR, but still it might be that - like it is done at Consorsbank - you get the dividends statement showing USD and the calculation to the amount paid in EUR to your account.

With regards to the "one-security-approach", I am not that convinced. As an example, if you have Apple shares in two different portfolios (EUR and USD) you can create two securities (EUR and USD) and have the correct prices and bookings per security and currency in the corresponding portfolio.

As I removed the change to the importer, the current proposal improves the matching 

1. to lookup by name, if e.g. ISIN has a duplicate and the name can resolve this, and 
2. to match by ISIN and currency, if ISIN has a duplicate that can be resolved by currency.

chirlu:
> BTW: gibt es eine Möglichkeit alle Tests einfach vorab lokal in Eclipse zu starten?

Passiert bei mir automatisch. Hast du Infinitest installiert?

chirlu:
> With regards to the "one-security-approach", I am not that convinced. As an example, if you have Apple shares in two different portfolios (EUR and USD) you can create two securities (EUR and USD) and have the correct prices and bookings per security and currency in the corresponding portfolio.

Disadvantages of this approach:
* There is no comprehensive overview of the share’s performance. It’s artificially divided into “USD shares” and “EUR shares”.
* You need to do everything twice: Add two shares to the same classifications, add twice the same links, set up historical quotes twice etc.
* It’s fully possible to buy shares on XETRA, in EUR, then sell those same shares on NYSE, in USD.
* Conversely, you may have shares bought in EUR and shares bought in USD in the same portfolio. They should not be considered separate for the Trades report.

inv-trad:
> Passiert bei mir automatisch. Hast du Infinitest installiert?

Inifinitest hatte ich nicht installiert. Ist jetzt nachgerüstet. Danke.

Issue #1951, opened 2021-01-08
Calculation trail for unrealized gains is wrong in connection with transfer


chirlu:
In certain cases (description and example below), the “calculation trail” tooltip for unrealized capital gains in the “Performance › Calculation” view is wrong. The result shown in the tooltip doesn’t match the value it is supposed to explain, which is correct.

---

Consider the example file contained in [sell-before-transfer.zip](https://github.com/buchen/portfolio/files/5789381/sell-before-transfer.zip). It contains the following transactions for the Allianz stock:
* Buy 10 shares in portfolio A.
* Sell 5 of the shares from portfolio A.
* Transfer the remaining 5 shares from portfolio A to portfolio B.

Now switch to the “Performance › Calculation” view. Under the (unrealized) capital gains, Allianz is listed with gains of 35.65; this is correct. Next, move the mouse pointer to this value, making a tooltip appear. It contains this calculation:
```
08.01.2021 Statement of assets B   5               EUR 1028.75
04.01.2021 Inbound delivery       10  EUR 1986.20
07.01.2021 Transfer from A to B   10               EUR 1986.20
           Difference                                           EUR –957.45
```
The result is obviously wrong (–957.45 instead of 35.65), as is the number of shares supposedly transferred (10 instead of 5). A line saying “5 of 10 shares …” is expected, but missing.

Variations:
* The calculation trail for the realized gains is correct.
* (Most surprising:) When only 2 shares are sold and then 5 transferred (leaving 3 to remain in portfolio A), the calculation trail is correct.
* When sell and transfer are interchanged (first transfer 10 shares, then sell 5 shares from portfolio B), calculation trails both for unrealized and realized gains are correct.

This happens on version 0.49.4 as well as on the current, unreleased, master branch.

PR #1955, opened 2021-01-09
Fix error in trail if whole position is transferred (#1951)

pascalhein:
Even if the entire position in one portfolio is removed,
we should create a "fractional" entry in the trail
in order to determine a correct initial valuation.

This commit fixes the error seen in #1951; I'm not entirely
certain whether it actually covers all edge cases.

(Additionally, minor spelling fix in trail calculation.)

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculation.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculation.java
@@ -146,13 +146,13 @@ import name.abuchen.portfolio.snapshot.trail.TrailRecord;
                                         .convert(forex, convert2forex.getRate(item.date, termCurrency)) //
                                         .convert(back, converter.getRate(t.getDateTime(),
                                                         t.getSecurity().getCurrencyCode()))
-                                        .substract(startTrail);
+                                        .subtract(startTrail);
                     }
 
                     realizedCapitalGains.addCapitalGains(Money.of(termCurrency, end - start));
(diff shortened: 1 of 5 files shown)


buchen:
Thanks. Good one.

chirlu:
Thank you for taking this on!

As an aside, for next time:
> (Additionally, minor spelling fix in trail calculation.)

Writing “Additionally, …” in the description of a change is a foolproof indicator that it is a technically unrelated change that should actually be a separate commit. Mixing unrelated changes in one commit makes it unnecessarily more difficult to read: For every changed line, you need to pause and ask yourself, “Is this just the trivial `s/substr/subtr/`, or is something more happening here?”

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculation.java
```diff
@@ -340,15 +334,15 @@ public void finish(CurrencyConverter converter, List<CalculationLineItem> lineIt
...
             forexGainTrail = forexGainTrail
                             .convert(back, converter.getRate(valuationAtEndDate, getSecurity().getCurrencyCode()))
-                            .substract(startTrail);
+                            .subtract(startTrail);
         }
 
         unrealizedCapitalGains.addCapitalGains(Money.of(termCurrency, end - start));
-        unrealizedCapitalGains.addCapitalGainsTrail(endTrail.substract(startTrail));
+        unrealizedCapitalGains.addCapitalGainsTrail(endTrail.subtract(startTrail));
         unrealizedCapitalGains.addForexCaptialGains(Money.of(termCurrency, forexGain));
```
chirlu, line 346: ;-)

Issue #1904, opened 2020-12-23
Interner Zinsfuß - NaN%


CK1-CK:
https://forum.portfolio-performance.info/t/wertpapier-performance-izf-nan/7919/3

Programmversion 0.49.4
Ansicht Berichte > Performance > Wertpapiere

IFZ als NaN% (Berichtszeitraum 1, 2 oder 3 Jahre)

Zur Verdeutlichung habe ich eine Beispiel XML angehängt. Ich hoffe man kann den Fehler nachvollziehen.
Bug.xml Download:
https://forum.portfolio-performance.info/uploads/short-url/eR5kd7Xh9CTa7AsRumK24QYRHij.xml

Ich kann ausschließen:
- kurze Haltedauer
- Fehlbuchung

Laut Excel müsste ein interner Zinsfuß von -99,x% herauskommen.

Highlife1911:
I can somewhat confirm the Problem as my IRR is too high:
![image](https://user-images.githubusercontent.com/7109398/103768984-c985d200-5023-11eb-80b6-a6f3feecec36.png)

CK1-CK:
> version 0.49.4
> I'm seeing the same issue in Performance => Trades:
> ![imagen](https://user-images.githubusercontent.com/10577978/103685661-e66fc600-4f8d-11eb-8946-2bd022ca019b.png)
> Only 2 trades for that stock, without currency exchange:
> ![imagen](https://user-images.githubusercontent.com/10577978/103685780-1323dd80-4f8e-11eb-8c27-a1b71af403c2.png)

I have the NaN Problem at the following reports

- Ansicht Berichte > Performance > Wertpapiere
- Ansicht Berichte > Performance > Trades

My expectation would be that something with -99.x% would come out. But not NaN

bosix:
I can confirm the bug in the current version. Here a minimum example to reproduce the bug:
![image](https://user-images.githubusercontent.com/26360670/107690732-c9c05f80-6caa-11eb-9b19-f80a1ac5ebc0.png)

The bug also destroys the "Performance-Diagramm" since the overall performance will be -100% forever no matter which trades are done after that.

![image](https://user-images.githubusercontent.com/26360670/107690916-0f7d2800-6cab-11eb-8624-7f7127b4d9a4.png)

chirlu:
@bosix: Your diagram issue is unrelated; the diagram doesn’t even use IRR.

CK1-CK:
[...]

PR #1959, opened 2021-01-10
Perform short pre-optimisation for IRR

pascalhein:
When computing IRR, the NPV curve may be very degenerate (such as in #1904):
curve
In this example, the Newton iteration only converges if the initial guess x0 is to the left of the extreme point (which is at approx. x = -0.9975), but currently, the initial guess is -0.95. (Note that the code actually computes "IRR+1", hence the initial guess x0 = 0.05.)

This commit attempts to find an initial guess in the interval (-1,0) first, via interval halving.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/math/IRR.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/math/IRR.java
@@ -2,6 +2,7 @@ package name.abuchen.portfolio.math;
 
 import java.time.LocalDate;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import name.abuchen.portfolio.math.NewtonGoalSeek.Function;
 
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/math/IRR.java
```diff
@@ -11,10 +12,39 @@ private IRR()
...
+
     public static double calculate(List<LocalDate> dates, List<Double> values)
     {
         Function npv = new NPVFunction(dates, values);
         Function derivative = new PseudoDerivativeFunction(npv);
-        return NewtonGoalSeek.seek(npv, derivative, 0.05d) - 1;
+
+        // find a crude initial guess in interval (0,1)
+        // npv(0) is undefined, but diverges with sign given by last term
+        double fLeft = Double.POSITIVE_INFINITY * values.get(values.size() - 1);
```
chirlu, line 42: For better readability, you could consider using `Math.copySign()`.

name.abuchen.portfolio/src/name/abuchen/portfolio/math/IRR.java
```diff
@@ -11,10 +12,39 @@ private IRR()
...
+        // npv(1) is the sum of undiscounted flows
+        double fRight = values.stream().collect(Collectors.summingDouble(f -> f));
+        // if they have the same sign, we hopefully don't have a very extreme case, so just guess 5%
+        double guess = fLeft * fRight < 0 ? halving(npv, 0, 1, fLeft, fRight) : 1.05;
```
chirlu, line 46: In a similar way, using `Math.signum()` might make the condition more readable.

chirlu (approved):

chirlu:
@buchen: I’d like to remind you of this PR. Any chance it can get in?

CK1-CK:
Unfortunatly the Bug is still present. 

see https://github.com/buchen/portfolio/issues/1904#issuecomment-885721821

PR #1961, opened 2021-01-10
Convert currencyCode combo to a text entry with autocomplete

pascalhein:
Feature Request #1876. Text field suggests matching
currencies first, then all available ones.

What I don't like about this solution is that
* the proposals aren't opened on click
* user can enter anything, requires post-validation
* tab to the next field doesn't work while the popup is open

Hence this is not perfect, but maybe a starting point for
further improvements.


buchen:
I like the proposal.

I did some minor changes:
* The input field is not now only the three digit code, the label displays the currency next to it
* It popups up if you click with the mouse into the field
* It also searches the description

Bildschirmfoto 2021-01-31 um 17 24 20

PR #1968, opened 2021-01-13
Fix Comdirect PDF import for dividends

inv-trad:
Issue: https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/104


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/comdirect/ComdirectPDFExtractorTest.java
```diff
@@ -1483,29 +1483,29 @@ public void testDividendeAusSteuermitteilung3()
     public void testDividendeAusSteuermitteilung4()
     {
         ComdirectPDFExtractor extractor = new ComdirectPDFExtractor(new Client());
-
+    
         List<Exception> errors = new ArrayList<>();
-
+
```
chirlu, line 1488: Whitespace in empty lines (happens 8 times in this file).

inv-trad: Fun fact: this "feature" can be reproduced. I didn't really touch that piece of code, only moved it in the "Outline" view, what seems to create those whitespaces in all empty lines of the block moved.

PR #1975, opened 2021-01-14
Adds quirion Kontoauszüge PDF import capability

christen90:
It works for 90 % of the bookings.

Some work regarding other currency and taxed might need to be fixed. Description of observed bugs can be found within the code.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
@@ -1,0 +1,377 @@
+package name.abuchen.portfolio.datatransfer.pdf;
+
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import name.abuchen.portfolio.datatransfer.Extractor.BuySellEntryItem;
+import name.abuchen.portfolio.datatransfer.Extractor.NonImportableItem;
+import name.abuchen.portfolio.datatransfer.Extractor.TransactionItem;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Block;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.DocumentType;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Transaction;
+import name.abuchen.portfolio.model.AccountTransaction;
+import name.abuchen.portfolio.model.BuySellEntry;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.model.Transaction.Unit;
+import name.abuchen.portfolio.money.Money;
+
+public class QuirionPDFExtractor extends AbstractPDFExtractor
+{
+    public QuirionPDFExtractor(Client client)
+    {
+        super(client);
+
+
+        addBankIdentifier("Quirin Privatbank AG"); //$NON-NLS-1$
+
+        addPeriodenauszugTransactions();
+    }
+
+    @SuppressWarnings("nls")
+    private void addPeriodenauszugTransactions()
+    {
+        final DocumentType type = new DocumentType("(Kontoauszug)", (context, lines) -> {
+            Pattern pCurrency = Pattern.compile("(Vermögensverwaltungskonto in )(\\w{3})\\n");
+            // read the current context here
+            for (String line : lines)
+            {
+                Matcher m = pCurrency.matcher(line);
+                if (m.matches())
+                {
+                    context.put("currency", m.group(1));
+                }
+            }
+        });
+        this.addDocumentTyp(type);
+
+//Kontoübertrag 1197537 28.05.2019 28.05.2019 3.000,00 EUR
+//Ref.: 86991330
+        
+//Sammelgutschrift 19.12.2019 19.12.2019 5.000,00 EUR
+//DE00000000000000000000
+//Ref.: 105892216
+//AWV-Meldepflicht beachten - Tel.-Nr.08001234111
+        
+//Interne Buchung 31.01.2020 31.01.2020 2,84 EUR
+//Ref.: 110934428
+//Kulanzzahlung
+        
+//Überweisungsgutschrift Inland 27.12.2019 27.12.2019 2.000,00 EUR
+//Ref.: 106509889
+//Konto Inhaber DE00000000000000000000
+//Verwendungszweck
+        
+        Block depositBlock = new Block(
+                        "^((Kontoübertrag (\\d+))|(Sammelgutschrift)|(Interne Buchung)|(Überweisungsgutschrift Inland)) (\\d+\\.\\d+\\.\\d{4}) (\\d+\\.\\d+\\.\\d{4}) ([\\d.]+,\\d{2}) (\\w{3})");
+        type.addBlock(depositBlock);
+        depositBlock.set(new Transaction<AccountTransaction>().subject(() -> {
+            AccountTransaction t = new AccountTransaction();
+            t.setType(AccountTransaction.Type.DEPOSIT);
+            return t;
+        })
+
+                        .section("valuta", "amount")
+                        .match("^((Kontoübertrag (\\d+))|(Sammelgutschrift)|(Interne Buchung)|(Überweisungsgutschrift Inland)) (\\d+\\.\\d+\\.\\d{4}) (?<valuta>\\d+\\.\\d+\\.\\d{4}) (?<amount>[\\d.]+,\\d{2}) (\\w{3})")
+                        .assign((t, v) -> {
+                            Map<String, String> context = type.getCurrentContext();
+                            t.setCurrencyCode(asCurrencyCode(context.get("currency")));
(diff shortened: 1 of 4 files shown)


buchen:
Nice @christen90 - thanks for adding the importer

Can you add some test cases?
https://github.com/buchen/portfolio/tree/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf

They all work after the same principle: load an anonymized extracted text and see if the expected transactions are extracted.

Without test cases, it is all to easy to break these importer when I have to introduce changes

christen90:
@buchen - thanks for shareing your project code. It would be a nightmare to program it from scratch.

I hope this update meets your expectations regarding test cases :)
The known bugs from the importer are prepared for activating via toggle comment. Let me know if you find the problem, why sellBlocks without taxes interfere with dividendBlock with taxes.

BR Christian

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
```diff
@@ -0,0 +1,377 @@
...
+                                            //.subtract(t.getMonetaryAmount()).toMoney();
+                                          t.addUnit(new Unit(Unit.Type.TAX, taxes));})
+                        .wrap(t -> new TransactionItem(t)));
+        
+
+//Wertpapier Kauf, Ref.: 85249245 08.05.2019 10.05.2019 -235,86 EUR
+//SPDR MSCI Wrld Small Cap U.ETF Registered Shares o.N.
+//IE00BCBJG560, ST 3,648
+        
+        Block sellBlock = new Block("(Wertpapier Verkauf, Ref.: \\d+) (\\d+.\\d+.\\d{4}+) (\\d+.\\d+.\\d{4}+) ([\\d.]+,\\d{2}) (\\w{3}).*");
```
inv-trad, line 306: Maybe adding an "endsWith" string could help here:
`
Block sellBlock = new Block("(Wertpapier Verkauf, Ref.: \\d+) (\\d+.\\d+.\\d{4}+) (\\d+.\\d+.\\d{4}+) ([\\d.]+,\\d{2}) (\\w{3}).*", ".* Ref\\.: .*");`

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
```diff
@@ -0,0 +1,377 @@
...
+                        .assign((t, v) -> {//t.addUnit(new Unit(Unit.Type.TAX, asAmount(v.get("tax"))));
+                                            Money taxes = Money.of(asCurrencyCode("EUR"),
+  //TODO other currencys not working                         Money.of(asCurrencyCode(v.get("currency")), asAmount(v.get("tax"))))))
+                                                            asAmount(v.get("tax")));
+                                            //.subtract(t.getMonetaryAmount()).toMoney();
+                                          t.addUnit(new Unit(Unit.Type.TAX, taxes));})
+                        //.assign((t, v) -> t.addUnit(new Unit(Unit.Type.TAX, asAmount(v.get("tax")))))
+//                                        Money.of(asCurrencyCode(v.get("currency")), asAmount(v.get("tax"))))))
+
+                        .section("tax", "currency").optional()
```
inv-trad, line 289: Wouldn't it make sense to read KEST and SOLI at the same time because it is always printed in one line?

Something like:

`.section("kest", "soli", "currency").optional()`
...
`.match("^KEST: (?<currency>\\w{3}) -(?<tax>[\\d\\.]+,\\d{2}), SOLI: (?<currency>\\w{3}) -(?<tax>[\\d\\.]+,\\d{2})(.*)$")`

Then reading both values and adding them up.

christen90: please see example in Kontoauszug01.txt line 80. If KSST is very low SOLI is not in the text.
`KEST: EUR -0,01`

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
christen90: Thanks for the hint, but this is causing that the block is cutted right behind the _Wertpapier Verkauf_ and not all bookings were find in the test. The easiest approach for me would be to limit the following lines in numbers.
Another approach would be to match with positive lookahead. Do you have an idea how to match a new line? I already tried:
`(.*\\n.*)+(?=(Sammelgutschrift)|(other booking types))`

inv-trad: The matching logic searches line by line, not the entire text at once. That's why it is a bit tricky.

I think, it would make sense to change `PDFParser.java` in line 151 to this:

`endLine = findBlockEnd(lines, startLine + 1, endLine);`

because searching for the endsWith should start in the next line instead of the same as startsWith.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
inv-trad: Is this regex better? I don't know, if it passes back an empty value for soli, in case it is not given.

`^KEST: (?<currency>\w{3}) -(?<kest>[\d\.]+,\d{2})(, SOLI: (\w{3}) -)?((?<soli>[\d\.]+,\d{2})|(.*))?$`
 
Otherwise, a `oneof` statement might make sense?!

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
buchen: I like this proposal. I would not allow for one-line blocks anymore, but I guess that is also fine. One line blocks are still possible if you match the next block on the next line.

buchen: I think I have to correct myself. The way "endsWith" is implemented today, it only looks within the **current** block and if it does not find the **endsWith** it actually skips the transaction. So if a new block with ```".* Ref\\.: .*"``` is actually defined, then nothing is found. Needs more thinking....

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
buchen: I believe both approaches are valid - match one line in a block multiple times (once for KEST and once for SOLI) or match it with one block.

the ```oneOf```  would only work if we are guaranteed to always have at least KEST or SOLI.

the problem I have is that there are no exchanges rates in the document. See below

buchen:
Right now I see the problem with information as this:
```
Erträgnisabrechnung, Ref.: 87991231 27.06.2019 27.06.2019 11,54 EUR
I.M.-I.S&P 500 UETF Reg.Shares Dist o.N.
IE00BYML9W36, ST 106,167
KEST: USD -2,83, SOLI: USD -0,15
```

The way PP expects the transactions is this:
* Either it has **one** currency. The it must not have a GROSS_VALUE unit. And the Unit for taxes and fees never have a foreign currency
* Or it has **two** currencies. Then there must be a GROSS_VALUE entry that provides the gross value in transaction currency and in forex and the exchange rate. The Units for taxes and fees either have no forex (then must be in transaction currencies) or have a forex, but then must include the exchange rate and the local currency.

I did not find the exchange rate with which the KEST and SOLI have been converted.
@christen90 - did you see something?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
buchen: I have no implemented a "setMaxSize" method which limits a block to a maximum of this number of lines.

buchen:
Okay, I merged the first part.

For me open is still the question of the exchange rates for the taxes.
If the exchange rates used is not available in the document, we can use the exchanges rates that PP knows.
At the moment, they are not available to the parser, put for sure we can enhance it and pass them to the parser.

christen90:
> I did not find the exchange rate with which the KEST and SOLI have been converted.
> @christen90 - did you see something?

I doublechecked the original PDF. There is no exchange rate contained in the document. The exchange rate of PP is better than now ;)
Just saving the USD as EUR might be ok for the start but maybe there are other currencies out there which might have more then 30% error. :D

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirionPDFExtractor.java
inv-trad: > I have no implemented a "setMaxSize" method which limits a block to a maximum of this number of lines.

Thank you for that idea/feature, Andreas. That will be helpful in many other cases as well.

PR #1979, opened 2021-01-15
Make quote precision configurable

pascalhein:
With the recent change to 8-digit quotes, values displayed in tables often have excessive precision (especially averages, such as "average purchase price"). Since not everyone needs this amount of detail, this commit allows configuration of the desired number of digits - the same way as for shares.

Since quotes were, until now, using a ColumnLabelProvider (unlike shares, which used SharesLabelProvider), they did not repaint on every tick, missing the settings update. This is changed here so new settings are applied immediately, rather than after a page refresh. It also requires a more low-level (though functionally unchanged) implementation of 
NumberColorLabelProvider, which can no longer rely on ViewerCell.

I hope that I found all such instances of "Quote-like" columns and replaced them with the new provider, but if any are missing they should be easy to adapt to the new pattern.
Unfortunately, there's a bit of repetition (`FormatHelper.format` / `Values.Quote.format`), but avoiding this would probably require a fair bit of restructuring in `Values.Quote` (whose default behaviour we also want to keep for tooltips).
The helper class `QuotesLabelProvider.Quote` is only required since the regular `Quote` does not allow null-currencies. This could be changed but might have further consequences that would need to be verified.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -919,6 +919,7 @@ public class Messages extends NLS
     public static String PrefLabelNote;
     public static String PrefLabelProxyHost;
     public static String PrefLabelProxyPort;
+    public static String PrefLabelQuoteDigits;
     public static String PrefLabelSharesDigits;
     public static String PrefLabelUseIndirectQuotation;
     public static String PrefMsgConfigureUpdates;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
@@ -98,6 +98,11 @@ public interface UIConstants
          */
         String FORMAT_SHARES_DIGITS = "FORMAT_SHARES_DIGITS"; //$NON-NLS-1$
 
+        /*
+         * Preference key for display precision of quote values
+         */
+        String FORMAT_QUOTE_DIGITS = "FORMAT_QUOTE_DIGITS"; //$NON-NLS-1$
+
         /**
          * Preference key to use indirect quotation ("Mengennotierung") when
          * displaying exchange rates.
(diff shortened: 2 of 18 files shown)


chirlu:
I would like to distinguish between actual quotes and calculated values (averages, cost-divided-by-shares in transactions, etc.). For the calculated values, I don’t care much about the precision – a few digits is enough, and I don’t need it to be configurable how many. For the real thing, however, I definitely want to see the exact value, so I would always set any configuration option to “as many digits as necessary”.

buchen:
> With the recent change to 8-digit quotes, values displayed in tables often have excessive precision (especially averages, such as "average purchase price"). Since not everyone needs this amount of detail, this commit allows configuration of the desired number of digits - the same way as for shares.

I agree to that. I just did not get around to adopting the formatting options.

And maybe because of the Bitcoin Nerds, we need a configuration option.

> I would like to distinguish between actual quotes and calculated values

Agree. The quotes in the history should show all digits, the rest as configured.


> Since quotes were, until now, using a ColumnLabelProvider (unlike shares, which used SharesLabelProvider), they did not repaint on every tick, missing the settings update. This is changed here so new settings are applied immediately, rather than after a page refresh.

I am not sure if we really want the OwnerDrawLabelProvider - and that only because we want to immediately to refresh.

For the shares, I did not want to display and digits after the comma. For shares that is relevant, because at least I am trading in whole numbers and I do not want the noise of the zero digits. But I do not think that is the case for quotes.

Then I don't like that there is another (inner) class ```Quote``` with the equal name. When reviewing the code on Github, I got confused and was missing the fullFormat function. Let's name that differently.

I'll cherry pick your changes to the preferences and then will have a go at it.

buchen:
cherry picked and merged.

pascalhein:
Hello @buchen,

thanks for your feedback.

> Agree. The quotes in the history should show all digits, the rest as configured.

I have updated the PR to only affect calculated fields, and not values taken immediately from the price history.

What I don't like about this is that if your security has only *some* prices with more than two digits, the numbers don't align and are hard to read - both in the SecuritiesListView / Historic Quotes tab, and in other tables such as Performance:

image

Maybe we can find another solution for this in a later PR.

PR #2012, opened 2021-01-24
Add Binance Quote Feed

NicolaiSchmid:
In addition to Bitfinex, Binance is also a big exchange for crypto "currencies". Since Bitfinex doesn't carry a lot of the tokens traded on Binance I added a second QuoteFeed to pull quotes from Binance as well.

Since I'm not well experienced with creating Java UI applications, I'm pretty sure I overlooked some things that still need to be changed. Please let me know, so I can incorporate the missing changes before this can get merged.

Thanks, @buchen, for your great work!

### DIFF
--- a/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
+++ b/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
@@ -1,6 +1,7 @@
 name.abuchen.portfolio.online.impl.ManualQuoteFeed
 name.abuchen.portfolio.online.impl.AlphavantageQuoteFeed
 name.abuchen.portfolio.online.impl.BitfinexQuoteFeed
+name.abuchen.portfolio.online.impl.BinanceQuoteFeed
 name.abuchen.portfolio.online.impl.FinnhubQuoteFeed
 name.abuchen.portfolio.online.impl.EurostatHICPQuoteFeed
 name.abuchen.portfolio.online.impl.KrakenQuoteFeed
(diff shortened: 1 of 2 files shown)


buchen:
Looks good. And Binance.com publishes the API documentation and talks about public API which indicates they are okay with this.

buchen:
@NicolaiSchmid 

What symbols would I have to provide?

I used ```BTCUSD``` but for this there seems to be a 400 http error:
```
https://api.binance.com/api/v3/klines?symbol=BTCUSD&interval=1d&startTime=5336474011822303232 --> 400
```

For the Symbol ```BNBBTC``` I get data, but I have troubles matching it to what I see on the web page. Or is there a delay?

Bildschirmfoto 2021-01-26 um 21 55 31

NicolaiSchmid:
> Looks good. And Binance.com publishes the API documentation and talks about public API which indicates they are okay with this.

Yes, even though I couldn't find a licence document specifically, the endpoints I'm using are public and documented.

> What symbols would I have to provide?
On Binance you don't trade with USD directly, but with either USDT or USDC stable coins:

```
https://api.binance.com/api/v3/klines?symbol=BTCUSDT&interval=1d
```
or EUR:
```
https://api.binance.com/api/v3/klines?symbol=BTCEUR&interval=1d
```

I'm however not sure about the timestamp, it should be Epoch milliseconds, where did you get `5336474011822303232` from?

buchen:
> I'm however not sure about the timestamp, it should be Epoch milliseconds, where did you get 5336474011822303232 from?

I found that URL in the log when testing the BinanceQuoteFeed.

buchen:
> 5336474011822303232

That is the current date minus 2 months (the preview pane is only reading the last two months.

With the two symbols your provided above it looks good. I have rebased the change to HEAD and pushed to master. Thanks!

NicolaiSchmid:
Thank you for merging! Looking forward to the update!

PR #2013, opened 2021-01-25
Fix investment plans with negative values for larger amounts

inv-trad:
Issue: https://forum.portfolio-performance.info/t/sparplan-erzeugt-negative-stuecke-und-kurse/13952

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
@@ -417,8 +417,8 @@ public class InvestmentPlan implements Named, Adaptable, Attributable
                             converter.with(targetCurrencyCode).getRate(tDate, security.getCurrencyCode()).getValue());
         }
 
-        long shares = Math
-                        .round(availableAmount * Values.Share.factor() * Values.Quote.factorToMoney() / (double) price);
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
```diff
@@ -417,8 +417,8 @@ else if (planType == AccountTransaction.class)
                             converter.with(targetCurrencyCode).getRate(tDate, security.getCurrencyCode()).getValue());
         }
 
-        long shares = Math
-                        .round(availableAmount * Values.Share.factor() * Values.Quote.factorToMoney() / (double) price);
+        long shares = Math.round((double) availableAmount * Values.Share.factor() * Values.Quote.factorToMoney()
+                        / (double) price);
```
chirlu, line 421: The second cast actually is redundant now.

chirlu (approved):

name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
buchen: Looks like it is still good enough with ```double```. At other places, I have converted calculations with shares and prices to ```BigDecimal``` because both have 8 digits precision now.

Thanks for fixing.

PR #2016, opened 2021-01-25
AlphaVantage: use API to get and use adjusted_close (instead of regul…

verybadsoldier:
…ar close)

Currently PP is using the regular close price. Wouldn't it be better to use adjusted close? If people really would like to use regular close then also an option could be added to switch between the two.


chirlu:
It never makes sense to use prices adjusted for dividends in PP, ever.

For stocks that have never paid dividends, but have had splits, it may be somewhat useful; however, in my opinion it is preferable to represent splits in a lossless way by outbound and inbound delivery of the shares, until a proper implementation of lossless splits is available in PP.

verybadsoldier:
Ok, I agree, dividends do spoil the fun. Didn't had them in mind, sorry. Closing this since it won't help. Thanks for clearing that up.

PR #2018, opened 2021-01-26
Support Trade Republic tax refund in PDF import

lahr:
Upon increasing the annual tax allowance ("Freistellungsauftrag") Trade Republic issues a dedicated tax refund document ("Steuerabrechnung") if the previously set tax allowance amount was already exceeded. There is also a tax refund entry in the quarterly account statement.

Current behavior: Portfolio Performance doesn't recognize the tax refund document and ignores the entries in the account statement.

Desired behavior: Portfolio Performance should create tax refund transactions for both document types.


buchen:
Thank you very much for your contribution 😄  Looks good!

Issue #2063, opened 2021-02-07
Change scale in diagram


simphide:
**Is your feature request related to a problem? Please describe.**
I wanted to see my accumulated dividends in a diagram, but the scale is too big for my numbers

**Describe the solution you'd like**
I think it would be great if it would only automatically scale to thousands (k) when there is at least one number that is greater than 1000

**Describe alternatives you've considered**
A way to change the scale manually would also work

**Additional context**
![image](https://user-images.githubusercontent.com/8183517/107157051-3a513e80-6982-11eb-9134-354b666ed425.png)

PR #2084, opened 2021-02-13
Change "Earnings & Expense" to "Income & Expenses" 

Ragas13:
by implementing optionally trades

In memory of PR #1380 and the case that values from trades are based on calculations rather than fixed bookings and recently repitly asked by the community I've recreated my former PR with few options:;
1. User could decide if they would like to use the calculated earnings from the trade view. Depend on the decission, the icon option were enabled or disabled.
EnabledDisabledTradeIcon
New Flags

2. As the user claimed the k-amount scalling for small amounts I've slighly changed the formatting more intuitive
AdjustedScalling

3. Finally (unfortnately I didn*t find the forum thread) there was asked to consolidate the retired securities to get only the active once. This is available as option as well.
RetiredSecurities1
RetiredSecurities

Skalierung der Grafik:
https://forum.portfolio-performance.info/t/skalierung-der-y-achse-im-diagramm-vermoegensaufstellung/1892
https://forum.portfolio-performance.info/t/skala-bei-ertraege-akkumuliert-veraendern/8382

Trades unter "Einnahmen & Ausgaben" aka "Erträge & Ausgaben"
https://forum.portfolio-performance.info/t/gewinne-verluste-aus-trades-unter-ertraege-und-ausgaben/6922
https://forum.portfolio-performance.info/t/anzeige-von-veraeusserungsgewinnen/86

Warm regards
Marco

#2063

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -549,6 +549,10 @@ public class Messages extends NLS
     public static String LabelEarningsPerYear;
     public static String LabelEarningsSelectStartYear;
     public static String LabelEarningsExpenses;
+    public static String LabelEarningsUseConsolidateRetired;
+    public static String LabelEarningsConsolidateRetired;
+    public static String LabelEarningsUseTradeProfitLoss;
+    public static String LabelEarningsTradeProfitLoss;
     public static String LabelExpenses;
     public static String LabelError;
     public static String LabelEurostatRegion;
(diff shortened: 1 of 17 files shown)


buchen:
> In memory of PR #1380 and the case that values from trades are based on calculations rather than fixed bookings and recently repitly asked by the community I've recreated my former PR with few options

I like the change towards the formatting of the Y-Axis and the consolidation of inactive securities.

The only remark for me would be: the summary line "inactive securities" should only show up if there are inactive securities that actually have values. Adding a line with just zeros does add noise to the UI.


Then about the discussion we already had in #1380:

I still do not think there is super clear case to mix these numbers. But I understand the argument one wants to look at all "Kapitalerträge" and that includes dividends, fixed income, and realized capital gains. That is valid so let's add the realized gains from closed trade (you got me 😉).


Now let's talk about the implementation:

What I do not like is that we have now *two* things: the button in the toolbar *and* the entry in the drop down menu.
Why don't we add the "Realized gains" as an element to the toolbar, and include it always in the "sum" element?

buchen:
Merged.

There is a new tab "closed trades" now. I have removed the menu item in the configuration menu. That addresses the point I mention above: having both a tool bar and a menu entry.

PR #2087, opened 2021-02-15
Add TradeCalendar Toronto Stock Exchange

Nirus2000:
Hallo @buchen

hinzufügen des Toronto Stock Exchange

Grüße
Alex

PR #2091, opened 2021-02-17
Add calendar for Toronto Stock Exchange.

roshanshariff:
Also add test suite for some tricky cases. See https://www.tsx.com/trading/calendars-and-trading-hours/calendar.


roshanshariff:
Wow, I didn't notice the almost-identical pull request #2087. The rules around Christmas and Boxing day in that patch are not quite correct; see the test suite included here. The advantage of that patch compared to this one is that it includes translations into more languages. Let me know how you would like to proceed.

Nirus2000:
No problem...  :-)

Nirus2000:
Thx @roshanshariff 
i closed my PR 👍🏻

PR #2121, opened 2021-03-03
Fix date off-by-one in performance heatmap widget

chirlu:
As I had [mentioned on the forums](https://forum.portfolio-performance.info/t/januar-fehlt-in-monatlichen-heatmaps/14219/12), March was missing from the performance heatmap on March 1st (but reappered the next day). This was due to an off-by-one bug, where the data interval was compared to monthly intervals of the form (2021-03-01, 2021-03-31) which did not actually include March 1st, because of the half-open nature of `Interval`s.

In addition to the (trivial) fix, I refactored the representation of months and years in the heatmap widget to use `YearMonth`s and `Year`s, instead of abusing `LocalDate`s such as 2021-01-01 to stand for the year 2021 or the month 2021-01, depending on context.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
@@ -93,7 +93,7 @@ public class PerformanceHeatmapWidget extends AbstractHeatmapWidget<Double>
                  month.getYear() == year.getValue();
                  month = month.plusMonths(1))
             {
-                if (actualInterval.intersects(Interval.of(month.atDay(1), month.atEndOfMonth())))
+                if (actualInterval.intersects(Interval.of(month.atDay(1).minusDays(1), month.atEndOfMonth())))
                     row.addData(calculatePerformance.applyAsDouble(month));
                 else
                     row.addData(null);
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
@@ -2,6 +2,8 @@ package name.abuchen.portfolio.ui.views.dashboard.heatmap;
 
 import java.time.LocalDate;
 import java.time.Year;
+import java.time.YearMonth;
+import java.time.temporal.TemporalAdjusters;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
```diff
@@ -87,18 +89,19 @@ public PerformanceHeatmapWidget(Widget widget, DashboardData data)
...
 
             // monthly data
-            for (LocalDate month = LocalDate.of(year.getValue(), 1, 1); month.getYear() == year
-                            .getValue(); month = month.plusMonths(1))
+            for (YearMonth month = YearMonth.of(year.getValue(), 1);
+                 month.getYear() == year.getValue();
+                 month = month.plusMonths(1))
             {
-                if (actualInterval.intersects(Interval.of(month, month.plusMonths(1).minusDays(1))))
+                if (actualInterval.intersects(Interval.of(month.atDay(1).minusDays(1), month.atEndOfMonth())))
```
chirlu, line 96: The `.minusDays(1)` here is the bug fix, changing the starting date from 2021-03-01 to 2021-02-28.

Ragas13:
The basic problem is might be a little bit more complex and impact several calculations. Root cause is that the Interval never consider the first day at all per follow condition:

https://github.com/buchen/portfolio/blob/c66f26b4e062b8d987ccce4ce6d2b4701bde2bd7/name.abuchen.portfolio/src/name/abuchen/portfolio/util/Interval.java#L59

I guess `other.isAfter(start)` should be `!other.isBwfore(start)` ?!

chirlu:
No.

1. The `contains` method isn’t even used here; it’s `intersects`.
2. Both methods are completely correct. `Interval` is defined and well documented to exclude the start date; i.e., `Interval(2021-03-01, 2021-03-31)` is absolutely intended to mean “from the 2nd to the 31st”.

Nirus2000:
Hallo @chirlu und @Ragas13
seit der Änderung habe ich in dem Dividenden-Widget keine Daten mehr im Monat März.
Ich habe heute die erste und einzige Buchung importiert.
Könnt Ihr euch das nochmal anschauen?

grafik

chirlu:
Das `EarningsHeatmapWidget` benutzt eine veraltete Kopie dieses Codes. Das ist zwar prinzipiell schlecht, heißt aber, daß sich dort durch die Änderungen hier nichts getan haben kann.

Nirus2000:
@chirlu und @Ragas13 
hmm... das versteh ich. Aber wo ist den dann auf einmal das Feld verschwunden?
Und wie bekommen wir es wieder?! Sorry... in swt hab ich mich noch nicht eingearbeitet.

PR #2130, opened 2021-03-05
Fix regular expressions used for Sutor PDF extract

arthurschreiber:
It looks like the `REGEX_AMOUNT_AND_SHARES` and `REGEX_AMOUNT_AND_SHARES_NEW_FORMAT` regular expressions are invalid and thus never match. 😅


Nirus2000:
> It looks like the `REGEX_AMOUNT_AND_SHARES` and `REGEX_AMOUNT_AND_SHARES_NEW_FORMAT` regular expressions are invalid and thus never match. 😅

Thats wrong.
The regEx looks like 4 digits in a group.
https://www.vogella.com/tutorials/JavaRegularExpressions/article.html

arthurschreiber:
Mmh, you're right - they're not seen as invalid by the Regex engine in Java.

But I'm still confused by those regexps. What's the point of the `+` in `(\\d+.\\d+.\\d{4}+)`? There is nothing that the `+` quantifier can refer to in this position, is there?

Other regular expression engines (e.g. the one used by Node.js/V8) complain about an expression like this:

```
$ node
Welcome to Node.js v14.15.2.
Type ".help" for more information.
> /(\d+.\d+.\d{4}+)/
Uncaught SyntaxError: Invalid regular expression: /(\d+.\d+.\d{4}+)/: Nothing to repeat
```

Nirus2000:
@arthurschreiber
The regEx is match, but i think the `+` is a copy fault and has in this part no effect. I see this in the past in other importers to... so however.

buchen:
It is funny, because [in December we a similar discussion](https://github.com/buchen/portfolio/pull/1912#pullrequestreview-558931529)... 

The gist: if you look at [Possessive Quantifiers (and Atomic Groups)](https://www.loggly.com/blog/five-invaluable-techniques-to-improve-regex-performance/), then the greedy operator is a (tiny) performance improvement

PR #2131, opened 2021-03-05
Filter trades

keszocze:
I added the support to filter the trades for open and closed trades. (This was a big request in [the forum](https://forum.portfolio-performance.info/t/trades-nach-offen-vs-geschlossen-filtern/11208/5))

The introduced changes are quite nonintrusive except for an additional `isClosed()` method to the trade class.

If there are any requirements (e.g. code formatting) that need to be met, please do not hesitate to tell me.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -184,6 +184,7 @@ public class Messages extends NLS
     public static String ColumnFix;
     public static String ColumnForeignCurrencies;
     public static String ColumnGrossDividend;
+    public static String ColumnGrossProfitLoss;
     public static String ColumnGrossValue;
     public static String ColumnHeight;
     public static String ColumnHoldingPeriod;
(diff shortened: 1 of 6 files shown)


keszocze:
I further added
* filtering for profits/losses
* column for gross profits/earnings

buchen:
Hi @keszocze,

I have squashed your changes into one commit and merged into master.

Then I did a couple of changes in a follow-up commit:

* First, I ordered the filter button in such a way that the filter is always in the same place: first filter, then export, then configuration icon

* Second, the view already had a filter: if the user selected a set of trades on the dashboard, the view would open with a limited list of trades. That was also a filter. And that meant there were two filters now. I merged them into one.

* Third, I changed the behavior of the filter to work like in the other views: "only open trades", "only profitable trades" ... See below a couple of screenshots from other views.


Bildschirmfoto 2021-03-13 um 20 30 50
Bildschirmfoto 2021-03-13 um 20 39 35
Bildschirmfoto 2021-03-13 um 20 32 41

metal450:
Would be really useful if there were a third option: Only profitable trades, Only lossmaking trades, and also **Only non-neutral trades**. This would filter out trades that have no gain or loss.

The reason is, liquidating money market funds - which are always valued at $1 - can pollute this view a lot with non-impactful sales that don't need to be reported.

PR #2133, opened 2021-03-06
Fix comdirect finanzreport

christen90:
importer missed Übertrag with +amounts


christen90:
sry for the many commits. now I'm satisfied and could successfully import the last 4 years without delta.

buchen:
> sry for the many commits

No Problem. I'll squash them into one commit.

I believe, if you do an "interactive rebase" and squash the commits locally and push them to the same branch, then the PR is updated with the newly squashed commit.

PR #2135, opened 2021-03-06
Fix INGDiba PDF-Importer with dividends exchange

Nirus2000:
Issue https://forum.portfolio-performance.info/t/pdf-import-von-ing/7114/30
Issue https://forum.portfolio-performance.info/t/pdf-import-von-ing/7114/32
Fixed issue #2126 and #2143  can be closed
Add nameContinued in dividend transaction
Clear some regEx quantifier like \\d{4}+ -> \\d{4} or \\w{3}+ -> \\w{3}


buchen:
I can't remember in which issue we discussed whether to use ```\d{3}+``` or ```\d{3}```. From what I understood, it does not make a semantical difference, but using the greedy operator ```+``` can make the regex slightly faster because the parser will not back up and hence fail faster. When I have time, I'll try to start a README for the PDF extractors and link the discussion there. It's not a big thing, but I don't think we should create commits just to change the greedy operator.

Issue #2106, opened 2021-02-26
Quotes from JSON (Timezones Not Treated As UTC)


metal450:
I'm pulling crypto price quotes from a JSON API.  It works, except for one issue: the dates it receives are always "ahead" by a day, thus it's not actually showing the latest quote - it's showing the quote from one day prior.

My assumption is that this is an issue of timezone: although the JSON API is providing standard times in UTC, the software seems to be treating them as "local time."  As a result, it thinks the latest quote is one day in the future, and instead displays yesterday's quote for today.

I've experienced this issue with 3 different JSON APIs, so it definitely seems to be on the side of PP.  It's true whether the API returns the timestamp as a string (i.e. "2021-02-26T06:19:00Z") or epoch (i.e. "1614297600").

Here's an example of an API you can try as-is, without needing a key:
```
URL: https://min-api.cryptocompare.com/data/v2/histoday?fsym=EWT&tsym=USD&limit=5
Path To Date: $.Data.Data[*].time
Path To Close: $.Data.Data[*].close
```

My local system is in Pacific Standard Time.  The chart at the bottom of the "Historical Quotes" dialog shows all the correct quotes - but the latest quote is shown as tomorrow's date, and the quote actually used by the software is always one day behind (aka the second quote in the list).

metal450:
Addendum: This is true with the newly-added "Binance" quote provider, too.  The "Historical Quotes" dialog shows today's quote with tomorrow's date (all dates are shifted forward by a way), and the software uses yesterday's quote instead of today.

Ragas13:
Did you have an idea how to solve? There are mostly no UTC zone shift information inside JSON, without this information it's impossible to readjust the date.

metal450:
You don't need any zone shift information in the JSON - if we just assume that APIs provide dates in UTC (which to my knowledge, they all do), simply get the current system's timezone & add that as an offset.

Ragas13:
This behaviour isn't true, at least Binance and above referred JSON feed has such information.

metal450:
As stated in my last reply, you don't need any information from the API to assume UTC by default.  Currently PP seems to assume "local system time," which will clearly not be correct (unless you happen to live in UTC).

metal450:
[...]

PR #2153, opened 2021-03-16
Bugfix relating to latest security price not getting updated.

RomanLangrehr:
Could be related to #2106 .


chirlu:
Uh, “Bugfix” as a description doesn’t really give an idea of what the issue is and how this fixes it, does it?

PR #2158, opened 2021-03-16
Port recent heatmap changes to earnings heatmap

chirlu:
Port recent changes made to the `PerformanceHeatmapWidget` over to the `EarningsHeatmapWidget`, which has a lot of duplicated code. (Ideally, of course, the duplicate code would be extracted into an abstract class for monthly heatmaps or similar.)


buchen:
> Ideally, of course, the duplicate code would be extracted into an abstract class for monthly heatmaps or similar.

Couldn't agree more. If you want to give it a shot, I am happy to merge 😄

PR #2159, opened 2021-03-16
Fix typo in a label string name

chirlu:
Add the *l* in *Consolidate* that was missing in some instances of `LabelEarningsUseConsolidateRetired`.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_pt.properties
```diff
@@ -1117,8 +1117,6 @@ LabelEarningsSelectStartYear = Ganhos a partir do ano:
 
 LabelEarningsTradeProfitLoss = Negocia\u00E7\u00F5es fechadas
 
-LabelEarningsUseConsoidateRetired = Consolida\u00E7\u00E3o de t\u00EDtulos inativos
-
 LabelEarningsUseConsolidateRetired = Consolida\u00E7\u00E3o de t\u00EDtulos inativos
```
chirlu, line 1120: This is the correctly named string.

buchen:
Thanks... interesting how this slipped through with the tooling...

PR #2196, opened 2021-04-06
Fix Degiro PDF Import

matthesvoss:
Change redex to exclude "Abgelehnte Auszahlung" from transactions


buchen:
Hi @matthesvoss - thanks for your contribution! For bigger changes, it is helpful to have a test case. Otherwise it is all to easy to break the regex with the next update. I have merged anyway as this seemed rather small and I don't want to discourage small contributions. Thanks .

Issue #2215, opened 2021-04-14
JSON quote feed: Time in seconds instead of mili seconds or days


KP7984:
**Is your feature request related to a problem? Please describe.**
Thank you for making it possible in 0.51.2 / 14. März 2021 to specify the date format when configure the JSON import for historical quotes. Nevertheless, I've got a **JSON response which** isn't defining the time in mili seconds or days since 1970 as **it's defining the time in seconds**. 

This is not possible for now right?

**Describe the solution you'd like**
I would love that to be possible :D.

**Describe alternatives you've considered**
Write in the tooltip that this is not possible :( or explain how it is possible :)

**Additional context**
Example of JSON response for time in seconds:
[onvista.de](https://api.onvista.de/api/v1/instruments/FUND/107918/eod_history?idNotation=120539918&range=M1&startDate=2021-04-01)

Thank you!

KP7984:
I got a little confused since in [this bug](https://github.com/buchen/portfolio/issues/2106#issue-817083824) it seemed to work with seconds.

RomanLangrehr:
That is strange, PP should handle epoch time in seconds as well (the tool tip is bad, I will update it). Can you maybe provide an example where this is not working.

PP treats numbers greater than `7258118400` as “epoch in milliseconds”, numbers smaller than `84.006` as “epoch in days” and everything in between as epoch in seconds.

KP7984:
I'm afraid that I've used your feature in a wrong way? However, [your documentation](https://help.portfolio-performance.info/kursdaten_laden/#json) seems easily understandable...

The epoch is 1617278400 (1.617.278.400) so it should fit in between milliseconds and days.


I've tested all paths with [your recommended tool](http://jsonpath.herokuapp.com/) which found the correct arrays with equal lengths. Again, here is the link for an example JSON response: [onvista.de](https://api.onvista.de/api/v1/instruments/FUND/107918/eod_history?idNotation=120539918&range=M1&startDate=2021-04-01)
![JSON-Quotes_will_not_refresh](https://user-images.githubusercontent.com/48592359/115009442-95387200-9eac-11eb-8ccc-72e23b721737.png)

My paths:
- $.datetimeLast
- $.last
- $.low
- $.high

Could the point of a decimal separator in the quotes be an issue?
Or that the time in the epochs are GMT: 12:00:00 o'clock? 
Probably not :D

[...]

PR #2217, opened 2021-04-15
Fixed security prices not getting updated in timezones "behind" UTC (and improved JSON date format tool tip).

### LINKED ISSUE #2106: Quotes from JSON (Timezones Not Treated As UTC)

RomanLangrehr:
Fixes #2106 (hopefully). The issue, that a security price is not getting updated, might still exist, when a quote provider uses a different timezone then UTC. Removing this "is the date in the future"-check would solve this (and be more robust), but I don't know if this check is important for something.

Clarify JSON date format tool tip as mentioned in #2215.

*Update:* I noticed there were a "few" (193 to be precise) other occurrences, that used the local time zone. I changed it consistently to UTC.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/checks/impl/CrossEntryCheckTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/checks/impl/CrossEntryCheckTest.java
@@ -8,6 +8,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -265,7 +266,7 @@ public class CrossEntryCheckTest
 
         applyFixes(client, issues);
 
-        ClientSnapshot.create(client, new TestCurrencyConverter(), LocalDate.now());
+        ClientSnapshot.create(client, new TestCurrencyConverter(), LocalDate.now(ZoneOffset.UTC));
     }
 
     private void applyFixes(Client client, List<Issue> issues)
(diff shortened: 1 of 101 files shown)


buchen:
See my comment in #2106 - I am not so sure that making everything UTC is the solution.

We could also change the Generic JSON importer to convert the time into the local time zone:
```
return LocalDateTime.ofEpochSecond(object, 0, ZoneOffset.UTC).atZone(ZoneId.systemDefault()).toLocalDate();
```

I will cherry-pick the tooltip change though

RomanLangrehr:
> I will cherry-pick the tooltip change though

Sorry, I should have created two pull requests. Epically, since this one got really unreadable.

Issue #2165, opened 2021-03-17
Rebalancing is completely broken


RomanLangrehr:
Either the rebalancing feature in Portfolio Performance is conceptually broken or it was meant to be used in an entirely different way then I would intuitively expect.

**The problem**
Let me give you a simple example:
We have a portfolio of 100 €. Half of it is invested in Fund A, that invests 90% in Developed Markets and 10% in Emerging Markets. The other half is invested in Fund B, that invests only in Emerging Markets. For simplicity, one share of each fund is worth 1 €. The classification looks as follows:
![Classification](https://user-images.githubusercontent.com/10776736/111515558-39ff4c80-8753-11eb-92bb-70b9ba117b3e.PNG)

Now the user actually wants to invest 70% in Developed Markets and 30 % in Emerging Markets. The Rebalancing view than looks as follows:
![Rebalancing Wrong](https://user-images.githubusercontent.com/10776736/111515741-716df900-8753-11eb-8159-d0be5135e9af.PNG)
Besides the fact that it is a bit odd that PP suggests both to buy 27.778 shares and sell 250 shares of Fund A, if the user would execute what is suggested by the Delta Shares column, he would end up with a short position of about -122 shares of Fund A and 25 shares of Fund B. The portfolio is even less balanced than before. The correct solution would be to buy 27.778 shares of Fund A and sell 27.778 shares of fund B.
In my opinion, it is also a bit odd that shares get split according to the weight.

It gets even more ridiculous if Fund B does not exist. Obviously, the portfolio is impossible to balance in this case, but PP still suggests to sell -11.11 shares and buy 100.
![Rebalancing Impossible](https://user-images.githubusercontent.com/10776736/111517077-f60d4700-8754-11eb-9f0b-b1fe51fce57e.PNG)

**My proposal for a sound rebalancing approach:**
- In the Rebalancing View we have another column called “Rebalancing Factor” next to the “Allocation” column. In the “Rebalancing Factor” column you can enter a percentage, just like in the allocation column. The Rebalancing Factor column is only editable for the individual security entries and not for the classification categories. (Remember that the “Allocation“ column is only editable for the classification categories and not for the security entries.)
- The rebalancing factor determines how much each security contributes to the rebalancing. The following constraints are placed on the rebalancing factor:
- For each lowest level category, the rebalancing factors of the security entries within this category must add up to 100%
- Each security can have at most one entry with a non-zero rebalancing factor. (Important for securities that are part of multiple categories)

**Advantages**
[...]

PR #2225, opened 2021-04-28
New Rebalancing based on an linear equation solver

RomanLangrehr:
New Rebalancing based on an linear equation solver (Appache Commons Math library is added for this).

First part towards #2165 

### Summary:
The new rebalancing algorithm shows correct results even when securities are split among multiple categories and when we have categories that contain both securities and subcategories (the approach before could not handle this).
It also recognizes ambiguous results and when no exact solution is possible and marks this in the GUI.
The target value is now takes into account securities in the parent categories.
A new column Rebalancing (Amount) has been added, that shows the amount (instead of the number of shares) of the rebalancing result.

### Details:
- Securities split among multiple categories are handeled correctly:
Before: Contradicitng results for the split security:
Rebalancing Wrong
Now: One result per security, repeated results are shown in gray:
Rebalancing Wrong new

- When a classification contains both securities and sub-classifications, the securities are taken now into account when calculating the target values (and the rebalancing results become correct):
Before:
Rebalancing Mixed Categories
Now:
Rebalancing Mixed Categories new

- Ambiguous results are recognised and highlightey yellow (explanation in the tooltip):
Now:
Rebalancing Ambigous new

- When no exact solution is possible, an approximative solution is calculated and highlighted orange (explanation in the tooltip):
Before:
Rebalancing Impossible
Now:
Rebalancing Impossible new
The approximative result minimizes the mean squared error, but I noticed that this probably not the best idea, because the rebalancing results then sometimes sell more shares than they buy (and repeated rebalancing sells everything. Even though the portfolio is rebalanced when everything is sold, this is probably not what the user wanted ;)) I will try to come up with something like “minimize mean squared error but without changing the total amount”.


RomanLangrehr:
Closed because this PR is superseeded by #2268

PR #2233, opened 2021-05-01
Added ECB Statistical Data Warehouse as Quote Provider.

RomanLangrehr:
### Added ECB Statistical Data Warehouse as Quote Provider.
Currently the EONIA Index is the only supported data series. (I plan to add more like the Euribor-Indizes and Long-Term bond indizes in the future). I think this indices are useful as benchmark and a prerequisite to calculate some key portfolio satistics like alpha, beta and R-squared (which I also plan to do at some point).

The Index provides daily rates of returns. These are converted into daily quote feeds (starting with 100 € at the start of the index) to fit into the PP model.

The wizard:
Add EONIA

EONIA as benchmark:
EONIA

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/InterestRateToSecurityPricesConverterTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/InterestRateToSecurityPricesConverterTest.java
@@ -1,0 +1,112 @@
+package name.abuchen.portfolio.online;
+
+import static org.junit.Assert.assertEquals;
+
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+
+import org.junit.Test;
+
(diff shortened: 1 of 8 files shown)


buchen:
> These are converted into daily quote feeds (starting with 100 € at the start of the index) to fit into the PP model.

I see the ECB always returns the very first value - so that works out.


Right now, all the downloading code is targeted for security prices. And because it was easier to re-use this code, I am also using this for the inflation rate. But it does not mean that we need to fit *everything* into security prices. Do you think it is worth to keep the original values reported by ECP and work with them instead?

(I'll have a look at the rest of the PR once I am back at my development machine)

RomanLangrehr:
I don’t think it is worth the affort to develop an alteranative to the security prices (that stores the returns instead).

The conversion to security prices is cause some very small rounding errors (and one would have to adapt the conversion a bit when the ECB API would no longer return the first values).
And the conversion to security prices has to be done at some point, unless a lot of things (calculation of return, volatility, graph plotting and probably a lot more) can handle the new format as well (which sounds like a lot of effort and possibile code duplication).

buchen:
merged. I plan to release a version tomorrow - let's see...

Issue #1913, opened 2020-12-27
Added: Eclipse Launch Configuration


buchen:
On the latest master, I have added [Eclipse Launch Configuration DSL](https://github.com/mduft/lcdsl).

The idea is to have a generic way of describing the launch configurations. At the moment, the launch configuration contains OS specifics - and because I am developing on macOS, it does not work out of the box on Windows.

Now:
* Install the plugin from the update site ```https://mduft.github.io/lcdsl-latest/```
* Open the view *Window -> Show View -> Other... -> Debug -> Launch Configuration*
*  Right-click to run *Eclipse Application -> PortfolioPerformance* to run

<img width="575" alt="Bildschirmfoto 2020-12-27 um 23 19 51" src="https://user-images.githubusercontent.com/587976/103180849-0a4e5000-489a-11eb-84b6-0023acf72162.png">

Please let me know if that works for you.
If yes, I will replace the old launch configurations.

buchen:
Anybody noticed that it sometimes takes quite a long time until the application or tests start?

chrisaut:
> Anybody noticed that it sometimes takes quite a long time until the application or tests start?

I did not notice a difference compared to before, I didn't time it though, but nothing noticeable on my end

andrericardo:
I only saw the launch configurations after installing the plugin "[Launch Configuration DSL](https://marketplace.eclipse.org/content/launch-configuration-dsl) - Launch configuration management". 

I'm new to Eclipse RCP, thought the plugins from Eclipse Marketplace were optional. Was confused on why the Launch Configuration did not now show up as documented in the CONTRIBUTING.md

After adding the plugin showed up nicely and was able to run on macOS.

PR #2238, opened 2021-05-02
Add org.eclipse.swt.win32.win32.x86_64 to JUnit lauch configurations.

RomanLangrehr:
On my machine (Windows 10 x64, Eclipse IDE for Eclipse Committers - 2020-12) this was necassary to get the JUnit test running. (I got a lots of java.lang.NoClassDefFoundError: org/eclipse/swt/…


buchen:
Actually... I want to delete those launch configurations. They are a pain in the back because they are platform-specific. As I am developing on macOS, I cannot create them properly for other developers. And I do not want to maintain multiple ones...

Can you check if these launch configurations work for you: #1913

The plugin generates the correct launch configuration. I am now using those for a couple of months.

RomanLangrehr:
Ah, okay. The launch configurations from #1913 work perfect for me :)

PR #2245, opened 2021-05-06
Allow setting securities active, even for multiple selection

### LINKED ISSUE #2226: Securities can't be activated / enabled with context menu.
When one security is inactive, I can't activate / enable it with context menu.
imagen

The context menu should show inactive / active based on the security state. Be careful, the user can select several securities at the same time.

PP 0.52.0

pascalhein:
Fixes #2226. In the security list, the context menu allows
the user to activate/inactivate the selected securites
based on their current state. To show accurate messages,
this commit adapts some messages (and translations) to
handle single/multiple securities.

Open for debate: if the user has marked 2 active and
1 inactive security, should the context menu and dialog
read "mark 3 inactive" (as implemented here), or
"mark 2 inactive" (i.e., only count those that will actually
change state)?

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -984,10 +984,11 @@ public class Messages extends NLS
     public static String SecurityMenuDebugGetHistoricalQuotes;
     public static String SecurityMenuDeleteAllPrices;
     public static String SecurityMenuDeleteLatestQuote;
-    public static String SecurityMenuDeletePrice;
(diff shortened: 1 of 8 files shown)


buchen:
> Open for debate: if the user has marked 2 active and
> 1 inactive security, should the context menu and dialog
> read "mark 3 inactive" (as implemented here), or
> "mark 2 inactive" (i.e., only count those that will actually
>change state)?

Personally, I tend to think the second option is more intuitive...
But the most important thing is that the UI makes it the behavior clear - and I think that works.


BTW, are you using the ResourceBundleEditor plug-in for Eclipse?
It should also show you all the languages and allows you to rename label (and thereby renaming it in all translation files at once).

PR #2246, opened 2021-05-07
Added log scale option to performance diagramm

### LINKED ISSUE #208: Feature request: Logarithmische Skalierung im Kursdiagramm
Option einer logarithmischen Skalierung der y Achse im Kursdiagramm wäre toll (aber angepasst auf min und max Werte im Darstellungszeitraum, nicht ganzzahlige Zehnerpotenzen)

### LINKED ISSUE #1449: Logarithmische Skalierung auch für Vermögensaufstellung
Nachdem die Logarithmische Skalierung für die Wertpapier-Diagramme eingeführt wurde, würde ich mich freuen wenn dies auch für die Vermögensaufstellung eingeführt wird.

Gruß
Sven

### LINKED ISSUE #1461: Log scale for the performance diagram
Similar to #208, I'd like to display the performance diagram with a logarithmic scale. https://forum.portfolio-performance.info/t/performance-diagramm-logarithmische-darstellung/875 suggested the same and asked whether maybe the performance diagram had just been forgotten (log scale has been made available for the stock price charts in the meantime).

Is there any particular reason why the performance chart didn't get the feature yet?

RomanLangrehr:
This feature was requested several times, but I not implemented because the graph drawing library swtchart “does not support log scale with negative values”. Well, that is not really a limitation of swtchart, the log function is simply defined only for positive values and using a log scale with negative/zero values makes mathematically no sense.

However, it makes sense when you start with 100% and not 0% (as long as you don’t lose everything or more than your invested amount). I implemented this, but tweaked the scale to start still with 0% in the GUI. The implemenation is in a very similar fashion   to the security chart.

Here is an example how it looks:
MSCI World Log Scale

and here are the GUI controls to activate the log scale:
Log Scale GUI
I think it would be nice to have more ticks/labels in the log scale, but found no way to do this. I think this is really a limitation of the swtchart library.

The users scale choice (linear or log) is currently not stored in the .xml file, because I don't know how to do this here. (In contrast to the security chart view, we dont have access to the `name.abuchen.portfolio.model.Client` object here.

Fixes #208
Fixes #1461
Fixes #1449

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PerformanceChartView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PerformanceChartView.java
@@ -3,15 +3,18 @@ package name.abuchen.portfolio.ui.views;
 import java.io.File;
 import java.io.IOException;
 import java.text.DecimalFormat;
+import java.text.FieldPosition;
 import java.text.MessageFormat;
 import java.text.SimpleDateFormat;
 import java.time.LocalDate;
+import java.util.EnumSet;
 
 import javax.annotation.PostConstruct;
 
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.IMenuListener;
 import org.eclipse.jface.action.IMenuManager;
+import org.eclipse.jface.action.MenuManager;
 import org.eclipse.jface.action.Separator;
 import org.eclipse.jface.action.ToolBarManager;
 import org.eclipse.jface.layout.GridDataFactory;
@@ -24,9 +27,11 @@ import org.swtchart.ISeries;
 
 import com.google.common.collect.Lists;
 
+import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.money.Values;
 import name.abuchen.portfolio.snapshot.Aggregation;
 import name.abuchen.portfolio.snapshot.PerformanceIndex;
+import name.abuchen.portfolio.snapshot.filter.ReadOnlyClient;
 import name.abuchen.portfolio.ui.Images;
 import name.abuchen.portfolio.ui.Messages;
 import name.abuchen.portfolio.ui.PortfolioPlugin;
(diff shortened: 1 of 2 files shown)


buchen:
> However, it makes sense when you start with 100% and not 0% (as long as you don’t lose everything or more than your invested amount).

Great idea!!

Additionally:
* We should not allow the log scale if configuration includes the "daily" data series - showing the daily changes - in the bar series, the "trick" will not work (because the bars are drawn from 0). That's why I also think it might now work to transform the data series all the time (```x -> x + 1d```).
* the performance chart on the widgets must also be updated (formatter, etc.)
* add a couple more labels to the y-axis - not sure why it shows to few
Bildschirmfoto 2021-05-09 um 08 47 08


Plus: I agree with your comments in the code that it makes sense to store the "show log scale". I also think it makes sense to store this per chart. The ConfigurationSet.Configuration is a pain in the back (I must have been drunk...). I propose to refactor this in a *separate* commit. Instead of the string parsing, we could serialize a Java object to JSON and have the chart configuration stored there.

@RomanLangrehr you want to have a go at the points above? I might find time for it this evening, but maybe not.

RomanLangrehr:
Thanks for your comments.
I haven’t thought about the daily data series and the widgets.
Regarding the labels: I agree. I already tried to do so, but I couldn't find any options to change the default behaviour of swtchart.

I probably won't have time today, so please feel free to implement already some points.

RomanLangrehr:
@buchen Regarding the daily data series: I experimented a bit and I think our best option is to use two Y-axis: One for the daily performance data series and one for the cummulated performance data series. The log scale will only apply to the cummulated performance data series (which solves the first issue you pointed out). It also solves the “issue”  that the daily data series is hardly noticable any more when you look at long time periods, because the daily bars get very tiny.

Here is an example how this would look like:
image

I don’t know whether to use this two-axis approach only for the log scale or although for the linear scale. I would like to hear your opinion on this.

Sn1kk3r5:
@RomanLangrehr @buchen 

This might have been fallen off. Do you guys see this coming by any change in the near future?

Sn1kk3r5:
One thing which came into my mind where this can be really useful as well is in the earning sections. 
German investors for instance which do have a stronger income between March and June would have a better view on the other (way lower months).
What do you guys think?

sig-noi:
I am not a programmer, would like to bump this as an important improvement. Thank you!

PR #2252, opened 2021-05-13
[PDF Import] Support importing zip files

mirkoruether:
Hello everyone,
first of all I'd like to thank you for this great piece of software. I really enjoy using it. :)

One little thing however always bothered me: When downloadling multiple documents (orders, etc.) from my DKB-Broker webinterface they come as a zip file containing the pdfs; In order to import them, I first have to extract it to some folder and delete the folder afterwards.

I've implemented a small fix for this, in the simplest and stupidest way I could come up with: 
When a zip file is selected during import, its contents (at least the contents which names end with .pdf) are extracted to a newly created folder in system temp dir. After the import is done, the files and the folder are deleted.

Please share your comments on this. We can also have a discussion on the forum in German if you like :)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/ImportPDFHandler.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/ImportPDFHandler.java
@@ -1,11 +1,19 @@
 package name.abuchen.portfolio.ui.handlers;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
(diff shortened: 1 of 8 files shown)


buchen:
Simple and stupid is good 😄 because then I and others also understand it in the future.
Thanks - very nice addition

buchen:
I have added one line: create possible sub-directories when unzipping the entries. I noticed that when creating a ZIP file on macOS, there might be sub-directories with meta-information about the file (the files cannot be imported, but at least the importer does not abort).

mirkoruether:
Glad you liked it :) Thanks for the quick feedback and merging.

PR #2261, opened 2021-05-24
Add a "latest trade" column to trades overview

### LINKED ISSUE #2260: add column "Last Trade" to Trades table
Due to tax regulations it's sometimes helpful to see, if all trades on this security are longer ago, then $date.

There is a nice overview about all active trades in the trades table (using the filter "only active trades"). But this table only shows "Start Date" and "End date" (by the way, the capital letters differs).

What I like to have is an additional column, showing the date of the last trade I've done on the individual security.
Actually you can use the hint on the "#transactions" column to see a list of individual trades. But this is a bit sub optimal, cause the information is hidden again, when your mouse is moved a bit. And you can only see one security at a time.

pascalhein:
Fixes #2260. For tax purposes, it may be helpful to know the
date of the most recent transaction in an open trade. With this
change, the information is more readily available (and sortable).


buchen:
Super. Thanks!

Issue #1495, opened 2020-04-17
Feature Request: Allow allocation on instrument


iwalucas:
Currently PP can only handle target allocation (%) on Levels. It would be nice to allow it to be set on an instrument itself.
Currently as a workaround, what I do is create a level with the name of the instrument, and place the instrument on it.

I added a SS to help visualize what I am talking about.
<img width="473" alt="image" src="https://user-images.githubusercontent.com/3654783/79588341-b153d880-80cb-11ea-9b33-f9531566f938.png">

buchen:
How would PP work? If the instrument has a target allocation, it would work just like a category. In your case, you would still have to create the "Europa Ocidental" category b/c the 44% allocation can be achieve with either of those instruments. It probably makes sense - you are not the first one to ask...

PR #2268, opened 2021-06-04
New Rebalancing based on a linear equation solver

RomanLangrehr:
This pull request superseds #2225 

The difference to #2225 is that here, when no exact solution is possible, a solution is proposed that minimizes the mean square error while respecting that the total sum of the rebalancing amounts should be zero.
This is much better than only minimising the mean square error, because this version guarantees idempotence (i.e., when you execute the rebalancing as suggest by PP all the rebalancing amounts change to zero indicating that no further rebalancing is possible.)

So the “inexact rebalancing” example from the description of #2225 now looks as follows:
image

And here is another, more interesting example:
image


buchen:
Hi @RomanLangrehr,

sorry for replying only now.

I like the idea to address the problem of securities being split across multiple categories. Up to now, the algorithm was simply ignoring this.

However, the new algorithms changes the way the categories are used - at least the way I am using the categories.

Take a look at this example (the kommer file + a couple of ETFs):

Bildschirmfoto 2021-06-13 um 09 07 38

The way I read this is:
* I want the category *Europe* to be 19 % of the Weltportfolio
* I can achieve this with a couple of ETFs available to me
* How much do I have to purchase from one of those to rebalance the category

Yes, I could purchase 7,06 from every of those ETFs, but most likely I only want to purchase from the one that I already own.

Sometimes I am changing the ETF I am using. I might not sell the whole position, but I will purchase new shares from the new ETF and rebalance by selling from the old one.

And then have a look if a position needs a sale in order to rebalance. As I do not own most of the ETFs, the sales information is not meaningful. Previously, the table also showed negative values for positions I do not own, but I would know how much to sell of an existing position to rebalance.

Bildschirmfoto 2021-06-13 um 09 12 28

Therefore I propose to keep the rebalancing on the level of **categories**.

Which brings back the questions what to do if a security is split across multiple categories. Can we calculate the value in such way that if the user would rebalance using this security only in the respective category?

Or do we change the behavior so that the user can mark which of the securities within a category shall be used for rebalancing? Be default, it would be those that have positions. But he could deselect securities he owns (in order to not change the position) or add new securities (that he plans to purchase).

buchen:
I converted the PR into a draft in order to reflect the discussion

zroug:
Maybe there could be the option to also assign an allocation not only to the categories but also to the securities. You could then allocate 100% to the stock that you are currently holding. If no allocation is assigned to the securities from a category it should default to equally allocated.

RomanLangrehr:
> Therefore I propose to keep the rebalancing on the level of categories.

I am not 100% sure what you are trying to say with this, but I would say that this PR still does “rebalancing on the level of categories”, because the new rebalancer also tries only to get the weights of the categories to match the desired weight (and not for each individual security). So the goal of the old and new rebalancer is the same. (There is one exception: When you have “mixed categories”, i.e. categories that contain both securities directly and sub-categories. This PR then lowers the target weight for the sub-cateogries by the volume of the securities in top-level category. The previous rebalancer could not handle “mixed categories”.)

I think the difference is how we want to handle ambiguities. This rebalancer computes the solution that minimizes the euclidean norm of the rebalancing demands, which typically (when securities are not split among multiple categories) means that the rebalancing demand is split equally among the securities in one category. And I think you want a solution that proposes multiple choices to the user instead (buy/sell one of these securities).

I completly agree an output like “buy one of these securities” would be much more helpful to most users, because you typically want as few transactions as possible. The problem is only that it is a little bit hard to generalize this to the case when securities are split among multiple categories.

One problem is, that you can not say any longer that exactly one security of each category should be used. Take for example this:
image
For both ETFs in the category “Schwellenländer” you have to perform a transaction in order to rebalance this.
In general, you can get only constraints like “you have to use 𝑘 out of these 𝑛 securities in the rebalancing” (but the 𝑛 securities do not have to coincide with a category) and a single security can appear in multiple of these constraints. Even worse, the rebalancing amount for the securities can depend on which securities are selected for the rebalancing. Furthermore, there are exponentially many valid ways to select securities for the rebalancing (take 2𝑛 securities that each appear in 𝑛 categories with non-zero weight. Then every set of 𝑛 securities can be used for the rebalancing, so we have 2𝑛 over 𝑛 solutions.) So it is infeasible to compute all these solutions and display them to the user in general.

So I would suggest a solution where the user can select which securities are used for the rebalancing (where “used” means that they are part of the rebalancing result. The amount currently invested in these securities is taken into account in any way). I plant to do exactly this with the rebalancing factors I described in #2165 (where a rebalancing factor of 0 would mean that a security is not used in the rebalancing). I originally had in mind the case the scenarion where someone has an ETF on an Index A and later switches to a new ETF on the same Index A (e.g. becaues it is cheaper) but does not want to sell the old ETF (e.g. because of transaction costs or tax reasons). I considered this an optional feature, that is why I already made a PR without it, but I see now that this could be also important in other use cases and should be added before merging this into the master branch.

My orginal proposal for rebalancing factors allowed arbitrary values between 0 and 1 and “not set”. Thinking again about this, in most cases it is probably sufficient to have only zero (for “do not use in the rebalancing”) or “not set” (for “use in the rebalancing”). This is not as powerfull, but a lot simpler (it can be easily implemented with a check box in the GUI) and easier to understand. That is why I think now this simplifed version of the rebalancing factor is probably better.

RomanLangrehr:
@zroug Thanks for poiniting this out. However, I would vote against such an option as described in #1495.
 
First of all, it can not solve the issue of ambigious solutions. When we specify the exact weight for each securiy directly, the rebalancing solution would be exact, of course. However, it is sometimes necassary to not specify a weight for each security. Take  for example the picture I posted above. There the solution is already unique (despite having multpile securities in one category). And if you specify a weight on each security, there might be no more exact rebalancing solution.

Second, if you really want a specific weight for one security, you can make a category that just contains this security.

I think the current classification model is very good. It is clean and simple, but also very powerful.  Categories allow you to specify what you want and you can than populate it with real-world securities and then the rebalancer tells you how much to buy of each security. In this model it does not make sense to have a weight for each individual security. The problem is only that the current rebalancer can only handle a fraction of all valid models (and 

I do not know why #1495 was proposed, it unfortunately does not describe *why* this would be helpful.

buchen:
> TLDR: [...] I propose that the user can select which securities are used for the rebalancing.

I can follow your argument. I agree.

It also fits to how the rebalancing works on some web sites: First pick the security for the category, then get the suggestions for the transactions (purchase/sale).

The ```Assignment``` already has a ```data``` object that you can use for storing the flag that a security is to be included in rebalancing. But I am also fine with adding a new attribute to the assignment class directly.

Let me know if you want me to do the UI parts (depending on how much you want to dive into SWT/JFace).

zroug:
@RomanLangrehr Yes, having individual weights on securities would essentially be a shorthand for a category with a single security. So it is not strictly necessary but I think it might be a common use case. It does not solve the problem of ambiguous solutions but it isn't meant to do that (I didn't propose it as an alternative to this PR in case that it sounded like that).

It is meant to solve among others the example that @buchen showed, multiple securities with a common category but where only one is allocated. If this feature was implemented (or if @buchen used the workaround for it), I think that the solver from your PR would already output the solution that @buchen wanted. That's why I answered here.

Independent from that I agree that it could be useful to further restrict the securities that are used for re balancing.

RomanLangrehr:
@buchen I implemented the “use for rebalancing” column now. Thanks for offering help with the GUI, but I figured out how to do this myself by looking at how the “inactive” column of the security overview works.
Regarding the storing: I don’t think it is a good idea to store this in the assignment, because it only makes sense to include/exclude a security (instead of an assignment) and a security can be used in multiple assignments. My implementation now stores the included/excluded securities all in the `data` object of the `TaxonomyModel`, to avoid potential incosistency problems. (In the GUI there is one check box for each assignment, but if you edit one checkbox, it will be changed for all assignments with the same security.)

@zroug Sorry, I misinterpreted your comment. Indeed, for conivience, one could add weights for securities as shortcut for a classification that contains only this security. But I think this is not a proper solution for @buchen ’s scenario and independent of this PR.

From my view, this PR is now ready to be reviewed/merged.

buchen:
@RomanLangrehr 

Okay, I finally got around to taking to reviewing this big change. It is merged now.

I added one commit that mostly completed the translations. Please review.

One particular thing I changed: I removed the ```Money ZERO = of("", 0);``` assignment with the empty currency code. All over the place, PP assumes that there is a currency. And also the additions and subtractions if one of the elements has an empty currency seem strange.

I replaced the only usage of ZERO with ```Money.of(snapshot.getCurrencyCode(), 0))```. All the actuals and targets are calculated in that currency, so I believe it should always be possible to add/subtract from this value. @RomanLangrehr - can you check?

Which brings me to the code of "currency" and "fallbackCurrency". I did not fully understand where you expect different currencies, but couldn't you use a currency converter to convert the values as needed? Or can you maybe explain how I can produce the MonetaryException thrown below?

```java
    // Better safe than sorry: Check if the user added everything in the same currency.
    private void verifyCurrency(Money money)
    {
        if(money.isZero())
        {
            // nothing to check, but if no non-zero money appears, we should use the currency code as fallback.
            fallbackCurrency = money.getCurrencyCode();
        }
        else
        {
            if (currency == null)
            {
                currency = money.getCurrencyCode();
            }
            else if (!currency.equals(money.getCurrencyCode()))
                throw new MonetaryException("Tried to add a constraint with currency " + //$NON-NLS-1$
                                money.getCurrencyCode() + "to the Rebalancer, but all " //$NON-NLS-1$
                                + "other constraints use the currency " + currency + "."); //$NON-NLS-1$ //$NON-NLS-2$
        }
    }
```

PR #2294, opened 2021-06-25
Bugfix for Issue: 1904 (IRR Calculation)

CK1-CK:
Bugfix for Issue: 1904
https://github.com/buchen/portfolio/issues/1904

IRR will now calculated right.
NewtonGoalSeek.seek() Function x0 Start-Value changed.


chirlu (commented):
I’m against this, because while it will solve a specific case, it will create new problems elsewhere. The better solution is #1959.

buchen:
I have finally merged #1959. Sorry.

PR #2297, opened 2021-06-26
Linebreaks in tooltip for metric completness.

RomanLangrehr:
The tooltip was wider than my screen, so I added some linebreaks.


RomanLangrehr:
@buchen Sorry, I didn’t see that you edited something before force-pushing. (The force push excluded the Security table, because their is no line break necessary and another view that seems to be unused.)

buchen:
rebased, merged

PR #2298, opened 2021-06-27
add sparse label mode to improve readability for charts with small width

buletti:
Added "sparse label mode" to the time line charts.
Sparser labeling has the goal to improve and guarantee readability by preventing overlaying or bleeding of the labels. 
It is used, when the chart width is below 320 pixel, i.e. for charts in narrow dashboard columns.
When sparse label mode is used, the labeling on the time axis is less dense. It uses fewer ticks and the ticks it does use have longer time intervals to compensate. 

Before:
pp_label_overlays
_^ label overlays on year-to-date, 2-years and 3 years diagrams_

After:
pp_sparse_labels
_^ no more label overlays with sparse-label-mode_

Edit: attached pp_dax_demo.zip, a copy of the included DAX portfolio with the demonstration dashboard of the above screenshots
[pp_dax_demo.zip](https://github.com/buchen/portfolio/files/6721185/pp_dax_demo.zip)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java
@@ -38,6 +38,12 @@ import name.abuchen.portfolio.ui.util.Colors;
 
 public class TimelineChart extends Chart // NOSONAR
 {
+
+    /**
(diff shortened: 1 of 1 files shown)


buchen:
Great idea. It was always annoying me, but not enough to do something.
Initially I thought the diagrams are quite empty (see my screenshot) but I like the very clean look.

Bildschirmfoto 2021-06-27 um 15 51 56

PR #2301, opened 2021-06-27
2 more trade calendars

RomanLangrehr:
This adds trade calendars for TARGET2 and the first of each month.

The TARGET2 calendar is necessary for the EONIA Index (and maybe other money market products).
The first of the month calendar is useful for monthly data, e.g. the HCPI indices.


buchen:
rebased and merged

PR #2302, opened 2021-06-27
More interest rate based indices

RomanLangrehr:
Blocked, because this depends on #2301 .

This adds several more interest rate based indices from the ECB statistical data warehouse, e.g. Euribor, Libor and Gouverment Bond Benchmarks, which I think is useful as benchmark for bonds and bond-based products.

It also slightly changes the EONIA Index in PP, namely it now computes only one quote per TARGET2 day (which are the days the EONIA interest rate is updated), instead of one quote for every day.


buchen:
rebased and merged

PR #2303, opened 2021-06-27
Unexpected quotes table

RomanLangrehr:
Added a table to the data quality pane, that shows unexpected quotes (quotes on holidays or weekends, that should not be there according to the selected trade calendar).

image


buchen:
Thanks. I have changed the layout slightly and played the tables next to each other:

Bildschirmfoto 2021-07-10 um 17 54 56

Issue #2228, opened 2021-05-01
Report interval: dynamic "last month" option


pat-s:
In the report interval selection there are some predefined intervals likes "this month" or "this year".

Others can be created by going back in time by a specific amount of days/months/years.

However, these are all relative to the current date, i.e. if one selects "last 1 month" this will show the last 30 days.

I find myself often in the situation in which I one want to see the performance of the last month, agnostic from the day of the current month.

I.e. if the current date is the 7th of May, I would want to have an easy option to see the performance from 01.04 - 30.04 without having to define a custom report interval for this period of the current year.

Doing so clutters the list of report intervals and one would always need to delete this interval again after creating it to keep the list clean.

mierin12:
Done since [release 0.54.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.54.0) with https://github.com/portfolio-performance/portfolio/commit/2073f0983fc548550d40a0d57734f5102e25a261

PR #2304, opened 2021-06-27
Add reporting periods: last week, last month, last quarter, last year

buletti:
This PR adds support for additional, "non-rolling" reporting periods. It adds 
- last week
- last month
- last quarter
- last year

It addresses the issue #2228 of @pat-s.

This PR also contains a related commit to reorder the reporting period dialog. It puts the "current X" items before the "last X" items and the "misc/other" items.
pp_reordered_periods


chirlu:
My two cents:
- Of course you are excited about your new entries, but the most important ones are still “since …” and “from … until …”, so those shouldn’t be hidden somewhere at the end of a (too?) long list.
- I’m not a native speaker, but I think “last” may be ambiguous and “previous” might be clearer. Or perhaps referring to “last calendar year” etc.
- This approach isn’t particularly flexible, and also you are running through letters pretty fast … What if someone next needs the most recent (rolling) four calendar months, say, or the current half-year? A more generic way would be to allow any number of calendar {weeks, months, quarters, (possibly half-years), years}, including or not including the current one.

buletti:
> My two cents:

Thanks for providing your feedback
 
>     * Of course you are excited about your new entries, but the most important ones are still “since …” and “from … until …”, so those shouldn’t be hidden somewhere at the end of a (too?) long list.

I don't really mind where the "misc block" is located - at the top or at the bottom. My intentions were merely to group/sort the current and last items chronologically in ascending order.

>     * I’m not a native speaker, but I think “last” may be ambiguous and “previous” might be clearer. Or perhaps referring to “last calendar year” etc.

I'm not a native speaker, either, but I think you are right. I actually used "previous" for the non English translations. I kept "last" for the English version as the UI (and the code) is already using it.

>     * [...] you are running through letters pretty fast …
Yep. The `CODE` chars I used are random and no longer hint to the type. All meaningful chars are already in use. I think the type information should be refactored to use an `Enum`.

> What if someone next needs the most recent (rolling) four calendar months, say, or the current half-year? A more generic way would be to allow any number of calendar {weeks, months, quarters, (possibly half-years), years}, including or not including the current one.

Well, yes semesters could indeed be useful, but all the rest is probably far fetched.
You are right about this not being a flexible approach, though. Unfortunately I do not understand how you would make these intervals more generic. Maybe you can share a little more insight into how to actually implement your approach.

buletti:
Add a commit to reorder the items in the dialog to have the "misc" section on top as suggested by @chirlu.

pp_reordered_periods_misc_on_top

buletti:
I have added a few more commits to address some of the other concerns of @chirlu 

One of them introduced a type `Enum` `ReportingPeriodType` to replace the currently used type `char`. This was needed as we reached the limit of meaningful type `chars`. (e.g. '`M`' was used for "current **m**onth").
However, introducing a type `Enum` requires a major change as the type `char` is used in the serialization format. So serialization has to be changed to make use of the type `Enum` instead. 
In my first draft of the change, I simply switched the type `char` in the serialized format with the `Enum` value. Then I realized "_Dude, if you have to touch the serialization and change it, then you don't have to stick with this old format. You can make it better_". 
So I switched the serialization from the `char`-based custom format to an object binding format with JSON serialization.

Following I'll try to summarize what changed and compare a typical read cycle:

* The serialized `String` is read from the Eclipse key-value preference store where it is persisted.
* The application tries to deserialize the `String` so it can be used to e.g. display the corresponding dialog
* The serialized `String` is passed to a factory method `ReportingPeriod#fromString()`

At this point the change happened.
Before:
- `ReportingPeriod#fromString()` determines the type by extracting the type `char` from the serialized `String`. It's the first `char` in the sequence.
- Once the type is known, it calls a type-specific constructor.
    - if the type is a simple type and has no data (like e.g. `CurrentMonth`), then a new object is instantiated via the default constructor
    - if the type is a complex type with data (like e.g. `FromXToY`, which includes two `LocalDate` objects), then the serialized object is passed to be parsed. The contained data is segmented and extracted from the `String`. It uses various `indexOf()` calls on separator `char`s combined with `substring()` calls for partial extraction. Once all the data is extracted and instantiated, the object is constructed.
- `ReportingPeriod#fromString() `returns the created object to the caller.

After:
- `ReportingPeriod#fromString()` determines the type calling the static `ReporingPeriodType.fromString()` factory method.
- `ReporingPeriodType.fromString()` tries to find the type
   - first it tries to read the `type` field of the JSON object that contains an `Enum` value.
   - if that fails, it checks the type `char` for backward compatibility as before
- Once the type is known, `ReportingPeriod#fromString()` calls a type-specific constructor or a type-specific `fromString()` factory method with the serialized `String`.
   - if the type is a simple type and has no data (like e.g. `CurrentMonth`), then a new object is instantiated via the default constructor.
   - if the type is a complex type with data (like e.g. `FromXToY`, which includes two `LocalDate` objects), then a type-specific, static `fromString()` method is called to do the parsing.
      - if it recognizes that the `String` is in JSO notation, it will deserialize it with a simple `GSON.fromJson(serializedString, SpecificType.class)` call. The heavy lifting is done by GSON.
      - if it recognizes that the `String` starts with a type `char`, then it uses the old parsing code as described above. This ensures full backward compatibility.
- `ReportingPeriod#fromString() `returns the created object to the caller.

Serialization on the other hand is now a one liner. It is identical for all types and implemented once in the abstract super class `ReportingPeriod`:
```
    public final void writeTo(StringBuilder buffer) 
    {
        buffer.append(JClient.GSON.toJson(this));
    }
```
I have adjusted and extended the unit tests accordingly.

buletti:
Oh and I also replaced "_last_" with "_previous_" for occurrences in the UI as well as code relating to the reporting period.

buchen:
Thanks @buletti!

Particularly thanks taking the time to code the fallback. Otherwise 50k users would lose their configured reporting periods.

About the UI: Maybe we can condense the dialog by grouping "Current" and "previous" into one line.
```
Current
( ) week ( ) month ( ) quarter ( ) year YTD
Previous
( ) week ( ) month ( ) quarter ( ) year
```

About the labels: I agree with the *previous week, month, quarter, year* labelling. It should be previous.

But is it really "previous X years and Y months"?
*Previous* to what? My gut feeling is still: "last X years" - "letzten 2 Jahre".
But I am not a native speaker either.

(BTW, that is the reporting period I use most - I might move it up again ;-)

buchen:
BTW, this is how it looks now in the settings file. I will check if it is easier to store the whole thing as JSON (instead of JSON fragments separated by semicolon).

```
AbstractHistoricView={"years"\:1,"months"\:0,"type"\:"PREVIOUS_X_YEARS_Y_MONTHS"};{"years"\:2,"months"\:0,"type"\:"PREVIOUS_X_YEARS_Y_MONTHS"};{"years"\:3,"months"\:0,"type"\:"PREVIOUS_X_YEARS_Y_MONTHS"};{"years"\:10,"months"\:0,"type"\:"PREVIOUS_X_YEARS_Y_MONTHS"};{"type"\:"CURRENT_WEEK"};{"type"\:"CURRENT_MONTH"};{"type"\:"CURRENT_QUARTER"};{"startDate"\:"2020-01-01","type"\:"SINCE_X"};
```

buchen:
Hi @buletti,

alright, I have included your changes. But I cherry-picked the changes here and there. That's why it is new commit.

I have taken over:
* the previous week, month, quarter, year classes
* the translations
* the test classes

I have adopted:
* the enum; it really makes it more readable to understand what types are available; and there is an easy test included that there is no duplicate code used

```java
    public enum Type
    {
        LAST_X_DAYS('D', LastXDays.class), //
        LAST_X_TRADING_DAYS('T', LastXTradingDays.class), //
        LAST_X_YEARS_Y_MONTHS('L', LastX.class), //
        PREVIOUS_WEEK('A', PreviousWeek.class), //
        PREVIOUS_MONTH('B', PreviousMonth.class), //
        PREVIOUS_QUARTER('C', PreviousQuarter.class), //
        PREVIOUS_YEAR('Z', PreviousYear.class), //
        CURRENT_WEEK('W', CurrentWeek.class), //
        CURRENT_MONTH('M', CurrentMonth.class), //
        CURRENT_QUARTER('Q', CurrentQuarter.class), //
        YEAR_TO_DATE('X', YearToDate.class), //
        SINCE_X('S', SinceX.class), //
        FROM_X_TO_Y('F', FromXtoY.class), //
        YEAR_X('Y', YearX.class);
    }
```

However, I did not take over the GSON serialization. Why? For reasons of backward compatibility, the old code anyway has to remain. So it did not feel it actually simplified things to have two serialization implementations available (yes, the writing of the old code could go, but not the reading). And the code for reading was now doing an indexOf search for the enum type instead of looking at the first character - does not seem more efficient. Furthermore, the reporting periods are embedded into many strings: to store the last selected periods, the period selected for specific columns, the period selected for a dashboard widget. I would have to make sure that all of them do not collide with the JSON characters, etc. All of this seem to make it easier to keep the one letter code (which it is for most of the reporting periods anyway).

I apologize that I was not more active watching the discussion in the PR. Too busy at the moment. Otherwise I would have pointed it out earlier. Next time. But please continue contributing features. That is very much appreciated.

buchen:
BTW, I placed the week/month/quarter/year options in one line:

Bildschirmfoto 2021-07-11 um 10 37 00

pat-s:
Thanks @buletti and @buchen for your work on this, highly looking forward to this feature!

PR #2305, opened 2021-06-27
Hcpi trade calendar

RomanLangrehr:
Blocked, depends on #2302 (I messed a bit up, logically it should only depend on #2301)

Use first of the month trade calendar for HCPI indices.


buchen:
rebased and merged. Thanks.

PR #2313, opened 2021-07-04
time machine: add 'previous trading day' shortcut

buletti:
Small change to add a drop down entry for the "previous trading day" in the time machine of the Statement of Assets view.
It's mainly for convenience reasons. Date pickers suck and "yesterday" is a common enough use case to warrant a shortcut IMHO. 

pp_tm_ptd


buchen:
squashed and merged


I did two minor changes: first, if the previous trading day is selected, then the calendar should be switched to ```Images.CALENDAR_ON``` (e.g. show the user that some day other than today is selected). second, by default an action is active so no need for the ```ifPresentOrElse``` statement (changed it to ```ifPresent```).

PR #2319, opened 2021-07-08
Optional linear interpolation of security quotes

RomanLangrehr:
Lets the user select how to deal with gaps in historical quote data (take last and linear interpolation are available for now).
“Take last” (which is used right now implicitly) stays the default option.
I think this is can be usefull when rather few quotes are available, e.g., for the HCPI indices (that are only updated once per month) or maybe for real estate (I saw that some people are adding this to PP: https://der-finanzfisch.de/immobilien-in-portfolio-performance-abbilden/ )

In the GUI I added this to the Historical Quote Provider Page:
image

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/SecurityPriceLookupTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/SecurityPriceLookupTest.java
@@ -6,6 +6,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import java.time.LocalDate;
 
 import name.abuchen.portfolio.SecurityBuilder;
+import name.abuchen.portfolio.model.SecurityPriceInterpolator.LinearSecurityPriceInterpolator;
 
 import org.junit.Before;
 import org.junit.Test;
@@ -39,6 +40,7 @@ public class SecurityPriceLookupTest
         assertThat(security.getSecurityPrice(LocalDate.parse("2014-11-06")).getValue(), is(5L));
     }
 
+
     @Test
     public void testLookupIfHistoricQuotesContainGaps()
     {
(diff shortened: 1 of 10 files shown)


buchen:
First about the **feature** itself: I kind of understand the feature with respect to the inflation rate. But I also have to admit that I have not seen it done this way anywhere? Do you have other sources (web sites, articles, papers, applications) that do it that way?
I am not so sure for other assets: wouldn't you want to see if there are gaps in the underlying data and then be able to fix it? Do you see this impacting other metrics of the asset such as volatility?


Once we discussed and agreed on the feature, then we can talk about the **implementation**:

I added two HVPI benchmarks - one with linear interpolation. What I noticed is that the two data series are not identical at the end of the diagram. Intuitively, I would expect the two data series to touch on every first of the month because that's where no interpolation is happening.

Bildschirmfoto 2021-07-18 um 15 01 16

Looking at the UI elements, I wonder if it would make more sense to have a) a checkbox and b) place that checkbook to the right (above/below the "show server response" button). Already today, the JSON data provider has a lot of configuration parameters and with this change, no table of prices is visible anymore.

Bildschirmfoto 2021-07-18 um 15 03 34

Then about the serialization: I try to avoid that the implementation leaks into the XML (it did it before and that therefore I know how difficult that is to change). Right now, the XML contains the implementation class. I propose to add an explicit attribute (either boolean or enum). Typically, I would attach this as "SecurityProperty" to the security, but in this case it is very performance sensitive as the lookup of prices is called very often.

```xml
    <security>
      <uuid>e0a12b7a-d901-4ffb-9ea2-4112dd9780c1</uuid>
      <name>Deutschland (HVPI) - linear</name>
....
      <interpolator class="name.abuchen.portfolio.model.SecurityPriceInterpolator$LinearSecurityPriceInterpolator"/>
    </security>
```

And finally because it changes the serialization format, you need to bump of the version number in Client and adopt the migration code in ClientFactory (empty)

RomanLangrehr:
I agree with your points about the implementation and GUI, but let’s discuss the feature itself first:

> I kind of understand the feature with respect to the inflation rate. But I also have to admit that I have not seen it done this way anywhere? Do you have other sources (web sites, articles, papers, applications) that do it that way?
At least most websites show their charts using linear interpolation (e.g. Stiftung Warentest, Morningstar) when they have only monthly data. But I think most platforms use only EOD data or only EOM data (and not both mixed), which already avoids some issues.

> I am not so sure for other assets: wouldn't you want to see if there are gaps in the underlying data and then be able to fix it?
Yes, but maybe there is no data available for the gaps (like for the HCPI indices). And for assets that have relativly stable returns (money market funds, short term bonds, etc...), I think it is better to use linear interpolation than using the last quote.

> Do you see this impacting other metrics of the asset such as volatility?
Yes, linear interpolation will decrease volatility when there are missing quotes (no quote for a day where there should be a quote according to the trade calendar). But I think this would be desirable.

buchen:
Hi @RomanLangrehr,

okay, I get your argument: ideally the data is consistent - EOF, EOM. The interpolation is one way to make it more consistent - alas, one needs to know the implications. So let's add this feature.

Do I see this correct that the implementation is now in #2320 and we can close this PR?

PR #2321, opened 2021-07-08
Annualized TTWROR.

### LINKED ISSUE #427: Annuallisierter TTWROR (Performance Übersicht / Vermögensaufstellung)
> PS: Ebenfalls könnte man zusätzlich auch den TTWOR noch annualisiert [...](Period Return, Annualized Return)

RomanLangrehr:
Add widget for annualized TTWROR.
Changed label for existing TTWROR widget to make clear that it is cummulative.

I think it is quite confusing that PP shows IRR annualized but TTWROR cmmulative. This PR adds the possiblity to show TTWROR in an annualized version, which is much better for comparison.

Fixes #427 
Has also been requested more recently here: https://forum.portfolio-performance.info/t/annualized-ttwror/7064

I also implemented a “cummulative IRR” widget, for symmetry/completness, but I decided not to make a PR for this, because I think that number doesn’t make much sense.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/PerformanceIndex.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/PerformanceIndex.java
@@ -163,6 +163,13 @@ public class PerformanceIndex
         return accumulated != null ? accumulated[accumulated.length - 1] : 0;
     }
 
+    public double getFinalAccumulatedAnnualizedPercentage()
+    {
+        double accumulatedPercentage = getFinalAccumulatedPercentage();
+        long days = getActualInterval().getDays();
+        return Math.pow(1 + accumulatedPercentage, 365d / ((double) days)) - 1;
+    }
+
     public double[] getDeltaPercentage()
     {
         return delta;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -722,6 +722,7 @@ public class Messages extends NLS
     public static String LabelTransfer;
     public static String LabelTransferTo;
     public static String LabelTTWROR;
+    public static String LabelTTWROR_Annualized;
     public static String LabelTTWROROneDay;
     public static String LabelUnknownVersion;
     public static String LabelUnnamedFile;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
@@ -1414,7 +1414,9 @@ LabelSuffix_HICP = \ (HICP)
 
 LabelSuffix_PreTax = \ (before taxes)
 
-LabelTTWROR = True Time-Weighted Rate of Return
+LabelTTWROR = True Time-Weighted Rate of Return (cumulative)
+
+LabelTTWROR_Annualized = True Time-Weighted Rate of Return (annualized)
 
 LabelTTWROROneDay = Last Day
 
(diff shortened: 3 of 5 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/PerformanceIndex.java
```diff
@@ -163,6 +163,13 @@ public double getFinalAccumulatedPercentage()
         return accumulated != null ? accumulated[accumulated.length - 1] : 0;
     }
 
+    public double getFinalAccumulatedAnnualizedPercentage()
+    {
+        double accumulatedPercentage = getFinalAccumulatedPercentage();
+        long days = getActualInterval().getDays();
+        return Math.pow(1 + accumulatedPercentage, 365d / ((double) days)) - 1;
```
Nirus2000, line 170: hmm... and 366 days? (Leap year)
i know... but the little tiny thinks... ;-)

buletti:
```java
int daysOfYear = LocalDate.now().lengthOfYear();
```
Should provide the correct number of days. 365 or 366 depending on the result of `LocalDate.now().isLeapYear()` has.

RomanLangrehr: @Nirus2000 @buletti Thanks for your comments. Actually, I did use 365 days fixed intentionally for the following reasons:

1) The IRR is calculcated also with 365 days fixed. So to make it comparable, it should be the same.
2) It is more common to use a fixed amount of days (usually 365 days or only 360 days) to annualize returns/interest rates.
3) It is more meaninful to annualize using a fixed number of days, because in a leap year you would also expect a slightly larger return than in a non-leap year due to the one additonal day.
4) It is quite hard to come up with a sound definiton how to annualize using the actual number of days when the reporting periode is not completly within one year. (E.g. when the periode is 01.02.2019 – 01.02.2020, how many days would you use per year? 1/12 * 366 + 11/12 * 365 because one month was in a leap year? But that has the strange effect that, assuming a constant return per day, the annualized return gets lower the more days of the reporting periode are in a leap year. Or do you annualize every day's return with respect to the actual number of days in it’s year? That has the strange effect that returns in a leap year have lower weight. (The return within 01.02.2019 – 31.12.2019 has more weight in the time periode 01.02.2019 – 31.01.2020 than in the time periode 01.01.2019 – 31.12.2019))

So you should always annualize with respect to a fixed number of days per year. I suggest using 365 days, because that has been used also for IRR. Alternatives are 360 days (common in Eurozone) or `365.242` days (average days per year).

Nirus2000: Hallo @RomanLangrehr 
A really very good explanation for the decision of a fixed number 365. Thx.

buletti: Oh ok. This whole topic of annualization seems way more complex than I imagined. I quickly checked [investopedia](https://www.investopedia.com/) and there are lots of details and nuances about that topic.
One major point is that annualization is often not used to sum up a past years performance, but to project the future performance of the current year. It's used to interpolate the performance at the end of the current year based on already known data of the year. In this case you always have well established time intervals that start on the 1st of January and end on 31th of December. There would be no intervals that cross the year boundary like 01.02.2019 – 01.02.2020 as mentioned above.

Here's a few interesting [investopedia](https://www.investopedia.com/) links
- [Annualize](https://www.investopedia.com/terms/a/annualize.asp)
- [Annual Basis](https://www.investopedia.com/terms/a/annual-basis.asp)
- [Annualized Total Return](https://www.investopedia.com/terms/a/annualized-total-return.asp)
- [Annual Percentage Yield](https://www.investopedia.com/terms/a/apy.asp)

Also [TIL about GIPS](https://www.investopedia.com/terms/g/gips.asp):
It seems there is a standards document called "Global Investment Performance Standards" that might have a definition for a "TTWOR with a yearly time interval that can be used for comparing past performance". Might be worth a look, if it is available.

RomanLangrehr: @buletti Thanks for your links! I didn’t know about the GIPS standard before, quite good to know that it exists.

Unfortunately, it seems that none of the article discusses the specific issue of how many days to use.

buchen:
Rebased, merged.

I have added the annualized return also to the security performance and the statement of assets views.

PR #2325, opened 2021-07-10
New quote data provider for CoinGecko (cryptocurrencies)

kasch-code:
Hi,

A added a new quote data provider for CoinGecko (https://www.coingecko.com).
It allows to fetch the historical prices for a wide range of crytocurrencies.

### DIFF
--- a/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
+++ b/name.abuchen.portfolio/META-INF/services/name.abuchen.portfolio.online.QuoteFeed
@@ -2,6 +2,7 @@ name.abuchen.portfolio.online.impl.ManualQuoteFeed
 name.abuchen.portfolio.online.impl.AlphavantageQuoteFeed
 name.abuchen.portfolio.online.impl.BitfinexQuoteFeed
 name.abuchen.portfolio.online.impl.BinanceQuoteFeed
+name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed
 name.abuchen.portfolio.online.impl.FinnhubQuoteFeed
 name.abuchen.portfolio.online.impl.EurostatHICPQuoteFeed
 name.abuchen.portfolio.online.impl.ECBStatisticalDataWarehouseQuoteProvider
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/CoinGeckoQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/CoinGeckoQuoteFeed.java
@@ -1,0 +1,213 @@
+package name.abuchen.portfolio.online.impl;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.text.MessageFormat;
+import java.text.ParseException;
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Optional;
+
+import org.json.simple.JSONArray;
+import org.json.simple.JSONObject;
+import org.json.simple.JSONValue;
+
+import name.abuchen.portfolio.Messages;
+import name.abuchen.portfolio.PortfolioLog;
+import name.abuchen.portfolio.model.LatestSecurityPrice;
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.model.SecurityPrice;
+import name.abuchen.portfolio.online.QuoteFeed;
+import name.abuchen.portfolio.online.QuoteFeedData;
+import name.abuchen.portfolio.util.WebAccess;
+
+public final class CoinGeckoQuoteFeed implements QuoteFeed
+{
+    public static final String ID = "COINGECKO"; //$NON-NLS-1$
+
+    private HashMap<String, String> tickerIdMap;
+
+    @Override
+    public String getId()
+    {
+        return ID;
+    }
+
+    @Override
+    public String getName()
+    {
+        return "CoinGecko"; //$NON-NLS-1$
+    }
+
+    @Override
+    public Optional<LatestSecurityPrice> getLatestQuote(Security security)
+    {
+        QuoteFeedData data = getHistoricalQuotes(security, false, LocalDate.now());
+
+        if (!data.getErrors().isEmpty())
+            PortfolioLog.error(data.getErrors());
+
+        List<LatestSecurityPrice> prices = data.getLatestPrices();
+
+        if (prices.isEmpty())
+            return Optional.empty();
+
+        Collections.sort(prices, new SecurityPrice.ByDate());
+
+        return Optional.of(prices.get(prices.size() - 1));
+    }
+
+    @Override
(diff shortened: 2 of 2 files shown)


buchen:
Cool. Thank you very much.

kasch-code:
Hi @eFFecTM,

Thanks for your feedback!

I actually was thinking about using the Coingecko ID, but this would not be very intuitive to use (as everybody in the crypto world is used to the tickers like BTC and ETH).

The first thing which comes to my mind to solve this, would be to add the Coingecko ID as an additional field (reusing the ISIN or WKN field?).

Then the following logic would apply:
- If the Coingecko ID is given for the Security, it will be used directly to retrieve the Quotes (without any mapping)
- Otherwise, the ticker symbol will be used as it is implemented at the moment

This way, both the common and edge cases (your BIT example) would be supported. 

One thing I don't like about this approach is misusing the ISIN or WKN field for storing the Coingecko ID. A possible other approach would be to save the ID in a custom attribute on the security.

What's your opinion on this?

eFFecTM:
Using one of those fields is a decent option, I did not consider that. I wonder though whether there is still a good reason to have both fields exist? I found out that WKN (only exists in Germany) and is a part of the ISIN already: https://deutsche-boerse.com/dbg-en/our-company/know-how/glossary/glossary-article/WKN-246456

Another thing is that ISIN does not exist for coins as far as I know, so we could indeed reuse that without causing any conflict. That would be the easy dirty fix.

And for a more clean fix, maybe like you say we need a custom attribute for this, I do not think it will look clean if there is a Coingecko ID shown in either of the first 2 tabs (as it is not applicable for other securities), so maybe we can have this custom attribute only shown on the Historical/Latest Quote tab when selecting Coingecko as the quote feed provider? Although, I am not sure if it is easily possible to have the GUI dynamically change depending on the selected provider.
Filling in this field would then allow the user to provide the Coingecko ID which takes priority over the symbol.

kasch-code:
Implementing this on the "Latest Quote" actually seems to be a good idea. There already is dynamically generated content on this screen based on the selected provider.

For example:
image

I will have a look on how this is implemented and try to do the same for the Coingecko provider.

eFFecTM:
Oh I was too lazy to check if the other providers had something dynamic going on already, but yea that is exactly what I meant! Thanks for creating the issue.

rbaeumler:
Also stumbled upon this issue today. Is there any progress already on the proposed solution with the custom fields for the quote provider?

kasch-code:
Sorry, I was quite busy the last weeks. But I will start working on it in the next days. 😃

eFFecTM:
Based on your screenshots I think this is exactly what I was looking for. Thank you for implementing it, feel free to make the PR already. I trust it will work fine based on the screenshots 👍

rbaeumler:
Hi @kasch-code, thanks for your implementation. I've tested it with my case and the data was pulled as expected.

I seem to have an issue with display of the tooltip when hovering with the mouse over the coin ID input field. An exception popup is shown, Not sure of this an issue caused by my build (I was building with maven on the command line, as I don't have an Eclipse environment ready for building). The stacktrace shows this:

```
java.lang.IllegalArgumentException: Argument cannot be null

	at org.eclipse.swt.SWT.error(SWT.java:4874)

	at org.eclipse.swt.SWT.error(SWT.java:4808)

	at org.eclipse.swt.SWT.error(SWT.java:4779)

	at org.eclipse.swt.graphics.GC.stringExtent(GC.java:4950)

	at name.abuchen.portfolio.ui.util.SWTHelper.stringWidth(SWTHelper.java:87)

	at name.abuchen.portfolio.ui.util.InfoToolTip.createToolTipContentArea(InfoToolTip.java:60)

	at org.eclipse.jface.window.ToolTip.toolTipShow(ToolTip.java:345)

	at org.eclipse.jface.window.ToolTip.toolTipOpen(ToolTip.java:469)

	at org.eclipse.jface.window.ToolTip.toolTipCreate(ToolTip.java:334)

	at org.eclipse.jface.window.ToolTip$ToolTipOwnerControlListener.handleEvent(ToolTip.java:609)

	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)

	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)

	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1060)

	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4054)

	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3626)

	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)

	at org.eclipse.jface.window.Window.open(Window.java:799)

	at name.abuchen.portfolio.ui.views.SecuritiesTable$AbstractDialogAction.run(SecuritiesTable.java:1135)

	at org.eclipse.jface.action.Action.runWithEvent(Action.java:474)

	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:580)

	at org.eclipse.jface.action.ActionContributionItem.lambda$4(ActionContributionItem.java:414)

	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)

	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)

	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1060)

	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4054)

	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3626)

	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1150)

	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)

	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1041)

	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:155)

	at org.eclipse.e4.ui.internal.workbench.swt.E4Application.start(E4Application.java:168)

	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)

	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)

	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)

	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:401)

	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:255)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)

	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

	at java.base/java.lang.reflect.Method.invoke(Method.java:566)

	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:659)

	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:596)

	at org.eclipse.equinox.launcher.Main.run(Main.java:1467)
```

kasch-code:
Hi @rbaeumler, I added a tooltip to the field. So this should be fine now.

rbaeumler:
Hi @kasch-code , I tested it with the latest code and it worked. Many tanks again for your feature implementation!

PR #2327, opened 2021-07-11
Cumulative return widget

RomanLangrehr:
This PR adds a widget that shows the cumulative return.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -583,6 +583,7 @@ public class Messages extends NLS
     public static String LabelInvestedCapital;
     public static String LabelInvestmentPlans;
     public static String LabelIRR;
+    public static String LabelCumulativeReturn;
     public static String LabelLanguage;
     public static String LabelLanguageAutomatic;
     public static String LabelLayout;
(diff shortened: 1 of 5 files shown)


buchen:
If I read the code right, this change defines "cumulative return" as
> **cumulative return = delta at end of reporting / investments at end of reporting period**

or
> **cumulative return = (valuation - investments) / investments**

This performance metric does **not** take transferals into account that happened during the reporting period (which is something that TTWROR and IRR do). Therefore I think the metric is not meaningful. The metric will depend very much on the point of investments during the reporting period. For example, it will be much too high if there is a withdrawal of funds at the end of the period.

I have quickly created a simple example to demonstrate the effect:

Two transactions:
Bildschirmfoto 2021-07-17 um 21 33 39

Because of the withdrawal, the cumulative return is very high:
Bildschirmfoto 2021-07-17 um 21 34 20

This is the sample file: [example2327.zip](https://github.com/buchen/portfolio/files/6835207/example2327.zip)


Can you share more details why you think TTWROR and/or IRR do not show/reflect what you want to see?

I am open to add new metrics - as the saying goes: if you ask 10 economists for an opinion on the right performance metric, you get 11 answers...

RomanLangrehr:
Thanks for your helpful comments.

The idea behind the cummulative return widget was that most banks show the cummulative return in their online banking interface (but usually just for the open trades) and I wanted to have a version that also takes into account closed trades. But, as you pointed out, this number doesn’t make much sense if you sell an asset and don’t reinvest the money. Thus I now think, it is better not to have this as performance measure in PP.

PR #2328, opened 2021-07-11
Fix invested capital widget

RomanLangrehr:
The portfolio value at begin of the reporting period is no longer counted as invested capital (for the reporting period).

Given the title “Invested Capital _for reporting period_”, I don’t see why investments before the reporting periode (and their returns) should count towards this.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
@@ -88,7 +88,7 @@ public enum WidgetFactory
                                     .with(Values.Amount) //
                                     .with((ds, period) -> {
                                         long[] d = data.calculate(ds, period).calculateInvestedCapital();
-                                        return d.length > 0 ? d[d.length - 1] : 0L;
+                                        return d.length > 0 ? d[d.length - 1] - d[0]: 0L;
                                     }) //
                                     .withBenchmarkDataSeries(false) //
                                     .build()),


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -88,7 +88,7 @@
                                     .with(Values.Amount) //
                                     .with((ds, period) -> {
                                         long[] d = data.calculate(ds, period).calculateInvestedCapital();
-                                        return d.length > 0 ? d[d.length - 1] : 0L;
+                                        return d.length > 0 ? d[d.length - 1] - d[0]: 0L;
```
buchen, line 91: The method #calculateInvestedCapital returns an array that cumulates the inbound/outbound transfers (=investments) of the given period of time. Therefore I think the widget does what it says: report the investments for the given reporting period.

Technically this changes nothing, because the first day (index 0) of the array will never have investments because the calculation starts with the valuation at the end of the first day.

I propose the following: can you add an example that shows what is currently displayed and why you expect a different value? Then open the PR again.

RomanLangrehr: > Technically this changes nothing, because the first day (index 0) of the array will never have investments because the calculation starts with the valuation at the end of the first day.

This is not correct, because the first day is the first day of the reporting periode and there might be transactions before the reporting periode.

Here is an example where it makes a difference: 1000 € where invested on 2020-08-03:
image

Now the invested capital widget shows on 2021-08-05 for the past 1 year (so the reporting periode does not contain the buy transaction) that 1000 € where invested (which is not true, there was no investment in the reporting periode):
image

With this PR, it shows the correct value (0 €):
image

RomanLangrehr:
@buchen I want to reopen this PR, but it seems that I can not do so.

chirlu:
You think you are “fixing” a bug, but it is intended to work as it does. You are changing the semantics.

chirlu:
I made a new, separate widget that does what you want in #2381.

RomanLangrehr:
@chirlu A okay, thanks. I think the name of the widget is a bit missleading, bit I also don’t have a better suggestion.

RomanLangrehr:
@chirlu A okay, thanks. I think the name of the widget is a bit missleading, bit I also don’t have a better suggestion.

RomanLangrehr:
@chirlu A okay, thanks. I think the name of the widget is a bit missleading, bit I also don’t have a better suggestion.

Issue #2329, opened 2021-07-14
SelfWealth PDF-Importer


flywire:
One transaction per file. What more can I say?

```
PDF author: ''
PDFBox Version: 1.8.16
-----------------------------------------
SelfWealth Limited ABN: 52 154 324 428 AFSL 421789 W: www.selfwealth.com.au E: support@selfwealth.com.au
This trade was executed and cleared by OpenMarkets Australia Ltd ABN 38 090 472 012,
AFSL 246 705, Market Particpant of ASX, CHI­X and NSX.
Buy Confirmation
MR JOHN DOE Account Number: 1234567
JOHN DOE A/C Reference No: T20210701123456­-1
1 LONG ROAD Trade Date: 1 Jul 2021
SYDNEY NSW Settlement Date: 5 Jul 2021
2000, AUS Market: ASX
WE HAVE BOUGHT ON YOUR ACCOUNT
Quantity Security Code Security Description Price +1 Consideration Currency
25 UMAX BETA S&P500 YIELDMAX 12.40 $312.50 AUD
Brokerage* $9.50 AUD
Adviser Fee* $0.00 AUD
Net Value $322.00 AUD
GST included in this invoice is $0.86
The confirmation is a tax invoice ­ please retain for tax purposes. If this confirmation does not correspond with your records please contact us immediately at
support@selfwealth.com.au
Settlement Instructions All consideration and any information or documents required by OpenMarkets must be provided to OpenMarkets by 9am AEST on the Settlement
Date. This transaction will be settled from your linked cash account or in accordance with your instructions on the Settlement Date.
Contract Comments
Ex Dividend
* Inclusive of GST
+1 Standard Financial Rounding Applied (if applicable)
This confirmation is provided to you by each of SelfWealth and OpenMarkets. The Brokerage and Adviser fees set out in this confirmation are charged by SelfWealth.
OpenMarkets has not charged you any fees for the above transaction(s). The above transaction(s) and this confirmation are issued subject to the directions, decisions
and requirements of the operator of the relevant Market, ASIC Market Integrity Rules, the operating rules of the relevant Market, and, where relevant, the Clearing
Rules of the relevant Clearing Facility and the Settlement Rules of the relevant Settlement Facility, the customs and usages of the relevant Market and the correction of
errors and omissions. If this confirmation relates to multiple transactions, those transactions may have been completed on ASX or CHI­X.”
Generated At: 5 Jul 2021 16:30:01 PM Page: 1 of 1  
```
[...]

PR #2340, opened 2021-07-17
Create SelfWealthPDFExtractor.java

flywire:
SelfWealthPDFExtractor in development.

See Issue #2329

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SelfWealthPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SelfWealthPDFExtractor.java
@@ -22,6 +22,7 @@ public class SelfWealthPDFExtractor extends AbstractPDFExtractor
         super(client);
 
         addBankIdentifier("SelfWealth"); //$NON-NLS-1$
+        addBankIdentifier("SelfWealth Limited ABN: 52 154 324 428 AFSL 421789 W: www.selfwealth.com.au E: support@selfwealth.com.au"); //$NON-NLS-1$
 
         addBuySellTransaction();
     }
@@ -35,7 +36,7 @@ public class SelfWealthPDFExtractor extends AbstractPDFExtractor
     @Override
     public String getLabel()
     {
-        return "SelfWealth"; //$NON-NLS-1$
+        return "SelfWealth Limited ABN: 52 154 324 428 AFSL 421789 W: www.selfwealth.com.au E: support@selfwealth.com.au"; //$NON-NLS-1$
     }
 
     private void addBuySellTransaction()
@@ -66,7 +67,7 @@ public class SelfWealthPDFExtractor extends AbstractPDFExtractor
                 })
 
                 // JOHN DOE A/C Reference No: T20210701123456­-1
-                .section("note")
+                .section("note").optional()
                 .match(" Reference No: (?<note>.*)$")
                 .assign((t, v) -> {
                     t.setNote(asNote(v.get("note")));
@@ -95,36 +96,12 @@ public class SelfWealthPDFExtractor extends AbstractPDFExtractor
                 })
 
 
-                // Brokerage* $9.50 AUD
-                .section("brokerage_fee")
-                .match("^Brokerage Fee//* //$(?<brokerage_fee>.*) [\\w]{3}$")
-                // Adviser Fee* $0.00 AUD
-                .section("adviser_fee")
-                .match("^Adviser Fee//* //$(?<adviser_fee>.*)$")
-                
-                fees = brokerage_fee + adviser_fee
-                .assign((t, v) -> {
-                    t.setAmount(asAmount(v.get("amount")));
-                    t.setCurrencyCode(asCurrencyCode(v.get("currency")));
-                })
-
-
                 // Net Value $322.00 AUD
                 .section("amount")
                 .match("^GST included in this invoice is //$(?<amount>.*) [\\w]{3}$")
                 .assign((t, v) -> {
                     t.setAmount(asAmount(v.get("amount")));
                     t.setCurrencyCode(asCurrencyCode(v.get("currency")));
-                })
-
-
-                // GST included in this invoice is $0.86
-                .section("gst")
-                .match("^GST included in this invoice is //$(?<gst>.*)$")
-                .assign((t, v) -> {
-                    t.setAmount(asAmount(v.get("amount")));
-                    t.setCurrencyCode(asCurrencyCode(v.get("currency")));
-                })
 
 
                 .wrap(BuySellEntryItem::new);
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SelfWealthPDFExtractor.java
```diff
@@ -0,0 +1,184 @@
...
+                .assign((t, v) -> {
+                    if (v.get("type").equals("Sell"))
+                    {
+                        t.setType(PortfolioTransaction.Type.SELL);
+                    }
+                })
+
+                // JOHN DOE A/C Reference No: T20210701123456­-1
+                .section("note")
+                .match(" Reference No: (?<note>.*)$")
```
Nirus2000, line 70: a note is everytime optional...
so change the line 
`.section("note")` to `.section("note").optional()`

change the line to
`.match("^.* Reference No: (?<note>.*)$")`

flywire:
> If you want, I can also create the importer for you. But for this I need more PDF debugs for buying, selling, dividends, and so on.

Good idea. SelfWealth only have buy or sell transactions, no dividends. Both added to repo.

Also note: `2000, AUS Market: ASX` "ASX" is Australian exchange.

Nirus2000:
So close this pull-request an put your PDF-Debug-Files in the issue #2329

flywire:
> Hallo @flywire
> So close this pull-request an put your PDF-Debug-Files in the issue #2329

Sorry, your communication quoted above is confusing.

I understood "so start in draft-request" meant Pull Request (??).
I do not know German or Java so it is what we call "pushing shit up-hill" (hard with little chance of success).

With your help this can be a good English example for others to follow.
Like https://leafletjs.com/examples.html written for non-programmers to follow.

If you email me I can send you a link which explains Australian reports.

flywire:
It's a bit had without knowing German or Java but with a bit of a tidy up this must be pretty close.

ismogal12:
This looks good start for our ASX(Australian Stock Exchange). 
I hope one can be created for CommSec PDF as well. I am not sure if this is the right place to put feature requests. 

Just a few features for ASX specific:
1. CGT discount if the holding period is more than 12 months. 
2. Ability to calculate LIFO, I think currently it only does FIFO
3. There is a split event but no Consolidation (which may be specific to ASX - ex: 10 shares are consolidated into 1).

flywire:
> I hope one can be created for CommSec PDF as well. I am not sure if this is the right place to put feature requests.

No, you need a new issue with sample buy/sell/dividend dumps.

Raise the ASX specific features on https://forum.portfolio-performance.info/c/english/16

ismogal12:
> > I hope one can be created for CommSec PDF as well. I am not sure if this is the right place to put feature requests.
> 
> No, you need a new issue with sample buy/sell/dividend dumps.
> 
> [#2329 (comment)](https://github.com/buchen/portfolio/issues/2329#issuecomment-881902163)
> 
> > For the structur in addBuySellTransaction or other transactions....
> > 
> > 1. security with security currency
> > 2. date
> > 3. amount and currency
> > 4. sometimes forex calculation if security not in amount currency
> > 5. notes (...)
> 
> Raise the ASX specific features on https://forum.portfolio-performance.info/c/english/16

Thanks, I have raised the feature request - https://github.com/buchen/portfolio/issues/2351, and a topic on the forum is not yet visible. 

Here is an example of the PDF how it looks like: https://www.coursehero.com/file/63931397/Contract-98505191pdf/

buchen:
Alright, thanks @flywire  for the contribution and thanks @Nirus2000 for commenting on the change and helping.

I have now picked up the code and fixed it as good as I could understand it.

A couple for remarks:

* I remove the GST line - it apparently was already included in the other line items
* The "Sell" file was somehow corrupted - it had strange space characters which led to the patterns not working. I did remove them, but if the PDF actually creates this space characters, then we need updated regex pattern.

Bildschirmfoto 2021-07-25 um 21 56 44

* The reason I fixed the sample file and not the regex expressions was that I had the impression the "sell" file was manually created. At least in Germany - if you sell something - you get the proceeds **without** brokerage fees. :smile: Therefore I have updated the file accordingly. If this different in Australia, then PP cannot treat "brokerage fees" as fees we would have to add them to the original proceeds.
```
Sell Confirmation
[...]
397 WPL WOODSIDE PETROLEUM 21.88 $8,686.36 AUD
Brokerage* $9.50 AUD
Adviser Fee* $0.00 AUD
Net Value $8,695.86 AUD
```
* I registered the importer in the ```PDFImportAssistent``` because otherwise it is not applied to PDF imports

I am happy to merge more contributions. Please make sure the Github Actions workflow compiles.

flywire:
Hmm

> A couple of remarks:

* gst - tax, I think will be revisited another time 😄
* The "Sell" file was somehow corrupted - yes, I wasn't aware of that problem - **see below**. (Buy dump unintentionally fixed when deidentified.)
* in Germany - if you sell something - you get the proceeds **without** brokerage fees. 😄 I made a bad mistake - correct way is sell for $100.00, fees $10.00 get $90.00.

---

Portfolio Performance
Version: 0.54.0 (Jul. 2021)
Platform: win32, x86_64
Java: 11.0.4+11-LTS, Azul Systems, Inc.
Locale: AU

The corrupted dump characters (`c2 a0` instead of `20`) occur with any dump of buy or sell transactions (tested 20 files including buy and sell). Notepad++ v8.1.1 (64bit), File, Open, PlugIns, Hex-Editor, View in Hex:

image

I can open pdf file in Microsoft Edge Version 92.0.902.55 (Official build) (64-bit) Win10, select/copy/paste into Notepad++ and viewed as above has no strange characters (same 20 files).

flywire:
Interestingly:
- `java -jar pdfbox-app-1.8.10.jar ExtractText PrivateBuy.pdf` uses `a0` as space character
- `java -jar pdfbox-app-2.0.24.jar ExtractText PrivateBuy.pdf` uses `c2 a0` as space character, same `v1.8.16`

chirlu:
Unicode character U+00A0 is the no-break space. In UTF-8, it is represented as the byte sequence C2 A0. In ISO 8859-1 and other (older) one-byte encodings, it is represented as A0.

flywire:
The SelfWealth PDF files have been decoded and issue investigated: https://issues.apache.org/jira/browse/PDFBOX-5247

Indeed every space is a [Non-breaking_space](https://en.wikipedia.org/wiki/Non-breaking_space)

> if you don't like it, do your own postprocessing. It's unusual but valid.

@buchen I think this is a good case for post-processing rather than have hidden codes in the test files. What do you think?

buchen:
> @buchen I think this is a good case for post-processing rather than have hidden codes in the test files. What do you think?

PP has PDFBox version 1.8.16 embedded.

If I understand it, the "purchase" has regular spaces, the "sale" has non-breaking spaces - is that also what you see when converting the PDF documents using the "File -> Import -> Debug: Create Text from PDF" feature?

From my point of view, there are two options:
* adopt the regular expressions (currently looking for regular spaces) - but this is cumbersome if one file has regular spaces, the other non-breaking spaces as it duplicates the regular expressions
* do a post-processing and replace the non-breaking spaces with regular spaces - would we do this on the text output *before* passing it to any converter, or would this be a pre-processing specific to SelfWealth?

flywire:
> If I understand it, the "purchase" has regular spaces, the "sale" has non-breaking spaces - is that also what you see when converting the PDF documents using the "File -> Import -> Debug: Create Text from PDF" feature?

No, they both have non-breaking spaces, see `Updated: SelfWealthBuy01.txt SelfWealthSell01.txt` above your post using hex editor.

My preference is `do a post-processing and replace the non-breaking spaces with regular spaces` and given they are not normally visible (unless in hex) I suggest do it on all files. Would also give same result as select/copy/paste from pdf file.

PR #2342, opened 2021-07-18
Delete or disable unnecessary code in the PDF importers.

### LINKED ISSUE #2341: Delete "Unnecessary Code" in PDF Importer
Eclipse reports that a couple of methods in the category of "Unnecessary Code":

Bildschirmfoto 2021-07-18 um 08 40 13

@Nirus2000 - Can you delete / refactor the code?

Nirus2000:
Delete or disable unnecessary code in the PDF importers
Fixes #2341

This source was only disabled for later implementation of PDF-Debugs.
When we need this, it can be easily activated.


buchen:
Actually, I would prefer if we delete the code. It is not gone forever - it remains in the git history. I feel commented-out code is a code smell - it deteriorates fast because it is not refactored anymore, not formatted right, hides the actual code that I want to see.

BTW, the methods look very identical to me. Maybe it makes sense to have them as generic methods in the super calls (or a helper class)?

PR #2344, opened 2021-07-18
Modify JustTrade PDF-Importer to support new sell transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-justtrade/10853/21


buchen:
merged 😄 

I will now start preparing a release... but I will only come around to actually publishing it later tonight

PR #2349, opened 2021-07-24
Allow searching securities when Yahoo symbol search fails

### LINKED ISSUE #2346: Searching instrument doesn't work
**Describe the bug**

I followed [this](https://indexfundinvestor.eu/2019/06/27/how-to-track-your-investments-with-portfolio-performance/) tutorial, but I have an issue with finding IWDA instrument. If I type in search instrument 'IWDA', it doesn't find instrument as [here](https://indexfundinvestor.eu/wp-content/uploads/2019/06/add_instrument.png).

**To Reproduce**
1. Go to Securities
2. Click on New Instrument
3. Type 'IWDA'
4. See error in screenshot

**Expected behavior**

**Screenshots**

Screenshot 2021-07-22 at 10 04 12

 - OS: macOS
 - Version 0.54.0

alvarobp:
Fixes #2346 

It looks like the service used to get the symbols for securities (s.yimg.com) is no longer working, or at least it hasn't been working for a few days.

I was able to add new securities to my portfolio by compiling the application with this patch. In my opinion, not being able to fetch the symbol (image) of a security should not prevent a user from adding a new security.

This is just a workaround, for the time being, while this service is not working. I think a longer term solution would be to find whether this service will be back or how to get the symbols in a more reliable way, maybe from another service.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooSymbolSearch.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooSymbolSearch.java
@@ -13,6 +13,7 @@ import name.abuchen.portfolio.model.ClientSettings;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.online.SecuritySearchProvider.ResultItem;
 import name.abuchen.portfolio.util.WebAccess;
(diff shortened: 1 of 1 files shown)


buchen:
Thanks. The error handling is better. I'll have to check how to update the search code - it seems the method is not available easily anymore

PR #2357, opened 2021-07-29
Improvements to data quality pane

chirlu:
A few improvements to the data quality pane:
* Say which trade calendar was used for determining completeness and missing/unexpected dates.
* Name the holiday that causes a quote to be unexpected, instead of just giving the date.
* Make the missing and unexpected tables equal size, and use the whole width. Especially important since they can’t be resized by the user, unlike columns in a single table.

This is what it looks like:
data quality pane with improvements
For review, I recommend looking at the individual commits.


buchen:
Thanks 😄

PR #2366, opened 2021-08-03
fix hello bank pdf import

wegwart:
- Abrechnung Handel / Geschäftsart: Kauf
- Abrechnung Dauerauftrag / Geschäftsart: Kauf aus Dauerauftrag


buchen:
@wegwart  thanks for the contribution

BTW, if I find time, I'll remove the launch configurations. Those are always OS-specific and as I am on macOS, that's what I maintain. In my development environment, I only use the launch configuration with the plugin #1913

PR #2369, opened 2021-08-04
Various fixes for parsing of FintechGroup PDF documents

rumpeltux:


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
```diff
@@ -172,206 +174,134 @@ private void addBuySellTransaction()
...
-                        BigDecimal exchangeRate = asExchangeRate(v.get("exchangeRate"));
-                        exchangeRate = BigDecimal.ONE.divide(exchangeRate, 10, RoundingMode.HALF_DOWN);
+    private void invertFeesForRefund(BuySellEntry t) {
+        if (t.getPortfolioTransaction().getType() == PortfolioTransaction.Type.SELL) {
+            t.setType(PortfolioTransaction.Type.BUY);
+        } else if(t.getPortfolioTransaction().getType() == PortfolioTransaction.Type.BUY) {
+            t.setType(PortfolioTransaction.Type.SELL);
+        }
+        List<Unit> invertedUnits = t.getPortfolioTransaction().getUnits().map((Unit u) -> u.split(-1))
+                        .collect(Collectors.toUnmodifiableList());
```
buchen, line 269: ```toUnmodifiableList``` is only available with Java 10, but I am keeping this bundle compatible with Java 8 so that I can downport any changes to the PDF importer also to the x86 branch of PP. But it this case it is also okay to use the normal ```toList```

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
```diff
@@ -172,206 +174,134 @@ private void addBuySellTransaction()
...
-                    {
-                        Money taxRefundFX = Money.of(asCurrencyCode(v.get("currency")), asAmount(v.get("taxRefund")));
+                    invertFeesForRefund((BuySellEntry) t);
+                });
+    }
 
-                        BigDecimal exchangeRate = asExchangeRate(v.get("exchangeRate"));
-                        exchangeRate = BigDecimal.ONE.divide(exchangeRate, 10, RoundingMode.HALF_DOWN);
+    private void invertFeesForRefund(BuySellEntry t) {
+        if (t.getPortfolioTransaction().getType() == PortfolioTransaction.Type.SELL) {
```
buchen, line 263: I can't see the build logs at the moment, but I expect check style will complain about the curly braces to be on the new line

buchen:
In general, the change looks good to me - just this two remarks.
I'll also give @Nirus2000 a chance to comment as he was busy on this part of the code 😄

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
```diff
@@ -172,206 +174,134 @@ private void addBuySellTransaction()
...
+            t.setType(PortfolioTransaction.Type.BUY);
+        } else if(t.getPortfolioTransaction().getType() == PortfolioTransaction.Type.BUY) {
+            t.setType(PortfolioTransaction.Type.SELL);
+        }
+        List<Unit> invertedUnits = t.getPortfolioTransaction().getUnits().map((Unit u) -> u.split(-1))
```
buchen, line 268: This is not the indented use of the split method. PP assumes the values are always positive. It is the type of the transactions (or the unit) that tells what to expect. Either TAX or TAX_REFUND - but both values are positives. I know it could be implemented differently, but, hey, that's the way it is now.

Which also unfortunately means the "Storno" is not easy to handle for PP. Usually I would say: just delete the previous purchase order. But this kind of action is not supported by the import. @Nirus2000 do you know how we handled that with other importers?

Nirus2000: "Stornos" are currently not processed by any importer.

Nirus2000:
Hallo @rumpeltux and @buchen 
I think the changes are actually quite good. Some small fixes...

Would this be an approach to extend the PDFExtractorUtils.java and move the "processTaxEntries" and "processFeeEntries" with it? This could then relieve some importers of these lines of source code.

With the reduction of the Tax-Refund, I am not so sure how one could move these into the PDFExtractorUtils.java.

A more recent idea was to extend the PDFExtractorUtils.java as an auxiliary class... ;-)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
rumpeltux: Installed checkstyle and fixed the violations.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
rumpeltux: This is obsolete now :)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java
rumpeltux: Reverted the last commit so this now throws an exception asking the user to delete the original booking.
(Before it would just import a refund order as an actual order…)

rumpeltux:
Thanks for the reviews!

I’m super new to this code base, so I’ll leave further refactoring decisions to you.

Nirus2000:
Hallo @rumpeltux 
No no... ;-)
We are happy about everyone who has new ideas and good implementations, so let the keys jingle. I like it!

buchen:
merged :-)

PR #2373, opened 2021-08-07
Linux CSV importer column packing (fixes #1723 / #1536)

### LINKED ISSUE #1723: CSV Import Spalten außerhalb des Fensters nach Wechsel zu Komma
Datei -> Importieren -> CSV

Zu Beginn nach Dateiauswahl:

Nach Wechsel zu Komma:

Warum ist Komma nicht als Standard ausgewählt? Warum landen die Spalten außerhalb, wenn ich das Trennzeichen ändere?

**Auffallend**
Nur die erste Spalte wird extrem breit. Der Rest passt wieder.

**Lösung**
Fenster muss größer gezogen werden und die erste Spalte wieder verkleinert werden.

rumpeltux:
Previous fixes omitted column packing entirely, therefore the table
was invisible upon changing any setting in the dialog. It only became
visible again when the window was resized.
Re-packing the first column seems to fix this.
Hopefully once-and-for-all!


buchen:
> Hopefully once-and-for-all!

let's hope :-) If you come across more Linux glitches, I would be really happy. I am developing on macOS, there are many user on Windows, Linux is the platform that does not get enough attention

PR #2381, opened 2021-08-08
Add a widget for performance-neutral transfers

chirlu:
Create a new widget that displays the savings done during the reporting period, i.e. the performance-neutral transfers. It is similar to the invested capital widget, but does not include the starting value of the portfolio.


buchen:
Dank Dir!

PR #2387, opened 2021-08-15
Add a Saving mode to the Payments (formerly Income & Expenses) report

chirlu:
This does three things:
- Rename “Income & Expenses” to the more general “Payments”, and rename “Earnings (accumulated)”. [See forums discussion.](https://forum.portfolio-performance.info/t/9531) See also the commits for detailed rationale.
- Add a Saving mode to the (newly renamed) Payments report that tabulates inflows/outflows of money (performance-neutral transactions), i.e. the balance of deposits, withdrawals, and inbound/outbound deliveries, to get an overview about how much you are saving (or withdrawing). [See feature request on the forums.](https://forum.portfolio-performance.info/t/5484)
- Some cleanup: remove long unused strings.

I am unsure about dealing with preferences keys, see below.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsView.java
```diff
@@ -36,13 +36,13 @@
...
-    private static final String KEY_YEAR = EarningsView.class.getSimpleName() + "-year"; //$NON-NLS-1$
-    private static final String KEY_MODE = EarningsView.class.getSimpleName() + "-mode"; //$NON-NLS-1$
-    private static final String KEY_USE_GROSS_VALUE = EarningsView.class.getSimpleName() + "-use-gross-value"; //$NON-NLS-1$
-    private static final String KEY_USE_CONSOLIDATE_RETIRED = EarningsView.class.getSimpleName()
+    private static final String KEY_TAB = PaymentsView.class.getSimpleName() + "-tab"; //$NON-NLS-1$
+    private static final String KEY_YEAR = PaymentsView.class.getSimpleName() + "-year"; //$NON-NLS-1$
+    private static final String KEY_MODE = PaymentsView.class.getSimpleName() + "-mode"; //$NON-NLS-1$
+    private static final String KEY_USE_GROSS_VALUE = PaymentsView.class.getSimpleName() + "-use-gross-value"; //$NON-NLS-1$
+    private static final String KEY_USE_CONSOLIDATE_RETIRED = PaymentsView.class.getSimpleName()
                     + "-use-consolidate-retired"; //$NON-NLS-1$
```
chirlu, line 46: Here and in a few other places, the key for storing/looking up preferences changes because it is based on the class name, which itself is changed because the view has long been about far more than just earnings. This means that existing settings won’t be found anymore. What to do about it? I see these options:
- Ignore the issue – users will (once) need to re-set their preferences for the Payments report.
- Convert the old preferences keys to the new ones on loading the settings file. (I don’t know how that could be done.)
- Instead of deriving the keys from the (changing) class name, hardwire the old names.
- Refrain from renaming the classes. That would be transparent to users, but may be confusing to developers as the code would still speak of an “Earnings view” that no longer exists in that form.

buchen: In other places I did option 3: I converted the key from *class.toString* to a plain string that contains the old class name. In other cases, I opted for option 1 and require a re-configuration by the user. For this particular change, both 1 and 3 look good to me

Option 2 is too much effort, and option 4 is not sustainable

PR #2390, opened 2021-08-16
Fix SelfWeath PDF-Importer to support hidden blank characters

Nirus2000:
Fix SelfWeath PDF-Importer to support hidden blank characters
https://forum.portfolio-performance.info/t/pdf-import-from-selfwealth/17399/9

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/selfwealth/SelfWealthBuy02.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/selfwealth/SelfWealthBuy02.txt
@@ -1,0 +1,33 @@
+PDF author: ''
+PDFBox Version: 1.8.16
+-----------------------------------------
+SelfWealth Limited ABN: 52 154 324 428 AFSL 421789 W: www.selfwealth.com.au E: support@selfwealth.com.au
+This trade was executed and cleared by OpenMarkets Australia Ltd ABN 38 090 472 012,
+AFSL 246 705, Market Particpant of ASX, CHI­X and NSX.
+Buy Confirmation
+MR JOHN DOE  Account Number: 1234567
+JOHN DOE A/C Reference No: T20210701123456­1
+1 LONG ROAD Trade Date: 1 Jul 2021
+SYDNEY NSW Settlement Date: 5 Jul 2021
+2000, AUS Market: ASX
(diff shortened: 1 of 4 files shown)


flywire:
**Seems not the preferred solution.** Preferred solution previously discussed would strip the nbsp characters from PDFBox output before applying regex.

---

No reply to my previous unanswered [query asking if I can test code without compiling](https://github.com/buchen/portfolio/pull/2386#issuecomment-896356395) so I assume this is just a regex change. It will retain [invisible] NBSP characters in the file and only cause ongoing issues for users.

Findings of investigation in https://github.com/buchen/portfolio/pull/2340#issuecomment-888269417 with PDFBox developer recommendation:
> if you don't like it, do your own postprocessing. It's unusual but valid.

https://github.com/buchen/portfolio/pull/2340#issuecomment-892377110
@buchen
> From my point of view, there are two options:

https://github.com/buchen/portfolio/pull/2340#issuecomment-892384513
> My preference is do a post-processing and replace the non-breaking spaces with regular spaces and given they are not normally visible (unless in hex) I suggest do it on all files. Would also give same result as select/copy/paste from pdf file.

Nirus2000:
I think this will be the last pull request for me for the SelfWeath importer. That is the end of the subject for me. You have now replied to the same commit multiple times, each time making the same links to the same comments, and so on. We have read this several times. If the PDF is read out correctly using the PDF debug, then we see the problems.
It is strange that the "hide" spaces do not exist and that with this pull request, the test files that "hide" spaces are present.
You could have solved the problem yourself in pull request # 2340, but you didn't. Instead, I am trying to work out a solution and so far all of the PDF debug you have created have gone through correctly.
What are correct and what are wrong PDF debugs ?!
We can only work with correct PDF debugs, but according to you, you only provide incorrect PDF debugs? So...?!

Kind regards
Alex

flywire:
@Nirus2000 I realise you have gone to a lot of effort to fix and I thank you for that. You have delivered a solid workaround for csv's. Unfortunately, the two-way communication has been ineffective and the technical issues I raised regarding &NBSP characters haven't been given enough consideration.

> You could have solved the problem yourself in pull request #2340

How? @buchen identified &NBSPs, made a quick fix which unfortunately turned out wrong and closed the commit. Surely my communication has been clear to strip the &NBSP characters and leave the regex.

buchen:
I have slightly adopted the change and implemented the generic removal of horizontal spaces from all files.

PR #2392, opened 2021-08-18
Change securities filter to show non-zero holdings

chirlu:
In lists of securities, there are two corresponding filters: show only entries with no holdings, or entries with positive holdings. They are not exact complements of each other in that negative holdings aren’t shown when either filter is active. This can mask errors in transactions; furthermore, there are users who do short selling and therefore intentionally have negative numbers of shares, even though this is not officially supported.

Change the “number of shares held is greater than zero” filter into “number of shares held is not zero”. This doesn’t change anything as long as the numbers are non-negative, but for users with negative holdings, those are no longer hidden from view.

[Feature request](https://forum.portfolio-performance.info/t/15545)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesPerformanceView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesPerformanceView.java
@@ -90,7 +90,7 @@ public class SecuritiesPerformanceView extends AbstractFinanceView implements Re
 {
     private class FilterDropDown extends DropDown implements IMenuListener
     {
-        private final Predicate<SecurityPerformanceRecord> sharesGreaterZero = record -> record.getSharesHeld() > 0;
+        private final Predicate<SecurityPerformanceRecord> sharesNotZero = record -> record.getSharesHeld() != 0;
         private final Predicate<SecurityPerformanceRecord> sharesEqualZero = record -> record.getSharesHeld() == 0;
 
         private ClientFilterMenu clientFilterMenu;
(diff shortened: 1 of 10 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesPerformanceView.java
```diff
@@ -123,7 +123,7 @@ public FilterDropDown(IPreferenceStore preferenceStore)
 
             addDisposeListener(e -> {
                 preferenceStore.setValue(SecuritiesPerformanceView.class.getSimpleName() + "-sharesGreaterZero", //$NON-NLS-1$
-                                recordFilter.contains(sharesGreaterZero));
+                                recordFilter.contains(sharesNotZero));
```
chirlu, line 126: Similar issue as in the other PR: I’m not changing the key for this setting even though “sharesGreaterZero” is no longer accurate, so that existing settings will still load.

flywire: I assume labels will reflect logic `Shares held <>0`

chirlu: You might want to look at the diff instead of assuming anything …

flywire: I scanned the diffs, didn't see it (it might be there) and noticed comment about other PR. If you expect more research than that for a quick comment you'll be disappointed.

buchen:
LGTM - thanks for the labels and translations

PR #2393, opened 2021-08-19
Wrong dutch translation GICS taxonomy

wimme002:
Found a branch which was not correctly translated from German to Dutch


buchen:
Thanks!

Unfortunately, this will only apply to newly created taxonomies. The reason is that when creating the taxonomy, PP copies the values - from then on the user can edit the texts as they need

PR #2418, opened 2021-09-07
Added Receiving dividends in different currency than the security for IB

opt-natter:
Issue: https://forum.portfolio-performance.info/t/receiving-dividends-in-different-currency-than-the-security/15638
When a US stock is held in EUR, but the Dividend is payed in USD the IB flex query import fails.

Change: Calculate the EUR dividend with the supplied exchange rate, if:
- a security in the base currency is available
- no security in the dividend currency is available


buchen:
Thanks @opt-natter for the contribution - see the additional commit on what I needed to change to merge this change

PR #2428, opened 2021-09-19
fixes issue with new "show limits" feature when empty limit attribute assigned to security

OnkelDok:
When a limit price attribute is assigned to a security, limit is not filled and the "show limits" option is active, a NullPointerException was thrown.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -949,8 +949,8 @@ public class SecuritiesChart
     private void addLimitLines(ChartInterval chartInterval, ChartRange range)
     {
         this.security.getAttributes().getMap().forEach((key, val) -> {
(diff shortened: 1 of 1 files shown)


buchen:
Good to have that merged before I release later today.

BTW, a ```val instanceof LimitPrice``` could have been also an option - and instanceof returns false on null values.

chirlu:
It would be good if you could use a correct and well-formed mail address for your commits, instead of `OnkelDoc@` without anything. Use the GitHub-provided dummy name if you must.

OnkelDok:
> @buchen wrote:
> BTW, a `val instanceof LimitPrice` could have been also an option - and instanceof returns false on null values.

yes, you are right. Thank you for that hint.


> @chirlu wrote:
> It would be good if you could use a correct and well-formed mail address for your commits, instead of `OnkelDok@` without anything. Use the GitHub-provided dummy address if you must.

When committing in eclipse I was a little bit scared to use my private mail. So I thought it is smart to use an invalid mail (eclipse did not complain). Thank you very much for the hint about the dummy address.

PR #2431, opened 2021-09-22
Delete PDF Author functions in all pdf extractors

Nirus2000:
Sometimes a 'BankIdentifier' is recognized as PDF author from the PDF extractors.
This can cause the wrong PDF extractor to be selected and errors to occur.
Since the PDF author is not used when extracting PDF's, we do not set it.
The same is true for the help function when PDF debug is created.

The PDF Author say "Consorbank" but this dokument is Smartbroker/DAB-Bank
grafik
https://forum.portfolio-performance.info/t/pdf-import-smartbroker-dab-bank/7684/160


Ragas13:
@Nirus2000 
Sure this is causing trouble if you keep the header from the PDF in the test  cases. This was initially not the intention.  It was a 100% match identifier to determine the fitting filter. So basically a home made bug, so go say.

PR #2435, opened 2021-09-23
Fix French translation

leemannd:
There is a small glitch in the french translation.
When you import historical quotes, the field is mapped with "Cotisation" which means in english "Contribution".
So in this case `s/Cotisation/Cotation`


buchen:
Thanks!

PR #2439, opened 2021-09-26
Optimization of markers and font in security chart

Nirus2000:
The dividend texts can be stacked up to three times.
As soon as an overlapping of the texts is possible, no text is written or is suspended until there is free space for the text again.

The respective offset value of the markers has been removed and adjusted by the respective font size and/or by the marker size.

The position of the text of the highest and lowest price has been swapped.

grafik
grafik
grafik
grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -1117,7 +1117,7 @@ public class SecuritiesChart
                         Point textExtent = event.gc.textExtent(label);
 
                         event.gc.setForeground(Colors.theme().defaultForeground());
-                        event.gc.drawText(label, x - (textExtent.x / 2), y + 10, true);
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1184,16 +1184,53 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
...
                         int y = yAxis.getPixelCoordinate(values[index]);
 
                         String label = getDividendLabel(dividends.get(index));
                         Point textExtent = event.gc.textExtent(label);
 
                         event.gc.setForeground(Colors.theme().defaultForeground());
-                        event.gc.drawText(label, x - (textExtent.x / 2), y - 22, true);
+
+                        if (((x - (textExtent.x / 2)) - lastWriteLabelLevel1) <= 0)
+                            freeSpaceForLabelLevel1 = false;
```
leemannd, line 1207: Why not using a ternary operator instead of doing an assignement to `true`and maybe switching it to false into a for loop?

PR #2442, opened 2021-09-29
Fixes StatementOfAssetsView model update issue

robertron:
This fixes the issue where the model of the StatementOfAssetsView (Vermögensverwaltung) is not updated properly after switching  between views (Ansicht) created by the user in the toolbar. Switching between multiple custom views on the StatementOfAssetsView currently results to blank columns after repeating the clicking through the custum views. As a workaround, the model is updated - and the blank columns disapear - after switching to (and back from) a different finance view e.g. security view (Alle Wertpapiere).

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsView.java
@@ -205,7 +205,10 @@ public class StatementOfAssetsView extends AbstractFinanceView
         assetViewer.setToolBarManager(getViewToolBarManager());
 
         updateTitle(getDefaultTitle());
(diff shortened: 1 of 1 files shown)


chirlu:
Somewhat related: https://forum.portfolio-performance.info/t/vermogensaufstellung-diagramm-wird-nicht-korrekt-aktualisiert/17564

buchen:
Are you on Linux?

Background: I cannot reproduce the problem. And switching the views usually does not necessitate to recalculate all values because the underlying data actually has not changed.

robertron:
Yes, this occours on Ubuntu 21.04. I can provide a Screenshot if needed.

buchen:
> Yes, this occours on Ubuntu 21.04. I can provide a screenshot if needed.

I don't think I need a screenshot. But I will call #notifyModelUpdated only if we are on Linux. The other platforms do not need a recalculation.

buchen:
merged. Thanks!

PR #2448, opened 2021-10-03
Update messages_de.properties

the-please-is-mine:
For Mac Users they would need to press the Option key instead of the Alt key. I added this little remark to the text file. Hope that it is helpful. Feel free to change it as you wish.


buchen:
Thanks!

PR #2449, opened 2021-10-03
Show totals at the top & bottom of Statement of Assets

metal450:
Statement of Assets is one of the most commonly-used screens. However, as soon as you have a decent number of positions, it requires scrolling all the way to the bottom to check your total.  Not only is this a bit cumbersome to do each time you visit that view, but it's inconsistent with how all the other categories are listed (for which the totals are at the top). This PR will let you view your totals either at the top or bottom of the page, eliminating the need to scroll down each time.

https://github.com/buchen/portfolio/issues/2409


buchen:
rebased and merged. Thanks.

I think having both lines is a "low-tech" solution to the problem that it is not possible to "fix" bottom lines that are always visible.

PR #2469, opened 2021-10-20
Rebuild OnVista PDF-Importer

Nirus2000:
Merge buy and sell transaction
Merge delivery_inbound and delivery_outbound transaction
Fix missing time in dateTime
Fix dateTime with year of two and four digits
Fix incorrect extraction of the name from the security
Fix calculation of withholding taxes
Fix of incorrect regEx for shares and bonds
Fix some other wrong transactions
Deleting duplicate testcases and testfiles
Delete addDepositTransaction()
https://forum.portfolio-performance.info/t/pdf-import-von-onvista-bank/2076/83
Integration in the PDFExtractorUtils.java (fee's and tax's)

**Information:**
Deletion of the "deposit statement" (Depotauszug) function.
This results in the recorded transactions all being posted to the date the deposit statement was created. e.g. (deposit statement as at 31.12.2010). So if PDF's are imported where individual transactions such as purchase, sale (...) are recorded, double bookings will occur.
This PDF is not suitable for the use of PP and leads to false entries in total.
For an example look in [Depotauszug](https://github.com/buchen/portfolio/blob/02aca91a3e160c733a97758257eebf20b350ddeb/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/onvista/OnvistaDepotauszug.txt)


ZfT2:
Hello @Nirus2000 and @buchen ,

I tried to import another larger Onvista Depot and was wondering why the inital import resulted in negativ stocks.
So after some debugging I found out that the addDepositTransaction() was deleted, as described by @Nirus2000 here in ths PR.

Of course I see that it is in general not useful to import these "static" statements, as importing all transaction statements should give the same result, but for the initial setup it is helpful and may save a lot of manual work.

Especially for Onvista it is useful, as they overtook the former Nordnet broker long time ago, which means that for these customers there are no Onvista PDF files for the inital purchases.

So could it be an agreement to add it again but with some warning message that users should usually not import it except they are initially setting up a depot?

Nirus2000:
Umm... no. These changes have been in place since **October 22, 2021**.

We have changed them on purpose.
Previously, transactions, purchases, sales and dividends were also read out in the periodic account statement. Although this works, it is not correct. No fees or taxes are recorded for these transactions.
We therefore limit ourselves to deposits and withdrawals, as well as any fees for the account. Purchases, sales and dividends are recorded in other documents.
We follow the same procedure in all other PDF importers, so please let us know if the other documents are all working correctly.

The same like Baader Bank, Scalable Capital and all the other importer...

ZfT2:
> Umm... no. These changes have been in place since **October 22, 2021**.

I know. But unfortunately now you get a different (incomplete) result when trying to import the same existing set of documents compared to doing the import with an PP version before October 22. So the outside behaviour of the importer is changed, which is confusing.

> We have changed them on purpose. Previously, transactions, purchases, sales and dividends were also read out in the periodic account statement. Although this works, it is not correct. 
> No fees or taxes are recorded for these transactions. We therefore limit ourselves to deposits and withdrawals, as well as any fees for the account. 
Agree. The account statement has not all the details and should not be used for transactions where a more detailed document exists.

> Purchases, sales and dividends are recorded in other documents. We follow the same procedure in all other PDF importers, so please let us know if the other documents are all working correctly.

As described, for the inital purchases there might be no documents, so a "one time" initial import of the deposit statement is usefull. 
I noticed another documment which is reflecting an "Umtausch" (money against securities) which is currently not imported but also worked earlier. I will investigate it further an maybe open an PR for that.

ZfT2:
Hi @Nirus2000 

I noticed that lot of existing test cases were changed, so all the documents handlig a security change transaction (Umtausch, Dividende mit Auslieferung, Kapitalerhöhung, Kapitalherabsetzung, freie Lieferung, Fusion, Einbuchung Rechte, Split etc.) are not covered anymore.
I don't see what speaks against it to import this documents? It saves a lot of manual work to get them imported automatically. If they are missing, the depot history is not complete.

Issue #1921, opened 2020-12-30
New UI Design proposal


usolved:
I asked in issue https://github.com/buchen/portfolio/issues/1858 that I like to make a suggestion for a new UI design. As @buchen said it might be not that easy to actually implement a more modern look because of Eclipse SWT. But I wanted to give it a shot anyway.

There are some other closed source / commercial tools with similar functionality but would be cool if an open source tool would also step up the design game a little bit :). 

**New UI / Example for portfolio**
<img width="1440" alt="Portfolio Performance New Design" src="https://user-images.githubusercontent.com/7000970/103381892-69f06980-4aed-11eb-806c-2e3e11f19ac2.png">

[Figma link to the project (read only mode)](https://www.figma.com/file/wqVCPuiwykiTGKsqd27sOe/Portfolio-Performance-(Public)?node-id=0%3A1)

### So what are the ideas behind this?

Generally speaking I concentrated mostly on the UI and _not_ the UX part. And it's also not complete (like showing the opened *.xml file). Things I addressed:

**1. New navigation menu**
The current one is already really nice but I wanted to bring in the macOS style to feel more at home like the other apps. Also it gives more focus to the actual entries and not the headlines.

**2. Add new portfolio button placed differently**
When I first used the application I was confused to have so many actions behind the plus icon on the right. I think it's better to see clearly how to add a new portfolio separated from the other actions.

**3. More spacing in table rows**
It's easier to read the table entries with a little more space added. On the contra side you have less rows on one screen but improved readability was more important to me that a little bit more scrolling.

**4. Added color scheme**
I don't know if you have a primary brand color (the buttons on the website are orange) so I used the blue tone from the logo because it's more subtile than the orange or green for this purpose.

**5. Actions menu**
Depending on the screen you're at you could show the mostly used actions directly (like buy and sell) and have the three dots for more actions (same as right click on the row). The advantage for the three dots is that you know without guessing that you can do some actions on the row.

**6. Show sub headline from selected row**
Depending on the screen it helps to have context by adding a sub headline (e.g. Comdirect). So when you look at the data from the selected entry you easily can make sure it's the right one.

**7. Show important information on top**
Like profit and loss in this example. That depends of course on the screen. There might be screens where you don't need highlighted information. In this example it's nice to have some overall information.

**8. Tab navigation**
Just different style and matches more the macOS style.

**9. Vertical line on the left**
Just a visual clue to signal that all the content belongs to the selected entry above and hold it more in place.

[...]

PR #2478, opened 2021-10-26
Basic (read-only) REST API

oxisto:
This PR, which is based on the work of https://github.com/buchen/portfolio/tree/feature_REST_API enables the usage of a simple REST API exposed via a local web server. The API itself is read-only for now and supports the following features.

| Endpoint                   | Description                                                                                                                                                                               |
| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `/api/v0/securities`       | Lists all securities, including their latest quote                                                                                                                                        |
| `/api/v0/accounts`          | Lists all accounts                                                                                                                                                                        |
| `/api/v0/accounts/{id}`     | Lists details about the specified account; currently the same as in list                                                                                                                  |
| `/api/v0/portfolios`        | Lists all portfolios, including a snapshot value                                                                                                                                          |
| `/api/v0/portfolios/assets` | List all assets within a given portfolio, including a snapshot of the security position as well as some information from `PerformanceSecurityRecord` such as purchase price and gain/loss |

The layout/structure of the API itself is probably to be discussed and changes. I followed the original approach by @buchen to version this api as `/api/v0`, so we are free to change it later.

This would be one pre-requisite to move in a direction of implementing #1921, possibly external to PP itself.

Please bare with me that I am not an export on Eclipse Plugin Development, so if there are any files committed that should not be there, I can easily change that.


buchen:
@oxisto quick feedback - I like the idea, I want to pursue it, but I do not find the time at the moment to review your change in detail (it is bigger). It is not lost.

oxisto:
> @oxisto quick feedback - I like the idea, I want to pursue it, but I do not find the time at the moment to review your change in detail (it is bigger). It is not lost.

Thanks for the quick, feedback! No worries, it is open source after all and we all can only dedicate the amount of time we can spare. I will try to keep this somehow more-or-less up-to-date w.r.t to the main/master branch in the meantime.

shivampaw:
I'd be quite happy to continue working on this and adding endpoints to allow the creation of dashboards with charts data etc. I'm struggling to get master merged into it and compiling this successfully at the moment though

zaubi787:
I have updated this old MR to the latest 0.71 version, migrated javax.* to jakarta.*  and recent jetty / jersey version. It works fine beside some minor compile issues (missing api changes, just comment them to try the REST api). It is available at https://github.com/zaubi787/portfolio/tree/rest-api and allows to execute e.g. the following request:

```
GET http://localhost:5712/api/v0/portfolios
Accept: application/json
X-PP-FILE: /home/zaubi/kommer.xml
X-PP-TOKEN: your-token
```

The rest server needs to be enabled in the settings, where the token also needs to be set. It is also necessary to open the requested X-PP-FILE in pp.

While this works in general, there is also a lot of complexity needed to handle the jersey and jetty dependencies which certainly will break in future and will introduce plugin dependency issues, classloading errors, etc which are rather hard to solve since the errors are easily swallowed by Eclipse. 

Therefore I was also trying an alternative approach by providing the  REST api via an embedded spring boot bundle with far less dependencies and having the full feature set and programming model of Spring Boot as additional advantage. Any thoughts?

Is there any feedback on chances for merging one of these approaches? My main use case for the REST api is to provide a dashboard with some charts similar to parq*et or get*uin using modern web technology. First POCs look promising

mierin12:
@zaubi787 : not a technical answer sorry, but maybe you could open a Draft Pull Request about it. Pull Requests even in draft will have much more visibility and should get reviewed. 

Also since the use case is dashboard with custom chart, maybe you could advertise your update on this thread on the forum https://forum.portfolio-performance.info/t/bessere-dashboard-engine-spaltenubergreifende-widgets-beliebige-visualisierungen/19834
They were also using the PP API.

buchen:
Closing in favor of #5870

PR #2482, opened 2021-10-28
Add the czech language

drozdikd:
Hi, here I made czech version of every message.properties.  So the PP is fully in czech, what remain in English are top level menu pane, dialog popup window when user does some action etc.


buchen:
Hi @drozdikd,

thanks for the Czech translation.

I have put your change into a separate branch: [feature_czech](https://github.com/buchen/portfolio/tree/feature_czech)

In this branch I have added czech to language configuration in the preferences (please check) and added the czech translations of the Eclipse libraries ([See here for details](https://download.eclipse.org/technology/babel/babel_language_packs/latest/index.php#cs) - roughly 60% translated to czech which is better than nothing). 

I propose to push any future changes to feature_czech branch (this PR can be closed).

Two things:

There are [more property files](https://github.com/buchen/portfolio/blob/master/portfolio-app/releng/translation-config.xml) in the project. Right now the translation is to be at 68%.

Bildschirmfoto 2021-10-30 um 10 13 28

And then one small thing: The capitalization does not seem to be consistent - see screenshot for "graf".

Bildschirmfoto 2021-10-30 um 09 49 05


But overall: great job! Thank! We are close.

buchen:
Merged :-)

PR #2489, opened 2021-10-31
Added czech files 2nd delivery & found a language label bug

drozdikd:
Looked for every .properties file and I translated it. Corrected capital letter 'graf' and few words.

Thank you for this [ XML file ](https://github.com/buchen/portfolio/blob/master/portfolio-app/releng/translation-config.xml) as guidance. 

I found out a bug in Help section. PP doesn't translate those 3 labels. I changed languages in current PP version 0.55 and for example in Spanish welcome board is all translated, but those words 'forum' 'How-Tos' and 'FAQ' remained. In Spanish, instead of forum should be Foro, instead of How-Tos Tutoriales, instead of FAQ Preguntas más frecuentes. 
howtos


buchen:
Merged :-)

PR #2495, opened 2021-11-02
Allow removing and adding sub/child filter elements, fixes #1375

### LINKED ISSUE #1375: 0.44.0: Managing Filters: no adding possible
With 0.44.0, managing Filters is possible, as in changing the name and deleting entries from a filter. Adding a new entry to an existing filter is not possible.

OnkelDok:
fixes #1375
- https://forum.portfolio-performance.info/t/bestehende-filter-erweitern/7355
- https://forum.portfolio-performance.info/t/filter-aendern-in-der-vermoegensaufstellung/9233

new with this PR:

* sub filter can now be added to main filter (see video in https://github.com/buchen/portfolio/pull/2495#issuecomment-962196719)
grafik
* subfilters can be deleted by context menu
* when deleting main/parent filter a confirmation message is shown
* confirmation message before deleting all user filters

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
@@ -196,7 +196,10 @@ public final class ClientFilterMenu implements IMenuListener
         manager.add(new SimpleAction(Messages.LabelClientFilterNew, a -> createCustomFilter()));
         manager.add(new SimpleAction(Messages.LabelClientFilterManage, a -> editCustomFilter()));
 
-        manager.add(new SimpleAction(Messages.LabelClientClearCustomItems, a -> {
+        manager.add(new SimpleAction(Messages.LabelClientClearCustomItems, a -> {           
+            if (!MessageDialog.openConfirm(Display.getDefault().getActiveShell(), Messages.LabelClientClearCustomItems, Messages.LabelClientClearCustomItems + "?")) //$NON-NLS-1$
+                return;
+            
             if (customItems.contains(selectedItem))
             {
                 selectedItem = defaultItems.get(0);
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
@@ -66,6 +66,11 @@ public final class ClientFilterMenu implements IMenuListener
         {
             return uuids;
         }
+        
+        public void setUUIDs(String uuids)
+        {
+            this.uuids = uuids;
+        }
 
         public ClientFilter getFilter()
         {
@@ -276,12 +281,10 @@ public final class ClientFilterMenu implements IMenuListener
 
     private Optional<Item> buildItem(String uuids)
     {
-        if (uuids == null || uuids.isEmpty())
+        if (uuids == null)
             return Optional.empty();
 
         String[] ids = uuids.split(","); //$NON-NLS-1$
-        if (ids.length == 0)
-            return Optional.empty();
 
         Map<String, Object> uuid2object = new HashMap<>();
         client.getPortfolios().forEach(p -> uuid2object.put(p.getUUID(), p));
@@ -299,11 +302,16 @@ public final class ClientFilterMenu implements IMenuListener
 
         String label = Arrays.stream(selected).map(String::valueOf).collect(Collectors.joining(", ")); //$NON-NLS-1$
 
-        String uuids = Arrays.stream(selected)
-                        .map(o -> o instanceof Account ? ((Account) o).getUUID() : ((Portfolio) o).getUUID())
-                        .collect(Collectors.joining(",")); //$NON-NLS-1$
+        String uuids = buildUUIDs(selected);
 
         return new Item(label, uuids, new PortfolioClientFilter(portfolios, accounts));
+    }
+    
+    public static String buildUUIDs(Object[] selected) 
+    {
+        return Arrays.stream(selected)
+                        .map(o -> o instanceof Account ? ((Account) o).getUUID() : ((Portfolio) o).getUUID())
+                        .collect(Collectors.joining(",")); //$NON-NLS-1$
     }
 
     public boolean hasActiveFilter()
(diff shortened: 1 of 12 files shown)


chirlu:
Could you explain more about this feature? In particular, I don’t understand what a “subfilter” is.

OnkelDok:
> 
> 
> Could you explain more about this feature? In particular, I don’t understand what a “subfilter” is.

With subfilter I mean an assigned account/depot of a user-created filter in the "Statement of Assets" view.
In current version in the "Statement of Assets"/"Vermögensaufstellung" there are filters for accounts and depots. There the user can create own filters by "New..." and assign accounts/depots to it. But in current version after the filter is created here is no possibility to change the assigned accounts/depots (remove account/depot from filter or assign account depot to filter).

I created a short video for the new functionality with this PR:

https://user-images.githubusercontent.com/90478568/140573602-5121e3ba-765e-4af7-9160-9fb234e9f70a.mov

chirlu:
OK, thanks for the explanation and demonstration. So, :+1: for the feature; definitely :-1: for the term “subfilter”, because that implies it is a filter when it isn’t – call it an element or a target of the filter or whatever, but not subfilter.

Some notes on the UI:
- Having “Hinzufügen” and “Löschen” like that in the same menu is confusing because a user will expect them to operate on the same object. I.e., the user will _either_ expect “Hinzufügen” to add another filter and “Löschen” to remove this filter (wrong for the first one and right for the second one) _or_ expect “Hinzufügen” to add more elements to the filter and “Löschen” to remove elements from the filter (right for the first one and wrong for the second one). So at the very least, “Hinzufügen” should become “Elemente hinzufügen” or similar.
- Then, however, it would still be strange that, when you want to remove an element from a filter, you start by selecting “Hinzufügen” before getting to the “Löschen” option for elements one step later. Reminds me of the jokes about Windows from the time when you needed to click “Start” in order to shut down the computer.
- Finally, the explanatory line (“Doppelklick zum Bearbeiten des Namens, Rechtsklick[,] um den Filter zu löschen”) should be adapted to the new options.

However, I think there might actually be a better way: Have a menu entry “Bearbeiten” next to “Löschen” (which is logical, because both operations refer to the filter as a whole) and then open a dialogue that is very similar to the one for adding a new filter, just with the filter name already filled in and with the elements preselected according to the existing filter. The user then could modify the name and select/deselect elements in the way already familiar from adding the filter.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
```diff
@@ -276,12 +284,10 @@ public Image getImage(Object element)
 
     private Optional<Item> buildItem(String uuids)
     {
-        if (uuids == null || uuids.isEmpty())
+        if (uuids == null)
             return Optional.empty();
 
         String[] ids = uuids.split(","); //$NON-NLS-1$
-        if (ids.length == 0)
-            return Optional.empty();
```
buchen, line 284: What is the reason to delete those checks?

buchen:
rebased, squashed, and merged

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterMenu.java
OnkelDok: With these checks the filter is implicit deleted when the user removes all child elements from main filter.
When deleting all child elements from main filter, saving file and restart PP the filter is not available in filter list anymore. In my opinion this behaviour will confuse the user, because they have no information (message box etc) why the main filter is missing.

The other question is: how often will this case (user deletes all child elements) happen? But my experience is: when there is even a small chance that something can happen, it will happen (sooner or later).

OnkelDok:
@chirlu: thank you very much for your detailed feedback. Even if the changes where already accepted and transferred to master I want answer:
I agree with your 1st and 3rd point. I changed the translation in the context menu (before "add" after "add elements"). Unfortunately I forgot to change the translation of the information label. I will create a new PR when I have more time.
But I don't understand your second point. What do you mean with:

> 
>
>start by selecting “Hinzufügen” before getting to the “Löschen” option
> 
>

OnkelDok:
@chirlu: here my suggestion of the explanatory line:
grafik
I used "weitere Aktionen" to make clear that the right click will not delete an item immediately. Do you think this hint descibes the possible action in the filter dialog sufficiently?

If this hint is ok, then I have to translate it and commit it. I don't know if its ok to commit the changes to this PR (its already closed) or if I should open a new one. I prefer committing to this one.

chirlu:
> I don't understand your second point. What do you mean with:
>> start by selecting “Hinzufügen” before getting to the “Löschen” option

Assume a user wants to modify a filter and remove account A from it. The course of action to accomplish that starts with selecting “add an account to this filter”, i.e. exactly the opposite of what is intended. That is not logical, and not obvious to a user.

Furthermore, I stand by my opinion that having a completely separate dialogue for modifying filters is bad UI. It could, and IMO should, look like the dialogue for creating filters:

> a dialogue that is very similar to the one for adding a new filter, just with the filter name already filled in and with the elements preselected according to the existing filter. The user then could modify the name and select/deselect elements in the way already familiar from adding the filter.

I have to say that the PR was merged prematurely.

(I don’t believe GitHub allows reactivating a closed PR, so any changes would require a fresh PR.)

OnkelDok:
Ok. I will create a seperate PR for changing the hint message.

> a dialogue that is very similar to the one for adding a new filter, just with the filter name already filled in and with the elements preselected according to the existing filter. The user then could modify the name and select/deselect elements in the way already familiar from adding the filter.

I understand your point with only one dialogue for creating main filter and changing (select/deselect) assigned elements.
But how it should be managed to select an existing main filter to open this "rename and change assignment"-dialogue?
Maybe this is possible when each user-filter gets an extra sub-context menü, with a menu-item "Verwalten...". Or do you have another idea?

> Assume a user wants to modify a filter and remove account A from it. The course of action to accomplish that starts with selecting “add an account to this filter”, i.e. exactly the opposite of what is intended. That is not logical, and not obvious to a user.

I have to admit that I still don't know what you mean with your 2nd point, from your first reply. The steps to delete an element from a filter are:
1. open filter menu
2. click "Verwalten..."
3. expand filter that should be deleted
4. right click on assigned element
5. click "Löschen"

PR #2504, opened 2021-11-06
Rebalancing via saving plans

RomanLangrehr:
This adds a new column to the rebalancing viewer that shows the required saving plan sums to rebalance the portfolio.

It also adds one line above the table where you can enter the saving plan sum (can be negative for a withdrawal plan) and the number of executions of the savings plan it should take to rebalance the portfolio. I planned to make this line disappear when the "savings plan sum" column is hidden, but unfortunately I didn't manage to implement this because I couldn't figure out a way to get notified when a column is shown or hidden.

image


chirlu:
The use case for that would be “I want to run a savings plan, and I’m not allowed to change the distribution over the assets for the next X months”? Hm.

So I don’t really see (yet) that being able to specify a number of months would be particularly useful. But a lump sum, with the provision that no assets are to be sold (or bought, for a withdrawal plan), certainly is – there were a number of requests for that on the forums: e.g. https://forum.portfolio-performance.info/t/rebalance-among-asset-classes-with-just-cash-and-not-selling-off-any-securities/17815, https://forum.portfolio-performance.info/t/rebalancing-mit-portfolio-performance/19/22, https://forum.portfolio-performance.info/t/entnahme-mit-rebalancing/12744

I’m somewhat worried people looking for a “lump sum feature” won’t find a “savings plan feature” with the special case months = 1. My suggestion would therefore be to omit the “months” option, which would allow renaming things. Unless, of course, there really is a use case for the savings plan.

------

Related: Some people [asked](https://forum.portfolio-performance.info/t/rebalancing-durch-zukauf/492/16) for the amount of fresh money that would be necessary to do a full rebalancing without selling anything. I’m not too convinced of the usefulness, though – even when I know that € 12345,67, I can’t simply draw that money from thin air …

chirlu:
The use case for that would be “I want to run a savings plan, and I’m not allowed to change the distribution over the assets for the next X months”? Hm.

So I don’t really see (yet) that being able to specify a number of months would be particularly useful. But a lump sum, with the provision that no assets are to be sold (or bought, for a withdrawal plan), certainly is – there were a number of requests for that on the forums: e.g. https://forum.portfolio-performance.info/t/rebalance-among-asset-classes-with-just-cash-and-not-selling-off-any-securities/17815, https://forum.portfolio-performance.info/t/rebalancing-mit-portfolio-performance/19/22, https://forum.portfolio-performance.info/t/entnahme-mit-rebalancing/12744

I’m somewhat worried people looking for a “lump sum feature” won’t find a “savings plan feature” with the special case months = 1. My suggestion would therefore be to omit the “months” option, which would allow renaming things. Unless, of course, there really is a use case for the savings plan.

------

Related: Some people [asked](https://forum.portfolio-performance.info/t/rebalancing-durch-zukauf/492/16) for the amount of fresh money that would be necessary to do a full rebalancing without selling anything. I’m not too convinced of the usefulness, though – even when I know that € 12345,67 is necessary, I can’t simply draw that money from thin air …

RomanLangrehr:
> Unless, of course, there really is a use case for the savings plan.

In my point of view, there is. At least I have monthly saving plans (and want to use them for rebalancing), but I don't want to adapt the savings plan sum every month (but rather every quarter or every 6 months), just because it would be to annoying to do this every month and there is no drawback in doing it a bit less often.

RomanLangrehr:
@buchen This PR hasn't made it into PP so far. May I ask why? Is there something that should be done before or is this feature just not desired?

auchri:
@buchen ping

RomanLangrehr:
@buchen I made this PR up-to-date. Any chance it can get in now?

buchen:
Hi @RomanLangrehr,

thanks for putting this back on the agenda. I think it slipped my mind because it feels like a niche issue. Additionally, I do not think that investment plans are rebalancing tools - part of rebalancing is also a sale of a position which an investment plan never could do. Of course, I get what you are trying to achieve - usually it is a purchase. Finally, the UI would always show an input field and pollutes space for every user.

What I can see is that PP offers to enter an "amount to invest" - along the lines of @chirlu pointed out above. I think for the investment plan use case it would be simple mathematics: Multiply the installment amount by the number of installments and then divide the calculated rebalancing amount by the number of installments.

For such a focused feature, I have created a spec #5877. Any thoughts / feedback?

Andreas.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/ReBalancingViewer.java
```diff
@@ -361,6 +447,88 @@ public String getToolTipText(Object element)
...
+
+                InvestmentVehicle investmentVehicle = node.getBackingInvestmentVehicle();
+                if (investmentVehicle == null || !getModel().getTaxonomy().isUsedForRebalancing(investmentVehicle))
+                    return null;
+                
+                if(savingsPlansSum == null || savingsPlanExecutions == null)
+                    return Messages.LabelNotAvailable;
+
+                Money rebalancingAmount = getModel().getRebalancingSolution(savingsPlansSum.multiply(savingsPlanExecutions)).getMoney(investmentVehicle);
+                rebalancingAmount = rebalancingAmount.divide(savingsPlanExecutions);
```
buchen, line 472: **No validation for 0 / negative executions**

getText only null-checks savingsPlanExecutions, never its range. Enter 0 and the code does savingsPlansSum.multiply(0) (amount 0) then .divide(0). Money.divide is Math.round(amount / (double) divisor), so it doesn't throw — it yields Long.MAX_VALUE (Infinity) or 0 (NaN), i.e. a nonsensical value in the cell rather than "n/a". Negatives silently invert. Fix: in onSavingsPlansChanged, treat <= 0 as invalid (set to null) so the existing LabelNotAvailable fallback kicks in.

Image

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/ReBalancingViewer.java
```diff
@@ -361,6 +447,88 @@ public String getToolTipText(Object element)
...
+                TaxonomyNode node = (TaxonomyNode) element;
+
+                InvestmentVehicle investmentVehicle = node.getBackingInvestmentVehicle();
+                if (investmentVehicle == null || !getModel().getTaxonomy().isUsedForRebalancing(investmentVehicle))
+                    return null;
+
+                if(savingsPlansSum == null || savingsPlanExecutions == null)
+                    return null;
+
+                if (!getModel().getRebalancingSolution().isExact(investmentVehicle))
```
buchen, line 488: **Color/tooltip describe a different solution than the value**

getText (471) uses getRebalancingSolution(sum × executions), but getBackground and getToolTipText call the no-arg getRebalancingSolution() — the zero-sum solution. So the red/warning background and the inexact/ambiguous tooltip can contradict the number shown. Use the same amount-based solution for isExact/isAmbigous. (Note: the pre-existing delta-shares column at 412/442 correctly uses zero-sum for both — this inconsistency is new to this column.)

Issue #2515, opened 2021-11-09
No data from CoinGecko API for valid symbol 


photobook-hive:
**Describe the bug**
Getting no data for a valid CoinGecko symbol

hive ist working 
hive_dollar OR HBD diddn't work   "HBD" should be the right SYMBOL 

Probably the UPERCASE is the problem?

Both request are valid and i'm getting data.... inside of the App hive_dollar dont work

https://api.coingecko.com/api/v3/coins/hive/market_chart?vs_currency=EUR&days=62&interval=daily
https://api.coingecko.com/api/v3/coins/hive_dollar/market_chart?vs_currency=EUR&days=62&interval=daily

Original JSON output from CoinGecko API "List Coins"
  },
  {
    "id": "hive",
    "symbol": "hive",
    "name": "Hive"
  },
  {
    "id": "hive_dollar",
    "symbol": "HBD",
    "name": "Hive Dollar"
  },
  {

photobook-hive:
I think i got it. It is a UPPER / LOWER CASE problem.

Symbol is "HBD" but the program change it to LOWERCASE "hbd" and this ist not Valid.

Tue Nov 09 12:03:30 CET 2021
Ticker-Symbol fehlt für Wertpapier hbd

java.io.IOException: Ticker-Symbol fehlt für Wertpapier hbd

	at name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed.getCoinGeckoIdForTicker(CoinGeckoQuoteFeed.java:186)

	at name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed.getHistoricalQuotes(CoinGeckoQuoteFeed.java:140)

	at name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed.getLatestQuote(CoinGeckoQuoteFeed.java:53)

	at name.abuchen.portfolio.ui.jobs.UpdateQuotesJob$1.run(UpdateQuotesJob.java:234)

	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

kasch-code:
Thanks for pointing this out.  I missed that there are currencies which include upper case letters in their ticker symbol.

I implemented to lower case conversion in the first place, because I prefer to have the ticker symbols in upper case letters (e.g. BTC for Bitcoin) in my lists. I suppose this may be the same for other people so I am not sure if it would be a good idea to completely get rid of the conversion.

[...]

Issue #2464, opened 2021-10-16
CoinGecko: Support for ambiguous ticker symbols


kasch-code:
At the moment, the CoinGecko quote provider uses the specified ticker symbol, e.g. BTC for Bitcoin or ETH for Ethereum, to retrieve the quote data. To do this, the common ticker symbol is mapped to the internal CoinGecko ID first, which has to be used for all following API calls.

This leads to a problem, when a ticker symbol is assigned to multiple coins, for example BIT.

Possible solutions would be:
1. Use the CoinGecko ID instead of the common ticker in the security master data
2. Give the user the option to specify the CoinGecko ID _in addition_ to the ticker symbol for these cases

Drawback of option 1 is that this would be not very intuitive. The users are used to the common ticker symbols of the coins and not the CoinGecko ID. Furthermore this would break existing files using the provider.

Option 2 therefore would be preferred. The additional ID could either be stored in one of the unused fields for crypto (WKN, ISIN) or as an additional provider attribute in the "Latest Quote" screen. The latter would be preferred as it is a cleaner solution.
When the CoinGecko ID is provided, it will be used instead of the ticker symbol to retrieve the quotes from CoinGecko (no mapping will be done in this case).

Also see the comments in #2325 for further information.

kasch-code:
I will have a look into this issue once I find some time. Please feel free to assign it to me.

photobook-hive:
Hey Karsten, 

i think this would also work for my bug.

Gruß
Arek

karlkurzer:
Hey Karsten,

thanks for looking into this.

The simplest fix would be to avoid `getCoinGeckoIdForTicker()`.

Since many crypto currencies carry the same symbol this can't work. Removing that function call should solve the problem, then it can just use the actual id (which can be easily found here: https://api.coingecko.com/api/v3/coins/list).
https://github.com/buchen/portfolio/blob/be8ebebcc2c0d2c1324819b2c5537e4773590adc/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/CoinGeckoQuoteFeed.java#L140

karanchopraa:
Team, 
I am facing the same issue, there are two coins with the same tickers, in metamask it's fetching the other ticker, help me to resolve it.

PR #2519, opened 2021-11-11
Add CoinGecko CoinID as parameter to provider

kasch-code:
With this change it is possible to specify the internal CoinGecko Coin ID as a parameter for the data provider. If this parameter is not specified, the ticker symbol will be used to fetch the data as before.

This is useful for fetching data in the following cases:
- Coin with an ambiguous ticker symbol - issue #2464
- Coin having upper case letters in their ticker (which is rather uncommon in the API) - issue #2515

The Coin ID can be found on CoinGecko website:
image


buchen:
@kasch-code looks good, thanks for the change.

I believe one thing is missing: in the ```HistoricalQuoteProviderPage``` a cache key is calculated that is used to cache the server response (in cased the user just switches back and forth). You need to add the goingeckoid there as well.

And thanks for working through the code: over the years, this class really deteriorated. But I never came around to cleaning it up....

kasch-code:
@buchen Thanks for pointing this out, I added the parameter to `HistoricalQuoteProviderPage` along with the ticker symbol (as the parameter is optional). Refreshing the quotes seems to be working fine.

No worries about the code, I've seen worse. 😉

PR #2522, opened 2021-11-12
fixes calculation of historical prices when applying split

### LINKED ISSUE #1368: Stock split feature not working as expected
Hi

Thanks again for developing this nice tool :)

On the date 01-01-2014 Novo Nordisk (NOVO-B.CO) had a stock split where 1 stock became 5 and prices went from around 1000 DKK to 200 DKK. I have tried to add this event to my portfolio but i think there is an error when the tool adjusts the stock prices prior to the split.

When splitting 5:1:
25 stocks -> 125 stocks (which is what i expected) but the price chart now looks like this:
Pricechart
All prices before stock split was divided by 5 and not multiplied by 5 (so opposite to what i was trying to do). This means that it looks like the stock had a price of 40 DKK before the split when it actually had a price of 1000 DKK

If i split 1:5 then 25 stocks become 5. I haven't checked the price chart when doing this, but my guess is that the prices are multiplied by 5.

Isn't there something wrong here? Maybe the price adjustment is wrong or the stock count adjustment is wrong?

Also i cannot find anywhere to remove the stock split event so it's kind of hard reverse.. It could be that i'm just looking in the wrong place.

Thanks for looking into this :)

Best regards Philip

OnkelDok:
https://forum.portfolio-performance.info/t/verbuchung-eines-aktiensplits-passt-kurse-falsch-herum-an/18581
https://forum.portfolio-performance.info/t/aktiensplit-buchen/11758/71

<details>
  <summary>Before Fix (wrong prices are marked red in screenshots)</summary>

grafik  
grafik
grafik
grafik
grafik

</details>
<details>
  <summary>With/After Fix</summary>

grafik
grafik
grafik
grafik
grafik

</details>

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/wizards/splits/StockSplitModelTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/wizards/splits/StockSplitModelTest.java
@@ -1,0 +1,27 @@
+package name.abuchen.portfolio.ui.wizards.splits;
+
+import static org.hamcrest.Matchers.is;
+
+import java.math.BigDecimal;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.Test;
+
+@SuppressWarnings("nls")
+public class StockSplitModelTest
+{
+
+    @Test
+    public void testCalculateNewQuoteAndStock()
+    {
+        StockSplitModel model = new StockSplitModel(null, null);
+        // 2 shares for 1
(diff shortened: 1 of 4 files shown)


OnkelDok:
I have the feeling this PR is not related to the issue  #1368 (Jan 2020).
Because the bug came in with 54b29dd (Oct 2021).

Who can remove the link between this PR and the Issue?

buchen:
Thanks @OnkelDok 

Yes, I added the bug when allowing split ratios with decimal places :sad:

I have removed the link to the wrong bug

PR #2536, opened 2021-11-19
Automatically show information pane when limit price attribute is selected in attribute settings view

OnkelDok:
The limit price settings are not that easy to find without any information: 
- https://forum.portfolio-performance.info/t/benutzerspezifische-farben-fur-attribute-vom-typ-limit/18485/2
- https://forum.portfolio-performance.info/t/prozentualer-absoluter-abstand-zum-limit-darstellen/8366/7

Therefore I added an automatism to show the information pane when user clicks on a limit price attribute in the settings view:

https://user-images.githubusercontent.com/90478568/142628347-5c72c10c-dc24-43be-b9de-5b3bd661029f.mov


buchen:
Thanks @OnkelDok

I am also thinking about redesigning the information pane. Right now it is possible to completely hide it. The little handle is only visible if you mouse over it. I wonder if it would help if at least the tabs are visible. Then the user knows that there is more information. When clicking on a tab, the pane is expanded and that tab is shown. One question is then how the user just as easily can hide the pane again.

PR #2559, opened 2021-11-28
Update messages_fr.properties

LucasC:
Fix small typo.


buchen:
Thanks. Every typo matters 😄

PR #2580, opened 2021-12-08
Fix sorting arrow in tables heads

### LINKED ISSUE #2576: Sort symbols wrong
**Describe the bug**
The arrow/symbol for sorting is wrong. An ascending sort shows the symbol for descending and the descending for ascending. This affects all tables. 

**Screenshots**
image
image

Correct would be:
image

Nirus2000:
Fixes #2576

Hallo @buchen 
Der Sortierungspfeil in den Tabellenköpfen ist beim sortieren immer verkehrt herum.
Ich bin mir nicht ganz sicher, ob diese Änderung ausreicht um den Sortierungpfeil in allen 
Tabellenköpfen richtigrum darzustellen.

Gruß
Alex

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ColumnViewerSorter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ColumnViewerSorter.java
@@ -369,7 +369,7 @@ public final class ColumnViewerSorter
             {
                 setupSortingContext();
 
-                int dir = direction == SWT.DOWN ? 1 : -1;
+                int dir = direction == SWT.UP ? 1 : -1;
 
                 if (element1 == null && element2 == null)
                     return 0;


buchen:
First thing: I agree. The sorting indicator in the tables is wrong - or at least not consistent with other applications (at least looking at macOS).

Then: I wonder if we can fix the sorting indicator without now changing all sorting directions. With this change, the sorting indicator stays, but the sort order suddenly changes. That means every one of our users now has to click once to get the expected sorting back. I'll do some researching on this...

OnkelDok:
As far as I understand the following code sets the sorting icon:
````
c.getParent().setSortDirection(direction)
````
When I comment out the two lines locally from following code block, then no sort direction icon is displayed.
https://github.com/buchen/portfolio/blob/239db78483d84ad82745fc81ace8f6975d7d791c/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ColumnViewerSorter.java#L342-L357

So maybe changing/switching the passed direction value would switch only the icon and not change the sorting itself.

````
c.getParent().setSortDirection(direction == SWT.DOWN ? SWT.UP : SWT.DOWN);
````

buchen:
> So maybe changing/switching the passed direction value would switch only the icon and not change the sorting itself.

Thanks for researching.

Although I have to say - that looks ugly and will be harder to maintain... maybe once everybody fixing the sort direction is better.

buchen:
Thanks @Nirus2000 @OnkelDok 

It took me some time until I got the right idea to have no impact on the user: I introduced new values when storing the table configuration (=existing columns + sorted column). When restoring, I can differentiate between the old values (and flip them to correct them) and the new values.

I tested it on my machine. Applying the change loads the existing files exactly the same - only the sort indicator is fixed.

PR #2583, opened 2021-12-10
Fix Non-terminating decimal expansion bug + optimization of multiplier calculation

### LINKED ISSUE #2581: Stock Split causes an error when the split price is a non-terminating decimal
**Describe the bug**
Stock Split causes an error when the split price is a non-terminating decimal.

**To Reproduce**
1. Right click on any stock
2. Click on Stock Split
3. Do a split 3 to 1 (since 3 can cause non-terminating decimal)

**Expected behavior**
Price to be divided by 3

**Screenshots**
image

 - OS: Windows
 - Version 0.56.2

OnkelDok:
fixes #2581
Reason for issue was missing scale and rounding mode when dividing with stock-multiplier for calculation of new quote.

I changed the calculation the following way:
For both calculations (new stock and new quote) an individual multiplier is used. These multipliers (stockMultiplier and quoteMultipler) were calculated when new shares or old shares are set. In this way the multipliers were only calculated twice and not in every iteration when all new stocks or new quotes were calculated.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/splits/StockSplitModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/splits/StockSplitModel.java
@@ -20,6 +20,8 @@ public class StockSplitModel extends BindingHelper.Model
     private LocalDate exDate = LocalDate.now();
     private BigDecimal newShares = BigDecimal.ONE;
     private BigDecimal oldShares = BigDecimal.ONE;
+    private BigDecimal stockMultiplier = BigDecimal.ONE;
+    private BigDecimal quoteMultiplier = BigDecimal.ONE;
 
     private boolean changeTransactions = true;
     private boolean changeHistoricalQuotes = true;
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/splits/StockSplitModel.java
```diff
@@ -93,22 +103,26 @@ public void setChangeHistoricalQuotes(boolean changeHistoricalQuotes)
...
-        return BigDecimal.valueOf(oldStock).multiply(getStockMultiplier())
+        return BigDecimal.valueOf(oldStock).multiply(stockMultiplier)
                         .setScale(0, RoundingMode.HALF_EVEN).longValue();
     }
     
     public long calculateNewQuote(long oldQuote)
     {
-        return BigDecimal.valueOf(oldQuote).divide(getStockMultiplier()) // when stock is multiplied, quote must be divided
-                        .setScale(0, RoundingMode.HALF_EVEN).longValue();        
+        return BigDecimal.valueOf(oldQuote).multiply(quoteMultiplier)
```
buchen, line 124: Do you really need to variables - stock and quote multipler - or couldn't you still call BigDecimal#divide but simply pass the math context?

OnkelDok: @buchen: Of course, using single field also works. I tried BigDecimal#divide(stockMultiplier, Values.MC) locally and the unittest succeeds.
Should I switch to single field?

PR #2585, opened 2021-12-15
Added: "unixtime" as possible <format> for TODAY

kingbtcvl:
Hi!

As discuessed in the Portfolio Performance Forum [here](https://forum.portfolio-performance.info/t/dynamische-kursdaten-urls/2929/78), I needed to generate a dynamic url which contained the current date as unix time. Since `DateTimeFormatter` does not contain a `<format>` that represents the date in unix time, I added the possibility to simply use `unixtime` as `<format>` for the `TODAY` Macro.

So with this pull request the users will not only be allowed to write `{TODAY:dd.MM.yyyy:-P3M}` in their dynamic urls which will create `15.09.2021`, but they will also be allowed to write `{TODAY:unixtime:-P3M}` which will create `1631664000`.

Hopefully one of the maintainers of thie project will merge this little request into the next update because I have litterally just created a github account and learned how to work with all that stuff here only to make this request.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/variableurl/macros/Today.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/variableurl/macros/Today.java
@@ -1,6 +1,8 @@
 package name.abuchen.portfolio.online.impl.variableurl.macros;
 
 import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
 import java.time.Period;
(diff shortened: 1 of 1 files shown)


buchen:
Thanks @kingbtcvl for the nice contribution. I polished the code slightly to remove the string literals where possible, added a test case, rebased it to master and merged it. Thanks. Will be published with the next version

kingbtcvl:
Thanks for your support and effort, @buchen !
Looking forward to the next version!

PR #2586, opened 2021-12-17
Experimental successor of etf-data.com

berndmr:
I implemented an, in this state experimental and beta, replacement for getting sector and country distribution for ETFs. It uses data from Mountain-View. The API Backend for performance portfolio is written and hosted by me. It uses caching to minimise data-access to the source data provider. The JSON result provides from the backend is identical to the one from etf-data.com (but only countries und sectors) and so the Java-Classes are refactored from the original implementation in portfolio performance. I tested the implementation, next step would be a broader Beta audience. In production, the API would be a paid solution (do not see a way to provide this data free of charge). Price should be evaluated in beta phase - I do not want to get rich with that, only cover the expenses. If something like that is not wished by the product owner, it is also ok for me. Then it was a nice coding practice and I will use it for my portfolio performance.


metal450:
Would love it if this worked for US equities as well - etf-data is only for European, making it unusable for myself & others :)

berndmr:
> Would love it if this worked for US equities as well - etf-data is only for European, making it unusable for myself & others :)

If you have one or two examples I can have a look if the data is available.

metal450:
VTI, VXUS, QQQ, SGOL, VNQ, etc - any ETF that trades on the US exchanges

berndmr:
Will have a look next week. Not too promising so far.

berndmr:
It will be possible to provide data for US equities. Not from the beginning as the data needs tweaking to fit into the scheme - but this can be done in backend. I will start implementing a soon as the PR is merged.

buchen:
Hi @berndmr,

did you think about this with @tfabritius.
I know that Thomas offers on portfolio-report already information about taxonomies.
I plan to support his API.
If we combine the data with the frontend code, then we can get a solution.

what do you think?

Andreas.

berndmr:
Hi @buchen, @tfabritius 

I heard the idea from Thomas a year ago but have not seen any progress so far. I think he wanted setup something like a "community editing" system - so if one fills it by hand the other can get it. I do not think that this can work with the mass of ETFs available.

I did quite an amount of research and came for me to the conclusion, that there is no way to provide the data free of charge. But it is possible to buy and resell the data for a fair price.

So at the end, it is your software and your decision:
1) wait for portfolio-report
2) add another paid provider
3) implement a generic json based fetcher

Number 3 is also an Idea from the forum:
https://forum.portfolio-performance.info/t/etf-stammdaten-von-etf-data-com/13363/75

So the user can decide between different solutions (paid, free, ...)
But I am not good enough in java to build it. 

Bernd

tfabritius:
Hi,

Portfolio Report already uses a paid source for quotes. I can imagine using a paid source for taxonomies as well as allowing users to enter taxonomies. Users could contribute either manually via UI or via script/API, e.g. parsing data from public reports etc.

The database currently contains some taxonomies already and the API to consume them is available.

Best,
Thomas

PR #2607, opened 2021-12-30
Support HTML import with different date format and price definition

vihtap:
(comes from hungarian)

Simple test is also added


buchen:
Rebased to master, fixed unused imports, merged, Thanks 👍

Issue #2400, opened 2021-08-25
View "Klassifizierungen - Rebalancing" - Add column/attribute "Soll % am GV"


0mynona0:
To enable direct comparison between as-is % allocation and target % allocation please add an attribute/column that shows the target % allocation related to the overall portfolio value (i.e. "Soll % am GV" that can be directly compared to "% am GV").

PR #2609, opened 2021-12-31
Resolves Issue #2400 opened on Aug 25 by 0mynona0

0mynona0:
Adds column "ColumnToBePctOfTotal" with title "To-be % of Total",
re-titles column "ColumnPctOfTotal" from "% of total" to "As-is % of
Total", adds column "ColumnDeltaPctOfTotal" with title "Delta % of
Total", adds respective menu labels


messages.properties
```diff
@@ -0,0 +1,2311 @@
...
+
+AccountFilterRetiredAccounts = Hide inactive accounts
```
Nirus2000, line 9: Coding during upload is wrong
(line breaks --> CRLF)

PR #2612, opened 2022-01-03
Fix inconsistent date format in CSV exports

metal450:
Fixes https://github.com/buchen/portfolio/issues/2077

Appearance of CSV export before (inconsistent date format, cannot be sorted chronologically):

With this fix (date format is consistent & sortable):

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TableViewerCSVExporter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TableViewerCSVExporter.java
@@ -1,5 +1,9 @@
 package name.abuchen.portfolio.ui.util;
 
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -15,6 +19,7 @@ import org.eclipse.swt.widgets.TableItem;
 
 import name.abuchen.portfolio.money.Values;
 import name.abuchen.portfolio.ui.util.viewers.SharesLabelProvider;
+import name.abuchen.portfolio.ui.Messages;
 
 public class TableViewerCSVExporter extends AbstractCSVExporter
 {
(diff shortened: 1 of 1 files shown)


buchen:
Hi @metal450,

I like the idea: the values in the CSV export should be "natively" readable by spreadsheet programs as dates.

However, I have a couple of thoughts:
* The parsing of the string value is locale-dependent (with German local it does not work for example). The current code works only for one locale. At least, we would need to use the DateTimeFormatter from the ```Values.Date``` or ```Values.DateTime``` which is used to generate the string.
* It picks up only the columns which are accidentally named StartDate and EndDate. There are more.

I wonder if it makes sense to create a ```DateLabelProvider``` (similarly to the SharesLabelProvider) and retrieve the raw value in the CSV exporter. It is a little tricky because it would have to support LocalTime and String (use for the "(open)" trades). What do you think?


Andreas.

metal450:
> German

I tried changing the language to test this, but I cannot get it to stay on anything but "Automatic." If I change the language to German, save, quit & re-start the software (via eclipse debugger), it's just back to Automatic again. So I'm not sure how to test this, but if it's an easy/quick fix, wanna just push the change?

> It picks up only the columns which are accidentally named StartDate and EndDate.

Not sure what u mean by "are accidentally named StartDate and EndDate"...?

> I wonder if it makes sense to create a DateLabelProvider (similarly to the SharesLabelProvider) and retrieve the raw value in the CSV exporter.

Probably beyond my capability at this point, in all honesty.  I was dying for this fix as the previous non-sortability of dates made it incredibly time-consuming to use the exports for my tax reporting, but given my extreme lack of familiarity w/ Java, it took me many hours just to figure this out. Feels like it's a small & self-contained enough fix so as not to require overengineering it too much, IMO.

buchen:
@metal450

**Kann denn Excel mit dem ISO Format etwas anfangen?**


Ich habe leider hier kein Excel, aber wenn ich den CSV Export jetzt in Apple Numbers importiere, ist das nur noch Text. Vorher konnte Numbers die Spalte problemlos als Datum erkennen, sortieren und auch damit rechnen (die dritte Spalte errechnet per Formal die Differenz). All das geht mit dem neuen Format nicht mehr.

Bildschirmfoto 2022-01-08 um 15 45 13

Ich habe auch Google Sheets ausprobiert. Google Sheets kann mit dem Datum etwas anfangen.

Ich möchte mit der Änderung keine anderen Szenarien "kaputt" machen.

metal450:
K, pushed a new commit, this output format will be recognized as dates.

(Previously, it would only be recognized by excel if there were dates with no times; dates with times were text-only. My primary goal was to get it to be consistent, so there's one format in that column, not one format for dates and a different format for dates+times.  This last push changes it to a format that Excel sees as date+time, tho).

buchen:
> K, pushed a new commit, this output format will be recognized as dates.

That format also works for Apple Numbers. Thanks. Will merge

PR #2613, opened 2022-01-03
Add Note column to the Trades view

metal450:
Add the ability to show a Notes column on the Trades view, which will show the note of the final (closing) transaction of the trade.

Why: PP is very useful for doing taxes - in particular, you can use the CSV export from the Trades view to see all closing trades, & help calculate taxes due.  However, in crypto, most transactions need context - thus, they're tagged with i.e. a link to the etherscan transaction for an ethereum gas fee, a mention of which exchange it was sold on, etc.  Without this context, it isn't really possible to use the CSV to reconcile taxes, as different "closing" transactions may be treated differently.  This lets that context be included in CSV exports, so it can be used for calculating crypto taxes.


buchen:
rebased, added tooltip and icon for consistency, merged

PR #2614, opened 2022-01-03
Securities Notes Usability Issues

metal450:
This PR fixes 3 usability issues related to Notes, when editing a Security:

1) When expanding the Edit Security dialog, the "notes" textbox does not expand to fill empty space (instead, a tiny notes textbox remains, & the expanded dialog area remains as empty background). Fixes: https://github.com/buchen/portfolio/issues/2051)

2) For very long notes, the text doesn't wrap, requiring the user to keep scrolling left & right to read it (or trying to manually manage the newlines). This fixes line wrapping, so that it's no longer necessary to scroll back & forth to read longer notes.

3) For multi-line notes, the main grid view itself actually shows newlines (rather than having one line per entry) - which results in the vast majority o vertical space being consumed by just the "notes" cell, & an impossible amount of vertical scrolling for those with very large Securities lists. This fix will show multi-line notes as a single line in the grid, making it possible to use multi-line notes, without blowing up the vertical space in the securities list :)

Screenshots:

Items 1 & 2, Before:
Peek 2022-01-02 21-41

Items 1 & 2, After:
Peek 2022-01-02 21-42

Item 3, Before:

Item 3, After:

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/NoteColumn.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/NoteColumn.java
@@ -32,7 +32,14 @@ public class NoteColumn extends Column
             {
                 Annotated n = Adaptor.adapt(Annotated.class, e);
                 if (n != null)
-                    return n.getNote();
+                {
+                    String note = n.getNote();
+                    if(note != null)
+                    {
+                        note = note.replace("\n", " "); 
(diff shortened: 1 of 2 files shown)


buchen:
> For multi-line notes, the main grid view itself actually shows newlines (rather than having one line per entry)

This seems to be Linux specific. On macOS, there is only one line shown for a multi-line note (but only the first line, not concatenated). Thanks for the fix.

rebased to master, merged.

PR #2615, opened 2022-01-03
improve sparse label mode of charts

buletti:
This is a follow-up of #2298. It further improves the label handling of narrow charts.
This commit mainly provides a check for whether or not a label should be placed. The check tries to "look around" and see if there is enough space to the left and to the right of the vertical line for a label to add. If the check fails and there is not enough space, then the label is omitted in order to avoid labels overlaying/bleeding/colliding.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java
```diff
@@ -297,33 +302,42 @@ else if (days < 1200)
...
             {
                 period = Period.ofMonths(1);
                 format = DateTimeFormatter.ofPattern("MMMM yyyy"); //$NON-NLS-1$
             }   
-            else if (days < 250)
+            else if (days < 400)
             {
                 period = Period.ofMonths(3);
                 format = DateTimeFormatter.ofPattern("QQQ yyyy"); //$NON-NLS-1$
-                cursor = cursor.plusMonths((12 - cursor.getMonthValue() + 1) % 3);
```
buchen, line 309: Why do you delete this line?

The "cursor" needs to be moved to the first occurrence of the time grid.
Let's say it is February and you want to paint quarterly lines, this needs to move to April (as the beginning of the second quarter).

I re-added the lines - otherwise the change looks fine and I will merge.

buletti: Yes, you are right of course. Without the cursor alignment the quarter lines are not placed correctly.

pp_quarter_line_regression

PR #2617, opened 2022-01-04
add support for reporting period 'since first transaction'

buletti:
This is an implementation of #2411.
This commit provides an option in the reporting period creation dialog to make a reporting period starting from the date of the first transaction.
This option is not introducing a new, specific reporting period implementation. Instead it is reusing the existing 'since-x' reporting period. The dialog merely uses a "macro" to bootstrap a 'since-x' reporting period with the date of the first transaction.
As such the option 'since first transaction' is not a pointer referencing the date of the first transaction in the ledger. It's not a dynamic functionality backed by the current state of a modifiable portfolio. It's based on a one-time, static snapshot of the portfolio state. When this option is selected in the reporting period dialog, it creates a 'since-x' reporting period based on the date of the first transaction at this point in time.
Due to the lack of dynamic transaction tracking this functionality obviously "breaks" when a new, first transaction is added in retrospect. Then a 'since-x' reporting period might have been created previously for the transaction that was considered the first one before. In this situation the user must update the 'since first transaction' reporting period manually. 
Despite of the need for a manual correction of the first transaction reporting period, there are no severe issues caused. The outdated state is handled rather graceful. When the user re-enters the reporting period dialog in such a situation, he'll notice that the 'since first transaction' option is no longer selected. Instead a 'since ${date of the previous first transaction}' is selected. An outdated 'since first transaction' deteriorates gracefully into a simple 'since-x' when a new first transaction is added in retrospect.
But despite of the described limitation, I think this is still a very useful feature. First transactions do not change (often).

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -671,6 +671,7 @@ public class Messages extends NLS
     public static String LabelReportingDialogMonths;
     public static String LabelReportingDialogQuarter;
     public static String LabelReportingDialogSince;
+    public static String LabelReportingDialogSinceFirstTransaction;
     public static String LabelReportingDialogTradingDays;
     public static String LabelReportingDialogUntil;
     public static String LabelReportingDialogWeek;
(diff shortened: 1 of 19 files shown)


OnkelDok:
Hello @buletti,

I noticed your PR and like idea behind that. You are right, normally the first transaction remains the first transaction when the refered dataseries are the same.
But what when the user switches the underlaying dataseries (e.g. by selecting a filter, or adding/removing dataserie to/from chart, ...)? In this case it is more likely that the first transaction will change (first transaction of ETF-Taxonomies can be different to first transaction of the joint/entire portfolio).
Or would your PR handle this correctly? If not, do you think you can add this kind of fully dynamic first-transaction determination?

PR #2643, opened 2022-01-15
Fix "Latest Quotes" being a day late for Coingecko

metal450:
Works around this issue: https://github.com/buchen/portfolio/issues/2106

For the CoinGecko provider


buchen:
I merged an slightly updated version.

PR #2644, opened 2022-01-15
Simpel pdf extractor

auchri:
Implement pdf extractor for sell/buy transactions of [Simpel S.A.](https://www.simpel.lu) ([Own360](https://www.own360.app/)).

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/simpel/SimpelBuy01.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/simpel/SimpelBuy01.txt
@@ -18,7 +18,7 @@ AT0000A1QA38 10.01.2022 5.123
 Girosammelverwahrung/AT
 Abrechnungsbetrag: 10.00 €
 Ausgabeaufschlag/Provision: 0,00%
-Auftrags-Nummer: 202201066000231886000000612345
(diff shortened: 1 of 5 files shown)


buchen:
Nice. I see the pull request is marked as "draft". Do you have particular questions?

The build fails because of code formatting. That is typically easy to fix within Eclipse because Eclipse should pick up the PP code style of the project automatically. If you reformat the file, it should be fixed.

And I propose to delete the commented out section of code.

@Nirus2000 can you also have a look at it?

auchri:
Hi @buchen ,

I also want to add the implementations for sell and dividend transactions, that's why I marked the PR as draft.

auchri:
Buy and sell transactions are implemented now, PR is ready for review @buchen :)

auchri:
I'm not sure if the buy transaction is correct.
In the pdf it is stated that 0,071 shares are buyed at a fund price of 140,59€ (which would be 9,982€ but it's a savings plan with 10€ and also 10€ are debited from the bank account).

So is it more correct to import the transaction with the amount of 9,98€ (like it is at the moment) or with the amount of 10€ and a different fund price?

auchri:
Hi @buchen, I removed the "Steuerlicher Anschaffungswert", the previous example is imported with 10€ now. I think that's correct.

buchen:
Thanks @auchri for the new PDF importer.
I have merged the changes :-)

PR #2649, opened 2022-01-17
Write backup files to a subdir

metal450:
Write backup xml files to a subdir next to the main db file, rather than into to the same dir itself.  Addresses this issue (https://github.com/buchen/portfolio/issues/1329) by allowing the backups to be relocated by then symlinking that /backup dir elsewhere.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
@@ -328,9 +328,18 @@ public class ClientInput
             // keep original extension in order to be able to open the backup
             // file directly from within PP
             String backupName = constructFilename(file, suffix);
-
+            
             Path sourceFile = file.toPath();
-            Path backupFile = sourceFile.resolveSibling(backupName);
+            
+            // Write backups to a subdir, so it can be symlinked to a different location
+            // than the main db if desired
+            Path backupSubdir = sourceFile.resolveSibling("backup"); //$NON-NLS-1$
+            File directory = backupSubdir.toFile();
+            if (!directory.exists()){
+                directory.mkdir();
+            }
+            
+            Path backupFile = backupSubdir.resolve(backupName);
             Files.copy(sourceFile, backupFile, StandardCopyOption.REPLACE_EXISTING);
         }
         catch (IOException e)
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
@@ -335,7 +335,8 @@ public class ClientInput
             // than the main db if desired
             Path backupSubdir = sourceFile.resolveSibling("backup"); //$NON-NLS-1$
             File directory = backupSubdir.toFile();
-            if (!directory.exists()){
+            if (!directory.exists())
+            {
                 directory.mkdir();
             }
             


buchen:
I support the idea of allowing the backups in a different directory.
But I don't think the way is to force all of the 50K installations to add a backup sub-directory.

What do you think about this approach:
* If the user configures a backup directory in the preferences, the backup files are stored there.
* If none is configured, then the backup files are stored - as today - next to the XML file

The ```ClientInput``` class already has access to the ```preferences``` to check whether to do backups at all. It should be fairly easy to retrieve whether a backup directory is configured.

The only problem I see: if the user has two files with the same name but in different directories. Then the backup files of those two files end up in the same location. We could workaround this by making the file name unique (for the internal perference files I am hashing the full path), but I think for the backups that is not necessary. If the user configures a backup directory, I needs to understand that the files end up there.

@metal450 - thoughts?

metal450:
> I don't think the way is to force all of the 50K installations to add a backup sub-directory.

It'll autocreate the backup sub-directory (just the same as it autocreates the backups currently) - nobody should be forced to do anything, right? i.e. the change should be invisible to all users, it's just a bit more organized & standard result :)

> What do you think about this approach:
>     * If the user configures a backup directory in the preferences, the backup files are stored there.
>     * If none is configured, then the backup files are stored - as today - next to the XML file

That'd be fine with me, it's just a more complex solution that IMHO didn't really introduce any benefit - figured it's better to keep the software simpler & avoid introducing extra options.  Aside from 'legacy reasons' (i.e. that the software currently uses the same dir for everything), is there a compelling reason to add the complexity of an option?  In my personal experience, putting backups in a subdir next to the data file is pretty much how it's always done (in the absense of an option).  I'm certainly not against an option. Just didn't seem like a top priority feature to add, as my 2c is that it isn't really necessary

> The only problem I see: if the user has two files with the same name but in different directories. Then the backup files of those two files end up in the same location.

Right - that'd only be an issue if adding a new feature of a user-configurable dir.  With this proposal, backups are always next to the main file that owns them, so there couldn't be conflicts.

auchri:
What about checking if a backup file exists beside the actual xml file, if yes, it is used and if not then the backup folder is created and the backup files are stored in the subfolder. This way, nothing should change for existing users and for new users (or new portfolios) the backup files are stored in the subfolder.

Personally, I don't like that the backups are stored in the same place as the XML file.

metal450:
> What about checking if a backup file exists beside the actual xml file, if yes, it is used and if not then the backup folder is created and the backup files are stored in the subfolder. This way, nothing should change for existing users and for new users (or new portfolios) the backup files are stored in the subfolder.
> 
> Personally, I don't like that the backups are stored in the same place as the XML file.

That works, & is certainly easier than a whole separate option. I just didn't really see that any change in location would even matter, as it's silent/transparent to the user.  But if there's a particular reason some people might be against having the backup files be in a subfolder, that could be  way to circumvent it.

auchri:
> is certainly easier than a whole separate option.

Yes. That could be a task for a further PR.

buchen:
> Aside from 'legacy reasons' (i.e. that the software currently uses the same dir for everything), is there a compelling reason to add the complexity of an option?

Not really.

I got some reports on macOS that new installations cannot create the backup file at all. I think it maybe because the application only gets access to those files that the user has explicitly given the application access to (via the open/save dialog). But this problem persists regardless of where the backups are stored - next to the file or in a separate folder.

Other than that, I don't think the "configure backup folder in preferences" option to be complex - or do you mean it as "every additional preference option increases the complexity"?

metal450:
> > Aside from 'legacy reasons' (i.e. that the software currently uses the same dir for everything), is there a compelling reason to add the complexity of an option?
> 
> Not really.
> 
> I got some reports on macOS that new installations cannot create the backup file at all. I think it maybe because the application only gets access to those files that the user has explicitly given the application access to (via the open/save dialog). But this problem persists regardless of where the backups are stored - next to the file or in a separate folder.
> 
> I don't think the "configure backup folder in preferences" option to be complex - or do you mean it as "every additional preference option increases the complexity"?

1) It just seemed overkill to me as it doesn't actually contribute to solving this particular issue (what I was setting out to do). Definitely it'd be a nicety to have in the future, since then you could move the files even without needing a symlink. It was just a feature beyond what I was actually trying to do.
2) I also had the sense from your response to https://github.com/buchen/portfolio/issues/1874#issuecomment-749989981 that  you generally prefer not to add options to the dialogs unless absolutely necessary ("I am hesitating a little bit because it make the code more complex and harder to maintain (and the dialogs are already complex). And if you know PP, you know how long I hesitated to make the decimal places for shares configurable"). So since an option wasn't actually needed to solve this issue, I figured why go to the extra effort to figure out how to do so :)

metal450:
Any chance of getting this in, & approaching the "full customization" feature farther down the road? :pray:

buchen:
Implemented with #2822

PR #2660, opened 2022-01-24
Preset time for new transactions

### LINKED ISSUE #1874: New Transactions Default To Current Time
Suggestion/request: Whenever you enter a new transaction (i.e. buy/sell), the dialog comes up with the date field preset to today's date, which is very handy.  However, the time is always preset to 12:00AM.  It would be much easier to enter transactions if it were preset to the current system time, so that inter-day trades would appear in the correct order (i.e. if I enter a transaction at 15:23, the dialog would come up to reflect that time - rather than coming up showing 12:00AM & then needing to be manually set to the current time for each transaction one enters).

Thanks! :)

tquellenberg:
The time value for new transactions can be preset with either 0:00h (the old standard) or the current time. The behavior can be configured in the preference.

Fixes #1874

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1029,6 +1029,11 @@ public class Messages extends NLS
     public static String PrefTitleDivvyDiary;
     public static String PrefTitleFinnhub;
     public static String PrefTitleQuandl;
+    public static String PresetsPrefPageDescription;
+    public static String PresetsPrefPageNow;
+    public static String PresetsPrefPageStartOfDay;
+    public static String PresetsPrefPageTime;
+    public static String PresetsPrefPageTitle;
     public static String SecurityFilter;
     public static String SecurityFilterSharesHeldEqualZero;
     public static String SecurityFilterSharesHeldNotZero;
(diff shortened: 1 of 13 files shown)


buchen:
The change looks good to me. Will merge once I did a test run.

I wonder about the "Start of Day" label though. I rather tend to think of it as "this transaction has not time". The "00:00" is not displayed in the columns. But I have to say, this way a transaction cannot happen at midnight (the data type LocalDateTime does not support having no time).

Anyway, this is only a small thing. The label is good enough. If folks are happy to have the time, so be it. Personally, I am not sure if it provides meaningful benefit...

tquellenberg:
At the end, the time filed is more about sorting same day transactions than about tracking the exact time of the transaction. I think, most users with one or two transactions per day do not need a time field at all. And users with more transactions need a stable and transparent ordering.

Maybe the options should be: time field yes or no. And if yes, the time is always the current time. And if no, the time field is hidden and the value is set to 0:00h in the background.

metal450:
> Maybe the options should be: time field yes or no. And if yes, the time is always the current time. And if no, the time field is hidden and the value is set to 0:00h in the background.

Being able to manually edit the time (like we can do currently) is still useful, because then you can come back and alter the order  after-the-fact, if desired.

metal450:
In any case, thanks for this change!! Can't wait until it gets released - then I can stop using the custom build I've been running for a few months just to override the 'no default time' issue, hehe :D

metal450:
@tquellenberg : It seems it remembers the time setting for some types of transactions, but some of them aren't working: i.e. Buy/Sell do indeed work, but not Interest/Dividend/Removal

PR #2674, opened 2022-02-03
Reverts same day transaction ordering to date, amount and type.

### LINKED ISSUE #2673: Revert new ordering of same day transactions
**Describe the bug**
#1874 changed the ordering of same day transaction. Before this change transaction were sorted by date, amount and type. The new sorting uses date and updatedAt.
For #1874 there is a second better solution (optional automatic time in addition to date) so that the sorting can be reverted without difficulties.

Two problems arise with the new sorting:
- Imports of same day transactions leads to consistency problems when sell transaction are ordered before buy transactions.
- Every change of a transaction modifies the updateAt and results in an annoying reordering of the transaction list.

tquellenberg:
Fixed #2673

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Transaction.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Transaction.java
@@ -153,18 +153,18 @@ public abstract class Transaction implements Annotated, Adaptable
             if (compareTo != 0)
                 return compareTo;
 
-            // Fallback to updateAt
-            compareTo = t1.getUpdatedAt().compareTo(t2.getUpdatedAt());
+            compareTo = Long.compare(t1.getAmount(), t2.getAmount());
             if (compareTo != 0)
                 return compareTo;
 
-            // Fallback to uuid
-            compareTo = t1.getUUID().compareTo(t2.getUUID());
-            if (compareTo != 0)
-                return compareTo;
+            if (t1 instanceof AccountTransaction && t2 instanceof AccountTransaction)
+            {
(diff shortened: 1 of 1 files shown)


Nirus2000:
Hallo @tquellenberg 
super, dass du dich so toll einbringst. Wenn es dir möglich ist, falls ein "bug" geschlossen wird durch einen PR, 
kannst du diesen schließen in dem du in die Beschreibung z.b. "Fixed #2673" schreibst. Dann wird dieser mit dem
merge von Andreas automatisch mit geschlossen 👍🏻 
Danke :-)

buchen:
Thanks.

If I think about it: we should separate the sorting for the UI from the (special) sorting to match trades (and I should a source comment so that I don't forget). I don't have to the time to check now, but will do so this weekend.

What are your thoughts on a "creation date"? We could add one, but a) it will add to the file size of the XML and b) I am not so sure it leads to the "stable" order that users expect - my transactions are not always entered in order - some are imported from PDF, some via CSV, some entered manually. My thinking is: if you want a specific order, add a time.

tquellenberg:
I don't think that a createdAt is helpful in this situation because it can't be influenced by the user. When I added transactions in the wrong order I have no chance to fix it.

For me, using the existing date/time property is the right way. All users with more than one or two transactions per day should activate the new flag in the preferences. (The main problem with "0:00h" as a default is, that there is no easy way to bring one transaction in front of other same day transactions, without changing the day.)

The other problem is the consistency check, which requires an exact order (and uses the same order as the UI). If the consistency check would be more relaxed and would treat all transaction of a day as exchangeable, half of the problems with the sorting wouldn't exist. Therefore, I think using a different approach for the transaction matching would be the right way.

PR #2711, opened 2022-02-27
PerformanceHeatmapWidget calculates wrong excess return for current month regarding HVPI benchmark 

knorx:
I always had issues with the displayed excess return rate in the monthly heatmap when using an inflation index (in my case DE HVPI). The values were just not true for the current month and jumped around when adjusting the Dashboard interval. The problem is as well described here in the forum: https://forum.portfolio-performance.info/t/falsche-berechnung-normalisierte-monatsrenditen-heatmap-bei-fehlenden-benchmark-werten/17481/2

I have debugged the issue and found that the performance computation in the heatmap UI class does not cover the case when neither the start, nor the end date have values inside the benchmark index. For me this is always the case for the most recent month since the HVPI has only monthly entries for the first of the month and gets published delayed. The current behavior of the code is to simply grab the first and last value from the benchmark regarding the Dashboard interval and calculate the cumulative performance. This is certainly not what you would expect for the current month so I assume this is a bug and not a voluntary heuristic. This also explains why the values differ if the interval is adjusted.

As a remedy I added a special code path to return zero performance in this case where no data is available yet, since neither start, nor end date of the interval are included in the performance index.

I refactored the content of the getPerformanceFor function out into a method of the PerformanceIndex to reduce duplicate code. I furthermore added several tests. I tried to cover all edge cases for the calculations fired from the heatmap UI class. I was unable to successfully run the whole project's tests since I am obviously missing resources some of them expect. Maybe you could add the necessary steps to the Readme? Feel free to relocate or adjust the code to your style.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/PerformanceIndexHeatmapCalculationsTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/PerformanceIndexHeatmapCalculationsTest.java
@@ -1,0 +1,172 @@
+package name.abuchen.portfolio.snapshot;
+
+import static org.hamcrest.number.IsCloseTo.closeTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import name.abuchen.portfolio.AccountBuilder;
+import name.abuchen.portfolio.SecurityBuilder;
+import name.abuchen.portfolio.TestCurrencyConverter;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.money.CurrencyConverter;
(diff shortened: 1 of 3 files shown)


buchen:
The fix looks good. Thanks!

(The build failure was due do formatting issues. I have fixed them and amended your commit)


> I was unable to successfully run the whole project's tests since I am obviously missing resources some of them expect. Maybe you could add the necessary steps to the Readme?

Interesting. I am not aware of any files missing in the git repository. I usually run all tests with a "mvn clean verify" from the command line. The same is done here in Github (although as an initial contributor I have to manually approve that the build runs).

buchen:
Hi @knorx,

due to this [bug report in the Wertpapier Forum](https://www.wertpapier-forum.de/topic/38306-portfolio-performance-mein-neues-programm/?do=findComment&comment=1517545), I have to slightly update the code. The zero performance is returned only if the requested interval does **not** intersect with the interval for which we have performance.

I believe it continues to address the bug reported here: the problem with the HVPI data was that for months _after_ or _before_ HVPI data was available, the full performance was used.

If you have time, it would be great if you can cross check. This is the commit: 6bba6ef6c2aadebbc3b729f5cd6703d6e82e1a7b

PR #2740, opened 2022-03-10
Update quotes for multiple selected securities

OnkelDok:
Currently it's not possible to update quotes for multiple selected securities.
It's only possible to update:
* all (in main menu)
* all active (in main menu)
* selected (only first security when user selects more than one) (in main menu)
* in security menu when one security is selected (in sub-menu "Quotes")

Or am I missed something?

That's why I added a menu action when more than one security is selected in the securities list view (also requested here https://forum.portfolio-performance.info/t/kurse-von-mehreren-selektierten-wertpapieren-aktualisieren/9088).

grafik

Greetings!


tquellenberg:
It's only a matter of taste but I would prefer this construct to cast the array of Objects:
`Arrays.stream(selection.toArray()).map(Security.class::cast).collect(Collectors.toList())`
Thus, the method parameter could be changed to a list of securities.

OnkelDok:
> It's only a matter of taste but I would prefer this construct to cast the array of Objects: `Arrays.stream(selection.toArray()).map(Security.class::cast).collect(Collectors.toList())` Thus, the method parameter could be changed to a list of securities.

Yes, I see the benefits when using a list of securites as parameter in the UpdateQuotesJob constructor with your approach: then no re-conversion to a list is required.
I have to admit I only searched at stackoverflow for casting an untyped array (selection) to a typed one and used this. If other implementation is better, then I can change this. Is there any other advantage of your suggestion, @tquellenberg? I want to learn. :)

My thoughts about using arrays were:
Try to use arrays whenever possible, when the number of the elements is constant and the logic only needs readonly access to the elements (no add/remove required).
I have to admit that I used this because I come from C# development and there it's possible to call "contains" (or similar function from core DLL) directly on arrays.

tquellenberg:
> Try to use arrays whenever possible, when the number of the ele...
I try to avoid using arrays in Java. I prefer List or ImmutableList (from Guava Lib). I don't no if this is an argument, but for me primitive types in Java do not seem 'object oriented'.

PR #2753, opened 2022-03-16
Modify DZ Bank PDF-Importer to support new transactions

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-unioninvest-volksbank-dz-bank/1686/82
Add missing testparts
Optimize regEx
Optimize source
Optimize testcases (Standardization)


Nirus2000:
Hmm?!
Wieso werden die tests nicht gestartet?!

PR #2789, opened 2022-04-06
Onvista PDF import: fee refund

fabtagon:
Parse refunds caused by a decision of Germany's Federal Court of Justice (BGH)


buchen:
looping in @Nirus2000 - LGTM

Nirus2000:
Passed, @buchen 👍🏻

PR #2805, opened 2022-04-09
Add gross value in security currency for some IBFlex transactions.

csprunk:
First attempt with redundancy in code branches and a TODO to understand more.

How to include a test case? The existing tests seem to not trigger the check since the matching with already imported securities happens somewhere else?


csprunk:
Thanks a lot for this great piece of software! I ran into an issue where a security was set up in EUR and after switching brokers is now receiving dividends in USD which trips the IB importer / a check for presence of gross value in security currency.
I am not expecting this to be merged as is, mainly hoping for someone to take a quick look to check if I spotted the correct issue and if that's the case happy to iterate on this to make it acceptable.

Thanks a lot for taking a quick look!

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
```diff
@@ -512,6 +512,8 @@ private void setAmount(Element element, Transaction transaction, Double amount,
                 {
                     fxRateToBase = new BigDecimal(1);
                 }
+                // TODO why are we dividing by the inverse rate instead of
+                // multiplying with fxRateToBase?
```
buchen, line 516: Agree. It should be good enough to multiply by fxRateToBase.

I assume this is a "path dependency" - the code was changed multiple time in the past...

buchen:
> How to include a test case? The existing tests seem to not trigger the check since the matching with already imported securities happens somewhere else?

If you want to test the extractor against an existing (=previsouly imported securities), then you can do this in the test code (for example the Deutsche Bank extractor does that):

```java
    @Test
    public void testDividende01WithSecurityInEUR()
    {
        Security security = new Security("CISCO SYSTEMS INC.REGISTERED SHARES DL-,001", CurrencyUnit.EUR);
        security.setIsin("US17275R1023");
        security.setWkn("878841");

        Client client = new Client();
        client.addSecurity(security);

        DeutscheBankPDFExtractor extractor = new DeutscheBankPDFExtractor(client);
```

buchen:
BTW, I converted this PR to a "draft" for now. Just push more changes (the test case) to the PR and then remove the "draft" flag. Thanks for your contribution!

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
csprunk: Thanks, consolidated in added commit.

csprunk:
Thanks a lot! I added a new test case in accordance with existing style (I think), let me know what you think.

Also let me know (once we reach that point) if it is better for me to squash the commits and force push over the existing ones.

csprunk:
(setting back from draft to ready for review as I assume this is for you to trigger taking a look, feel free to set it back to draft!)

buchen:
Looks good to me. Squashed and merged. Thank you for your contribution!

> if it is better for me to squash the commits and force push over the existing ones.

Typically I use the "squash and merge" feature from Github, so it does not really make a big difference. Just pushing the commits onto the PR. The only thing that makes it harder if one merges the master into the PR branch...

PR #2822, opened 2022-04-22
Move backups to different folder

### LINKED ISSUE #1329: Speicherort des Backupfiles ändern
* Der Pfad für das Backupfile müsste editierbar sein, es direkt neben dem "offenen" File zu legen ist im Zweifel nicht genug Schutz.

auchri:
Write backups to sub directory by default but fall back to same folder if backup already exists.

Closes #2649 

I also add an option to set the backup location in the GUI to fix #1329

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
@@ -331,26 +331,9 @@ public class ClientInput
             // file directly from within PP
             String backupName = constructFilename(file, suffix);
             Path sourceFile = file.toPath();
+            Path backupFile = getBackupFilePath(sourceFile, backupName);
 
-            File oldBackupFile = sourceFile.resolveSibling(backupName).toFile();
-            File backupFile;
-
-            if (oldBackupFile.exists() && oldBackupFile.isFile()) {
-                // Use the old variant (backup in same folder than original file)
-                backupFile = oldBackupFile;
-            } else {
-                // Write backups to a sub directory
-                Path backupSubdir = sourceFile.resolveSibling(DEFAULT_BACKUP_FOLDER);
-                File directory = backupSubdir.toFile();
-                if (!directory.exists())
-                {
-                    directory.mkdir();
-                }
-
-                backupFile = backupSubdir.resolve(backupName).toFile();
-            }
-
-            Files.copy(sourceFile, backupFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            Files.copy(sourceFile, backupFile, StandardCopyOption.REPLACE_EXISTING);
         }
         catch (IOException e)
         {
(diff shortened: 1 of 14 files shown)


auchri:
I've added two options for an relative and an absolute path but it's not really intuitive for now

auchri:
Hi @buchen, 

is it possible to get the currently opened ClientInput from the preference page (the path would suffice)?

auchri:
I changed the GUI, what do you think @buchen ?

The only thing that needs to be resolved is how to get the current opened file.

buchen:
I'll have a look tomorrow (today I spend to much time on getting the build to run again). But in general I like the idea

Nirus2000:
This is a fantastic extension and also a wish of the Comunitiy. Excellent! :star_struck:
There are still some translations missing in the source code (BackupsPreferencePage.java) and maybe I would change "backups" to "backup". (singular)

buchen:
@auchri As said, I like the change to allow more flexibility as to where backups are stored. 👍 

What I do not like is that the default is not to create a sub-folder "backups". At least I do not want PP to create folders. I do want the option to continue with the current behavior.

And the second thing is: if the backup file already exists, then the configuration in the preferences seems to be ignored (see ClientInput line 350). I find this confusing: say a user configures an absolute directory but the backups for existing files continue in the same place - but isn't that what the user wants to change?

I propose the following behavior:
* There are 3 "backup modes": in the same directory, in a sub-directory, or in an absolute directory (create an enum)
* By default, the mode is "in the same directory"
* The user can change the mode in the preferences (pick one
* If the user chooses "in a sub-directory", he can pick a name (default: backups)
* If the user chooses "in an absolute directory", she must pick directory
* If "absolute directory" is chosen, but for whatever reason no directory is configured, then fall back to "same directory"
* If the creation of the sub-directory fails, then fall back to "same directory"
* The code in ClientInput resolves the location of the backups according to the mode (it does not matter which existing backups are written somewhere else)

I think your pull request already goes some way

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/BackupsPreferencePage.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        Path currentDataFile = Path.of("C:\\temp\\bla.xml");
+        Path currentDataFolder = currentDataFile.getParent();
+        Path dest = currentDataFolder.resolve(newValue).normalize();
+
+        String infoText = MessageFormat.format(
+                        "Backups für die aktuelle Datendatei {1} werden in den folgenden Ordner gespeichert: {0}", dest,
+                        currentDataFile);
+        relativeFolderInfoLabel.setText(infoText);
```
buchen, line 126: Please remove the windows specific path here.

In the preferences, we do not have access to a specific file. Therefore, I believe you can only print the name of the sub-folder, e.g. "backups".

I believe in a separate change, we should add the information where backups are written - somewhere to the file. At the moment, there is no good place where to do this. Maybe we extend the tooltip with the name of the file (currently shows the full path of the file). Maybe we add a tab to the "settings" page where bookmarks and attributes are configured. Anyway, let's keep that to a separate pull request (this one is about the backup mode).

auchri:
> I think your pull request already goes some way, do you think you can enhance it with the behavior outline here?

Yes, I'll look into it

auchri:
Hi @buchen,

I changed the PR according to your comment. Can you please verify it works as expected?
Then I will add the missing translations.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
```diff
@@ -62,6 +62,49 @@
...
+        {
+            return title;
+        }
+
+        public String getPreferenceValue()
+        {
+            return getClass().getSimpleName() + "_" + name(); //$NON-NLS-1$
+        }
+
+        public static BACKUP_MODE byValue(String preferenceValue)
```
buchen, line 89: Why not use BACKUP_MODE.valueOf() method (the default in java)?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/BackupsPreferencePage.java
```diff
@@ -0,0 +1,123 @@
...
+            case ABSOLUTE_FOLDER:
+                backupFolderAbsoluteEditor.setEnabled(true, getFieldEditorParent());
+                backupFolderRelativeEditor.setEnabled(false, getFieldEditorParent());
+                return;
+            case RELATIVE_FOLDER:
+                backupFolderAbsoluteEditor.setEnabled(false, getFieldEditorParent());
+                backupFolderRelativeEditor.setEnabled(true, getFieldEditorParent());
+                return;
+            default:
+                throw new RuntimeException("Unsupported backup mode: " + mode.name()); //$NON-NLS-1$
```
buchen, line 120: For those internal errors (which indicate a programming error if they occur), I usually use the IllegalArgumentException (Sonar does not like the generic RuntimeException)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
```diff
@@ -62,6 +62,49 @@
     // compatibility: the value used to be stored in the AbstractHistoricView
     private static final String REPORTING_PERIODS_KEY = "AbstractHistoricView"; //$NON-NLS-1$
 
+    public static final String DEFAULT_RELATIVE_BACKUP_FOLDER = "backups"; //$NON-NLS-1$
+
+    public enum BACKUP_MODE
```
buchen, line 67: I propose to name the Enum with a normal Java name (BackupMode) and extract it from the ClientInput class (because it is used in other places as well). There is no need to have that the inner class of ClientInput.

buchen (approved):
Hi @auchri,

I have merged your change. And I have added another commit with the translations and some (smaller) code changes. Have a look at the changes and please create a pull request if stuff does not work as expected

buchen:
Looks nice. I have adopted the labels to something that the user can hopefully relate to (I am not sure if absolute and relative directory are terms that the user will understand - programmers will)

Bildschirmfoto 2022-04-24 um 16 52 29

PR #2826, opened 2022-04-25
Add support for BE number input

icecoldfire:
In some culture (eg. fr_be and nl_be) it is normal do use the group separator as the decimal separator in the input field. Other applications like Excel convert this automatically to the correct value. So we added code to check if there is only one grouping separator in the string. If so we we assume this is the decimal separator and parse the value.

The code is only executed when Belgium is selected as country.

The main issue is that in belgium we use the AZERTY keyboard with the "." on the numpad:
afbeelding
This is not the case in Germany:
afbeelding

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java
@@ -5,7 +5,6 @@ import java.text.DecimalFormatSymbols;
 import java.text.NumberFormat;
 import java.text.ParseException;
 import java.util.Locale;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import name.abuchen.portfolio.money.Values;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java
@@ -4,6 +4,7 @@ import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.NumberFormat;
 import java.text.ParseException;
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


buchen:
@icecoldfire 

One question: according to Java, the group separator for fr_BE is a space (" ") while it is for nl_BE a dot (".").

Therefore the PR does not work for fr_BE.
Why? Because the dot is not an allowed character for the number string and therefore a dot is not allowed.

Is this behavior also relevant for fr_BE?
Then we cannot use the DecimalFormatSymbols but explicitly go for the dot (but I think that is okay because we check explicitly for the country BE).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java
```diff
@@ -84,20 +90,47 @@ private long convertToLong(String part) throws ParseException
...
+            {
+                // In some culture (eg. fr_be and nl_be) it is normal do use the group separator as the decimal separator in the input field
+                // Other applications like Excel convert this automatically to the correct value
+                // So we check if there is only one grouping separator in the string
+                
+                String[] groups = strBefore.split(Pattern.quote(Character.toString(this.groupingSeperator)));
+                if(groups.length == 2)
+                {
+                    // We found only one grouping separator so we assume this is the decimal separator
+                    before = full.parse(groups[0]);
```
buchen, line 106: If the dot is the first character in the string (Example ".23"), then this parsing will fail.
You will have to manually check if ```groups[0].length() > 0```

icecoldfire:
Hi @buchen 

Thank you for the code review, my code was indeed not that good... So I took the test driven approach and added unit tests. 

For the grouping separators I now remove specific characters for all languages. I think this is a more defensive approach, do find this a good idea?

For the decimal separators I first try the default approach and then I do a fallback to the specific Belgium behaviour. 

With the update the fr behaviour is also fixed and tested.

buchen:
Hi @icecoldfire,

thank you very much for the contribution. It also helped to have the test cases added.

I have now refactored the code much more:
* First, I realized that 10 years ago (when I created the piece of code) I did not understand that one can instruct the DecimalFormat to parse to BigDecimal and therefore avoid the precision problems with Java doubles. That was the reason to separately parse the part before and after the decimal separator. Using the DecimalFormat directly simplifies the code significantly - and thanks to your tests I was able to do this with good confidence 😄 
* Second, I added special handling for the non-breaking space grouping separator. It makes sense to allow the input of both - the space and the non-breaking space. I found similar code in the JavaMoney library.

So I believe overall the work on this pull request made PP better. Please check for your Belgian locale - and create pull request if needed.

PR #2837, opened 2022-05-12
Update trade calendar for Germany

chirlu:
Three updates to the trade calendar for the German stock exchanges:
* Deutsche Börse has again decided to make Whit Monday and German Unity Day trading days for 2022, like it was already between around 2000 and 2014.
* German Unity Day in 2000 (2000-10-03) was wrongly classified as a trading day in PP; while this was Deutsche Börse’s original intention, trade unions prevented it from being implemented.
* Whit Monday in 2007 (2007-05-28) was wrongly classified as a trading day in PP, too. In that year, Whit Monday coincided with Memorial Day, so that NYSE was closed as well, and it didn’t seem meaningful (or lucrative) for Deutsche Börse to open on that day.


chirlu:
With these fixes, I’m rather confident that the trade calendar is correct and complete since at least 1995, and probably even back to the 1950s.

PR #2841, opened 2022-05-17
Modify UBS PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-ubs-ag/20627/2
Fix buy and sell transaction.
See in commit https://github.com/buchen/portfolio/pull/2796#issuecomment-1128420078


buchen:
Hi @Nirus2000,
I convert this pull request to a draft until we have feedback from @eztam-
Just changing the "tax" values in the test files fixes the calculation, but then we do not see the root cause anymore

eztam-:
Hi all,
I added a comment on https://github.com/buchen/portfolio/pull/2796#issuecomment-1128420078

Nirus2000:
Hello @buchen 
as I thought... I have rebuilt the UBS Importer to be the same in structure as the other importers. It wouldn't matter now if we adjust the fees in the PDF debug, or the net amount. The pull request will extend the importer anyway and adjust the test cases to the same standard. Thanks to @eztam- for the recheck. 👍🏻 

Greetings
Alex

eztam-:
Hi @Nirus2000 ,
thanks for all the good work on the UBS importer! I'm very happy to see the added support for dividends!
I've been using the importer now for a few months whereby I've discovered a few edge cases for rather exotic transactions, that would need some small improvements. I'll wait with submitting them until your changes are done.

Regards,
Matthias

Issue #2139, opened 2021-03-07
Importing dividends in different currency from security (Interactive Brokers Flex Query)


delianenchev:
I have the following error when importing a Flex Query: `Exchange rate of gross value is missing (transaction currency USD and security currency EUR)`. Flex Query have [`fxRateToBase`](https://guides.interactivebrokers.com/rg/reportguide/cash_transactionsfq.htm)
![Image](https://user-images.githubusercontent.com/1281108/110248099-1d575d80-7f78-11eb-9956-c769a576e647.png)

First import `Trades.xml`

`
<FlexQueryResponse queryName="Trades" type="AF">
<FlexStatements count="1">
<FlexStatement accountId="UXXXXXX" fromDate="20190826" toDate="20191231" period="LastBusinessDay" whenGenerated="20210307;043303">
<Trades>
<Trade accountId="UXXXXXX" acctAlias="" model="" currency="EUR" fxRateToBase="1" assetCategory="STK" symbol="SGN" description="SURGUTNEFTEGAS-SP ADR" conid="27346611" securityID="US8688612048" securityIDType="ISIN" cusip="" isin="US8688612048" listingExchange="SWB" underlyingConid="" underlyingSymbol="" underlyingSecurityID="" underlyingListingExchange="" issuer="" multiplier="1" strike="" expiry="" tradeID="2592275115" putCall="" reportDate="20191030" principalAdjustFactor="" dateTime="20191030;091642" tradeDate="20191030" settleDateTarget="20191101" transactionType="ExchTrade" exchange="TGATE" quantity="120" tradePrice="6" tradeMoney="720" proceeds="-720" taxes="0" ibCommission="-6" ibCommissionCurrency="EUR" netCash="-726" closePrice="5.965" openCloseIndicator="O" notes="" cost="726" fifoPnlRealized="0" fxPnl="0" mtmPnl="-4.2" origTradePrice="0" origTradeDate="" origTradeID="" origOrderID="0" clearingFirmID="" transactionID="11501005" buySell="BUY" orderReference="" volatilityOrderLink="" exchOrderId="N/A" orderTime="20191030;080542" openDateTime="" holdingPeriodDateTime="" whenRealized="" whenReopened="" levelOfDetail="EXECUTION" changeInPrice="0" changeInQuantity="0" orderType="LMT" traderID="" isAPIOrder="N" accruedInt="0" />
</Trades>
</FlexStatement>
</FlexStatements>
</FlexQueryResponse>
`

[...]

PR #2850, opened 2022-05-21
Fix IBKR import and value calculation of securities listed in other currencies

RadoslavGatev:
I was facing an issue that prevented me from importing a security listed in other exchanges (hence in different currencies). I guess the issue occurs when you hold the security from various exchanges. 

My setup:
- EUR security account
- USD security account

The security is originally listed in USD but also on German exchanges in EUR. At first, you can have the originally listed security in USD but at some point, you may decide to buy the one listed in EUR. Then the issue occurs because the EUR listed security resolves to the same security listed in USD (I guess because it has the same isin). 

Therefore, the gross amount of the portfolio transaction must be set. Because of the order of operations in the code and the absence of currency code on the transation, gross value was not set. 
I spotted also an error in the calculation of value because there is an assumption that the value of the investment is in the currency of the security.

Caution: I am not an expert in Portfolio Performance.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/IBFlexStatementExtractor.java
@@ -373,18 +373,11 @@ public class IBFlexStatementExtractor implements Extractor
                 }
             }
 
+            transaction.setSecurity(this.getOrCreateSecurity(element, true));
+            
             // transaction currency
             String currency = asCurrencyUnit(element.getAttribute("currency"));
-
-            // Set the Amount which is "netCash"
-            Double amount = Math.abs(Double.parseDouble(element.getAttribute("netCash")));
-            setAmount(element, transaction.getPortfolioTransaction(), amount, currency);
-            setAmount(element, transaction.getAccountTransaction(), amount, currency, false);
-
-            // Share Quantity
-            Double qty = Math.abs(Double.parseDouble(element.getAttribute("quantity")));
-            Double multiplier = Double.parseDouble(Optional.ofNullable(element.getAttribute("multiplier")).orElse("1"));
-            transaction.setShares(Math.round(qty.doubleValue() * Values.Share.factor() * multiplier.doubleValue()));
+            transaction.getPortfolioTransaction().setCurrencyCode(currency);
 
             // fees
             double fees = Math.abs(Double.parseDouble(element.getAttribute("ibCommission")));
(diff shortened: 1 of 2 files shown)


RadoslavGatev:
It likely fixes #2139

buchen:
Hi @RadoslavGatev,

thanks for taking the time to build this contribution.

The internal model which PP expects makes the following assumption: the currency of a transaction either matches the currency of the security *or* there is a gross value unit (including the exchange rate) that matches the currency of the security. If you have the security configured with the currency USD, then either the transaction must be in USD or there must be a gross value unit including the exchange rate from the account currency to USD.

> I guess because it has the same isin

Correct. The importer typically work this way: they try to identify/extract the ISIN. If the ISIN is found, the existing security is used. If it is not found, a new security object is created with the information in the statement.

I am a little confused because I would have expected that an ISIN also denotes the currency.

From your description, I assume you run into the following problem:
* Initially the security is imported with USD as currency (because it was traded on a USD exchange)
* Then you do additional trades for the same security in EUR using an account in EUR
* The imported transactions only contain EUR (there is no exchange rate) and then PP complains

There are two ways to address this: change the way PP works (big) or use the exchange rate of the day to create the appropriate gross value (smaller).

For the second option, it would require to inject the ```ExchangeRateProviderFactory``` in the constructor of the ```IBFlexStatementExtractor``` and then use it if the exchange rate cannot be extracted from the statement itself.

Let me ask about your PR (IB flex was contributed, I am no expert). The test cases fail because of this block:

```
String currency = asCurrencyUnit(element.getAttribute("currency"));
transaction.getPortfolioTransaction().setCurrencyCode(currency);
```

Up until now, the code was looking at the ```AccountInformation``` in the XML and extracting the currency there as the currency for the transaction (see the code in the ```setAmount``` method). With your change, there is now a different transaction currency which prevents importing taxes and fees. Do you know how Interactive Brokers is handling this? In what situation can the currency of the account differ from the currency of the transaction element?

Andreas.

RadoslavGatev:
Hi @buchen,

Thanks for the provided information!

What you described is exactly the issue I run into. It has to do with the absence of gross value. But it seems the way you proposed (using the `ExchangeRateProviderFactory`) is better. My current PR supplies the gross value in EUR while the security is in USD and that causes other problems. That's why I had to change the logic in `SecurityPosition` to respect the currency of the transactions.

I'll have to change my approach a bit and I'll see if unit tests still complain. To be honest, I haven't executed them locally.

> Up until now, the code was looking at the `AccountInformation` in the XML and extracting the currency there as the currency for the transaction (see the code in the `setAmount` method). With your change, there is now a different transaction currency which prevents importing taxes and fees. Do you know how Interactive Brokers is handling this? In what situation can the currency of the account differ from the currency of the transaction element?

The Flex statements are kind of messy, in my opinion. The concept of the base currency in Interactive Brokers is just for the sake of reporting. The same thing is applied to their XML reports - the `AccountInformation` element carries the base currency, and its child elements (Trades and others) can be of any currency, therefore they provide the `fxRateToBase` attribute. If a particular transaction is in the base currency, the rate is set to 1.

It's somewhat difficult to import those into PP. I track 2 pairs of cash and securities accounts: EUR and USD. Before I start importing, I open the XML, comment out the elements that are not in the base currency, and import to the respective account (that matches my base currency) in PP. Then, I change the account currency to whatever currencies I have traded in, in my case it's USD, and then uncomment the USD ones while commenting out the EUR (which were imported in the previous step).

PR #2857, opened 2022-05-24
New column "Delta latest price to SMAx"

OnkelDok:
Hello,

the percentage distance of the current price of a security to its SMA allows statements whether a security is weakening or strong in the short/long term (depending on the SMA selected).
Therefore I have added a corresponding column option which will display distances to different SMAs (following the available SMA intervals in the security charts) as a sortable column.
This allows the comparison of this indicator for the listed securities.

Here you can see this column used in the DAX sample file:

https://user-images.githubusercontent.com/90478568/170118071-9092841b-7680-4440-9bc3-7c616cefeca2.mov

Requests in forum to this topic:
* https://forum.portfolio-performance.info/t/abstand-des-aktuellen-kurses-zum-gd200-gleitender-durchschnitt/5752
* https://forum.portfolio-performance.info/t/sma-in-vermoegensaufstellung/4475/3
* https://forum.portfolio-performance.info/t/berechnung-eines-dip-finders/20734

Best greetings!


buchen:
Hi @OnkelDok,

nice change!

Two thoughts:
* There is already a class ```SimpleMovingAverage``` which is calculating the moving average. It is covered by tests and taking care of all the edge cases (not enough prices, prices series with gaps, etc.) It is used for the charts. At the moment it is used for the charts. I propose we (slightly) refactor that and use the same calculation logic.
* One can view the statement of assets for another date in the past (the "time machine" button in the tool bar). I would expect that the "distance to moving average" is then also calculated relative to the chosen date. I think one option would be to pass in a "date provider" into the QuoteToSmaDeltaColumnHelper that is either using today's date (LocalDate.now()) or the date from the model (model.getDate())

What do you think?

About #2503: I have seen the remarks about the color of the line. I understand that. I am all about accessibility. I will try to find time and have a look at the change.

OnkelDok:
@buchen: thank you for your feedback.
Yes, I completely agree to your points. I will see how I can implement it.

> 
> About #2503: I have seen the remarks about the color of the line. I understand that. I am all about accessibility. I will try to find time and have a look at the change.

All good. I can only guess how much effort it takes for you to work on this great software project and manage and review all the ideas, improvements and fixes.
I think first step would be to finish this PR. Afterwards the configurable SMA interval solution of #2503 maybe needs some changes.

auchri:
What's about EMA?

OnkelDok:
I made only some small changes to use the SMA calculation logic in `SimpleMovingAverage`.
My first thought to do this was to use fully the same method that is used by the chart for the SMA lines and pass an interval with same start end end date to get the SMA for a single date.
But I realized that this wouldn't calculate SMA for follwing cases:
* when the requested date is after the latest price date, no calculation is made: https://github.com/buchen/portfolio/blob/effa5b7baf9a918e1b5fe83942ddc480e0fd48b9/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SimpleMovingAverage.java#L68-L69
* when for the requested date no price is available (but prices available after this date), then the SMA would be calculated for the next date after the requested date where a price is available: https://github.com/buchen/portfolio/blob/effa5b7baf9a918e1b5fe83942ddc480e0fd48b9/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SimpleMovingAverage.java#L65-L66

For both cases, I think, the delta of price to SMA should be calculated with the last existing price before the requested date.
So even if for a security the latest price is far in the past, the value of the last sma-price-delta should be shown.
Do you have a different opinion on this?

So I decided to only refactor the part of code which is common (calculation of the SMA for a given amount of prices):
https://github.com/buchen/portfolio/pull/2857/files#diff-b86bc2226412765ddf0ce19ecee890acb280191ce7ecc7c5620dacc521e5c3a8R96-R100


I created a new method `getLatestNPricesOfDate` on the `Security` which returns the latest N prices for a given date (this should also allow easy implementing the planned change for the "time machine").
https://github.com/buchen/portfolio/pull/2857/files#diff-3ef6df0dd23c035a077044440dc3f5765b3ba9f5fae8a704cb1964369ce07a26R362-R382

With this I call the common SMA calculation function and afterwards calculate the relative distane from price to this SMA.
https://github.com/buchen/portfolio/pull/2857/files#diff-40ac3707c4fbb90c03d4c0d807814e4bf77d5a7ce372d5bbf4d2eef4c5149d4fR42-R48

Next step would be to connect the time machine with the logic.


> [auchri](https://github.com/auchri) wrote:
> What's about EMA?

@auchri, currently you are the first who requests also the delta to EMA. If @buchen agrees than it should be easy to also implement this. Maybe in a separate PR when the SMA is done?

buchen:
To be honest, I have troubles to understand how to define the *simple moving average* taking into consideration that PP cannot make any assumption that the historical prices are complete. This is independent of the change you propose, of course.  As you can see from the chart below, the SMA is taking the last X days (38 or 100 in this example) regardless of big gaps in the historical prices. Is this what you are expecting? (I do not know much about chart analysis).

I realized there was a discussion in the forum some years ago: assume the historical prices are complete. When looking at the chart, I can at least see this directly. But looking at the number in the table, it is difficult to see (but then again, one click and the diagram can also be displayed in the statement of assets). So let's leave it for now.

Bildschirmfoto 2022-05-28 um 09 58 36

---

About your change:

Overall, looks good to me. I propose a couple of minor changes (if you want, I can apply them later today as I am planning a release).

* Naming: I propose to name it "Distance from Moving Average" - see [here](https://school.stockcharts.com/doku.php?id=technical_indicators:distance_from_ma) and [here](https://www.leebullbear.com/post/distance-from-moving-average-dma-indicator)
* ```getLatestNPricesOfDate``` do not use Integer but int instead (no checking for NPEs, etc.)

> when the requested date is after the latest price date, no calculation is made

I agree. But I think there has to be some "cut off date". If the latest prices are months, years in the past, then it does not make sense to calculate the distance to the moving average. (That is easily possible if one just keeps updating the current price, but deactivates the historical prices)

> when for the requested date no price is available (but prices available after this date), then the SMA would be calculated for the next date after the requested date where a price is available

At least I would also say we need some kind of "cut off" here. But as a I user, I would find it confusing to see the distance calculated to a future date (compared to the price).

Assuming one generally looks at today's moving average, I think the distance to the moving average should be calculated using the most recent calculated moving average going back at most 10 days.

OnkelDok:
Regarding the missing prices: yes it's difficult to handle this. When the user want a correct calculation, then all prices should be present without gaps.
I think the current calculation (use the last x prices and calculate the average) is OK also for gaps.
For the new column we can give the user a feedback if there are gaps in the prices or if the date of the SMA is far away in the past. We can display an icon (e.g. exclamation mark) and display a tooltip with some notes.

Then it's like in the security chart where the user can display dates with missing prices and is able to estimate that the SMA is not quite correct.



> 
> > when the requested date is after the latest price date, no calculation is made
> 
> I agree. But I think there has to be some "cut off date". If the latest prices are months, years in the past, then it does not make sense to calculate the distance to the moving average. (That is easily possible if one just keeps updating the current price, but deactivates the historical prices)
> 
> > when for the requested date no price is available (but prices available after this date), then the SMA would be calculated for the next date after the requested date where a price is available
> 
> At least I would also say we need some kind of "cut off" here. But as a I user, I would find it confusing to see the distance calculated to a future date (compared to the price).

Both of my cases in my last comment were about the current calculation for the SMA chart lines.
Maybe I described the seconds one not good enough so it was a little bit confusing. What I meant was:  these index checks are there to find the start date for the first SMA chart dot calculation. So it is a date after the `interval.getStart()` date. Here, in the chart it makes sense to use this logic.
But for the SMA column it should not calculate the SMA for dates after the requested date (time machine).

Yes, we can define cut off "time spans" where no calculation should be made. If so, I suggest to display it to the user (icon with tooltip). Because then she or he knows why the field is empty.
Alternatively we can still calculate the distance from SMA and display some icon and tooltip, to notify that it's an outdated SMA.

Regarding your name proposal: Do you mean naming in sourcecode (classes, functions, ...), naming of the labels in the GUI or both?

buchen:
rebased, squashed, merged

PR #2860, opened 2022-05-29
Some improvements related to trade calendars

chirlu:
Some improvements related to trade calendars. The only user-visible change is in the trade calendar overview (in the Preferences), where holidays that fall on a weekend are marked with their day of week. Some exchanges don’t list them in their yearly trade calendars (e.g., Deutsche Börse [doesn’t list](https://www.xetra.com/xetra-de/handel/handelskalendar-und-zeiten) Labour Day or Christmas Day this year), so this may avoid some confusion. Example (with many holidays on Saturday and Sunday this year):
Preferences window with trade calendar overview
Apart from that, I’ve removed some redundant code, made weekends configurable for trade calendars, fixed a typo, etc.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TradeCalendarTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TradeCalendarTest.java
@@ -310,7 +310,7 @@ public class TradeCalendarTest
     @Test
     public void testEmptyCalendar()
     {
-        TradeCalendar calendar = TradeCalendarManager.getInstance(TradeCalendar.EMPTY_CODE);
+        TradeCalendar calendar = TradeCalendarManager.getInstance("empty");
 
         // we generate a random day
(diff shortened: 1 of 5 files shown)


chirlu:
@buchen, in the overview, I’d like to offer a choice which year to display. (This is why, in the last commit, I already made `year` a parameter to the method creating the holidays list.) Users could then have a preview of next year’s calendar, or even look up what the non-trading days were in 2007. However, I’m not sure how to add a control to a preferences tab *that is not treated as a setting and saved* (which obviously doesn’t make sense here).

buchen:
> However, I’m not sure how to add a control to a preferences tab that is not treated as a setting and saved (which obviously doesn’t make sense here).

If you do not want the UI controls mapped to preferences, you have to create them manually. Have a look at the [LanguagePreferencePage](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/LanguagePreferencePage.java) which creates the drop-down elements manually.

PR #2866, opened 2022-05-31
Extract Security exchange from DeGiro transaction PDFs and Display it

ah4c:
DeGiro includes the 3 character exchange of the security for BuySell transactions in the transaction PDFs.

Portfolio Performance does not extract the exchange, which I would like to see.

This pull request adds exchange to the Security model class and extracts the exchange by modifying the regexes of the existing 15 DeGiro extraction sections of DegiroPDFExtractor.

The exchange is displayed in a new column in the import dialog (ReviewExtractedItemsPage) and in a new column in the Securities view (SecuritiesTable).

Asserts for the exchange have been added to all JUnit tests for DeGiro Transaction PDFs in DegiroPDFExtractorTest.

This does not change the PDFExtractors for any other banks.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/degiro/DegiroPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/degiro/DegiroPDFExtractorTest.java
@@ -1605,24 +1605,28 @@ public class DegiroPDFExtractorTest
         assertThat(security1.getIsin(), is("DE000C21EMZ1"));
         assertThat(security1.getName(), is("ODX5 C11500.00 29MAR19"));
         assertThat(security1.getCurrencyCode(), is(CurrencyUnit.EUR));
+        assertThat(security1.getExchange(), is("ERX"));
 
         Security security2 = results.stream().filter(i -> i instanceof SecurityItem).skip(1).findFirst()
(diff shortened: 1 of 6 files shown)


buchen:
Hi @ah4c,

I am hesitant to add a property "exchange" directly to the security. It is one of those things that I can never undo (due to the nature of the XML). Plus: the way I see it, I can have *one* security that is traded at *multiple* exchanges (or markets or whatever you want to call it). There is a 1-N relationship. Adding the property to the security would have more implications. It is separate instruments. But it is possible to purchase a stock from one exchange (say Frankfurt) and sell it at another (say Stuttgart). If that is mapped to separate instruments, then this would not be possible.

(Already today the quote provider already runs into that problem. I take prices from one exchange, to reflect the value. But to get to a valuation, one has to pick one exchange. It doesn't add value to take an average from multiple exchanges or always the highest or lowest)

Then: the problem is, that you get the exchange of the first purchase or sale. But this exchange does not have to relate to the exchange for which you get prices. Therefore I do not see what the value is - you know what exchange you did the first trade?

If users want to maintain "exchange" information, they can create a custom property for securities and maintain it as they see fit.

As @Nirus2000 pointed out, another option is to keep it in the notes of the transaction (the transaction has no other structured free properties at the moment). Then you can see on which exchange this particular purchase or sale was made. I think that would be okay, although I am not sure how useful this information for many people.

I will close this PR. Just push another PR for the change to notes

PR #2875, opened 2022-06-04
Rename UN resolution from Turkey in all languages to Türkiye

Nirus2000:
The UN has confirmed the change of name of Turkey: The country will be called "Türkiye" in all languages from now on. The English name Turkey will no longer be used at the request of the government in Ankara.


chirlu:
I hope you mean this as a joke.

Nirus2000:
https://www.tagesschau.de/ausland/turkey-tuerkiye-101.html
https://rp-online.de/politik/ausland/tuerkei-uno-bestaetigt-namenswechsel-tuerkiye-statt-turkey_aid-70785171

chirlu:
I meant it’s a joke that PP should follow Erdoğan’s whims. But you seem to mean it …

Will you also rename the other names that don’t follow official policy, please? North and South Korea and Taiwan are badly wrong, among others.

Nirus2000:
For the time being I will refer the pull request exclusively to this. 
I will be happy to leave the other changes to you ;-)

buchen:
To be honest, I am here with @chirlu. I read this as a request to the UN to change the naming. And in English, there might even be an argument. But in German, "Türkei" continues to be used even in the articles about the request to name change. Also, [Wikipedia](https://en.wikipedia.org/wiki/Turkey) continues to use Turkey (but it reads on the page "_"Türkiye" and "Turkiye" redirect here._"). And it also replaces the Cyrillic name with the latin alphabet. If the future brings change, we'll also adopt. For now, I close this PR.

PR #2878, opened 2022-06-06
Add FinnHub.io as Security Search provider when adding instruments

ah4c:
Added FinnhubSearchProvider and FinnhubSymbolSearch to use the free Finnhub.io Symbol Lookup API to search for a security by symbol or ISIN when adding a new investment vehicle. This works in the same way as the existing YahooSearchProvider and provides additional results.

If the user has not configured a Finnhub.io API key then the search API is not called. This avoids 401 Unauthorized for those users.

I added a getSearchProvider(String searchId) method to Factory to make it possible to retrieve one particular search provider by ID.

The Finnhub API key is provided to FinnhubSearchProvider by Preference2EnvAddon in the same setFinnhubApiKey() method where it is provided to the existing FinnhubQuoteFeed.

https://finnhub.io/docs/api/symbol-search


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/FinnhubSymbolSearch.java
```diff
@@ -0,0 +1,170 @@
...
+public class FinnhubSymbolSearch
+{
+    private String apiKey;
+
+    public FinnhubSymbolSearch(String apiKey)
+    {
+        this.apiKey = apiKey;
+    }
+
+    public void setApiKey(String apiKey)
```
buchen, line 36: I don't think that method is used (as the API key is only set via the constructor).

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/FinnhubSearchProvider.java
```diff
@@ -0,0 +1,78 @@
...
+            FinnhubSymbolSearch.Result item = new FinnhubSymbolSearch.Result(Messages.MsgMoreResultsAvailable);
+            answer.add(item);
+        }
+
+        return answer;
+    }
+
+    private void addSymbolSearchResults(List<ResultItem> answer, String query) throws IOException
+    {
+        Set<String> existingSymbols = answer.stream().map(ResultItem::getSymbol).collect(Collectors.toSet());
```
buchen, line 73: I am not sure why this check for duplicate symbols is needed. I would propose to just return what Finnhub returns

(I think the code is inspired by the Yahoo symbol search. But the Yahoo search provider, does *two* queries and then tries to merge the results. In that case it makes sense to remove duplicate results).

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/FinnhubSearchProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+import name.abuchen.portfolio.Messages;
+import name.abuchen.portfolio.online.SecuritySearchProvider;
+
+/**
+ * Use the <a href="https://finnhub.io/">Finnhub</a> API to search for securities.
+ * 
+ * @see FinnhubSymbolSearch
+ */
+public class FinnhubSearchProvider implements SecuritySearchProvider
```
buchen, line 18: Why do you split the code into two classes: FinnhubSearchProvider and FinnhubSymbolSearch? The former is just calling the latter. I think it will be easier to read (and maintain in the future) if it is just one class.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/SecuritySearchProvider.java
```diff
@@ -9,6 +9,11 @@
 
 public interface SecuritySearchProvider
 {
+    /**
+     * Returns the technical identifier of the security search provider.
+     */
+    String getId();
```
buchen, line 15: I understand that you need to get a specific search provider to set the API key. But i would propose to do it with the class name (like it is done for the dividend feed). That is a compile time dependency that makes it easier to change the code later.

I know the quote feed does it that way. But... a) it is old code which I could consider to refactor at some time and b) the ID is needed to store the configured quote provider in the XML file as I do not want to have the implementation class name in the XML file (there are already some cases where I accidentally leaked the class name into the XML and it is pain in the back to change).

```
    public static <F extends DividendFeed> F getDividendFeed(Class<F> feedType)
```

buchen (changes requested):
Hi @ah4c,

the change is a nice addition - it works as expected when running from the UI. Nice.

I have proposed some change to the code. Can you be so kind and consider my comments?

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/FinnhubSearchProvider.java
ah4c: You are correct. I copied the code and forgot to remove the duplicate symbol search. I will remove it.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/FinnhubSearchProvider.java
ah4c: I will combine them. Again, I copied the Yahoo code and did not refactor it.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/SecuritySearchProvider.java
ah4c: Good idea. I was wondering why the DividendFeed lookup was not the same as the quote feed.

buchen:
Great. Merged. 👍 😄

PR #2899, opened 2022-07-04
Add DivvyDiary.com as Search Provider when adding new instruments

ah4c:
Added DivvyDiarySearchProvider to use the free DivvyDiary.com Symbols Lookup API to search for a security by ISIN when adding a new investment vehicle. This works in the same way as the existing FinnhubSearchProvider and provides additional results.

If the user has not configured a DivvyDiary.com API key then the search API is not called. This avoids authentication errors for those users. The API only accepts valid ISINs for the search API so queries that are not exactly 12 characters after trimming spaces are ignored. This avoids validation errors.

The DivvyDiary.com Symbols API returns:
- symbol
- name
- WKN
- ISIN
- exchange (4-character MIC)
- currency (3-characters)
- dividendCurrency (3-characters)
- dividendFrequency

The symbol does not include an exchange suffix. The getExchangeSuffixFromIsoMic() method uses the 4-character MIC exchange to get a suitable suffix for the symbol. 60 exchanges are currently mapped.

The DivvyDiary.com API key is provided to DivvyDiarySearchProvider by Preference2EnvAddon in the same setDivvyDiaryApiKey() method where it is provided to the existing DivvyDiaryDividendFeed.


ah4c:
Deleted this PR and replace it with a new one due to some Git issues.

PR #2908, opened 2022-07-11
Description dashboard widget

ma4nn:
This PR introduces a new dashboard widget to show a simple descriptive text next to the KPIs that is wrapped correctly and displayed in the regular font style.
By using the `StyledLabel` component the user can also use some simple markup like `<green>`, `<red>`, `<strong>`, `<em>` and `<a>`.


buchen:
Looks good to me. Thanks for your contribution.

buchen:
Bildschirmfoto 2022-07-17 um 11 28 26

I'll slightly tweak the context menu...

buchen:
I wonder if we should rename it to "Note" (Notiz). After all it is something that the user enters, not an existing description.

buchen:
rebased and merged

ma4nn:
> I wonder if we should rename it to "Note" (Notiz). After all it is something that the user enters, not an existing description.

@buchen yes, good idea 👍 But better leave the widget type now as it is and only change the labels/translations as otherwise the existing widget data will be lost, right? Shall I create a new PR for the renaming?

PR #2909, opened 2022-07-14
Group trade activity by year in dashboard widget

ma4nn:
This PR adds 2 new diagram types for the trade activity dashboard widget:

1. Count by Year
2. Sum by Year

This allows the user to group the trade activity by years which is especially useful for longer time periods.


buchen:
mreged.

PR #2912, opened 2022-07-16
Git ignore IntelliJ configuration files

ah4c:
I often use IntelliJ for running static code analysis and creating diagrams to aid my understanding of Portfolio Performance. Perhaps other people do as well. This pull request just ignores some standard IntelliJ configuration files by adding them to `.gitignore` without changing any other files.

This includes the following:
* `.idea`
* `*.iws`
* `*.iml`
* `*.ipr`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,9 @@ workspace
 .project
 .recommenders/*
 .DS_Store
+
+### IntelliJ IDEA ###
+.idea
+*.iws
+*.iml
+*.ipr
--- a/name.abuchen.portfolio.bootstrap/.gitignore
+++ b/name.abuchen.portfolio.bootstrap/.gitignore
@@ -1,5 +1,0 @@
-### IntelliJ IDEA ###
-.idea
-*.iws
-*.iml
-*.ipr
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -3,4 +3,10 @@ workspace
...
 .recommenders/*
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+### IntelliJ IDEA ###
+.idea
+*.iws
+*.iml
+*.ipr
```
buchen, line 12: I am fine to add the pattern, but I think it is enough to add the pattern _once_ at the root level. It will ignore any files in the sub-directories (just as the ```.DS_Store``` matches any such file in any subdirectory)

ah4c: Thank you. That is simple and elegant. I have committed and push the code.

buchen:
> creating diagrams

BTW, what diagrams do you mean there? I would be interested in new ways to understand the code

ah4c:
I use https://vanco.github.io/SequencePlugin/ to create a sequence diagram of a method. The sequence diagram can be exported as Mermaid, which can be embedded in Markdown files.

Github recently added support for Mermaid so it could be interesting to add some diagrams to explain how Portfolio Performance works.
https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

PR #2918, opened 2022-07-18
Browser dashboard widget

ma4nn:
This PR creates a new dashboard widget for simply integrating websites.
JavaScript is disabled in the browser for security reasons (to be discussed).

The widget can be used e.g. for adding news headlines or other external information (charts, data, etc.) to the dashboard. It could also be a way to develop additional (HTML) dashboard components that do not need to be implemented as Java code.


buchen:
Nice idea.

> It could also be a way to develop additional (HTML) dashboard components that do not need to be implemented as Java code.

The browser widget works good on macOS, okay on Windows, and not very good on Linux.
For the pie charts, I am using the browser component. Because it often does not work on Linux (and the charts just remain blank), there is an alternative pie chart implementation for Linux.

For some time, I was hoping for the Chromium SWT component. The intention was to embed Chromium into SWT. That way I could have delivered the same browser including the native binaries on all platforms. Unfortunately, the Chromium component was always only beta and eventually was removed from SWT again.

I am okay to having the browser widget. It is an optional component. But for now I do not intend to implement more widgets as HTML unless there is a stable solution for all platforms.

buchen:
rebased, squashed, merged

PR #2920, opened 2022-07-18
Distance from All Time High (ATH)

ma4nn:
This PR adds a new column "Distance from ATH" to both the securities table as well as the statement of assets viewer.
In addition a new dashboard widget is also available to show the ATH distance for one single security.

For each of these 3 elements the reporting timeframe for the ATH can be chosen individually, e.g. ATH in the past 3 years/months/days, etc.
 
Issue: https://forum.portfolio-performance.info/t/kurse-relativ-zum-letzten-allzeithoch/19050
Issue: https://forum.portfolio-performance.info/t/abstand-zum-hoechstkurs/6131


Nirus2000:
i like your ideas... :-)
hmm... is the distance taken here to the "all time high" in the reporting period or overall?

I think we should try to write out the abbreviations "ATH". It could be possible that some users, do not understand these abbreviations. A description would therefore be helpful via mouseover would be more helpful.

ma4nn:
Thanks for your feedback @Nirus2000!

Good point, the distance is taken to the high in the chosen reporting period, so perhaps "all time high" is not an ideal label. And it also works in combination with the "time machine" feature in the statement of assets view.

I also agree regarding the abbreviation, I just looked at the similar "distance to SMA" column and assumed that abbreviations were self-explanatory (also used e.g. for "GLD"), in addition I wanted to keep column names short.

What do you think about renaming it to something like "Distance to Previous High" / "Abstand vom letzten Hoch"?

Nirus2000:
> What do you think about renaming it to something like "Distance to Previous High" / "Abstand vom letzten Hoch"?

I think that will be up to Andreas to decide how he likes it better. It could lead to misunderstandings if a filter is in use. Maybe there is a possibility to show the "All-Time-High" of all existing data and the one which is available in the reporting period or from "first purchase".

buchen:
I wonder if the better approach is to have a "all time high" column - **without** configuration of a reporting period - which does exactly this: compare to the all time high until today (in case of the time machine, the date might be in the past).

I would assume this is what most users want (although those numbers are kind of depressing, because they can only be negative... 😉)

Then we could have a _separate_ column called "Distance to Previous High of given interval". One could use it to say: compare to the high of the last 12 months. But do we need this value? I just wonder if it is adding options that only a fraction of users will use and which have limited analytical informative value...

Pnda87:
Personally, I'd like to see both columns. And even extend the PR to all-time-low. There, obviously the column previous-low-of-given-interval is more important

use case examples (->given interval):
distance from pre-corona-crash high
distance from corona-crash low

ma4nn:
Thanks both for your feedback 👍 

> Then we could have a separate column called "Distance to Previous High of given interval". One could use it to say: compare to the high of the last 12 months. 

In my opinion the current approach **with** the configuration of the period allows for both of the mentioned use-cases without having 2 separate columns: one could simply create a new reporting period that covers the whole time (or a very big time interval) and a new one with the shorter time frame of e.g. 12 months.

> But do we need this value? I just wonder if it is adding options that only a fraction of users will use and which have limited analytical informative value...

Personally I'm interested in the all-time-high of the previous 3 years to have a better understanding of the current market situation. That's also what I understood from [the forum](https://forum.portfolio-performance.info/t/abstand-zum-hoechstkurs/6131/7) would help other users. 

But of course it could be that only a fraction of users might be interested in that (but the same may also apply to e.g. Bollinger bands, SMA, etc.).
Shall we collect more use cases in the forum or how do you generally estimate the demand/need for a feature?

> although those numbers are kind of depressing, because they can only be negative... 😉

Hehe good point, I haven't thought of that yet.
For this case we could make the ATH numbers in the grid more neutral (e.g. without any colors), what do you think?

> And even extend the PR to all-time-low.

Good idea 👍

buchen:
I merged the pull-request as is.
But I removed the coloring (it will be red all the time anyway).
And then I display the security in the information pane if one clicks on the value.

PR #2921, opened 2022-07-19
Latest security price widget

ma4nn:
This simple PR introduces a new dashboard widget for displaying the latest price of a security.

This can be useful for example to show the current [CAPE](https://en.wikipedia.org/wiki/Cyclically_adjusted_price-to-earnings_ratio) or [VIX](https://en.wikipedia.org/wiki/VIX) values on the dashboard to get a quick overview of different market situations.


buchen:
squashed, rebased, and merged 😄 

Thanks for your patience 👍

PR #2924, opened 2022-07-20
Add new Easybank AG PDF-Importer

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-easybank/21323/7
https://forum.portfolio-performance.info/t/pdf-import-von-easybank/21323/11
https://forum.portfolio-performance.info/t/pdf-import-von-easybank/21323/13
https://forum.portfolio-performance.info/t/pdf-import-von-easybank/21323/17


buchen:
👍

PR #2943, opened 2022-08-08
add bookmark for tracking differences

vv01f:
maybe you like to add trackingdifferences.com 

lookup ETF tracking difference data


buchen:
I like it. Thanks. Did not know about that site.

Please keep in mind that the bookmark is not available to users by default. It will be added to new files. And users can add it to existing files.

PR #2954, opened 2022-08-21
Extend time grid of StackedTimelineChart 

OnkelDok:
The time grid of the taxonomy area chart is different to other ones (e.g. in statement of assets historical chart). In the current version there is only a vertical line for each year. Additionally the color of the vertical line and the text has a color which is difficult to see. Reported here: https://forum.portfolio-performance.info/t/beschriftung-der-zeitachse-im-flachendiagramm-der-klassifizierung-fehlt/21607

So I changed the color and used the same time grid calculation in the ``StackedTimelineChart`` as in the ``TimelineChart``.
Here some screenshots from current version and same view with this PR:

**Period 3 years**
Current Version:
grafik
With this PR:
grafik

**Period 100 days**
Current Version:
grafik
With this PR:
grafik


buchen:
Hi @OnkelDok,

thanks for the contribution. I have added one more commit which is refactoring the painting of the time grid into a separate helper class. That reduces somewhat the code and ensures the behavior is identical (for example to not label lines if there is not enough space).

Issue #2893, opened 2022-06-20
Slovak translation


seg-on:
Hi, 
I would like to make translation to Slovak. Can you give me a advice what is best way to do it? Is there some tool for it?
Until now i was contributing in translation via 3rd party web solution like Transifex...

Nirus2000:
Join the translation team
1. [POEditor.com](https://poeditor.com/join/project?hash=4lYKLpEWOY)
2. Join translations
3. Login with your account
4. Start your translation in your language

Have fun....

seg-on:
I have done it. I don't know if anyone is checking.

Nirus2000:
Great, @buchen will be informed about this automatically.
Thanks 👍🏻

buchen:
@seg-on I have to apologize - I had too many things on my plate and I am not always watching all issues here in Github. Thankfully @Nirus2000 pointed me to your change.

I will build a test version for you to review. Which platform build do you need?

buchen:
And adding a word of explanation: the synchronization of translations between poeditor and github is something that I have to trigger manually. I have scripts for this (in the releng folder) but it is something I usually do before a release. Adding a new language is a completely new configuration. I do not do this automatically because I do think that users do not want partially translated applications - it is either English or the local language but ideally not a mix. That's why I missed the new language...

buchen:
And one more thing: Java supports special keywords such as ```choice``` if you want to pick one label based on something. Let's take ```LabelReportingPeriodMonths``` as an example. The original version is:
```
{0,choice,0#|1#1 month|1<{0,number} months}
```
telling PP to use the label "1 month" if the number is one, otherwise the label "{0,number} months". the values for "choice" must not be translated.

I'll try to fix them locally as I can more easily search and replace.

seg-on:
Hi @buchen, thanks for your work. I assumed that adding a new language would require more work on your side.
I translated it from English and then compared it with Czech. But I think that there will definitely still be a few mistakes.
I'm use PP on Windows.

seg-on:
> telling PP to use the label "1 month" if the number is one, otherwise the label "{0,number} months". the values for "choice" must not be translated. It should read

Are you mean, that I translated the word "choice" in POEditor? Is that a problem? I can not remember. It's been two months.
If yes, I'll be careful next time.
If I made more mistakes, write them to me so that I can avoid them in the future. :)

buchen:
> I'm use PP on Windows.

[...]

PR #2963, opened 2022-09-01
Add new Sedol security idenitification for new AJ Bell Securities Limited PDF-Importer

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-from-ajbell-youinvest/21679
https://forum.portfolio-performance.info/t/pdf-import-from-aj-bell-securities-youinvest/21679/3
Add new datetime format in PDFExtractorUtils.java

Add new "_Sedol_" security idenitification for PDF Importer (same like _ISIN_ or _WKN_)
[https://en.wikipedia.org/wiki/SEDOL](https://en.wikipedia.org/wiki/SEDOL)

Hello @buchen 
i for this PDF importer a new security identifier is needed. This I have hopefully been able to implement completely.  I add also some new testCase and improve some source code.
At least I tried :grimacing:

However, I failed with the implementation in the protobuf. I don't understand this one. 
If necessary, something would still have to be triggered via the client factory that updates the existing securities.

Greetings
Alex


buchen:
(grrr... I just typed up a long response and now the input field is empty again 😬... let me type up the short version)

I never heard about Sedol. I understand it as a country-local security identification number very similar to the WKN (which is the local identification number for Germany).

And reading the Wikipedia, one is also convertible:
> SEDOLs serve as the [National Securities Identifying Number](https://en.wikipedia.org/wiki/NSIN) for all securities issued in the United Kingdom and are therefore part of the security's [ISIN](https://en.wikipedia.org/wiki/International_Securities_Identification_Number) as well.

and
> For British securities, SEDOLs are converted to ISINs by padding the front with two zeros, then adding the country code on the front and the ISIN check digit at the end.

I am hesitant to add new fields. Maybe the mistake was already to add "WKN" (which is the German Wertpapierkennnummer). If we add something for the Britisch/Irish, then what about the French, Spanish, ...?


What can we do?

* We could "rebrand" the WKN as the "country-specific security identifier" (rebrand because we cannot rename it in the XML for compatibility reasons). If the users enters the Sedol in the WKN field, it will just work.
* We could construct the ISIN out of the Sedol and then use the ISIN to identify the security
* We could extend the attribute mechanism to allow for arbitrary local security identifier. The users could configure as many as they want (and even custom ones like the idNotation which is used by popular websites)

Nirus2000:
Hello @buchen

> (grrr... I just typed up a long response and now the input field is empty again 😬... let me type up the short version)

I know, this has also happened to me several times... :disappointed: :disappointed: :disappointed:

---

I understand your point of view.
In principle I know 6 major security identification systems.
- ISIN
- WKN
- CUSIP
- VALOR
- SEDOL
- FIGI

The CUSIP is used e.g. in IBFlexQuery or also in the Score Priority Corp. PDF importer. 
The valor numbers mostly in the Swiss PDF importers.... and so on.

Let's make a separate branch for this again, then I can do some of the work again. 👍🏻 

I like the idea that we convert the WKN into a country specific security identifier. 
I have also already searched the net to see if there is a source on how to convert these. 
([https://www.isindb.com/](https://www.isindb.com/))

What would we need to consider?
- How do we do the recognition of which system it is?

Ah... there I went again with a hit.... well, PP is developing globally and we should set the course for this.
The slovak translation is already finished... #2893 ... unbelievable how fast this is going forward.

Greetings
Alex

buchen:
> In principle I know 6 major security identification systems.

❤️ 

Thanks. That is a nice overview. This is something we can target. (And then we can throw in the notationId for good measure).

Before we implement, let's try to understand some boundary conditions. Can a security have a WKN (german identification) *and* and Sedol (British) at the *same time*? I am asking because the transformation rules to the ISIN would be different (the British would add the GB, pad some 0, and add the verification nu

jamescrowley:
@buchen to answer one of your questions - yes, it's possible for a security to have both a SEDOL and a WKN. I found the following from a random Google:

ISIN: LU0346392219
SEDOL: BNTC223
WKN: A0NGXR
Bloomberg: FFGINCY LX

(https://www.fidelityinternational.com/FILPS/Documents/en/current/ret.en.xx.LU0346392219.pdf)

I'm guessing it's common when listed on multiple exchanges.

Nirus2000:
Hello @buchen
I am absolutely with you on this.

As @jamescrowley mentioned, a security can have any or all of these identification numbers.
Another constraint is that the quotes can also be searched with each of these identification numbers.
I'm also thinking of a simple DropDown field where we set a HashMap and add an "_idNotation_" to it.
A pre-check, could be created by a syntax comparison and then check it. If a unique syntax is found e.g. ISIN.

**ISIN:**
Patter: `^(([A-Z]{2})([A-Z0-9]{9}))(\d{1})$` -> [Mathematical ](http://www.pruefziffernberechnung.de/I/ISIN.shtml)

**WKN:**
Patter: `([ABCDEFGHJKLMNPQRSTUVWZ0-9]{6})` -> [Mathematical ](https://www.isin.net/wkn/)
(Patter; excludes the letters "I" and "O")

**CUSIP:**
Patter: `^([A-Z0-9\*@#]{8})(\d{1})$` -> [Mathematical ](https://www.cusip.com/identifiers.html?section=CUSIP#/CUSIP)

**VALOR:**
Patter: `^([A-Z0-9]{5,9})$` -> [Mathematical ](http://www.pruefziffernberechnung.de/P/Paritaet.shtml)

**SEDOL:**
Patter: `^([A-Z0-9]{6})(\d{1})$` -> [Mathematical ](https://en.wikipedia.org/wiki/SEDOL)

**FIGI:**
Patter: `^([BBG]{3})([A-Z0-9]{7})[\d]{1}$` -> [Mathematical ](https://www.openfigi.com/assets/content/figi-check-digit-2173341b2d.pdf)

loucasal:
@Nirus2000 any reason this was closed? It would be great to have this work brought to fruition eventually.

Nirus2000:
@loucasal it is ready with https://github.com/portfolio-performance/portfolio/pull/3695 and will be released with the next version, as far as everything is ok.

Issue #2970, opened 2022-09-04
non cash-effective dividend payments


Pnda87:
Stock dividends and spin offs are usually booked as two separate transactions in PP: 
(1) Cash dividend + (2) Buy transaction using the same amount
This approach has one major caveat: in the payments report, these dividends are always included despite not being paid in cash.

similar issue: #1506

I therefore propose an option to mark dividend payments as non cash-effective
transactions marked as such can later be excluded in the payments reports.

![image](https://user-images.githubusercontent.com/34549632/188326451-dd237a6f-7bca-4bff-9ad9-a5afe603f65f.png)

**Payment report as is**

![image](https://user-images.githubusercontent.com/34549632/188326513-761b2d19-242b-45bd-b61a-d5c5bfc4db45.png)

**Payment report excluding non cash-effective transactions**

![image](https://user-images.githubusercontent.com/34549632/188326548-75d11813-d312-4f1d-9a1f-072b17e78329.png)

PR #2971, opened 2022-09-04
Adds option for non cash-effective dividend payments.

Pnda87:
Payments marked as such can be excluded in the payments
report to display only cash effective transactions.

issue: #2970


Nirus2000:
Hello @Pnda87 
would it be possible to add a few more TestCases and the missing translations?

If we look under all transactions and sort by dividends as an example, how are they identified? 🤔 

Greetings
Alex

Pnda87:
In my opinion, they dont need to be identifiable under all transactions as the option is not applicable to 99% of all transactions.
If the option is set or not can be seen in the edit dialog of the specific dividend transaction

rg. translations: EN+DE is there, the other languages shall ideally be done by natives.

Nirus2000:
Let's say I have 100 divide transactions and have disabled several of them but forgot which ones, that would be very tedious for the user to always call up 100 transactions to see if it has been disabled.

Pnda87:
I rebased the code on the current master and refined it a bit more. cf. #3610

PR #2983, opened 2022-09-17
Added option to display currency code also for the main currency.

RomanLangrehr:
I frequently change the main currency to get my portfolio performance measured in different currencies. To avoid confusion, I like to display the currency code also for the main currency.

This pull request adds an option to Preferences → Presentation to enable showing also the main currency symbol (default is disabled).


Morpheus1w3:
@RomanLangrehr 
The Germany umlaut are not UTF8 encoded. Could you kindly update your code, please? Maybe translation for all other supported languages via deeply our Google Translate are welcome. 
Ty

RomanLangrehr:
@Morpheus1w3 Thanks for your suggestion. I escaped the non-ASCI characters and added translations with deeply.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
```diff
@@ -122,6 +122,11 @@ interface Preferences // NOSONAR
          */
         String USE_INDIRECT_QUOTATION = "USE_INDIRECT_QUOTATION"; //$NON-NLS-1$
 
+        /**
+         * Preference key to display the currency code also for the main currency.
+         */
+        String DISPLAY_MAIN_CURRENCY_CODE = "DISPLALY_MAIN_CURRENCY_CODE"; //$NON-NLS-1$
```
auchri, line 128: DISPLALY_MAIN_CURRENCY_CODE Typo :)

auchri (changes requested):

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
RomanLangrehr: Thanks. I fixed it.

Nirus2000:
Hello @RomanLangrehr 
In order to create a consistent source code also for the definitions please change the name "main currency" to "base currency".

baseCurrency --> Base currency
termCurrency --> Foreign currency

It would be great if you rename the descriptions and variables therefore.

Alex 👍🏻

RomanLangrehr:
@Nirus2000 Thanks for pointing this out. I changed it.

I also adapted the formatting methods for Quotes (which I forgot before).

buchen:
I like the improvement :+1: Will merge.
Thanks not only to @RomanLangrehr but also @Morpheus1w3 and @Nirus2000 for reviewing.
Personally, it adds too much clutter to the displayed tables. I am also happy to experiment with alternatives: have the full value + currency displayed in the tooltip, additionally show the reporting currency in the sidebar, in the tab next to the file name, have the currency show when hovering over the table cell, ...

My 2 cents on the wording:
* base/term makes only sense if one has a currency **pair**. In that sense the "baseCurrency" in the Client is not consistent. It is the display currency (and therefore the "termCurrency" in conversions because this is the currency we want to display). I will not change the name "baseCurrency" in the client because it would be an incompatible change in the XML (so hard to fix mistakes).
* the variable name "skipCurrency" is unfortunate. I did not want to name it "currency" because the currency of the value is part of the money object. And "skipDisplayIfValueMatchesThisCurrencyCode" is too long. For now, I propose to leave it with "skipCurrency". The new variable name "baseCurrency" implies somehow some conversion which will never happen.

PR #2998, opened 2022-10-05
Selfwealth

flywire:
New variant of the contract note uses a different case which fails to import


Nirus2000:
Hello @flywire 
Please close this pull request and create a correct PDF debug in the [forum](https://forum.portfolio-performance.info/t/pdf-import-from-selfwealth/17399/20), then we will adjust it.
We always(!) need a correct PDF debug to process everything correctly later when the importer changes.

Thanks Alex

flywire:
See #2999

PR #3006, opened 2022-10-08
Additional percentage y axis in security chart

OnkelDok:
This PR introduces an additional y axis with percentage values on the left side of the diagram.
It shows percentage values of the prices compared to the first quote in chart or the purchase price (depending on the configured mode):

 ### Mode market trend
grafik

### Mode development to purchase price
grafik

In logarithmic mode this new y axis is disabled/invisible.

Also requested here: https://forum.portfolio-performance.info/t/prozentuale-achsen-beschriftung-im-kursdiagramm/6333


OnkelDok:
Converted to draft because of exception for securities without purchase price in development to purchase price mode.
I will convert back after fixing it...currently I have not time for it.

OnkelDok:
After feedback in the forum I added chart settings for show/hide the horizontal marker lines for both axes.
The user can now decide from which of the axes the horizontal lines should be displayed. In the current version, the lines of both axes can be displayed simultaneously.
Another possibility would be to make only the lines from one of the two axes displayable at a time (behaviour like radio buttons where only one item of a the two options can be active).
I can change the behavior to that if desired.

grafik

pfalcon:
Hey @OnkelDok, great work on improving the chart widget in PP (this and other patches, e.g. the measurement tool)!

Github currently says "This branch has conflicts that must be resolved", so would you mind to rebase this PR on the current master?

Also, I wonder if you saw @chirlu's comments on suggested patch workflow: https://github.com/buchen/portfolio/pull/3248#issuecomment-1492703119 (and following comments there) and agree with them? E.g.:

> Corrections should not be separate commits, but be squashed into the original commit they are fixing.

In the ideal world, following that would workflow allow patches to be reviewed/merged faster/easier. It would also allow others to test them in the meantime easier. (I'd be interested to try it when I have time for example.)

Thanks!

OnkelDok:
Thank you, @pfalcon, for pointing out about open conflicts. I haven't checked this PR for a long time. Are there automatic notifications when a PR moves into conflicting state after commits in the target branch.

I agree with @chirlu s points, but have to admit that I have not followed them in a lot of my past PRs. I think in some of my first PRs I asked @buchen if I should squash, and I think I remember he says that he could do this step easily before mergin the PR into target branch. But I cannot find the comment at the moment.

buchen:
Thanks @OnkelDok for the contribution.

A couple of thoughts:
* First I think it should not be the default to show two axis. Personally, I am fine with one axis and want to avoid the visual clutter. If we make it an options, users can decide on their own. I have added that.
* Second, I think it is very confusing to not have any horizontal lines at all (and I also find it confusing to have both - I already made a screenshot to share a "bug" here until I realized that is the design)

BTW, @OnkelDok are you using the [Resource Bundle Editor](https://marketplace.eclipse.org/content/resourcebundle-editor) plugin? It sorts the keys in the correct position, it escapes all special characters (and it just makes it easier to edit all translations of one key at once). You might want to check it out.

buchen:
merged :-)

pfalcon:
@OnkelDok:

> Are there automatic notifications when a PR moves into conflicting state after commits in the target branch?

Not that I know of. But Github gets updated all the time, so, who knows. Otherwise I guess can check from time to time, or rely on passers-by to ping about it ;-).

> I understand @chirlu s points, but have to admit that I have not followed them in a lot of my past PRs. This project is my first experience with github.

Ah, great work then again!

> Are there global github best practices of creating, expand and maintain PRs or is this more project specific?

I'd say it's more project-specific, though a lot of well-maintained projects seem to converge on similar set of guidelines, at least in spirit. I for one found @chirlu's recommendations very familiar and well-resonating with my own experince, that's why I brought them up ;-).

Btw, unrelated and offtopic, but did you see a situation when for a low-amplitude security, an Y scale is being selected poorly, so that the chart is shown as almost a flatline? I saw such cases, and interested to debug it, that's why I'm looking into charts-related patches, to see how other folks hacked on them.

PR #3012, opened 2022-10-21
Link from performance/statement of assets chart widgets back to original chart

ma4nn:
This small PR adds a link from the performance and statement of assets chart dashboard widgets back to the original charts.

Issue: https://forum.portfolio-performance.info/t/link-aus-diagramm-im-dashboard-auf-die-grossversion-unter-diagramme/22027/2

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ConfigurationStore.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ConfigurationStore.java
@@ -212,7 +212,7 @@ public class ConfigurationStore
         listeners.forEach(l -> l.onConfigurationPicked(active.getData()));
     }
 
-    private void activate(Configuration config)
+    public void activate(Configuration config)
     {
         listeners.forEach(ConfigurationStoreOwner::beforeConfigurationPicked);
         active = config;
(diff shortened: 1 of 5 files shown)


buchen:
I like the change - will merge with some minor changes (sorry for the delay).

One thing: I have extended it now to also set the reporting reporting exactly to the same interval that was used by the chart on the dashboard. When navigating, that is more intuitive because one sees the same data - but in big.

However, updating the reporting period of the chart changes the "global" interval selection in the upper right corner and then applies to all other views. If the user continues to navigate, it will be the reporting period "injected" via the dashboard. Until now, the reporting period only changes if the user explicitly changes it. This does not feel like good UX to me so I will deactivate it. But: what do you think of a "temporary" reporting period - it is activate for this chart view after navigating there, but does not persist (e.g. when moving to another view, it will be back to the previous version).


And then: I wonder if we should make the button only show when hovering over the widget. First it will remove some "noise" from the dashboard, second it does not reduce the space available for the title.

What do you think?

buchen:
Merged with 7bcdddf7b5fd3a307128d3d03dfcf94cc5675268

ma4nn:
@buchen nice, thanks for merging!

I like both ideas - temporary reporting period and display on hover only 👍 
If I have some free time I will have a look into the implementation.

Have a successful start in 2023!

Issue #3016, opened 2022-10-27
Windows installer does not set DisplayVersion when registering the uninstaller


l-c-g:
**Describe the bug**
Windows installer does not set DisplayVersion when registering the uninstaller. This is used by winget to determine the currently installed version and to compare it to the newest available version during `winget upgrade`. See [ItzLevvie's comment on winget issue 1356](https://github.com/microsoft/winget-cli/issues/1356#issuecomment-895409232).

**To Reproduce**
1. Use the installer on Windows (or run `winget install buchen.portfolio`)
2. Check in the registry for the DisplayVersion value in 'HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Portfolio Performance' 
3. Notice that there is no DisplayVersion value
4. (Alternatively, wait for a new Portfolio Performance release and notice the warning during `winget upgrade` about unknown version numbers. Checking the registry is faster.)

**Expected behavior**
DisplayVersion should exist and contain the currently installed version of the application.

**Desktop (please complete the following information):**
 - Windows 11 22H2
 - Portfolio Performance 0.59.2

Morpheus1w3:
Well, Portfolio Performance has a very well implemented self-update which checks every start, if the software is up to date. External update clients are omitted, as they have often have difficulties with Java compiled software.

CanePlayz:
@l-c-g The installer definition is located in `portfolio-product/installer/installer.nsi`. The version should be set in line 74 with this `WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\$APPNAMEFULL" "DisplayVersion" "${SOFTWARE_VERSION}"`. Now, I don't know where this variable is coming from, but there seems to be an issue with it

@Morpheus1w3 Would there be anything you could do now with this new information?

l-c-g:
@CanePlayz I know where this registry entry is set, I made [the pull request](https://github.com/portfolio-performance/portfolio/pull/3017) that added [the line](https://github.com/portfolio-performance/portfolio/blame/6608f7bbb80e6c995c8c8861e1b598c4ea3636f8/portfolio-product/installer/installer.nsi#L74).

In fact, when your comment came in, I was surprised to see this still open. @buchen had merged the code and my original issue with winget has been resolved ever since. I did not follow up on this issue after that but I am closing it now.

CanePlayz:
[...]

PR #3017, opened 2022-10-27
Set DisplayVersion in the uninstaller entry

l-c-g:
This is used by winget to compare the installed version with the latest online version to determine whether an update is available.

Issue: #3016


Morpheus1w3:
Does not make any sense, because the version written in the registry is those one at the installation date. While PP update itself, the registry is not updated, tough.

l-c-g:
> Does not make any sense, because the version written in the registry is those one at the installation date. While PP update itself, the registry is not updated, tough.

You may consider it pointless, for me it fixes an issue. I do not update Portfolio Performance from within itself but keep my applications up to date with winget. Setting DisplayVersion lets winget update the application via its own installer executable and since the installer keeps DisplayVersion updated as well, it continues to work.

But yes, for people who use the installer once and then update from within the application, it introduces the cosmetic issue that the version number in the uninstall entry refers to an older version.

Mattizin:
Könnte der Registry DisplayVersion Eintrag beim Selbstupdate von PP nicht auch aktualisiert werden?
Ich stimme zu, dass der Selbstupdater ausreichend funktioniert, ein Support für den de facto Standard winget wird in Zukunft aber sicher für immer mehr (User & Anwendungen) sinnvoll sein.

PR #3032, opened 2022-11-06
Trade calendar improvements

chirlu:
Various improvements to the trade calendars:
* Use one string for all state funerals, instead of having separate `FUNERAL_OF_PRESIDENT_XXX` strings.
* Add the funeral of Bush Sr. in 2018 as an NYSE holiday.
* Add the coronation of Charles III in 2023 as an LSE holiday.
* Historical SIX (SWX) calendar fixes: In the early 200x years, some holidays were observed or not observed that weren’t correctly reflected in PP.
* Change the names used for 25th/26th December in the SIX calendar, since they are normally not called first and second day of Christmas in Switzerland, but Christmas and St Stephen’s day.
* Update the Vienna trade calendar 2023: They will no longer observe Whit Monday as a holiday. (Strange choice, by the way, since it will fall on the last Monday in May that year, so the major exchanges all are closed for Memorial Day and Spring May Bank Holiday.)


Nirus2000:
Hello @chirlu 
Another great pull request from you.
Say, did you also notice that the translations of the holidays in Russian, Italian and Slovak are missing?
I also had a look in poeditor and all languages are missing.

Greeting
Alex

chirlu:
Yes, it seems those translations are still incomplete.

buchen:
merged. Thanks!

PR #3033, opened 2022-11-06
Add year selection to the calendar preview

chirlu:
On the calendar preference page, a preview of the holidays in the chosen calendar was already shown. It was, however, restricted to the current year. Since it may be relevant to see the holidays of the next year in particular, but sometimes even of other years, add a control through which the user may specify the year for the calendar preview.

New calendar preference page (in German), showing a preview of the LSE calendar for 2023


chirlu:
I had originally mentioned the idea here, half a year ago: https://github.com/buchen/portfolio/pull/2860#issuecomment-1140463573

buchen:
merged.

PR #3036, opened 2022-11-10
Add option to display "p.a." for annualized return rates.

RomanLangrehr:
This PR adds a preference („under Presentation”) to add "p.a." to annualized returns. By default, this is turned off.

This is similar to what I proposed in #2322 (which was later reverted because some people had strong objections), but this time entirely optional (and deactivated by default).


buchen:
rebased and merged

PR #3038, opened 2022-11-12
Display currency symbol in dashboard (when activated)

RomanLangrehr:
The option I added in #2983 does not work for the dashboard. This fixes it.


buchen:
What if we make the IndicatorWidget allow to use Object:
```public class IndicatorWidget<N extends Object> extends AbstractIndicatorWidget<N>```
and then return Money. Then we do not need any specific logic in "Values" nor create new Value Objects all the time.

buchen:
I did a slightly different implementation - but the result should the same

PR #3047, opened 2022-11-13
Ratio widget

RomanLangrehr:
Added a new widget that shows the ratio of two assets.

image

This required some changes to the `DataSeriesConfig` and the `IndicatorWidget`: They now can be configured to require an arbitrary number of data series, instead of just one normal data series and optionally one benchmark.


RomanLangrehr:
Please ignore this PR. I messed something up with „supportsBenchmark” feature. I will come back with a better solution.

PR #3048, opened 2022-11-13
Ratio widget

RomanLangrehr:
Added a new widget that shows the ratio of two assets, You can use it, for example, to show how much of your total assets you have invested in stocks, etc.

image

This required some changes to the `IndicatorWidget`: This can be configured now to require an arbitrary number of data series, instead of just one.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/AbstractIndicatorWidget.java
```diff
@@ -75,6 +107,28 @@ public Control getTitleControl()
     @Override
     public void update(D data)
     {
+        // construct label to indicate the data series (user can manually change
+        // the label later)
+        getWidget().setLabel(WidgetFactory.valueOf(getWidget().getType()).getLabel() + ", " //$NON-NLS-1$
```
buchen, line 112: This code will prevent that the user can edit the label b/c you override the edited version with the default title.

buchen (changes requested):
Hi @RomanLangrehr,

I like the Ratio widget, I got a couple thoughts on the change to the Indicator Widget.

When hovering over the value of the indicator widget, then a NPE is thrown (for example on "invested capital since first transaction"):
```
java.lang.NullPointerException
	at name.abuchen.portfolio.ui.views.dashboard.IndicatorWidget$Builder.lambda$1(IndicatorWidget.java:83)
	at name.abuchen.portfolio.ui.views.dashboard.IndicatorWidget.lambda$0(IndicatorWidget.java:212)
	at name.abuchen.portfolio.ui.util.InfoToolTip.createToolTipContentArea(InfoToolTip.java:54)
	at org.eclipse.jface.window.ToolTip.toolTipShow(ToolTip.java:345)
	at org.eclipse.jface.window.ToolTip.toolTipOpen(ToolTip.java:469)
	at org.eclipse.jface.window.ToolTip.toolTipCreate(ToolTip.java:334)
	at org.eclipse.jface.window.ToolTip$ToolTipOwnerControlListener.handleEvent(ToolTip.java:609)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)

```

Generally, I think this change introduces unnecessary complexity - and the only benefit is that one does not have to define a separate widget class for the RatioWidget (which might as well extend the IndicatorWidget). This is what I would propose to do.

Andreas.

buchen:
I have added a simplified implementation

ma4nn:
Really helpful widget, thanks 👍

PR #3058, opened 2022-11-26
Modify Baader Bank PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-baader-bank-scalable-capital/2057/114


buchen:
BTW, I am not sure if the seconds survive saving the the file. I think I only save hours and minutes.

PR #3069, opened 2022-12-08
Throw HTTP exception if URL is redirected

Morpheus1w3:
If HTTP Response Code is 200 and 301/302, trigger logging. Root cause @ https://forum.portfolio-performance.info/t/kursaktualisierung-uber-ariva-funktioniert-nicht-mehr/18970/11

Screenshot_20221208_193929_Chrome Beta

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/util/WebAccess.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/util/WebAccess.java
@@ -216,7 +216,7 @@ public class WebAccess
             response = client.execute(request);
 
 
-            if (response.getCode()(response.getCode() == HttpStatus.SC_MOVED_PERMANENTLY || //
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/util/WebAccess.java
```diff
@@ -215,7 +215,11 @@ private CloseableHttpResponse executeWith(Request function) throws IOException
             HttpUriRequestBase request = function.create(uri);
             response = client.execute(request);
 
-            if (response.getCode() != HttpStatus.SC_OK)
+
+            if (response.getCode()(response.getCode() == HttpStatus.SC_MOVED_PERMANENTLY || //
+               response.getCode() == HttpStatus.SC_MOVED_TEMPORARILY)
```
chirlu, line 220: I guess the double `(response.getCode()` needs to be removed onve.

buchen:
Maybe we should follow the redirects?

PR #3089, opened 2022-12-18
Ibflex dividend reversal

csprunk:
Hi,

For IBFlex import negative dividends for reversed transactions cause issues (import maps to dividend transaction with absolute value, negative dividends are not supported). Here's a draft to try to match such reversed dividend transactions during import and suppress them.

Alternative might be to do this in the actions on the ReviewExtractedItemsPage, but this would require introducing a new action type that has access to all transactions being imported at once, if I understood correctly, and I am not sure how much of an issue these dividend reversals are elsewhere.

Thanks a lot for taking a look!


csprunk:
(closing this one since IBFlex importer got reorganized in the meantime, will open a new one for easier merge)

PR #3091, opened 2022-12-19
Convert files with CR+LF line endings

chirlu:
This has a huge diff, but actually not much is happening. There were a number (53, to be exact) of text files in the repository that had Windows-style CR+LF (`"\r\n"`) line endings. Convert these to the Git (and Unix, and MacOS X) standard of LF-only (`"\n"`). Then, add a `.gitattributes` file to make sure that CR+LF line endings won’t sneak back in in the future.


chirlu:
Some background: https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

.gitattributes
```diff
@@ -0,0 +1,12 @@
+# default to automatic recognition whether a file contains text:
+* text=auto
```
chirlu, line 2: This would actually be enough – the heuristics used by Git correctly classify all files in PP.

PR #3112, opened 2023-01-07
reduce filter size of holdings pie percentage labels

schurteb:
Currently the percentage labels on the holdings pie chart are hidden if the position is smaller than 3%. For my sense that is a quite big range being filtered out.

I gave it a play and think we could reduce that filter's size to 1.5% without having issues with overlapping labels.

Here's some screenshots for you to get a feeling of what i mean 😃 

**Current situation:**
2.98% Position @ FullHD

**Proposal:**
Lower bounds - Small Screen

Higher bounds - Small Screen

Lower bounds - Larger Screen

Higher bounds - Larger Screen

Lemme' know what you think 😄

### DIFF
--- a/name.abuchen.portfolio.ui/META-INF/html/pie.html
+++ b/name.abuchen.portfolio.ui/META-INF/html/pie.html
@@ -186,7 +186,7 @@ var thePieChart = {
 
         this.valueLabels.enter()
             .append("text")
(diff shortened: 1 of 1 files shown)


buchen:
Agree.

(In the SWT Version of the charts, I am calculating the available space and typically get to roughly 1.5% as well).

Issue #3117, opened 2023-01-11
Incorrect equations in the descriptions of "Capital Gains" columns


TheJiahao:
**Describe the bug**
The descriptions/tooltips of capital gains columns in "Securities" view are incorrect (at least in English version, probably in other languages as well). In particular, the equations "Capital Gains = Purchase Value (FIFO/moving average) - Market Value" and "Capital Gains % = Market Value / Purchase Value (FIFO/moving average)" are incorrect. The actual calculations are still correct.

**To Reproduce**
1. Go to "Securities" view
2. Add columns "Capital Gains (FIFO, current holdings)", "Capital Gains % (FIFO, current holdings)", "Capital Gains (moving average, current holdings)" and "Capital Gains % (moving average, current holdings)" from "Performance" section.
3. Move mouse to the added columns and hover
4. See the error

**Expected behavior**
- Equation "Capital Gains = Purchase Value (FIFO/moving average) - Market Value" should be "Capital Gains = Market Value - Purchase Value (FIFO/moving average)". The current one gives negative capital gain in case of profit and negative in loss.

- Equation "Capital Gains % = Market Value / Purchase Value (FIFO/moving average)" should be "Capital Gains % = (Market Value - Purchase Value (FIFO/moving average)) / Purchase Value (FIFO/moving average)" or "Capital Gains / Purchase Value (FIFO/moving average)". The current one always give a positive value, even in case of loss (assuming that the market value and purchase price are both positive).

**Screenshots**
![capital_gains_ma](https://user-images.githubusercontent.com/42709836/211879569-cc2c03d7-ef63-433b-91c9-a9b923fa4c2f.jpg)
![capital_gains_ma_percent](https://user-images.githubusercontent.com/42709836/211879573-312ddcd5-1cc4-41c8-b5c3-84d2b8f70061.jpg)
![capital_gains_fifo](https://user-images.githubusercontent.com/42709836/211879575-b9b2d273-3553-4fc7-ac1c-c2c547869528.jpg)
![capital_gains_fifo_percent](https://user-images.githubusercontent.com/42709836/211879578-b2041c89-6a8c-4fad-9ab0-6e5e5516d1d5.jpg)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: 0.60.1

PR #3118, opened 2023-01-11
Fix English descriptions of Capital Gains columns

TheJiahao:
Partial fix (only for English version) to issue: #3117


buchen:
Thanks for fixing! Will also apply this to the other languages

Nirus2000:
Fixed #3118

PR #3119, opened 2023-01-12
Modify of PDF-Importers for non-importable PDFs

Nirus2000:
- [x] Fixes BaaderBank return not null in dividende transaction
- [x] Fix non-importable custom error translations
- [x] Fixes Onvista addAdvanceTaxTransaction() with non amount
- [x] Fixes an error, if no errors are generated during import, the errorTableViewer unmapAllElements() and the generated ImportActions are deleted.
- [x] Extends the PDF Import Wizard to identify non-importable transactions and additionally add the non-importable items as extraction errors.
- Maybe optimize the enlarge ErrorTableviewer if errors are identified. Otherwise output "No errors found".
- Maybe a test for MarkNonImportableTest
- Maybe opitmize visualization enhancement for users --> see below

---

**Show the error in ReviewExtractedItemsPage.java and the changes of non-importable transactions:**
([testfile_visu.zip](https://github.com/buchen/portfolio/files/10402316/testfile_visu.zip))

https://user-images.githubusercontent.com/45203494/212237642-4a5f8cb9-fe7a-40a9-8e61-b1ca1b05e094.mp4

---

**Open visualization enhancement for users:**
To improve the usability of the importer for PDFs that cannot be imported or that PP does not know yet, we should visualize this part better.

Error output for PDFs which are identified but cannot be imported. 
([testfile_patter_fault.pdf](https://github.com/buchen/portfolio/files/10408360/testfile_patter_fault.pdf), [testfile_unknow.pdf](https://github.com/buchen/portfolio/files/10408368/testfile_unknow.pdf))
- Scrollbar horizontal
- Output of the recognized PDF importer label

grafik
grafik


buchen:
I close the PR because the feature is merged with a different PR (with an updated/different implementation)

PR #3125, opened 2023-01-14
Fix error message (shown to the user!) when a date can not be parsed..

RomanLangrehr:
…while importing a PDF.

Before the error message 'invalid date ''{0}''' was shown to the user.


Nirus2000:
Hello @RomanLangrehr 
Great that you found the error! 😀 
Please add the change [here](https://github.com/buchen/portfolio/blob/3656371f01f4d3aa54689de3d7909fa13dce5156/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFExtractorUtils.java#L287) too

Regards
Alex

RomanLangrehr:
Done.
Thanks for pointing out that there where more occurences.

PR #3130, opened 2023-01-16
Add icons and join translation team in main menu

Nirus2000:
Add icons and "join translation team" in main menu

Add new contributing rules

Change some icons for better understanding

sorting the enum in the images.java data

(All icons by iconmonstr.com)

---

Please add Creative Commons CC0 Licence in PP Programm


buchen:
I merged the link to join the translation team (the first commit) and the update to the contributing readme about colors (the second commit).

About adding icons to every single menu entry: I am not a fan of adding icons. First, they are adding visual noise. As the eyes are reading the text, it always has to adjust to another icon. At least on macOS, the use of icons in menus is very limited and I want to stick to this style. Second, most icons are not self-explanatory anyway so it is not helping the user to understand the functionality. Plus: who anyway remembers that one was storing files on a diskette and what the icons means.

The other qualm I have about this change is that it is suddenly introducing a light blue color that in my opinion does not fit into the color schema used so far. We should use colors very careful.

Let's go through the menus:
* On the "file" menu we can highlight Import and Export with icons (done with the third commit)
   * Maybe we should consider moving up the "debug - extract text from pdf" up to the PDF import? But does it need an icon?
* The "view" menu has the icons from the sidebar navigation
* The "transaction" menu has no icons
* The "online" menu has the icons of external websites
* The "help" menu: if we want to highlight something, then maybe the "debug: reset UI" entry

I really appreciate the time you spend on this change. I suggest next time you want to propose big significant changes to the UI, let's discuss before executing the changes.

Nirus2000:
Hello @buchen 

Thank you for your detailed reply. Great!
Creating the icons was not difficult and did not take much time. I have fun with it.

---

### Then let's talk about the graphical changes in PP.

> I merged the link to join the translation team (the first commit) and the update to the contributing readme about colors (the second commit). 

>The other qualm I have about this change is that it is suddenly introducing a light blue color that in my opinion does not fit into the color schema used so far. We should use colors very careful.

Super. I think it's important so that other users who can't program Java have the opportunity to participate in the translation of PP. 
Through PoEditor we have this and can be accessed through PP. 
The other thing is that contributors now know which color palette is preferred.

> About adding icons to every single menu entry: I am not a fan of adding icons. First, they are adding visual noise. As the eyes are reading the text, it always has to adjust to another icon. At least on macOS, the use of icons in menus is very limited and I want to stick to this style. Second, most icons are not self-explanatory anyway so it is not helping the user to understand the functionality. Plus: who anyway remembers that one was storing files on a diskette and what the icons means.

I agree with this reasoning. 
If you (we) want to continue this way consistently, then I suggest that we remove some icons even from the current status.

**So if we follow this "clear design" then these icons should be removed:**
- Icons in the notes to the transactions and securities
- Icons in the bookmark 
- Icons in the attributes
- Icons in the filter function(s)

On the one hand, the icons at these positions really confuse or show exactly what they are not supposed to. 
So, for example, note in the filter reporting period. 
Or in the bookmarks. When a separator line is inserted.

In my personal opinion, I would take that step.

---

### What could you do with the existing and the created icons from me?!

To reverse the decision at some point and/or use icons after all, I would still leave them in the system and save them.
You never know when they might be needed. (Re-Design :wink: :wink: :wink: )

---

### Let's go to the point "menu"

I would definitely not make any changes here. :sparkling_heart:

---

### What changes would I apply from the pull request?

- Please give me the goofy green check mark (Ok) in the PDF import wizard ... :grin: :speak_no_evil: :see_no_evil:
- Storage of icons for possible later use 💯 
- Sorted images.java, maybe without one or the other comment inserted 🥇 

---

Sooo... that was it from my side and fun as always.

Greetings from the Padawan
Alex

buchen:
Okay, then let's work on the requests:

> Please give me the goofy green check mark (Ok) in the PDF import wizard

Done 737f49969f92ea906ea18808f9a1a4b7be5e2caa

buchen:
> Storage of icons for possible later use

That's why I kept this pull request around. Then it is super easy to pick additional icons as needed. I do not want to check them into the main branch because it should not contain unused code or other stuff.

I can also add it as on official branch into the repository - then you can delete the PR branch, if you prefer

buchen:
> Sorted images.java, maybe without one or the other comment inserted

It is always difficult to have a sort order. Right now it tries to add logos, then UX elements, then model elements (=icons representing assets in the file), buttons, views, other stuff, 3rd party logos.

In general, I am hesitant to just reorder them just for the sake of a new ordering. Reformatting the file makes *git blame* unusable. And does it make a difference? In code completion, the entries are sorted (this is what you use in the editor). And if you want to know which icons are available, open the folder in Finder/Windows Explorer to see the icons visually.

That said: I have fixed some ordering here: 67bd6d68fc171044b32a0a956c9d867877160dc2

buchen:
> If you (we) want to continue this way consistently, then I suggest that we remove some icons even from the current status.

> So if we follow this "clear design" then these icons should be removed:

> Icons in the notes to the transactions and securities
> Icons in the bookmark
> Icons in the attributes
> Icons in the filter function(s)


Let's go through them:

> Icons in the notes to the transactions and securities

My original idea was that the "note" column can be very small - and the icon indicates that one has entered a note. And then can use the tool tip to check what it is.

Bildschirm­foto 2023-02-26 um 13 54 21

I have to admit I am not using it that way. Therefore there is no specific need to have the icon. I am also not arguing to *not* have any icons at all. I was arguing that having icons for everything clutters the UI. For now, I would keep it.

> Icons in the bookmark
> Icons in the attributes

Fair. Those are not important views. Again, I am not arguing to have no icons at all, I was arguing to not have icons for each and every menu entry.

> Icons in the filter function(s)

I guess you mean these icons:

Bildschirm­foto 2023-02-26 um 13 59 06

We could remove them. But keep in mind that some entries have icons (such as accounts, securities, etc.). That will change the indentation. So in this example, "Aktien Schwellenländer" would further to the left than "Asset Allocation" even though the it is a child.
Bildschirm­foto 2023-02-26 um 14 04 04

buchen:
BTW, I have now created a branch [feature_spare_icons](https://github.com/buchen/portfolio/tree/feature_spare_icons). Then we can delete the pull request, but do not loose the icons at the same time.

PR #3142, opened 2023-01-21
Improve handle icons for better visibility

Nirus2000:
Improve handle_we and handle_ns icons for better visibility


buchen:
looks much better. Also in dark mode it is quite visible. thanks

PR #3144, opened 2023-01-22
Improve update bookmark list

Nirus2000:
Updating the bookmarks so that they work multilingually 
Fix sorting ignores upper and lower case


buchen:
merged

BTW, for sorting strings, I use now TextUtil#compare as it is using a text collator internally. That will take care of special characters such as accents, etc.

PR #3146, opened 2023-01-24
Fix issue with ampersands in description widget

ma4nn:
This PR removes the use of `TextUtil.tooltip()` from the description widget to avoid issues with the SAX parser.

Issue: https://forum.portfolio-performance.info/t/eingabe-eines-zeichen-im-widget-allgemein-beschreibung-fuhrt-zu-einem-internal-error/22944/3

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DescriptionWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DescriptionWidget.java
@@ -11,7 +11,6 @@ import name.abuchen.portfolio.model.Dashboard.Widget;
 import name.abuchen.portfolio.ui.UIConstants;
 import name.abuchen.portfolio.ui.util.Colors;
 import name.abuchen.portfolio.ui.util.swt.StyledLabel;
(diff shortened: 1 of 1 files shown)


chirlu:
Completely removing the escaping is not a good idea. Any ampersands should still be escaped in the tooltip.

ma4nn:
@chirlu thanks for your input! What do you recommend then? With the current logic for the tooltip this can only be solved by adding an additional label element to the description widget which does not make a lot of sense.

Also as the tooltip does not interpret e.g. HTML code in this case, IMHO the tooltip is not really helpful anyway for this kind of widget.

What do you think about extending the logic in `DashboardView::buildToolTip()` to additionally check for a new method `getTooltip()` on the `delegate` so that tooltips can be disabled or set individually for single widgets?

buchen:
merged.

I also added a validator to the multi-line input to make sure it is valid XML. One has to use &amp; Not really sure why this failed for the poster in the forum. In this branch, I am already on Java 17

ma4nn:
@buchen thanks for merging! It failed because `TextUtil.tooltip()` simply replaces `&` with `&&`, i.e. `&amp;` became `&&amp;` and that lead to the `IllegalArgumentException` before the fix.

PR #3152, opened 2023-01-27
Add a filter option to the AllTransactionView.

RomanLangrehr:
The user can select a type filter and a client filter to filter the transaction displayed in the all transaction view.

I think this is usefull, because the AllTransactionsView can get quite full over time.

image


buchen:
Rebased and merged. Small updates on top. Thanks - a very useful addition

buchen:
And, yes, the filter really have to get their own id (so that modifications do not change the id) and must be moved in the file. We'll get to there

PR #3154, opened 2023-01-29
Bugfix: Forex-unit not getting fixed in gross value fix procedure.

### LINKED ISSUE #3095: IBKR: Flex Query import total price of trade
I find out that importer (IBKR: Flex Query import) import to total value of stock's price with fee and then to the fee the fee price.
Total price of trade include 2 times fee.

Here is detail from PP
image

In XML is traded price of one share $14,05 total price $140,5 and fee $0,33125725

And one another thing. The conversion rate is different. But it is a separate FX trade automatically make by IBKR. Maybe it is too complicated to use conversion rate of the FX trade. Only note.

[AFQ__PP.xml.txt](https://github.com/buchen/portfolio/files/10288000/AFQ__PP.xml.txt)

OS: Windows
Version 0.60.0

RomanLangrehr:
Fixes #3095

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorWithFixGrossValueBuySellTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorWithFixGrossValueBuySellTest.java
@@ -108,7 +108,7 @@ public class IBFlexStatementExtractorWithFixGrossValueBuySellTest
 
         Unit grossValueUnit = entry.getPortfolioTransaction().getUnit(Unit.Type.GROSS_VALUE)
                         .orElseThrow(IllegalArgumentException::new);
-        assertThat(grossValueUnit.getForex(), is(Money.of(CurrencyUnit.USD, Values.Amount.factorize(844.95))));
+        assertThat(grossValueUnit.getForex(), is(Money.of(CurrencyUnit.USD, Values.Amount.factorize(844.62))));
     }
 
     private Extractor.InputFile createTempFile(InputStream input) throws IOException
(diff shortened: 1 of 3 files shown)


RomanLangrehr:
Unfortuntely. this solution is not very precise. Rounding errors lead to the gross value being off by +-0.01 quite often (that is also why the test fails). Any better ideas?

Nirus2000:
No, unfortunately not, but this solution is more accurate than the previous one. 
Before this was much larger +/-0.01.

RomanLangrehr:
I relaxed the test condition so that the test passes.

RomanLangrehr:
Hm, I just realize that this breaks a lot more test cases than the one for the IBFlex Importer, so maybe it is not a good idea to merge this. On the other hand, the `fixGrossValue()` methode not updating the forex value seems like a bug to me. What do you think, @Nirus2000 ?

Nirus2000:
Take a look at the PDF importers. Roughly this looks like this. First of all, this generally unifies the conversion problems with exchange rates, both with amounts and with taxes and fees. If a rounding error occurs, it is not larger than +/-0,01.
If it is then necessary, we set the fixed.

We work here with the "termCurrency" and the "baseCurrency", since also here is differentiated, how the broker indicates the exchange rate. So price or quantity quotation.
--> [Values we use in Contributing.md](https://github.com/buchen/portfolio/blob/master/CONTRIBUTING.md#naming-conventions-for-detected-values) <--

```
PDFExchangeRate rate = asExchangeRate(v);
type.getCurrentContext().putType(asExchangeRate(v));

Money gross = Money.of(asCurrencyCode(v.get("currency")), asAmount(v.get("gross")));
Money fxGross = rate.convert(asCurrencyCode(v.get("termCurrency")), gross);

checkAndSetGrossUnit(gross, fxGross, t, type);
```
or
```
PDFExchangeRate rate = asExchangeRate(v);
type.getCurrentContext().putType(rate);

Money fxGross = Money.of(asCurrencyCode(v.get("fxCurrency")), asAmount(stripBlanks(v.get("fxGross"))));
Money gross = rate.convert(asCurrencyCode(v.get("currency")), fxGross);

checkAndSetGrossUnit(gross, fxGross, t, type);
```

PR #3155, opened 2023-01-30
Add correlation widget

RomanLangrehr:
Adds a widget that show the correlation coefficient between two performance indices.

image


Morpheus1w3:
Side topic of https://github.com/buchen/portfolio/tree/feature_correlation_matrix ...

RomanLangrehr:
@Morpheus1w3 Thanks for letting me know, I wasn't aware of this.

I think the widget could still be useful. But if both get into the main branch, we should definitely de-duplicate some code.

buchen:
It will get there. Starting next month I should have more time

PR #3157, opened 2023-01-31
Refactoring PDFExchangeRate.class to ExchangeRate.class

Nirus2000:
With this pull request we rename the `PDFExchangeRate.class` to `ExchangeRate.class` and set it high. (Refactoring) 
The class becomes public.

With this we create the possibility of a better overview when used in the IBFlex-Importer and CSV-Importer.

Furthermore we remove unnecessary source code in the PDF importers.

No tests have been changed.

---

Hello @buchen 
I would like to move the `PDFExtractorUtils.java` to the `ExchangeRate.class` as well, 
since it already gets application in the CSV-Importer and later also in the IBFlex-Importer.
And of course renaming `PDFExtractorUtils.java` to `ExtractorUtils.java`.
Would you be on board with this change?

It would be helpful in #3154.

Regards
Alex


buchen:
merged

PR #3160, opened 2023-02-04
Update Comdirect monthly VISA fee support

shochmuth:
Updated ComdirectPDFExtractor fee section to support monthly VISA credit card fee statements that got previously ignored.

Added anonymized testcase using extracted statement.

The transaction note is truncated due to PDF extract (test file line 40/41) transaction description being row and cell distributed.


buchen:
👍

PR #3129, opened 2023-01-15
Add tradecalender test

Nirus2000:
Only completion of the open test dates of the trading days. No modification or deletion of tests.

The @chirlu for the latest changes and checks.
So that we prevent this I have added these tests.

PR #3167, opened 2023-02-05
Put back removed trade calendar tests

chirlu:
Fix the mess from #3129. Also, add some tests for Juneteenth because the headline was already there, even though the tests were’t.

(I absolutely *hate* it when something, like that PR, claims not to remove anything while doing exactly that.)

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TradeCalendarTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TradeCalendarTest.java
@@ -180,17 +180,16 @@ public class TradeCalendarTest
         assertThat(calendar.isHoliday(LocalDate.parse("2023-02-20")), is(true));
 
         // Good Friday
+        assertThat(calendar.isHoliday(LocalDate.parse("2020-04-10")), is(true));
         assertThat(calendar.isHoliday(LocalDate.parse("2021-04-02")), is(true));
         assertThat(calendar.isHoliday(LocalDate.parse("2022-04-15")), is(true));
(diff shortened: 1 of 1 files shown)


Nirus2000:
I'm a little saddened by your statement and really makes me think, as I was hoping the PR was good and gaining clarity.

It is of course questionable whether "Eastern" or "Easter Monday" is more correct than correct.
Also "friday" to "Friday"... the world doesn't end from that.
Or whether a "Washington's Birthday" becomes "Washington's Birthday".... What?!
"Chrismas eve russia" to "Christmas eve (Russia)" so comment?!
Also, the missing "t" in the comment is highly objectionable! Sorry, I wrote it the way I spoke it.

That the link in the testTradeCalenderMICEXRTS() is wrong is true.
I really forgot to correct it. It is still missing now... How?!
Here is the link: https://www.tradinghours.com/markets/moex


Why does it look different now? I have sorted the tests by date... From the beginning of the year to the end. Also I tried to always map the last three years. Special holidays or special days are then at the bottom.
For some TradeCalendars there were no tests before, I added them.

Let's see further... after your last cleanup, which I thought was really good, you didn't correct obvious things, but hey... keep calm.

To conclude with a statement to this sentence of yours....

> (I absolutely hate it when something, like that PR, claims not to remove anything while doing exactly that).

Show me one line that affects the "outcome" of the "calendar issue" and justifies indirectly "hating" someone for their performance and work?!
https://de.wikipedia.org/wiki/Hass

I find it very disturbing to deeply discredit someone with these sentences in a PR while presenting yourself in a light.
Have you ever asked yourself if you want to be treated this way with this abysmal contempt?

@buchen
I'm out...

PR #3169, opened 2023-02-06
Ibflex importer dividend reversal

csprunk:
Hi,

For IBFlex import negative dividends for reversed transactions cause issues (import maps to dividend transaction with absolute value, negative dividends are not supported). Here's a suggestion to try to match such reversed dividend transactions during import and suppress them.

Alternative might be to do this in the actions on the ReviewExtractedItemsPage, but this would require introducing a new action type that has access to all transactions being imported at once, if I understood correctly, and I am not sure how much of an issue these dividend reversals are elsewhere.

Thanks a lot for taking a look!

(Opened a new one for this for easier merge since ibflex structure reorganized in the meantime and I recalled a comment to please not rebase pull requests)


Nirus2000:
Hello @csprunk 
Please check if your PR is compatible with this #3177. 
Here I have created a complete revision.

Regards
Alex

csprunk:
Nice, thanks a lot for this rewrite!

I will take a closer look, most likely these are not directly compatible, I will adapt my change.

I did not include the taxes in the reversal since these were actually correctly imported at head, they cancel each other out.

pfalcon:
The sample XML file included in the PR doesn't show that, but generally, Flex XML should include "actionID" attribute which links related operations (e.g. original xact and its reversal). Did you consider somehow using them?

csprunk:
> 

Thanks a lot for that hint! I went back to IBKR and checked the FlexQuery settings and it seems they added the actionID attribute after I set up the queries (or for some other reason it was not selected for the query). Exported again and this nicely links the transactions, this will be a much cleaner way to go about this.

buchen:
@Nirus2000 @csprunk - what about this pull request. Is it now obsolete (because the changes are part of the bigger refactoring)? Can I close it?

Nirus2000:
Nope... it is still in progress at @csprunk 
I raised this issue here because the PR is significantly affected by mine. Currently, only the dividend was recorded and never the taxes on the dividend.

csprunk:
Sorry for the delay here, close this.

PR #3170, opened 2023-02-06
Promote OpenJDK instead of Oracle JDK

Luensche:
As I got in PP the message, that I need a newer JRE I was a little bit surprised to find a link to Oracles JDK. I suggest to change to OpenJDK.


chirlu:
Since it probably only or mostly affects Linux users by now, additional advice to first try the distribution’s package system might also be useful.

Related: https://forum.portfolio-performance.info/t/version-0-61-0-startet-nicht-braucht-java-17/23104

buchen:
thanks. The message needs some rework. On Windows und macOS Java is embedded (the Azul JRE), but on Linux it is not

buchen:
Thanks for taking the effort @Luensche. I have now rewritten the whole message and updated all translations

PR #3186, opened 2023-02-18
Support to open bookmark for multiple securities

OnkelDok:
Sometimes it would be helpful to be able to open bookmarks for several securities at the same time.

I think I have also read a similar request in the forum as sidenote of a post, but I cannot find it.

grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/BookmarkMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/BookmarkMenu.java
@@ -16,15 +16,20 @@ import name.abuchen.portfolio.ui.views.settings.SettingsView;
 
 public class BookmarkMenu extends MenuManager
 {
-    private final Security security;
+    private final List<Security> securities;
     private final Client client;
     private final PortfolioPart editor;
 
     public BookmarkMenu(PortfolioPart editor, Security security)
     {
+        this(editor, Collections.singletonList(security));
+    }
+
+    public BookmarkMenu(PortfolioPart editor, List<Security> securities)
+    {
         super(Messages.MenuOpenSecurityOnSite);
 
-        this.security = security;
+        this.securities = securities;
         this.editor = editor;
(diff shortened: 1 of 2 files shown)


chirlu:
Uh, this seems dangerous to me. You might accidentally get hundreds of browser windows open over each other …

When I first read the title, I actually thought it would open *one* bookmark that accepts a list of parameters, such as the Fondsweb comparison page: https://www.fondsweb.com/de/vergleichen/ansicht/isins/IE00B14X4Q57,IE00B3VTMJ91,LU1650487413

buchen:
> You might accidentally get hundreds of browser windows open over each other …

Shall we limit it to 10 securities max?
If the user accidentally opens the bookmarks, it is okay to close 10 tabs in the browser, but not 100

OnkelDok:
> Uh, this seems dangerous to me. You might accidentally get hundreds of browser windows open over each other …

Good point. We can display a confirmation message when the user has selected more than X (maybe 10 or 15?) securities.

> When I first read the title, I actually thought it would open one bookmark that accepts a list of parameters, such as the Fondsweb comparison page: https://www.fondsweb.com/de/vergleichen/ansicht/isins/IE00B14X4Q57,IE00B3VTMJ91,LU1650487413

Thats an interesting idea. Maybe as additional feature in the future? Do you know other pages that accepts multiple parameters for comparison? For this we would need additional bookmark or bookmark-placeholder with which the user can flag that multiple securities should be joined together into one bookmark with a configurable string as separator.

chirlu:
I don’t think everything that could be implemented should be implemented …

buchen:
> I don’t think everything that could be implemented should be implemented …

Which implementation proposal are you commenting? Opening multiple bookmarks is okay from my side (with a limit). Opening one bookmark with multiple parameters I would not do for now - we would have to have a meaningful example first; and then I am not sure if anybody would understand the pattern logic anymore

I'll add the limiter and merge

chirlu:
Yeah, I meant the one-bookmark-with-list-parameter idea.

PR #3187, opened 2023-02-18
Drag'n'Drop support for 'All securities' sidebar item

OnkelDok:
In the current version, to drag and drop a security from one file to another, you must drop the security onto a watchlist (https://forum.portfolio-performance.info/t/watchlisten-und-wertpapiere-zwischen-portfolios-kopieren/19894/19)

With this PR you can directly drop a security into the ``All securities`` side bar item.


buchen:
Great. Merged. Was really missing.

I did two minor changes in a follow-up commit:
* if there are no changes, then do not touch the client (e.g., moving existing security to "all securities" does not really change something)
* instead of the new marker class, check for SecurityListView

chirlu:
@buchen: Was it intentional in commit ac55b43d12a79db49fe6d80b5c68de9b607813ff to have those `System.out.println()` calls, or was that only for quick testing? If it is intentional, I would definitely prefer using `PortfolioLog` (and it shouldn’t be `"COPY :"`, but `"COPY: "`; same for the other string; and they would also need `//$NON-NLS-1$`).

PR #3188, opened 2023-02-18
Fix assertion with the wrong parameter order

### LINKED ISSUE #2948: Test Smell: assertion with the wrong parameter order
Hi！

Description:
we detect that some assertions in your test code have the wrong parameter orders.
For example, the test case named ''testExternalIdentifier()'' in ''SecurityTest.java'' wrties the ''assertEquals'' as 
image
However, referring to the API documentation of ''org.junit.Test'', the correct API usage of ''assertEquals'' is ''assertEquals(Object expected, Object actual)''.

Negative:
Once the test case fails, the ''assertEquals()'' assertion with the wrong parameter order will give the wrong log information.
The log information will say:’’ java.lang.AssertionError: expected [actual value] but found [ except value]’’
This is confusing, to say the least, and you shouldn't have to deal with a possible misdirection of that message.

Solution:
Generally, the excepted value should be a known value, such as a real number, a string, etc.
The actual value should be the result of the method-under-test.
Therefore, the parameter order of ''assertEquals'' should be exchanged.

We list the test cases with the same problem as follows:
testExternalIdentifier() in SecurityTest.java
testInvalidRGB() in ColorConversionTest.java
testToRGB() in ColorConversionTest.java
testToHSB() in ColorConversionTest.java
testDividendPerShareNoShares() in DividendTransactionTest.java....

shochmuth:
Fix #2948

Changed mismatching parameter order for actuals and expected values to match _org.junit.Assert.assertEquals_ semantics.
Changed _assertEquals_ checks for **not null values** to use _assertNotNull_

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/SecurityTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/SecurityTest.java
@@ -3,7 +3,6 @@ package name.abuchen.portfolio.model;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
-import static org.junit.Assert.assertEquals;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.beans.BeanInfo;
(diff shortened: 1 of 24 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/SecurityTest.java
```diff
@@ -199,22 +198,22 @@ public void testExternalIdentifier()
         Security security = new Security();
 
         security.setName("Apple ORD");
-        assertEquals(security.getExternalIdentifier(), "Apple ORD");
+        assertThat(security.getExternalIdentifier(), is("Apple ORD"));
```
buchen, line 201: I wonder: isn't that still the wrong order (if it is "expected, actual")?

buchen: Okay, got me confused... 

It is "assertEquals(expected, actual)" and "assertThat(actual, matcher)"

PR #3189, opened 2023-02-18
New context menu to duplicate single security

OnkelDok:
New context menu entry to duplicate an existing security.
If ``clone`` would be a better word for this function than ``duplicate`` then I can change this.

grafik

https://forum.portfolio-performance.info/t/wertpapier-duplizieren/8360/10


Nirus2000:
Hello @OnkelDok 
Hmm... in my opinion, this could cause a problem.
If now the security is duplicated 1 to 1, how do we identify e.g. in the PDF importer or in the CSV importer to which security the transaction is assigned?

`private static final Security DUPLICATE_SECURITY_MARKER = new Security();`

- [SecurityCacheTest.java](https://github.com/buchen/portfolio/blob/ac710e068c6a467ed0a30bfc45e12c463423b32b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/SecurityCacheTest.java)
- [CrossEntryCheckTest.java](https://github.com/buchen/portfolio/blob/ac710e068c6a467ed0a30bfc45e12c463423b32b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/checks/impl/CrossEntryCheckTest.java)
- ...

Regards Alex

chirlu:
Cloning only makes sense as an intermediate step, you will always want to change something in the clone. It’s the user’s responsibility to make appropriate changes.

buchen:
About the duplicate securities during import: I agree that the duplication can only be a first step. Already today, the importer will complain if it cannot identify a single security.

If we duplicate the security within one existing file, we should also duplicate the assignments in the taxonomies. That is not part of the "deepCopy" (and does not work across multiple files due to different UUIDs).

OnkelDok:
> If we duplicate the security within one existing file, we should also duplicate the assignments in the taxonomies.

Yes, that's correct. You already merged the PR. Should I work on also duplicating the taxonomies with a new PR or do you change this directly?

chirlu:
Commit 264d43a259cf6a1ef1aa3b5ecda2a79630bfa70f.

PR #3192, opened 2023-02-19
Remove debugging output to stdout

chirlu:
See https://github.com/buchen/portfolio/pull/3187#issuecomment-1435972131 (I guess it doesn’t have much of a chance to be read, on a closed pull request).


buchen:
ah, danke, slipped through

PR #3194, opened 2023-02-20
Fixes wrong dates when exporting transactions as CSV

OnkelDok:
Issue: https://forum.portfolio-performance.info/t/wrong-dates-when-exporting-security-transactions-as-csv/21226
Not visible table items get wrong datetime when exporting the table after re-order.

Don't know much about the background of the error. But an additional check is required to get the text updated. For items which are invisble after an re-order of the table (click on date column header), the ``getData()`` returns a valid object (not null) but ``getText()`` returns empty string. So also in case of empty string the additional call of ``getText()`` seems to be required.


OnkelDok:
@buchen: I realized that the new condition ``item.getText().isEmpty()`` in combination with following ``item.getText()`` is unnessesary. Because to let update the data it's enough to call ``item.getText()`` once. With my initial solution it's called twice (in if condition and when condition matches). So either we can call ``item.getText()`` in general, or we find another condition how to detect if a call of ``item.getText()`` is needed.

buchen:
@OnkelDok thanks!!!

I was never able to reproduce it - hence somehow I was in the dark. Have you been able to reproduce the problem?

PR #3195, opened 2023-02-21
Clarify that reporting periods don’t include day 1

chirlu:
It happens too often that someone sets a reporting period of “from X to Y” or “since X”, then complains on the forums or elsewhere that something that happened on day X is not shown in some report. To make it clear that the reporting period starts only after day X (with initial value as of the end of day X), change the reporting period dialogue to call the options “from X (excl.) to Y” and “since X (excl.)”.

example of reporting period dialogue with the change

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -592,6 +592,7 @@ public class Messages extends NLS
     public static String LabelCurrentWeek;
     public static String LabelDashboard;
     public static String LabelDataSeries;
+    public static String LabelDateExclusive;
     public static String LabelDateXToY;
     public static String LabelDefaultFontSize;
     public static String LabelDefaultReferenceAccountName;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/ReportingPeriodDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/ReportingPeriodDialog.java
@@ -125,6 +125,8 @@ public class ReportingPeriodDialog extends Dialog
         radioFromXtoY = new Button(editArea, SWT.RADIO);
         radioFromXtoY.setText(Messages.LabelReportingDialogFrom);
         dateFrom = new DatePicker(editArea);
+        Label lblFromExcl = new Label(editArea, SWT.NONE);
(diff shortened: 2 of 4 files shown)


chirlu:
I intentionally did not add (bad) translations for the languages I don’t speak. I believe translations should be done by the human translators, via POEditor. Until they get around to do it, a few untranslated English strings should be acceptable. Not having a string is a better indicator of “still needs to be translated” than having a string that unfortunately is wrong.

buchen:
No idea what got me to do this from (excluded) - to (included)... too late to change now. But making it self-explanatory helps. Thanks

chirlu:
Unrelated issue:
branch view from gitk
Release 0.60.2 is on a dead-end branch. The changes were separately included into the main branch, but I think it would be better to `git merge release_0.60.2` into master.

buchen:
> I intentionally did not add (bad) translations for the languages I don’t speak. I believe translations should be done by the human translators, via POEditor. Until they get around to do it, a few untranslated English strings should be acceptable. Not having a string is a better indicator of “still needs to be translated” than having a string that unfortunately is wrong.

Personally I do not like partially translated applications. And not all languages have active POEditors. Plus I think deepl and others are already pretty good. I think one can argue either way

buchen:
> Release 0.60.2 is on a dead-end branch. The changes were separately included into the main branch

Good point. Let me try this. For the 0.60.2 release, I cherry-picked the changes from the master branch because I felt the Java 17 part was not ready yet (but already in the main branch).

OnkelDok:
Would it lead to even less confusion if the period displayed at the very top of the selection dialog would show the corresponding correct period?
So in screenshot above from @chirlu the correct period to be displaced should be 01.01.2015 - 21.02.2023, correct?

> too late to change now

@buchen : Do you see any major drawbacks when changing that logic (if you would decide it should now be different)?
I would see two possibilities:
a) just change the reporting period interpretation and provide enough information about this change to the user. So the user have to change their periods manually.
b) migrate the old settings automatically on startup to get the same period the user had configured before.

But the question is: Is the current period handling suitable or not?

From my side, I have no problem with the current implementation.

chirlu:
> I have no problem with the current implementation.

Neither do I. I think we should leave it as it is.

> Would it lead to even less confusion if the period displayed at the very top of the selection dialog would show the corresponding correct period?
So in screenshot above from @chirlu the correct period to be displaced should be 01.01.2015 - 21.02.2023, correct?

I’m sure that would be **more** confusing. I also think putting the weight on the description “with initial value as of the **end** of day X”, as the additional hint does, makes it quite understandable.

PR #3196, opened 2023-02-23
SecurityChart: reset axes when no chart is displayed + place short message with no data reason into chart

OnkelDok:
https://forum.portfolio-performance.info/t/kursdiagramm-zeigt-50-jahre-zeitraum-an-wenn-man-einen-index-anzeigt/23283

First implementation (outdated after feedback):
~~The current solution displays all prices when no transaction is available to determine the holding period.
So with this solution also for normal securities without transaction all prices are shown now. I'm not sure if this is OK.~~

~~The alternative would be to first check if the security is an index and then call the new function ``getAllPrices``. What would be an indikator for an index? ``security.getCurrencyCode() == null``?~~

Current implementation:
Reset axes to a specific range to always have the same appearance when no data is displayed + display a message with the reason why no data is shown.
Screenshots in this post: https://github.com/buchen/portfolio/pull/3196#issuecomment-1450684727


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -190,17 +190,21 @@ public ChartInterval getInverval(Client client, CurrencyConverter converter, Sec
...
-                    else
-                        return new ChartInterval(prices.get(0).getDate(), prices.get(prices.size() - 1).getDate());
+                    return getAllPrices(now, security);
 
                 default:
                     throw new IllegalArgumentException();
             }
         }
+
+        private ChartInterval getAllPrices(LocalDate now, Security security)
```
chirlu, line 200: I don’t like the name `getAllPrices()` because it doesn’t actually get prices, it finds out for which interval prices are available.

OnkelDok: I agree with you. What would be a good name? ``determinIntervalFromPrices``?

chirlu:
I’m not convinced this is actually something that needs fixing. By choosing “H”, users indicate they only want to see the chart for when there were holdings. When there never were holdings, it is in my opinion correct to not show anything.

Where I see the issue is that the empty chart has a legend claiming that it displays 52 years of prices (and wrongly at that, because even for the period where prices are available, they are not shown). Perhaps we should show an actually empty chart there.

(BTW, this also affects the currency page, where there are many currencies without showable rates.)

kimmerin:
> I’m not convinced this is actually something that needs fixing. By choosing “H”,
> users indicate they only want to see the chart for when there were holdings. When
> there never were holdings, it is in my opinion correct to not show anything.

The common use case in my eyes is that you select "H" and then go through the entries of interest. When selecting an index you already forgot that you've selected "H" last week which might come with special cases and seeing an empty 50 year time frame from 1997 to 2047 is a bit confusing. I agree with you that "show nothing" is the right thing but currently it is doing something, namely showing a timeframe without diagram which is something worth "fixing" IMHO.

@OnkelDok wrote:
> The alternative would be to first check if the security is an index

I don't think that a check of an entry being an index is necessary or correct, because a user might have added a security to be able to use it as benchmark. That as well means that you don't actually hold anything which should lead to the same effect (haven't tried it but I'd bet other peoples body parts on it ;-)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -180,7 +180,7 @@ public ChartInterval getInverval(Client client, CurrencyConverter converter, Sec
                 case H:
                     List<TransactionPair<?>> tx = security.getTransactions(client);
                     if (tx.isEmpty())
-                        return new ChartInterval(now, now);
+                        return getAllPrices(now, security);
```
chirlu, line 183: We could just return `null` as the `ChartInterval` (here and below, where the other `ChartInterval(now, now)` is currently); it seems this one-day interval causes the chart to get that huge an interval. Then at https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java#L725-L738, we could catch the `null` and do the same as for a `null` range. We could even put the test for `null` into `ChartRange.createFor()` (return `null` when a `null` interval is passed) and wouldn’t have to do checking twice in this file.

chirlu:
A general issue with this whole security chart is that it has state – it doesn’t reset everything when you move to a different security, or to no security at all. At least the axes retain the ranges they last had if they aren’t explicitly reset. We should probably always reset it to its initial state: for me, right after PP starts, it shows nothing on the X axis and a 0 to 1 linear scale on the Y axis. That would probably be fine for an intentionally empty chart, too.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
OnkelDok: Implemented with [18138fe](https://github.com/buchen/portfolio/pull/3196/commits/18138fe0a2a34ae35f9ddce26b7be0d3441f96b7)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -712,6 +714,7 @@ private void updateChart()
 
             if (security == null || security.getPrices().isEmpty())
             {
+                chart.resetAxes();
```
chirlu, line 717: Last thing (actually, I just see a second thing …): I would call `resetAxes()` once, before any `if` blocks, so that we always start from the same, known state. So above, at [lines 702–711](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java#L702-L711), where everything else is cleared (marker lines, paint listeners etc.).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -733,6 +736,7 @@ private void updateChart()
             ChartRange range = ChartRange.createFor(prices, chartInterval);
             if (range == null)
             {
+                chart.resetAxes();
                 chart.redraw();
```
chirlu, line 740: So, (perhaps) very last thing: Those calls to `redraw()` are redundant, because the method is once again called from the `finally` block at [line 855](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java#L855). I don’t know if it really redraws twice, but if it does, it would be a waste of processing time. :) So just a `return` here (doesn’t even need the braces then), and setting the title before `return`ing above.

buchen:
> I agree with you that "show nothing" is the right thing but currently it is doing something

I propose we do not display an empty chart, but (also?) a message "no holdings". I also think if there are not prices, the maybe a message "no prices" or "no prices in selected time range" would explain better what is happening.

What do you think?

OnkelDok:
> I propose we do not display an empty chart, but (also?) a message "no holdings". I also think if there are not prices, the maybe a message "no prices" or "no prices in selected time range" would explain better what is happening.
> 
> What do you think?

Yeah, with a message the user always knows why no data is displayed.
Where do you want to place the message? In the middle/center of the plot area of the chart?

buchen:
> In the middle/center of the plot area of the chart?

Sounds reasonable.

OnkelDok:
With [8bee54d](https://github.com/buchen/portfolio/pull/3196/commits/8bee54de89539a4fd9378a4a807bfd6dfe3dbfa0) I followed @chirlu s reviews and ``resetAxes`` on default.
Additionally now a message is displayed when there are no holdings or nor prices or nor security selected.

First example:


Open points:
- setup translation of the messages
- tune style of the message (color, size, font)...I am grateful for any suggestions

kimmerin:
> tune style of the message (color, size, font)...I am grateful for any suggestions

Personal taste here: The box surrounding the text doesn't need solid edges, i.e. there should be one to make sure that the horizontal line in the background doesn't go through the text but with it being visible I think it draws too much attention to that part of the GUI.

On a side note: Is it lost when creating the screenshot or why is the selected timeframe not visible? Shouldn't the "H" be shown as selected?

OnkelDok:
@kimmerin: thank you for providing your thoughts. Maybe I can change it later today and give an example.

> On a side note: Is it lost when creating the screenshot or why is the selected timeframe not visible? Shouldn't the "H" be shown as selected?

That's an open bug in dark theme, reported [here](https://forum.portfolio-performance.info/t/selektierten-zeitraum-in-kursdiagrammen-hervorheben-und-beibehalten/17225/9?u=onkeldok).

OnkelDok:
With [05168ec](https://github.com/buchen/portfolio/pull/3196/commits/05168ecdb224948d26b0110e64f664df2cab1175) I committed the version you can see in my last post + translations. So I think it's ready again for a review.

buchen:
rebased, slightly changed, merged

PR #3197, opened 2023-02-23
Fixes serialization error after duplicate dashboard column

OnkelDok:
https://forum.portfolio-performance.info/t/nach-duplizieren-einer-spalte-im-dashboard-kann-xml-nicht-mehr-geparst-werden/23294

Sorry for including the error with #3175. I blindly followed the code change suggestion from eclipse. The ``.toList()`` creates an immutable list which leads to serialization problem.


buchen:
No problem. I ran into a similar problem. We really have to be careful about the "toList" usage.

Wherever we know that the immutable list will not do it, we could add ```// NOSONAR mutable``` at the end of the line. Then we a) do not get the suggestion from Sonar again and b) know why not to refactor this in 5 years again...

buchen:
Or we add a ```toMutableList``` collector ourselves - which is used in all places where we must have a mutable list

PR #3202, opened 2023-02-26
Rework Targobank PDF-Importer

### LINKED ISSUE #3092: Targobank PDF import no longer working
Targobank PDF import no longer working correctly.

I try to import both the Steuerbeilage and the Ertragsgutschrift at the same time. But the result is as follows:

Bildschirmfoto von 2022-12-20 17-19-10
[Ertragsgutschrift vom 2022-12-14.pdf.txt](https://github.com/buchen/portfolio/files/10270262/Ertragsgutschrift.vom.2022-12-14.pdf.txt)
[Steuerbeilage vom 2022-12-16.pdf.txt](https://github.com/buchen/portfolio/files/10270263/Steuerbeilage.vom.2022-12-16.pdf.txt)

Nirus2000:
Fixes #3092
Fixes postProcessing process
Integrate in ExtractorUtils,java
Importer converted to the contributing standards

Hello @buchen
please check again the postProcessing of the PDF importer.

Thanks
Alex


Nirus2000:
Hello @buchen 
> Do you have something specific in your mind what to check?

I have built the postProcessing() part completely different.
If you find time, you could have a look at the List<Item> part of the streams for sales + tax. ([L421 to L442](https://github.com/buchen/portfolio/pull/3202/commits/fc864f60ea749a66c2ff035ca43bd8281a5f9ab5#diff-502458d76325736befcaf0580168f5c141ba5837f9df8da497ca3719c854818cR421-R441))
I could not connect this as a stream analogous to dividends + tax.

Second would be the try-catch-retryCatch ([L471 to 495](https://github.com/buchen/portfolio/pull/3202/commits/fc864f60ea749a66c2ff035ca43bd8281a5f9ab5#diff-502458d76325736befcaf0580168f5c141ba5837f9df8da497ca3719c854818cR471-R494)) part.
I hope thats okay.

Greetings
Alex

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TargobankPDFExtractor.java
```diff
@@ -58,413 +68,529 @@ public String getLabel()
...
-                    if (a1.getUnit(Unit.Type.TAX).isPresent())
+                    // Which transaction is the taxes and which the sell?
+                    try
                     {
-                        a2 = (AccountTransaction) transactions.get(1).getSubject();
+                        sellTransaction = (BuySellEntry) transactions.get(0).getSubject();
+                        taxTransaction = (AccountTransaction) transactions.get(1).getSubject();
                     }
-                    else
+                    catch (ClassCastException e1)
```
buchen, line 477: It is not a good pattern to have the control flow via exceptions. Exceptions are expensive to create. And in this case, you could easily check with "instanceof" - you know there are only those two types available.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TargobankPDFExtractor.java
```diff
@@ -58,413 +68,529 @@ public String getLabel()
...
-                    // combine document notes and mark transaction 2 to be
-                    // deleted
-                    a1.setSource(concat(a2.getSource(), a1.getSource()));
-                    a1.setNote(concat(a2.getNote(), a1.getNote()));
-                    a2.setNote(TO_BE_DELETED);
+                        // Combine at notes
+                        sellTransaction.setNote(concat(sellTransaction.getNote(), taxTransaction.getNote()));
+
+                        // Set note that the tax transaction will be deleted
+                        taxTransaction.setNote(TO_BE_DELETED);
```
buchen, line 511: You also have access to the Item itself (e.g. 474). Therefore instead of using the note to make it as deleted and iterate over the items later, you could also say ```items.remove( ... )``` here

buchen (changes requested):
I have looked only at the #postProcessing method in details.

I propose to change at least the use of the ClassCastException - use ```instanceof``` instead.

Maybe the algorithm could be easier, if
* you do not put sell and tax transactions in the same map. If you keep the item types in two maps, you always know which type the elements are and you avoid casting. You just check if both maps have exactly one element (the equivalent to the size = 2 check)
* use a composite key, e.g. Pair<LocalDateTime, Security>, to avoid a map of maps. Could also make the loop easier.

PR #3203, opened 2023-02-26
[DRAFT] Refactored non-importable items to use failure message instead

buchen:
This commit replaces the use NonImportableItem with a regular item that has a failure message. This way all item types can be marked as failure while presenting each type specific data to the user.

Extractors can use this feature by storing a FAILURE message in the transaction context and setting the message when wrapping the item.

Issue: #3119


buchen:
Hi @Nirus2000,

here a proposal for an alternative implementation of #3119. Instead of duplicating the transaction item classes, this proposal adds a "failure message" to the Extractor.Item. Instead of using the "note" field to mark items as non-importable, this change use then the failure message. Extractors can use the transaction context to remember if a transaction fails.

For now, I have only adopted the Baader PDF extractor. Once we agree on the concept, I can update the commit to change the existing usages of NonImportableItems. And a second commit could add other enhancements from #3119 for non-importable items

buchen:
merged + fixes remaining test cases

Issue #2359, opened 2021-07-31
CSV Import custom date format


vwkd:
I tried to import a CSV for historical data whose dates had the format "MMM DD, YYYY", where the months used english words instead of numbers. I found the option "tt-mmm-yyyy (Englisch)", however it would have required a different order and dashes.

<img width="1019" alt="pic" src="https://user-images.githubusercontent.com/33468089/127753318-8cc92761-9a1d-443e-b20e-2368c5dd15b8.png">

Since there are so many different combinations to write dates, it would make sense to allow the user to customize the format. Google Sheets allows something similar

<img width="512" alt="google sheets date and time formats" src="https://user-images.githubusercontent.com/33468089/127753387-9ae9aef1-a135-4a42-9440-29572e4dc3cf.png">

I recognize this issue isn't pressing. As a workaround, one can import the CSV to Google Sheets, change the format to one that Portfolio Performance supports, and re-export it. Since importing historical data is already rare, adding yet another manual step doesn't hurt too much. It probably is more valuable to use the time to make portfolio-report.net more complete, such that manual imports are even more rarely necessary.

flywire:
Date field is probably over specified. Options are:
1. dmy, mdy or ymd
2. 2 digit day, 2 digit/ 3 character/ word month, 2 or 4 digit year
3. space/slash/dash/dot/none for separator
4. various other special formats including day of week and time
5. language
5. of course, there are only so many days in a month and months in a year

User should be able to confirm format suggested by code supported by locale.

flywire:
The default should be based on locale.

> 1. dmy, mdy or ymd

_If_ an option is required this should be the only one.  There is no need to specify `. / -` separators, just accept all.

If four-digit years are used that should help determine the format, and the install has a default language. If locale default isn't an option it's really only a choice between dmy or mdy with dmy the default between these two.

flywire:
Related #2612

PR #3209, opened 2023-03-05
Adding a new money and date format in the CSV importer

### LINKED ISSUE #2944: CSV import, special number series
So here one issue I come across:
image

The format of numbers currently only identifies
0.000,00
0,000.00
0’000.00
but 0 000,00 not yet.

This new format would be useful for me at least as it then always just imports 1 if I use 0.000,00, but the value was 1 000.

### LINKED ISSUE #2945: CSV import, new number format
So here one issue I come across:
image

The format of numbers currently only identifies
0.000,00
0,000.00
0’000.00
but 0 000,00 not yet.

This new format would be useful for me at least as it then always just imports 1 if I use 0.000,00, but the value was 1 000.

### LINKED ISSUE #3156: Number format without thousands separator in CSV import
Is it possible to add a number format for the number of stocks without a thousands separator and with a comma to separate tenths? For example 123456,123
Here use in CSV import:
image

Nirus2000:
Fixex #2359
Fixes #2944
Fixes #2945
Fixes #3156
Add new money format 0 000.00
Add new date format MMM dd, yyyy
Delete the unused label PDFdbLabel


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
```diff
@@ -347,6 +348,8 @@ public static class AmountField extends CSVImporter.Field
                                         NumberFormat.getInstance(Locale.GERMANY)),
                         new FieldFormat("0,000.00", Messages.CSVFormatNumberUS, //$NON-NLS-1$
                                         NumberFormat.getInstance(Locale.US)),
+                        new FieldFormat("0 000.00", Messages.CSVFormatNumberFrence, //$NON-NLS-1$
```
buchen, line 351: typo

buchen:
fixed and merged

PR #3210, opened 2023-03-05
Improve translation TestUtilities

Nirus2000:
This PR is used to check in the translations whether they start with a newline or end with a newline.
Output of the key in case of an error

Delete unused PDFdbLabel label

_I know that the list "skippedKeys" contains only one value so far.
This is as preparation for possibly further keys_ :innocent:


buchen:
Merged.

Changes
* the test utilities trim the terms for all whitespaces
* refactored code that relied on leading or trailing whitespace
* update the releng tools to automatically trim terms https://github.com/buchen/portfolio-releng/commit/3d71975c63b6aa321326f53f2af2b3031e067bfe

PR #3219, opened 2023-03-12
Added mutli-national currencies

RomanLangrehr:
All multi-national currencies (with an X__ ISO 4217 code) were missing, so I added these. I also included XDR (IMF special drawing rights), even though this currency can not be obtained by individuals, because its value represents a basket of the most important currencies and thus is useful when you want to measure performance with respect to a currency basket.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_de.properties
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_de.properties
@@ -337,6 +337,16 @@ VUV = Vatu
 
 WST = Samoanischer Tala
 
+XAF = Zentralafrikanischer CFA-Franc
+
+XCD = Ostkaribischer Dollar
+
+XDR = IWF-Sonderziehungsrechte
+
+XOF = Westafrikanischer CFA-Franc
+
+XPF = CFP-Franc
+
 YER = Jemen-Rial
 
 ZAC = S\u00FCdafrikanische Cents
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_fr.properties
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_fr.properties
@@ -337,6 +337,16 @@ VUV = Vatu
 
 WST = Tala
 
+XAF = Franc CFA (CEMAC)
+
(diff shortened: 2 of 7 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_de.properties
```diff
@@ -337,6 +337,16 @@ VUV = Vatu
 
 WST = Samoanischer Tala
 
+XAF = Zentralafrikanischer CFA-Franc
+
+XCD = Ostkaribischer Dollar
+
+XDR = IWF Sonderziehnungsrechte
```
chirlu, line 344: 1) Should have a hyphen instead of a space. 2) Has an additional n (Zieh**n**ung).

name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_fr.properties
```diff
@@ -337,6 +337,16 @@ VUV = Vatu
 
 WST = Tala
 
+XAF = Franc CFA (CEMAC)
+
+XCD = Dollar des Cara\u00EFbes orientales
+
+XDR = FMI Droits de tirage sp\u00E9ciaux
```
chirlu, line 344: French is a language where determiners are put behind the main word and often require a preposition. (Can be seen in “inverted abbreviations”: e.g. NATO, North Atlantic Treaty Organization, is OTAN in French, l’organisation du traité de l’Atlantique du Nord; or UNO, United Nations Organization, is ONU, l’organisation des nations unies). So I suggest “Droits de tirage spéciaux du FMI” here.

name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_de.properties
RomanLangrehr: Thank you. I fixed it.

name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies_fr.properties
RomanLangrehr: Thank you. I fixed it.

PR #3220, opened 2023-03-17
Taxes column for deposit account transactions pane

rcjan:
https://forum.portfolio-performance.info/t/steuern-direkt-zu-zinsen-erfassen/6158/13


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/panes/AccountTransactionsPane.java
```diff
@@ -364,6 +366,31 @@ public Color getForeground(Object element)
...
+            }
+
+            @Override
+            public Color getForeground(Object element)
+            {
+                return colorFor((AccountTransaction) element);
+            }
+        });
+        ColumnViewerSorter
+                        .create(element -> ((TransactionPair<?>) element).getTransaction() instanceof AccountTransaction
```
buchen, line 386: This will result in a ClassCastException because the items are actually of type AccountTransaction

buchen (commented):
Hi,

thanks for your contribution!

I like the idea of adding the additional columns. However, taxes and fees are not directly stored with the AccountTransaction in case of purchase and sale. Therefore, I needed to enhance the logic how to lookup the taxes. It was easier to just do the change, but I added the both of us as authors.

buchen:
Adapted and merged

PR #3221, opened 2023-03-17
Write backup-after-open even with active autosave

chirlu:
Normally, immediately after a portfolio file has been successfully opened, a copy of the file with infix `backup-after-open` is created. This serves to always retain one backup with a known good state (for otherwise opening the file would have resulted in an error). However, this behaviour was suppressed when autosave was active, with the original rationale that the autosave file would immediately be written with the same content.

However, as recent events have shown, the autosave functionality is not a suitable replacement for backup-after-open; in case of a bug, the autosave file may become corrupted in the same way as a file saved manually. The “known good state” is no longer guaranteed. Therefore, change the condition for writing a backup-after-open file to be independent of whether autosave is active or not.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
@@ -368,8 +368,7 @@ public class ClientInput
 
     public void createBackupAfterOpen()
     {
-        if (clientFile != null && preferences.getBoolean(UIConstants.Preferences.CREATE_BACKUP_BEFORE_SAVING, true)
-                        && preferences.getInt(UIConstants.Preferences.AUTO_SAVE_FILE, 0) == 0)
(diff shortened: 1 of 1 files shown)


chirlu:
The original discussion was here:
https://github.com/buchen/portfolio/pull/1559#discussion_r429349401
It appears there was a misunderstanding at work. When @buchen said

> Aber da wird ja schon ein Backup geschrieben ("after-open"). Ich denke das kann weg, oder?

he probably meant the annotated code line (writing an immediate autosave backup) should go; however, @inv-trad seems to have understood it as “writing an after-open backup should go” (when autosave is active), and that was what he then implemented.

The commit message of the subsequent cleanup commit (c05fcddcb54ec6a1953a38bb8dd240babd1966eb) shows that @buchen wasn’t aware of this “no after-open backup when autosave is active” change:

> removed creation of autosave file upon opening the file (**there is already a backup file created**)

(My highlighting; the highlighted statement actually was no longer true.)

PR #3223, opened 2023-03-19
Added option to specify a factor quotes get mutliplied with for JSON quote provider.

RomanLangrehr:
I have to work with a quote provider that multiplies all quotes with `1000000000000000000` to avoid decimal representation. To import this, I added a field that allows the user to sepecify a factor (like `0.000000000000000001`) for the quotes.

image


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
```diff
@@ -237,6 +243,16 @@ protected List<LatestSecurityPrice> parse(String url, String json, String datePa
                                 Messages.MsgErrorNumberOfDateAndCloseRecordsDoNotMatch, dates.size(), close.size())));
                 return Collections.emptyList();
             }
+            
+            BigDecimal factor;
+            if(factorString.isPresent())
+            {
+                factor = new BigDecimal(factorString.get());
```
OnkelDok, line 250: What if the user entered a wrong decimal format which will lead to a ``NumberFormatException``?
Maybe I missed the check. But I would suggest to check if the entered value can be parsed without error in the ``onJsonFactorChanged`` and display an indicator that the value is not correct.

RomanLangrehr: If it the user enters something invalid, the respective error message (e.g. “Not a digit“ or “Character array is missing "e" notation exponential mark“) is shown in the quote preview pane, as it is down for all error messages that might appear when parsing quotes. I think this is fine.

PR #3230, opened 2023-03-26
Fix IndexOutOfBoundsException thrown when importing quotes via CSV, b…

RomanLangrehr:
…ut no valid quote could be found.


buchen:
thanks, rebased, changed to is isEmpty

Can you me do a favor and format the code in PP style?
It should be pre-configured in Eclipse. You can also configure a "save action" which is automatically formatting the code.
Thanks!

Issue #3237, opened 2023-03-28
add new feature "dividend YoC"


Marfir:
**current situation**
I like to develop for myself the new feature "Dividend yield on cost". Currently it's not possible to show the dividend yield based on the past purchase price. But this can have a huge difference to the current dividend yields that are available in PP, if the current (stock) price and your past buying rate are significant different.
see discussion in the forum:
https://forum.portfolio-performance.info/t/personliche-dividendenrendite/335/2

**goal**
Implement a dividend YoC, no additional things that was discussed in the forum (link above).

chirlu:
Duplicate of #2266.

Marfir:
Hm not sure if this is an exact duplicate. I had planned to show the dividend YoC for each security in the portfolio. To do the same over the complete portfolio is an additional feature. But maybe I can do it also.

pfalcon:
Hmm, clicking around, I can see following (in master):

Reports -> Performance -> Securities, then select a particular security, then select Transactions in the below pane. This won't show a standard transaction table as in other places, but rather adhoc table with additional columns, among which are "Dividend Yield" and "Dividend Yield (moving average)". The description of both are "Dividend yield = dividends / purchase price using FIFO method" and "Dividend yield (moving average) = dividends / purchase price using moving average method". The shown percentages are "raw" (one-time). I figure PP doesn't know the frequency of dividends, so one needs to "mentally" multiply by 4/12/whatever to get familiar "per anum". Ah, and it's based on "Gross dividend" (pre-tax), someone may want to know both rates. (Generally, I find it confusing where in PP dividends are shown gross or net. IMHO, it should be explicitly mentioned every time in a description/tooltip.)

Marfir:
@buchen fyi

PR #3238, opened 2023-03-28
WIP: #3237: extends gitignore file

Marfir:


buchen:
The project configuration files are checked-in on purpose. It is common practice to submit the project files for Eclipse RCP applications (see for example the [Eclipse Platform UI](https://github.com/eclipse-platform/eclipse.platform.ui)).  Some setup can be shared. For me the most important part is that the code formatting in PP style is automatically included. And it does make little sense to use IDEs other than Eclipse because of use of SWT, Eclipse RPC, the run configuration framework, the target platform, etc.

I know that the SonarCloud configuration may be annoying (as you cannot setup a connection the cloud project to share the same configuration). It is "on my list" to remove this from the main branch but I haven't figured out which parts exactly to exclude so that the local setup continues to work (happy to merge such a PR).

Marfir:
I'm not sure. It could be part of the settings folder or inside of the .project file. But I never heard that people share the auto generated class path file. Thats not an recommended way to work and make also no sense because it's an generated file. I can reduce the gitignore file a little bit.

Yes Sonar did not works. And it's worse because eclipse print every second a warning/error in the console output and because of this it's impossible to work with the other views (search etc.). I removed the sonar lint plugin from eclipse and then I was able to work.

But why you close the pull request? I'm not done with the feature!

Marfir:
> @Marfir: I recommend you to convert/set your PR to `draft` so that the PR is not accidentally reviewed when you are still working on it. And (sometimes) a bit of description will help to understand what you are planning to do.

ok thx, I will do so.

PR #3251, opened 2023-03-31
Completion of the PDF-Importer test after #3119

Nirus2000:
Add missing security tests (Ticker Symbol, WKN, ISIN) 
Add some notes
Adjustment of the tests after the implementation of #3119 
Add new message "MsgErrorInvalidWKN"
Delete unused translation "MsgErrorMissingDate"
Fix some typo in classnames in tests
Rename some testfiles to contributing standards


Nirus2000:
Yes, because so often the TickerSymbol is not read out, because it simply does not exist in the PDF debugs.
I added these tests because these are options with which variables the security is generated.
This completes the tests for generating securities for transaction.

> I do not want to keep you busy with things that just lengthen the code. ;-)

Don't worry about that... right now you know I can't do much. It's important to me that the tests are really complete with all variables, since you already told me that you often program against the tests ;-)

PR #3255, opened 2023-04-01
Improve dividend and investment markers

Nirus2000:
**Dividend markers:**

1. In this optimized code, the commonly used symbolSize, fontMetrics, labelHeight, and halfLabelHeight values are calculated outside the loop and stored in variables.
2. The dividend label is concatenated using a StringBuilder. Instead of concatenating dividend labels using the + operator, we use a StringBuilder to improve performance.
3. The other variables are stored outside the loop to avoid repeated computation Number of loop repetitions reduced and performance improved.

**Investment markers:**

1. I call textExtent directly on the Graphics object to avoid creating a new Point object for each loop execution.
2. I have replaced the comparison of (textExtent.x / 2) with textExtent.x / 2 to save a division.
3. If there is no transaction, skip

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -31,6 +31,7 @@ import org.eclipse.swt.events.PaintEvent;
 import org.eclipse.swt.events.PaintListener;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.graphics.FontMetrics;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.widgets.Composite;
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1246,59 +1255,50 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
...
             {
                 customPaintListeners.add(event -> {
+                    FontMetrics fontMetrics = event.gc.getFontMetrics();
+
+                    StringBuilder labelBuilder = new StringBuilder();
+
+                    // Three levels of the label
+                    int[] yPosLabels = new int[3];
+                    int[] lastWriteLabels = new int[3];
+                    Arrays.fill(lastWriteLabels, 0);
```
buchen, line 1265: In Java, int arrays are always initialized with zeros. No need to do this here.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1246,59 +1255,50 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
             if (chartConfig.contains(ChartDetails.SHOW_DATA_LABELS))
             {
                 customPaintListeners.add(event -> {
+                    FontMetrics fontMetrics = event.gc.getFontMetrics();
+
+                    StringBuilder labelBuilder = new StringBuilder();
```
buchen, line 1260: The use of the StringBuilder does not make sense here. It will be only filled by exactly one string (the one returned by getDividendLabel.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1246,59 +1255,50 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
...
+
+                    // Three levels of the label
+                    int[] yPosLabels = new int[3];
```
buchen, line 1263: You do not need to store the Y position - it is fixed. Anyway, you only write into the array just before reading out of the array. Therefore it can be a local variable.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1246,59 +1255,50 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
...
+                        boolean labelPlaced = false;
 
-                        if (((x - (textExtent.x / 2)) - lastWriteLabelLevel2) <= 0)
-                            freeSpaceForLabelLevel2 = false;
-
-                        if (((x - (textExtent.x / 2)) - lastWriteLabelLevel3) <= 0)
-                            freeSpaceForLabelLevel3 = false;
-
-                        if (freeSpaceForLabelLevel1 || freeSpaceForLabelLevel2 || freeSpaceForLabelLevel3)
+                        for (int level = 0; level < 3 && !labelPlaced; level++)
```
buchen, line 1287: Instead of the boolean labelPlaced, I think you can just call "break" and you will jump out of the loop.

buchen:
> I have replaced the comparison of (textExtent.x / 2) with textExtent.x / 2 to save a division

I do not understand that sentence. Where do you save a division? (And I trust the compiler to optimize repeated terms and unnecessary brackets).

Otherwise I adapted the changes I mentioned in the code, rebased it to master, and merged it.
It is a great simplification to work with a loop instead of the multiple variables for each level

PR #3257, opened 2023-04-02
Fix NumberFormat France in CSV-Importer

Nirus2000:
https://forum.portfolio-performance.info/t/csv-imported-transactions-were-rounded-down-on-import/23539/
Testfiles:
[Test_Numberformat_france_whitespace_coma.csv](https://github.com/buchen/portfolio/files/11132566/Test_Numberformat_france_whitespace_coma.csv)

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
@@ -349,8 +349,12 @@ public final class CSVImporter
                                         NumberFormat.getInstance(Locale.GERMANY)),
                         new FieldFormat("0,000.00", Messages.CSVFormatNumberUS, //$NON-NLS-1$
                                         NumberFormat.getInstance(Locale.US)),
-                        new FieldFormat("0 000.00", Messages.CSVFormatNumberFrance, //$NON-NLS-1$
-                                        NumberFormat.getInstance(Locale.FRANCE)),
+                        new FieldFormat("0 000,00", Messages.CSVFormatNumberFrance, () -> { //$NON-NLS-1$
+                            DecimalFormatSymbols unusualSymbols = new DecimalFormatSymbols(Locale.FRANCE);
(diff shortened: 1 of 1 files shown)


chirlu:
Could you describe what you believe the problem is that you are trying to solve?

Nirus2000:
Hello @chirlu 
Absolutely! You're absolutely right that I didn't include a description here of the problem property and how to solve it.
I apologize for this and thank you for this.

**The following problem:**
https://forum.portfolio-performance.info/t/csv-imported-transactions-were-rounded-down-on-import/23539/7
https://forum.portfolio-performance.info/t/csv-imported-transactions-were-rounded-down-on-import/23539/14
An error occurred in PR 64177023b9df4d9ad197548830d93e7716fde966, which is now corrected.
The correct format is `0 000,00` and not `0 000.00`.

**Format**
Locale [fr_FR](https://www.localeplanet.com/java/fr-FR/index.html) or [en_FI](https://www.localeplanet.com/java/en-FI/index.html) (1 234,56)
`getGroupingSeparator() = " "`
`getDecimalSeparator() = ","` (comma)
`Amount = "1 234.56"`

grafik
grafik

Nirus2000:
@buchen
I somehow have the impression that the CSV test cases don't really work well and should be revised. The locales (formatting of the numbers) are not checked. I think we should set up test cases for this to check the different number formats to prevent these errors.
@chirlu What do you think?

buchen:
I think the main problem is that the locale use the "non-breaking space" as grouping separator - which then does not match the CSV import (which typically just has a space)

chirlu:
Yes, it makes more sense to me now than the original version with multiple different French modes.

Generally, though, I still don’t understand why there are three definitions for each mode, one of them translatable, and why the non-translated definition is shown in the UI while the translated definition is used for interpreting the CSV.

buchen:
> Generally, though, I still don’t understand why there are three definitions for each mode, one of them translatable, and why the non-translated definition is shown in the UI while the translated definition is used for interpreting the CSV.

Agree. That does not make sense. I think I messed this up when I introduced the label and then @Nirus2000 was just extending the same pattern. Will fix it.

PR #3259, opened 2023-04-04
Refactoring: JEP 394: Pattern Matching for instanceof casts

Nirus2000:
JEP 394: Pattern Matching for instanceof casts


buchen:
I have adapted the remaining statements as well (see second commit).

While doing it, I was thinking whether it is worth to do these bulk updates. On the one hand, it simplifies the code. On the other hand, it is all too easy to introduce errors, changes the commit history unnecessarily, produces big commits which are hard to read. I think there are arguments for change the code only if one touches it as part of a functional change.

For now, I would prefer *not* to have bulk commits for the other (possible) changes, for example in the switch statements or collectors.

PR #3262, opened 2023-04-05
Fix missing ING DiBa bankIdentifier in PDF-Importer

Nirus2000:
Fix missing ING DiBa bankIdentifier in PDF-Importer


buchen:
Intersting. I thought it is only "ing" nowadays - but apparently the PDF documents still contain the diba... let's for how long. In general, I think the bank identifier is not really needed anymore (it was an ideas to short circuit the detection).

PR #3272, opened 2023-04-08
Fix calculations for interest transactions with taxes

chirlu:
In #1190, it was made possible to enter taxes with interest transactions, but nothing was done to ensure that the backend calculations could actually deal with that. :-1: So here are a number of fixes to the `ClientIndex`, `WithoutTaxesFilter` and `ClientClassificationFilter`, with tests; all the tests failed in current PP. Only `ClientPerformanceSnapshot` happened to work correctly.

Some of the issues were described in https://forum.portfolio-performance.info/t/23589/2 and https://forum.portfolio-performance.info/t/6158/3.


buchen:
Thanks @chirlu for the fix. Thanks also for making it so easy to review because of the "precision commits" :-)

PR #3273, opened 2023-04-09
Update Contributing.md

Nirus2000:
With this PR, we are expanding the Contributings to include the following:

- The table of contents was created automatically. (Contributing.md)
- Formatting the table
- Change of the example file for postProcessing() from Comdirect to Targobank
- Removal of some spaces and typos
- Adding the Interactive-Flex-Query importer
- Adding the trading calendar
- Adding the label descriptions (variable names in *.properties) for translations

I did some work here and tried to find an intersection between the existing and the future source code.
Of course it's not quite perfect, but again another step towards standardization for other developers.

@chirlu @OnkelDok
If you have time, you can have a look at it and give hints, comments and changes.

@chirlu 
After the last discussion (see PR 3237) could you take the time to create a point(s) on the named missing information...!
That would be absolutely super and would help!

The last word to the Contrubing rules has of course the dear @buchen.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,32 +2,44 @@
 
 ## Contents
 
-- [Development Setup](#development-setup)
-	- [Install Eclipse](#install-eclipse)
-	- [Install Plug-ins](#install-eclipse-plugins)
-	- [Configure Eclipse](#configure-eclipse)
-- [Project Setup](#project-setup)
-	- [Source Code](#source-code)
-	- [Target Platform](#setup-target-platform)
-	- [Launch Application](#launch-portfolio-performance)
-	- [Build with Maven](#build-with-maven)
-- [Contribute Code](#contribute-code)
-- [Translations](#translations)
-- [Images, Logo & Colors](#images-logo-and-color)
-	- [Color code & status](#color-code-and-status)
-	- [Format and size](#format-and-size)
-- [PDF Importer](#pdf-importer)
-	- [Source Location](#source-location)
-	- [Imported Transactions](#imported-transactions)
-	- [Anatomy of a PDF Importer](#anatomy-of-a-pdf-importer)
-	- [Naming Conventions for Detected Values](#naming-conventions-for-detected-values)
-	- [Auxiliary classes](#auxiliary-classes)
-	- [Formatting of PDF Importer](#formatting-of-pdf-importer)
-	- [Test Cases](#test-cases)
-	- [Regular Expressions](#regular-expressions)
 
+  - [Development setup](#development-setup)
+    - [Install Eclipse](#install-eclipse)
+    - [Install Eclipse plug-ins](#install-eclipse-plug-ins)
+      - [Install from the simulatenous release update site](#install-from-the-simulatenous-release-update-site)
+    - [Configure Eclipse](#configure-eclipse)
+  - [Project setup](#project-setup)
+    - [Source code](#source-code)
+    - [Setup Target Platform](#setup-target-platform)
+    - [Launch Portfolio Performance](#launch-portfolio-performance)
+    - [Build with Maven](#build-with-maven)
+  - [Contribute code](#contribute-code)
+  - [Translations](#translations)
+    - [Label descriptions](#label-descriptions)
+  - [Images, Logo and color](#images-logo-and-color)
+    - [Format and size](#format-and-size)
+    - [Color code and status](#color-code-and-status)
+  - [Interactive-Flex-Query Importer](#interactive-flex-query-importer)
+    - [Source location](#source-location)
+    - [Imported transactions](#imported-transactions)
+    - [Naming conventions for detected elements](#naming-conventions-for-detected-elements)
+    - [Test cases](#test-cases)
+  - [PDF importer](#pdf-importer)
+    - [Source location](#source-location-1)
+    - [Imported transactions](#imported-transactions-1)
+    - [Anatomy of a PDF importer](#anatomy-of-a-pdf-importer)
+    - [Naming conventions for detected values](#naming-conventions-for-detected-values)
+    - [Auxiliary classes](#auxiliary-classes)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -357,19 +397,64 @@ Beside general good practices for regular expresions, keep in mind:
...
+
+
+## Trade calendar
+
+The trade calendar takes into account the weekends and regional holidays when there is no exchange trading. 
+The non-trading days of the stock exchange itself are also included.
+
+### Source Location Trade calendar
+
+Trade calendar Manager: `name.abuchen.portfolio\src\name\abuchen\portfolio\util\TradeCalendarManager.java`
```
chirlu, line 425: Backslashes are a Windows peculiarity.

CONTRIBUTING.md
```diff
@@ -357,19 +397,64 @@ Beside general good practices for regular expresions, keep in mind:
...
+The test cases include trade-free days of the trade calendar.
+
+The structure of the test cases is as follows:
+
+* Starting from the first day of the year with at least three test checks each, e.g. 01.01.20xx, 01.01.20xy, 01.01.20xz until 31.12.20xy of the respective year.
+* The regular trading-free days are followed by the one-time trading-free days.
+* The respective trade-free day to the test is named in the comment.
+* Samples 
+	* Trade calendar: [New York Stock Exchange](https://github.com/buchen/portfolio/blob/master/name\abuchen\portfolio\util\TradeCalendarTest.java) - see `testTradeCalenderNYSE()`
+	* Trade calendar: [London Stock Exchange](https://github.com/buchen/portfolio/blob/master/name\abuchen\portfolio\util\TradeCalendarTest.java) - see `testTradeCalenderLSE()`
```
chirlu, line 460: Definitely wrong in GitHub URLs.

CONTRIBUTING.md
```diff
@@ -357,19 +397,64 @@ Beside general good practices for regular expresions, keep in mind:
...
+	* `description` --> Display label of trade calendar, e.g., *New York Stock Exchange*
+* Days of weekend
+	* `weekend` --> Sets the default days for the weekend in the Trade Calendar, e.g., *Saturday and Sunday*
+
+
+### Test Cases Trade calendar
+
+The test cases include trade-free days of the trade calendar.
+
+The structure of the test cases is as follows:
```
chirlu, line 453: I don’t really understand this section, neither what it is trying to say nor the purpose.

chirlu:
Generally, I don’t care much about this file and mostly ignore it. I don’t think basic contributing information should be that detailed, and it will fall out of sync with the code anyway.

CONTRIBUTING.md
Nirus2000: done! 👍🏻

CONTRIBUTING.md
Nirus2000: done! 👍🏻

CONTRIBUTING.md
Nirus2000: I have rewritten the complete part of the Trade Calendar again.

CONTRIBUTING.md
```diff
@@ -193,28 +235,98 @@ Designation basic format as an example:
...
+
+An XML-compliant document is a file that is structured and formatted according to the rules of the eXtensible Markup Language (XML). It consists of a set of tags and attributes that contain information about the structure and content of the document.
+
+A valid XML document must meet the following conditions:
+
+* It must have a root element that contains all other elements.
+* Each element must be properly closed.
+* Attributes must be defined within a start tag and their values must be specified within quotes.
+* All elements must be properly nested, i.e. they must contain each other and not overlap.
+* There must be no comments or processing instructions outside the root element.
```
buchen, line 261: I don't think this belongs into the contributing document. This is a common sense explanation of XML. The contribution guidelines should be concise and short so that a new contributor is able to get the gist in 10 minutes

CONTRIBUTING.md
```diff
@@ -193,28 +235,98 @@ Designation basic format as an example:
...
+* `multiplier` --> The multiplier of the contract traded.
+* `proceeds` --> Calculated by multiplying the quantity and the transaction price. The proceeds figure will be negative for buys and positive for sales.
+* `reportDate` --> The date of the statement.
+* `tradeDate` --> The date of the execution.
+* `tradeTime` --> The time of the execution.
+* `dateTime` --> The date and time of the statement.
+* `netCash` --> Net cash is calculated by adding the proceeds plus tax plus commissions. (buy and sell)
+* `amount` --> Amount for account transactions
+* `currency` --> Currency of the total amount.
+* `fxRateToBase` --> The conversion rate from asset currency to base currency.
```
buchen, line 295: Again, I do not think the belongs into a contribution guide. If at all, it belongs to the IBFlex class. But even then, let's be careful to copy documentation which will soon become outdated. The link to the IBK guides sounds good enough to me

buchen: BTW, if you feel that is valuable, then we can also add a README.md to the "pdf" or "ibflex" folders

CONTRIBUTING.md
```diff
@@ -357,19 +477,71 @@ Beside general good practices for regular expresions, keep in mind:
...
+
+Holiday name: `name.abuchen.portfolio/src/name/abuchen/portfolio/util/HolidayName.java`
+
+Tests: `name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TradeCalendarTest.java`
+
+
+### Anatomy of a trade calendar
+
+The structure of the trade calendars is as follows:
+
```
buchen, line 524: Again, I think at most we can point out to common things to contribute (example: trade calendar). This just blows up the contributing guide. And if somebody does not understand the code, they'll also not be able to contribute meaningful code anyways

buchen:
I have merged the file now because I think it is okay for the time being. Every once in a while we should prune the file to keep it short and sweet.

PR #3277, opened 2023-04-11
Remove unuset translations

Nirus2000:
With this PR unused translations are removed.

- CSVColumn_Transferals
- CSVImportSecurityExists
- LabelApplyClientCalendar
- MsgErrorNoCrumbFound
- MsgErrorsConvertingValue
- MsgErrorUnexpectedStatusCode
- MsgErrorUnexpectedZipEntry
- MsgMissingResponse
- MsgNoResults
- MsgUnexpectedSymbol

---

I am puzzled by these two labels. One is in use and the other is not. 
Therefore, I have not made any changes here.
```
    public static String LabelTransferAccount; <-- not in used
    public static String LabelTransferPortfolio; <-- in used
```
---

Sorry for the sort order changes in `Messages.java`. Eclipse does this automatically for me.

---

I wanted to check the `Messages.java` in the UI, but I get this error and could not check it.
grafik


buchen:
> I wanted to check the Messages.java in the UI, but I get this error and could not check it.

I typically check with the File Search function.
All symbols you mentioned above are not used anymore (keep in mind that some "properties" files are used differently. Then there is no Messages class to check against)

buchen:
> Sorry for the sort order changes in Messages.java. Eclipse does this automatically for me.

I typically go into the Resource Bundle editor and delete the key. That deletes it in all files (including Messages) but does not reformat the file.

buchen:
fixed typo and added missing key into commit message --> merged

PR #3278, opened 2023-04-11
Add tests for ExtractorUtils.java

Nirus2000:
Add ExtractorUtilsConvertToNumberBigDecimalTest
Add ExtractorUtilsConvertToNumberLongTest
Add ExtractorUtilsDateParserTest

Remove the `value = value.trim().replaceAll("\\s", "")` from each importer and moved to the ExractorUtils.java.

The group separator for language format French is a space followed by the decimal separator comma.
[fr_FR](https://www.localeplanet.com/java/fr-FR/index.html) --> `1 234,56`

If the amount has the group separator space, the string can still have the following format.
[xh_ZA](https://www.localeplanet.com/java/xh-ZA/index.html) --> `1 234.56`

To simplify the two variants, we remove the spaces in the string, bypassing the generation of an additional identification via the locale.

The problem is that we can only pass one main formatting per importer.
e.g.
Importer: [PostfinancePDFExtractor.java](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostfinancePDFExtractor.java)
Test file:[ Kauf01.txt](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/postfinance/Kauf01.txt) vs. [Kauf04.txt](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/postfinance/Kauf04.txt)

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ExtractorUtilsConvertToNumberBigDecimalTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ExtractorUtilsConvertToNumberBigDecimalTest.java
@@ -1,0 +1,63 @@
+package name.abuchen.portfolio.datatransfer;
+
+import static org.junit.Assert.assertEquals;
+
+import java.math.BigDecimal;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Locale;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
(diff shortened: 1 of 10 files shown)


buchen:
Looks good (at first reading in the browser only), but I wonder shouldn't all of this implicitly tested by all the PDF imports anyway? And if not, why is there code in the ExtractorUtil which is not used by extractors. I don't want to curb your enthusiasm... but code also comes with future maintenance costs

Nirus2000:
I have expanded the pull request, changed the top post to better explain the tests and the changes.

PR #3291, opened 2023-04-17
Small changes of measurement tool

OnkelDok:
First of all: Thank you @buchen, for taking over the measurement tool draft and made the improvements. This first version in PP looks great (in GUI and code).

Nevertheless, I have some suggestions for changes:
- 9cf80834372fb1d1475f282a14301d8898eb9928: In the current version on master branch, when de/activiating the measuretool, the button still remains inactive. Maybe there are more elegant solutions to solve this.
- 324929b438764b9f7620ab719a6e95567dcbdefa: It's only some refactoring and a non-functional change to reduce code complexity (eclipse marked the function ``handleEvent`` as too complex).

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
@@ -4,9 +4,11 @@ import java.time.Instant;
 import java.time.LocalDate;
 import java.time.ZoneId;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
 
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.ToolBarManager;
+import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.PaintEvent;
 import org.eclipse.swt.graphics.Color;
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
```diff
@@ -91,22 +94,34 @@ public void setColor(Color color)
...
-                            a.setImageDescriptor(isActive ? Images.MEASUREMENT_ON.descriptor()
-                                            : Images.MEASUREMENT_OFF.descriptor());
+                            ImageDescriptor image = isActive ? Images.MEASUREMENT_ON.descriptor()
+                                            : Images.MEASUREMENT_OFF.descriptor();
+
+                            // update image of button
+                            if (doUpdateButton)
+                                buttons.forEach(b -> {
+                                    b.setImageDescriptor(image);
+                                    b.setChecked(isActive);
```
buchen, line 121: This ```isChecked``` has no effect unless we set the style bit ```IAction.AS_CHECK_BOX``` on the action.

If we do this, then it looks like this:

Bildschirmfoto 2023-04-18 um 18 23 22

I do not like that form the UI/UX design (and the only reason I have enabled it for the time periods "1Y", "2Y" etc. is that it is not possible to change the text color).

However, what I do not know if there is any benefit for screen readers. Do you know?

For now, I will remove the setChecked and merge the proposal. LGTM

OnkelDok: > This `isChecked` has no effect unless we set the style bit `IAction.AS_CHECK_BOX` on the action.

Oh, I have not consiered that, thank you for the clarification, @buchen.
I don't know about if screen reader can handle this.

By the way, I was thinking of looking for another icon. I think it looks like a hair comb. Because the lines of the ruler look like gaps.

PR #3294, opened 2023-04-18
Modify Interactive Broker to support new transactions

Nirus2000:
https://forum.portfolio-performance.info/t/interactive-brokers/276/241 
https://forum.portfolio-performance.info/t/interactive-brokers/276/244

Add new transaction "WAR" for Warrants

Improve some notes

Fix typo contributing rules for IB-Flex-Query Importer


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -263,16 +276,35 @@ else if (this.ibAccountCurrency.equals(s.getCurrencyCode()))
             setAmount(element, transaction, amount);
 
             // Set note
-            transaction.setNote(element.getAttribute("description"));
+            if (!transaction.getType().equals(AccountTransaction.Type.DIVIDENDS))
```
pfalcon, line 279: @Nirus2000 : What's the reason for this change (not recording original description for dividend transaction) ? Nothing in the commit message explains that. IMHO, this description is one of the useful which could be recorded, as it includes high-precision dividend per share, type of dividend and even ISIN (not easily seen in IBKR data otherwise).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -263,16 +276,35 @@ else if (this.ibAccountCurrency.equals(s.getCurrencyCode()))
...
-                transaction.setNote("Trade-ID: " + element.getAttribute("tradeID") + " | " + transaction.getNote());
+            if (transaction.getNote() == null)
+            {
+                if (!element.getAttribute("tradeID").isEmpty() && !element.getAttribute("tradeID").equals("N/A"))
+                    transaction.setNote("Trade-ID: " + element.getAttribute("tradeID"));
+            }
+            else
+            {
+                if (!element.getAttribute("tradeID").isEmpty() && !element.getAttribute("tradeID").equals("N/A"))
+                    transaction.setNote("Trade-ID: " + element.getAttribute("tradeID") + " | " + transaction.getNote());
```
pfalcon, line 291: @Nirus2000: Another note: it's certainly useful to record tradeID/transactionID to cross-reference to the original datasource in case it ever needed, but I don't think that's the most useful/most frequent information a user may want to see. Thus, it would make sense to put it at the end of the note, not at the beginning.

Nirus2000: This is a personal view and you can change the note yourself at any time. 
The ID's should be at the beginning to keep the column widths small for country descriptions.
You can identify what it is by the transaction type.

Example:
"Transaction-ID: 12345679 | USD IBKR MANAGED SECURITIES (SYEP) INTEREST FOR NOV-2022"
Transaction type is interest

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
pfalcon: @Nirus2000 : Thanks for promptly looking into this!

> but you are the first of many thousands of users to notice this. 

Well, that's fairly recent change for many to try it. Then not everybody will notice it (at least for a while). Then of those who notice it, only a few will report. I for one wanted to as about it for a week and kept fogeting...

> The importer was revised because it did not include many new features.

Sure, that work is much appreciated. My question was specifically why `if (!transaction.getType().equals(AccountTransaction.Type.DIVIDENDS))` line was added, i.e. why description was started to not be recorded for dividends and only dividends.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
pfalcon: > This is a personal view and you can change the note yourself at any time.

Sure, and I'm not calling for an immediate change, just providing feedback and discussion of the work you're doing, to both show appreciation and in the hope to make further small improvements.

> The ID's should be at the beginning to keep the column widths small for country descriptions.

Hmm, I'm not sure I understand this. Let me elaborate my point. If user selects to show the Note column, it's width certainly will be limited (by other columns or physical screen width), so given your example, a user could see for example "Transaction-ID: 12345679 | USD". I'm not sure that seeing a transaction ID per se would be useful to end user most of the time. IMHO, it would be better to show "USD IBKR MANAGED SEC", much more informative.

PR #3295, opened 2023-04-18
Added a missing space before the benchmark suffix

Fruchtzwerg94:
grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dataseries/DataSeries.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dataseries/DataSeries.java
@@ -124,7 +124,7 @@ public final class DataSeries implements Adaptable
 
     public String getLabel()
     {
-        return isBenchmark() ? label + Messages.ChartSeriesBenchmarkSuffix : label;
+        return isBenchmark() ? label + " " + Messages.ChartSeriesBenchmarkSuffix : label; //$NON-NLS-1$
     }
 
     public void setLabel(String label)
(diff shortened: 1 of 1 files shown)


Nirus2000:
If specific spaces are required between special characters and a string part for the language (e.g. French), then this should be adjusted in the *.properties and not hard in the source code.

chirlu:
So, we need a way to allow trailing/leading spaces in some cases while keeping them out when they are the unintentional result of sloppy editing. Perhaps an intentional space at the beginning or end could be indicated on POEditor by an underscore or in some other explicit way? When updating strings from POEditor, it would have to be automatically converted.

Fruchtzwerg94:
I agree. Sorry, I was not aware that this was caused by a previous commit and effects also other suffixes.
Shall I close this PR or is it better to leave it open until the issue is resolved in another way?

Nirus2000:
Please keep open... but convert to draft pull-request.

buchen:
As @chirlu pointed out, in recent change we removed all leading or trailing spaces in order to generally check that labels are properly formatted (when translators add terms/translations on poeditor.com, it is apparently easy to add spaces and line breaks undetected). As such, you proposal is completely in line and fixes the missing space. I will merge it.

@chirlu writes:
> Generally, I’d think it is preferable to not have the whitespace hardcoded, but have it in the strings: There may be languages for which a space before the suffix is undesirable.

I also agree to this point. I wonder if the better solution is to *not* translate the suffix, but translate the complete term, say ```{0} (suffix)```. Then we have both things: no space concatenation in the code, translators can - depending on the language add a space as necessary.

buchen:
And: welcome a new contributor :-)

PR #3299, opened 2023-04-19
Small changes of measurement tool

Nirus2000:
- Changing the measurement tool icon
- Add separator in the toolbar

With the addition of separators, the tools are clearly divided into: 
Tools | TimeSeries | Settings

Furthermore, the crosshairs can then also be added to the tools.


Nirus2000:
@OnkelDok

I will use this PR to mention the following problems with the measuring tool.

1.) The labels are overlaid.
grafik

2.) When displaying with marker lines, the label is also overlaid
grafik

3,) Draw speed
I have an 8 Core CPU with 32GB (AMD Ryzen™ 7 5800X, processor).
The more that has to be drawn, e.g. values etc., the more speed is lost when drawing.
I have the fear that there will be increased excitement and grumbling here if the measurement tool does not work very smoothly. 

https://user-images.githubusercontent.com/45203494/233150591-731da1cb-fda4-47fc-848c-3adf5a35eb80.mp4

OnkelDok:
@Nirus2000 My first guess why measure tool is drawn in background is because the measure tool is not the last registered ``customPaintListener``.
I will try something around to handle this.

Ok, the performance drop is bad. I don't know if we can keep the current drawing with ``customPaintListener``. Or if we need to draw the measurement tool in a different way. Because it seems the ``customPaintListener`` is not designed for "realtime" drawings.

Nirus2000:
@OnkelDok 
I've installed your "Crosshair" for fun and tried... if that is still active, then nothing works. 👎🏻

OnkelDok:
> I've installed your "Crosshair" for fun and tried

What do you mean with "your" Crosshair? Don't know that I provided a crosshair which can be installed. But I assume I misunderstand your comment. Can you please explain with more detail?

OnkelDok:
> [#3004 (comment)](https://github.com/buchen/portfolio/pull/3004#issue-1401579833) this one...

Ah ok, I understand. And what is nor working? Do you have a screenshot or video? Maybe it's due to some logical changes/improvements @buchen has made.

buchen:
> Regarding the perfomance, maybe we have to draw on a separate canvas/area and not on the chart plot area. I believe the tool tip is also drawn as separate element. When I have time the next days, I'll take a look at it. Maybe I'll come up with something.

What I can think of: the chart is drawn into a an image. And then the drawing of the measurement is taking the image, drawing the measurement, and drawing it on the plot area. But of course this is a substantial change - possible also to SWTChart.

I believe ´the tooltip is a separate UI element of the operating system - a different "window". Then the OS takes care.

buchen:
BTW, just to mention on my Mac (Apple M1, 32GB) the painting is super smooth.

https://user-images.githubusercontent.com/587976/233770530-34beb6fe-505d-4bc4-b177-2593ff8f2d85.mp4

buchen:
Okay, I merged this PR.

Unfortunately, this closes the PR and hence "closes" the performance discussion. Shall we move it to a issue ticket (I assume it is not easy to solve)

OnkelDok:
> BTW, just to mention on my Mac (Apple M1, 32GB) the painting is super smooth.
> Bildschirmaufnahme.2023-04-22.um.09.43.26.mp4

@buchen can you please test again with a bigger time interval? Start date around 2005, like in video from @Nirus2000.


> Unfortunately, this closes the PR and hence "closes" the performance discussion. Shall we move it to a issue ticket (I assume it is not easy to solve)

Yes, a separate ticket would be good.

PR #3301, opened 2023-04-19
Let measure tool draw top most

OnkelDok:
Bugreport here: https://github.com/buchen/portfolio/pull/3299#issuecomment-1515096166

Reason: measure tool was not the last registered PaintListener. Also the measure tool was registered as ``ICustomPaintListener`` instead as ``PaintListener``. It seems the former will be notified first, and then the latter.

grafik


buchen:
Thanks for the bugfix @OnkelDok.

However, I do not like that now **all** charts need to know when exactly to initialize the measurement tool. That should be something internal - between the measurement tool and the timeline chart.

We have two types of PaintListener. The SWT ones (where SWT is managing the order) and the CustomPaintListener (where SWTChart is managing the order and they are painted within its own SWT PaintListener). I added now a function #addPlotPaintListener to the TimelineChart that makes sure the paintListener are added as custom paint listener and therefore will always be executed *before* the measurement tool

PR #3305, opened 2023-04-20
Fix sonar issues related to class Optional (java:S3655)

NSCarvalho:


buchen:
Hi @NSCarvalho  - looks good to me (@Nirus2000 to comment on the pull request as well).

Maybe we should rename the method slightly to make sure that the tax units are compared (as you pass in the transaction, it could also be the transaction amount). Or pass the Optional units.

Nirus2000:
My plan is that I want to touch the ComDirect Importer again in the next weeks to set the PostProcessing() analogous to the Targobank PDF Importer.

There are still 14 open PDF debugs on my ToDo list. 

The tests will also be done analog to the Targobank PDF importer. 

The goal is:
1. every PDF should be treated exactly as it was created for on the part of the bank.
2. the problem is here also that in the ComDirect still some "manipulated" PDF debugs are and the tests say "Ok", but are not actually. 

3. also tax documents were misused for dividend transactions or sales. 

4. also withholding taxes appear in the regular dividend transactions, but no longer in the tax treatment.

By the way, this was also the reason why I took ComDirect out of the Contributings as an example for PostProcessing and replaced it with Targobank.

If you don't agree with this, please give me a sign.

buchen:
First, I think the proposal by @NSCarvalho does not contradict any plans with the importer; it just fixes the access on the Optional without checking whether this is present. Therefore I will merge the change.

> the problem is here also that in the ComDirect still some "manipulated" PDF debugs are and the tests say "Ok", but are not actually.

> also tax documents were misused for dividend transactions or sales.

Okay, I was not aware of those two issues. Let me create an issue to keep track of these Comdirect issues (once I close this PR, then the discussion is hidden from sight)

buchen:
FYI, I have added the missing methods to the Money class to compare to values with 8f8f728ae92867e6000ae2a7dbe5c1ea336b4897
The code was a little confusing, because the comment was checking for a1 < a2 but the code was check for a2 >= a1. Which is the same, but tripped my brain (also because the proposed method then was is called isA1LowerThanA2

buchen:
Okay, I have adopted the contribution (and because it was quite changed, I put the two of us as co-authers there). The code really tripped me because the comment was indicating "less than" but it was actually checking for "less or equal to". And indeed the test cases were failing if the both units had a value of zero. I have now used the newly added methods as well - so code and comment are in sync.

I'll change one more thing and remove the double negation - replace ```!a1.getUnit(Type.TAX).isPresent()``` by ```a1.getUnit(Type.TAX).isEmpty()```.

PR #3307, opened 2023-04-21
Modify FlatEx PDF-Importer to support new transactions

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/197

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/fintechgroupbank/FinTechGroupBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/fintechgroupbank/FinTechGroupBankPDFExtractorTest.java
@@ -804,12 +804,12 @@ public class FinTechGroupBankPDFExtractorTest
         List<Item> results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "biwAGKontoauszug01.txt"), errors);
 
         assertThat(errors, empty());
-        assertThat(results.size(), is(1));
+        assertThat(results.size(), is(2));
         new AssertImportActions().check(results, CurrencyUnit.EUR);
 
         // check transaction
         Iterator<Extractor.Item> iter = results.stream().filter(TransactionItem.class::isInstance).iterator();
-        assertThat(results.stream().filter(TransactionItem.class::isInstance).count(), is(1L));
+        assertThat(results.stream().filter(TransactionItem.class::isInstance).count(), is(2L));
 
         Item item = iter.next();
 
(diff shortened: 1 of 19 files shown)


buchen:
Hi @Nirus2000,

as always, thanks for the contribution.

I am not sure I understand the logic of this NonImportable item. The message says: "Transaction is processed through another document." Shouldn't we just return nothing here - and then the other importer processes the document. Or is the transaction never processed here? Then what is the reason to not import the transaction here?

And: what is it about this Kontoauszug03? Was is accidentally a DeGiro document in Flatex? And the Flatex importer actually detected it?

Nirus2000:
Hello @buchen

**First point:**

There was a small sorting error here.
Since FinTechGroupBank is merged several times,
we sort the documents by bank.
This is determined in the first line of the PDF debug.
- flatex Bank AG
- flatexDEGIRO Bank AG 
- biw AG
- FinTech Group Bank AG

FlatExDegiroKontoauszug02.txt is the new...

FlatExKontoauszug03 was wrongly assigned and is correct FlatExDegiroKontoauszug01.txt and vice versa....
These were simply twisted.

Sorry, for the short confusion, but without a correct sorting after the 4th merger it does not work... this is analogous also as with the Degiro... which actually also belongs to FinTechGroup, but is separate.
In Degiro we sort according to the national languages....

---

**Second point:**

        // @formatter:off
        // This transaction is not imported.
        // The exchange rate is not shown once the security is held in foreign currency.
        //
        // We import these fees via the depot account statement.
        // @formatter:off

We identify the document so we don't create an error.
If the document is read in one at a time, without any other PDF's we create an "empty" table... this is then rather more confusing.

This document is "never" processed.... but should give the hint that this is captured via another document, namely the Depot Statement Transaction.
I wanted to keep the hint general, but can specify it.

However, if you want, you can specify it so that it can be used for other importers and documents that we will "never" import.
The message I would customize that reference the other document that processes the transaction.

Make your choice ;-)

Here is my explanation:
https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/198

buchen:
> This document is "never" processed.... but should give the hint that this is captured via another document, namely the Depot Statement Transaction.
> I wanted to keep the hint general, but can specify it.

Okay, understood. Personally, I would keep the hint very specific and explain why this particular transaction is not imported. The hint in the code says it all. Otherwise I think users will start threads in the forum to understand why.

I have rephrased the messages. Please have a look.

I also realize it is a pain in the back to externalize strings in these PDF extractors (because there are so many to ignore). I opened the dialog, then selected all (Ctrl-A), clicked internalize, and then had to search for the one label to translate. How do you do this?

buchen:
merged

PR #3310, opened 2023-04-23
Extended file import filter mask with uppercase *.PDF

EtherGraf:
On my linux system with XFCE Environment the file system is case sensitve. 
When I try to import pdfs then only '*.pdf' are shown in the file selection dialog, not the files with the uppercase '*.PDF'. 

This patch extends the file selection mask with the uppercase variants.


buchen:
Thanks. I have updated all other "open file" filter extensions as well.
The filter extensions are platform specific...

PR #3323, opened 2023-05-04
Enable PDF Import for Merkur Privatbank that also uses the DZ Rechenzentrum format

cubinet-code:
Adding the BankIdentifier addBankIdentifier("Am Marktplatz 10 · 97762 Hammelburg") enables the import of the MERKUR PRIVATBANK PDFs that uses the same format as the DZBankGruppePDFExtractor.

Unfortunately, there is no bank name in the extracted text, so I had to use the street address instead, just like the code line before.

By the way, thank you very much for the detailed "How to develop" instructions. I never used Ecplise or Java, but it worked right away and I was up and running within 30 minutes.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dzbankgruppe/DZBankGruppePDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dzbankgruppe/DZBankGruppePDFExtractorTest.java
@@ -387,50 +387,6 @@ public class DZBankGruppePDFExtractorTest
     }
 
     @Test
-    public void testWertpapierKauf09()
-    {
-        // Merkur Privatbank
-        DZBankGruppePDFExtractor extractor = new DZBankGruppePDFExtractor(new Client());
-
-        List<Exception> errors = new ArrayList<>();
-
-        List<Item> results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Kauf09.txt"), errors);
-
-        assertThat(errors, empty());
-        assertThat(results.size(), is(2));
-        new AssertImportActions().check(results, CurrencyUnit.EUR);
-
-        // check security
-        Security security = results.stream().filter(SecurityItem.class::isInstance).findFirst()
-                        .orElseThrow(IllegalArgumentException::new).getSecurity();
-        assertThat(security.getIsin(), is("IE00BJ0KDQ92"));
-        assertThat(security.getWkn(), is("A1XB5U"));
-        assertNull(security.getTickerSymbol());
-        assertThat(security.getName(), is("XTR.(IE) - MSCI WORLD REGISTERED SHARES 1C O.N."));
-        assertThat(security.getCurrencyCode(), is(CurrencyUnit.EUR));
-
-        // check buy sell transaction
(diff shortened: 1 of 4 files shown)


chirlu:
A test document would be desirable.

cubinet-code:
Thank you. I have added a test case for this document.

Nirus2000:
Hello @cubinet-code 

No, this is not correct and should not be included in the master.

**Reason:**
Merkur Privatbank does not belong to the DZ-Group!

No banks should be mixed, because later changes are very complicated and insanely time-consuming.
You also assume that the document design is the same, and it may be a coincidence that the two look "similar".

If you look at Merkur's imprint, they have branches in other cities as well. (Düsseldorf, Fulda etc.)

So this is the "Merkur Privatbank KGaA". (see [Contributing rules](https://github.com/buchen/portfolio/blob/master/CONTRIBUTING.md#pdf-importer))
You have to create a separate PDF importer for this.
_The importer is to be created based on the test cases and should only contain functions that are also validated by the test cases._

Regards
Alex

cubinet-code:
> **Reason:** Merkur Privatbank does not belong to the DZ-Group!
> 
> No banks should be mixed, because later changes are very complicated and insanely time-consuming. You also assume that the document design is the same, and it may be a coincidence that the two look "similar".
> 
> If you look at Merkur's imprint, they have branches in other cities as well. (Düsseldorf, Fulda etc.)
> 
> So this is the "Merkur Privatbank KGaA". (see [Contributing rules]
Dear Alex,

Thank you for your feedback.

Please compare [here](https://www.merkur-privatbank.de/rechtliches/nutzungsbedingungen.html).

> Für wertpapierbezogene Informationen arbeitet die Bank mit der **DZ Bank AG**, der Atruvia AG und anderen Anbietern zusammen.

I have also validated with them personally that they 1:1 use the same Services as Volksbank/Union Investment/VR-Bank/GLS Bank, etc. which are already covered by this importer. They use the same PDF forms with only their address header information replaced just like with the other banks covered in DZBankGruppePDFExtractor.

Of course it would be possible to create a seperate importer class, but that would just lead to a duplicated code base.

The test case has been added as well.

please let me know your thoughts.

best regards,

Ollie.

Nirus2000:
I understand what you wrote.
This is not about the software, but about the banks.

Maybe you read over that. (see 1st line in [Contributing rules](https://github.com/buchen/portfolio/blob/master/CONTRIBUTING.md#pdf-importer))
>Importers are created for each supported bank and/or broker. 

You have to create a separate PDF importer for this.

cubinet-code:
Sure. I will create a separater importer then.

I was just confused, as [this](https://github.com/buchen/portfolio/blob/master/CONTRIBUTING.md#pdf-importer) clashes with the fact that this importer already covers multiple banks (Volksbank/Union Investment/VR-Bank/GLS Bank).

best regards,

Oliver.

Nirus2000:
> I was just confused, as [this](https://github.com/buchen/portfolio/blob/master/CONTRIBUTING.md#pdf-importer) clashes with the fact that this importer already covers multiple banks (Volksbank/Union Investment/VR-Bank/GLS Bank).

I understand. 
The difference is that these banks belong "directly" to a banking group. 


Merkur only works with DZ Bank AG for securities, but not with the banking group. 
> Für wertpapierbezogene Informationen arbeitet die Bank mit der DZ Bank AG, der Atruvia AG und anderen Anbietern zusammen.

PR #3324, opened 2023-05-05
Enhancement ex-date for dividende transaction (The final chapter StepByStep)

Nirus2000:
_1st pull request_
- Adding the ex-date model
- Adding the translations + correction
- Added the visibility of DatePicker
- Adding the DetectDuplicatesAction for Ex-Date
- Test of the InsertAction + extension

---

Hello @buchen and everyone else,
started several times, never finished. :rofl:

Since I have a little more time at the moment (due to health), I have now decided to implement the "ex-date" for dividend transactions, which is requested by the community and called for in many issues.
So that this happens in an orderly manner, this will be done StepByStep.

@OnkelDok @chirlu
Can you have a look if I have forgotten something here?
Also it would be great if you can give me some stick points so that we implement this feature correctly and completely.

Until @buchen gives an OK. to the "ex-date" model, all far steps stand on "OnHold/InDraft".

-----

### To-Do List:
- [x] Adding the ex-date model
   - [ ] Wait for release of the model (@buchen)
- [ ] Decision of the implementation variant
- [ ] ????
- [ ] ????
- [ ] Implementation in the PDF importer <-- This is the very last step
- [ ] Implementation in the CSV importer <-- This is the very last step
- [ ] Implementation in the IB-Flex importer <-- This is the very last step
- [ ] Link all open issues and draft pull requests that can be closed
- [ ] Finish "The final chapter" ex-date for dividende transaction


Nirus2000:
Hello @chirlu 
I have also thought about this and see two variants:

**Variant 1:**
The user will have the option to enable "Calculation with ex-date" (example) in the main settings. 
This will then be applied to all calculations or if disabled then the current variant will apply. 
It must be a global setting. 
So depending on the selection, we look at the appropriate date or ex-date for dividend transactions.

**Variant 2:**
We introduce the ex-date as an attribute, which is a pure information value and has no impact on the real transactions, calculations or performance.

What are your thoughts on this?

buchen:
I have to admit I cannot fully wrap my head around how to use the ex date best.

Let's recap what we have today
* transactions with the payment date
* dividend events with ex-date and payment date (in the event section; from DivvyDiary; not yet manually editable - see discussion #3248)

I see the dividend events helpful because
* one could have a better benchmarks (right now we only compare the prices - but exclude the dividends - kind of "adjusted price" would be needed
* one could predict the upcoming payments (because the shares at the moment of the ex date define the payment)
* have proper dividend returns (only using the payments is very misleading)

For all of this we do not need the ex date in the transaction.


> The user will have the option to enable "Calculation with ex-date" (example) in the main settings.

I am against two ways to implement the performance calculation. It will be very hard to explain to users. If the exDate is maintained then it would be to be used for the performance calculation.

But I am not sure how to implement this. I tend to think it would have to be "pseudo accounts" like @chirlu suggested. At the "ex date" one would add the money to the account, at the payment date the money would be removed. The pseudo account would be performance relevant. Maybe we would call it "accrued dividends".

Besides the challenge on how to display that to the user, I also see the challenge that the performance is retroactively "changed". Typically the dividend statements are important via PDF and only available once the payment actually happened. At that time, we correct the performance of the last X weeks (because the accrual was earlier).

I'd have to do some prototyping/thinking about it, but maybe the AccountSnapshot would have to have multiple AccountPositions similarly to the PortfolioSnapshot having multiple SecurityPositions.

My gut feeling: this is a big complicated change. Before we do this, would it make sense to *manually* check if this actually makes a meaningful difference in the performance calculations?

chirlu:
> Let's recap what we have today: transactions with the payment date

A number of people (in particular Thomas, and also me, among others) [strongly recommend](https://forum.portfolio-performance.info/t/buchung-von-dividenden-am-ex-tag-oder-bei-gutschrift/1718) using the ex-dividend date as the transaction date. That’s also what I personally do in my portfolio; I put a note into the transaction giving the payment date, for reference.

With that, the performance calculation is completely correct: On the same day the money disappears from the company (drop in value of the shares), it reappears on the associated account. Performance-wise, there is no difference if the dividend amount is lying around on the definite account somewhat early or on an “accrued dividends” special account.

The issues with that are 1) that many people use the PDF importers to enter dividend transactions and that some? many? all? PDF importers enter the payment date instead of the ex-dividend date, and 2) that the account balance is out of sync with reality for a few days. Issue 1 would obviously best be dealt with by fixing the affected importers. Issue 2 is less relevant with German stocks, as the payment generally happens just two days after the ex-dividend date; however, for US stocks, among others, the interval can be much longer. So I see that, in those cases, it can be confusing or simply “look wrong”.

> Variant 2:
> We introduce the ex-date as an attribute, which is a pure information value and has no impact on the real transactions, calculations or performance.

I could imagine this, but the other way round; i.e. adding an informational field “payment date”, while the regular date field would hold the ex-dividend date. Essentially, this is the same as putting the payment date into the note. As I said above, the (unmodified) performance calculation that simply ignores the new field would remain correct.

Regarding Variant 1, I fully agree with @buchen:
> I am against two ways to implement the performance calculation.

In particular since using the payment date (even as an option) would produce *wrong* results.

chirlu:
So, in summary, there would be three different levels of implementation that might make sense:

* Leave as is; the payment date can be put into the note.
* Add a new “payment date” field to dividend transactions, but only informational.
* Add a new field and consider it when calculating account balances. The balance increases only on the payment date; from the ex-dividend date to before the payment date, the money is on a virtual account “accrued dividends” – either a collective virtual account for all dividends, or possibly separately for each stock.

Obviously, the complexity increases from top to bottom, as does the potential usefulness. The question is how worthwhile each option is.

Nirus2000:
Hello @chirlu @buchen 
Very good thoughts!

In the course of setting up the entire importer, Andreas clearly communicated that the valuta date should be used for the dividend transactions and not the ex-date. Likewise, the ex-date is not shown or avoided in every dividend transaction (PDF + CSV + IBFlex). This excludes the correction of all importers.

Implementing an additional field in which the ex-date can be entered, i.e. also manually if the ex-date is not available, I see as a possible solution. This would then be an optional date field.

Thosvi:
Hello together,

To demonstrate the difference between ex and payment date I prepared 3 PP files with a japanese security, where the gap between these dates is approximately 3 months. Each file uses EUR as main currency and the security is denominated in JPY.

Unfortunately I cannot upload these files here. If you are interested, let me know, how I can share them. Attached are screenshots of the default performance dashboard.

In the file payDate the dividend payments use the payment date as transaction date with the exchange rate from my bank
payDate

In the file exDate an additional account for dividends is used. This account is denominated in the main currency. Dividends are processed in two steps:
- On ex date the dividends are transfered to this extra account using the payment data from my bank including taxes
- On payment date the amount in transferred from the extra account to the cash account
exDate

The file exDate_forex also uses an extra account for dividend payments, but in this case it is denominated in the currency, in which the dividends are declared (which is not necessarily identical to the denomination of the security or payment). 
Dividends are processed in three transactions:
- On ex date dividends are transferred to the extra account without taxes
- on payment date the gross amount is tranferred from the extra account to the cash account using the exchange rate from my bank
- on payment date an additional transaction for tax deduction is used
exDate_forex

This last one would be my favorite method. Currently it has an issue, which prevents me from using it: The forex gains/losses between ex date and payment date on the extra account are not linked to the security. On the portfolio level it is OK, but the security level it does not reflect the real world.


Kind regards

PR #3338, opened 2023-05-08
Add dependency management via Github's dependabot

sheim-dev:
This PR adds dependency management via Github's dependabot to Portfolio Performance.

With this PR, pull requests for outdated dependencies of PP will we raised automatically. Those can subsequently be merged at the maintainer's convenience.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


buchen:
Thanks @sheim-dev for the proposal.

One question: Most of the project dependencies are defined inside the target platform ([for example here](https://github.com/buchen/portfolio/blob/master/portfolio-target-definition/portfolio-target-definition.target#L55-L76)) - looking at the documentation of dependabot, it is not clear if that is picked up. Do you know? In the pom.xml are only the build plugins.

Morpheus1w3:
@buchen 
If the file is named `pom.xml` the answer is yes, otherwise not.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#maven

> Dependabot doesn't run Maven but supports updates to pom.xml files.

auchri:
I don't think so: https://github.com/dependabot/dependabot-core/issues/4682

PR #3339, opened 2023-05-08
Enable/disable measurement tool by GUI setting "experimental features"

OnkelDok:
As I wrote [here](https://github.com/buchen/portfolio/issues/3311#issuecomment-1537383238): it's easier to enable/disable the measurement tool from the GUI.


buchen:
Good point. I did not use the expermintal feature flag because @Nirus2000 mentioned that anyway lots of folks have that enable by default (hence it is not hiding much). But I agree it makes it more accessible (and we want feedback)

PR #3342, opened 2023-05-09
Improve the ListSelectionDialog to speed up the outputs

Nirus2000:
With larger table contents, the speed is reduced considerably.

**CoinGecko before**

https://github.com/buchen/portfolio/assets/45203494/35781b02-c8ac-42ed-9860-fb49a7b107cd

**CoinGecko after**

https://github.com/buchen/portfolio/assets/45203494/95baa4a9-4625-488a-83b1-3fc593823772

snapshot

**VisualVM 2.16**
[snapshot-1683638272212.zip](https://github.com/buchen/portfolio/files/11431526/snapshot-1683638272212.zip)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/ListSelectionDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/ListSelectionDialog.java
@@ -172,11 +172,14 @@ public class ListSelectionDialog extends Dialog
 
         elementFilter = new ElementFilter();
 
-        int style = SWT.BORDER | SWT.FULL_SELECTION;
+        int style = SWT.BORDER | SWT.FULL_SELECTION | SWT.VIRTUAL;
         if (isMultiSelection)
             style |= SWT.MULTI;
+
         tableViewer = new TableViewer(tableArea, style);
+        tableViewer.setUseHashlookup(true);
         CopyPasteSupport.enableFor(tableViewer);
+
         final Table table = tableViewer.getTable();
         table.setHeaderVisible(false);
         table.setLinesVisible(false);


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java
```diff
@@ -113,13 +113,19 @@ public void menuAboutToShow(IMenuManager manager)
                         @Override
                         public String getText(Object element)
                         {
+                            if (element == null)
+                                return null;
```
buchen, line 117: How can this element be null?!?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/ListSelectionDialog.java
```diff
@@ -172,11 +172,14 @@ protected Control createDialogArea(Composite parent)
 
         elementFilter = new ElementFilter();
 
-        int style = SWT.BORDER | SWT.FULL_SELECTION;
+        int style = SWT.BORDER | SWT.FULL_SELECTION | SWT.VIRTUAL;
         if (isMultiSelection)
```
buchen, line 176: At least for other VIRTUAL tables, I also needed this config:

```java
tableViewer.setUseHashlookup(true);
```

Otherwise there were strange error messages on some platforms. So I propose to add it here as well.

buchen:
BTW, as we discussed, the dialog is super fast on my machine. Somehow I cannot think this is a macOS vs. Windows thing...

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java
Nirus2000: äh... i think i have an other problem...
https://bugs.eclipse.org/bugs/show_bug.cgi?id=570553

Please remove this lines

Nirus2000: The variable `query` in the CoinGecko search provider is null and thus an error is generated, causing "`item.getSymbol()`" to generate an error.

Nirus2000: I can't reproduce the error again at this time, but it was written to the error log on my end.
Currently, however, this is flooded by an error after update of Eclipse. I would have made a screenshot.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=570553 👎🏻

buchen (changes requested):
I think we should add the setUseHashlookup

And let me understand why the items can be null

Nirus2000:
That was my mistake...
I somehow had a bug in Eclipse (https://bugs.eclipse.org/bugs/show_bug.cgi?id=570553).
I have reinstalled Eclipse on me and now the error is also gone.

PR #3343, opened 2023-05-09
Add PWP Leeway as new rest api quote provider

Nirus2000:
Add PWP Leeway as new rest api quote provider
Add new column currencyCode in SecuritySearchProvider

https://leeway.tech/

grafik


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewaySearchProvider.java
```diff
@@ -0,0 +1,186 @@
...
+        public String getSource()
+        {
+            return NAME;
+        }
+
+        @Override
+        public boolean hasPrices()
+        {
+            return true;
+        }
```
Nirus2000, line 111: I am not sure if I implemented this part correctly.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/SecuritySearchProvider.java
```diff
@@ -46,7 +51,8 @@ default boolean hasPrices()
 
     public enum Type
     {
-        ALL(Messages.LabelSearchAll), SHARE(Messages.LabelSearchShare), BOND(Messages.LabelSearchBond);
+        ALL(Messages.LabelSearchAll), SHARE(Messages.LabelSearchShare), BOND(Messages.LabelSearchBond), //
+        ETC(Messages.LabelSearchETC), ETF(Messages.LabelSearchETF), FUND(Messages.LabelSearchFund);
```
buchen, line 55: I propose **not** to add additional labels ETC, ETF, FUND, etc.

First, it is not used to filter the resource anyway (see the screenshot where I picked to search for "ETF" but it returns everything. Also the Leeway implementation ignores these labels for searching (you can set an arbitrary type in the result if you just want to display it - continue to use the translations, but to not add the enum here).

Bildschirmfoto 2023-05-10 um 19 45 14

Second, if we add new types, we should at least support the primary quote feed "Portfolio Report". Otherwise users choose to search for ETF and then are asked to sign on to Leeway. These providers come and go. I want to support primarily Portfolio Report where we can improve the backend as well.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewaySearchProvider.java
```diff
@@ -0,0 +1,214 @@
...
+
+        if (Isin.isValid(query))
+            addISINSearchPage(answer, query.trim());
+
+        return answer;
+    }
+
+    private void addISINSearchPage(List<ResultItem> answer, String query) throws IOException
+    {
+        String html = new WebAccess("api.leeway.tech", "/api/v1/public/general/isin/" + query) //$NON-NLS-1$ //$NON-NLS-2$
```
buchen, line 169: naming: it is returning JSON, not HTML

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewaySearchProvider.java
```diff
@@ -0,0 +1,214 @@
...
+
+        for (int i = 0; i < jsonArray.size(); i++)
+        {
+            JSONObject item = (JSONObject) jsonArray.get(i);
+
+            if (item.get("Code") == null || item.get("Exchange") == null || item.get("Name") == null
+                            || item.get("Type") == null || item.get("ISIN") == null || item.get("previousClose") == null
+                            || item.get("previousCloseDate") == null || item.get("countryName") == null
+                            || item.get("currencyCode") == null)
+                continue;
```
buchen, line 207: I propose to move this into the "Result.from" method. That method can already return "null" if no valid item is created. Now you duplicate the "get" logic + you have to places where the code needs to know what properties are available.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewayQuoteFeed.java
```diff
@@ -0,0 +1,221 @@
...
+        return "PWP Leeway UG"; //$NON-NLS-1$
+    }
+
+    public void setApiKey(String apiKey)
+    {
+        this.apiKey = apiKey;
+    }
+
+    @Override
+    public Optional<LatestSecurityPrice> getLatestQuote(Security security)
```
buchen, line 53: Because you do essentially the same call to get the latest quote as for the historical quotes, I propose to add caching (see portfolio report). Otherwise we eat into the API quota very fast. Probably means last setting #mergeDownloadRequests to true

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewayQuoteFeed.java
```diff
@@ -0,0 +1,221 @@
...
+
+        try
+        {
+            @SuppressWarnings("nls")
+            WebAccess webaccess = new WebAccess("api.leeway.tech", //
+                            "/api/v1/public/historicalquotes/" + security.getTickerSymbol().trim().toUpperCase()) //
+                                            .addParameter("api_token", apiKey) //
+                                            .addParameter("from", start.toString());
+
+            String html = webaccess.get();
```
buchen, line 126: naming: the content is JSON, not HTML

buchen (changes requested):
And one more thing I noticed: the list of exchange rates does not seem to be complete. If I choose the US exchange in the search results, then the drop down remains empty and no prices are loaded.

Bildschirmfoto 2023-05-10 um 19 51 45

Bildschirmfoto 2023-05-10 um 19 51 40


In general the code looks great. Thanks for the contribution

Nirus2000:
Hello @buchen

Thanks first of all for all the responses.

I'll go through the points in order:

1. https://github.com/buchen/portfolio/pull/3343#discussion_r1190233606

The API works exclusively with the ISIN. If the ISIN is not valid, no query is performed.
There is also the possibility to search by ticker, but I did not want to implement this, because the number of API calls is too high (Limit 50).
_I have left a note in the settings for this._

2. https://github.com/buchen/portfolio/pull/3343#discussion_r1190210852
Done.

3. https://github.com/buchen/portfolio/pull/3343#discussion_r1190233606
Lechway only publishes markets via the API where historical rates in the respective currency are available.

4.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190236899
That is should be done and also other items at the portfolio report search provider.
- The markets are separated by commas. This should be separated.
- If the security is decoupled from Portfolio Report, it cannot be recoupled. 
   - Maybe it has to be this way, but it is annoying that you have to delete the security with all transactions to reconnect this to Portfolio Report.
- The currency from the market should be implemented in the Currency column. I don't know how?! :sob::sob::sob:

5.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190240415
The point I understand, but then why still the selection itself? Portfolio Report supports "shares", "index", "fund" and "bond".
To make Portfolio Report a favorite, it would be good to mark it with an orange background, similar to the CoinGecko Search Provider.

Why not add "Fund", "ETF", "ETC" and "Crypto"?
grafik
Yahoo and the other providers also output this information during the query, but so far they have not processed it further.
The selection and search for the individual types also works for me. 
How can we move towards this function being used and used correctly?

6.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190241256
Done.

7.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190242246
I have implemented this differently. It is no longer necessary to check here, because Lechtech already provides the data correctly. 
(No stock exchange with historical prices available = Array empty)

8.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190244790
I think I could implement this successfully 👍🏻 
Thanks for the hint. :smiley:

9.) https://github.com/buchen/portfolio/pull/3343#discussion_r1190245631
Done.

10.) A note about the ticker symbol in the AbstractQuoteProviderPages.
There are some things that don't work here.
Whether the ticker symbol is needed has not been checked at all. Neither at Yahoo nor at Leeway.
I have tried to implement this as best I can.
Thereby I noticed several errors.
- Alpha Vantage does not check if the symbol is selected. (Symbol or ticker symbol !?)
- Bitfinex does not check if the symbol is selected. (Symbol or ticker symbol !?)
- Binance does not check if the symbol is selected. (Symbol or ticker symbol !?)
- CoinGecko doesn't check anything and spend anything?
- EOD does not check if the symbol is selected. (Symbol or ticker symbol !?)
- Finnhub does not check if the symbol is selected. (Symbol or ticker symbol !?)
- Kranken does not check if the symbol is selected. (Symbol or ticker symbol !?)
- Yahoo Finance does not check if the symbol exists.
- Yahoo Finance (Adjusted Close) does not check if the symbol exists.

https://github.com/buchen/portfolio/assets/45203494/dfdfccac-a873-43bc-825b-d81304ac7988

- With Quandle the graphic (deco.show) is not removed again, as soon as one selects another supplier. This is then permanently displayed.
- With Quandle the setStatus() is not overwritten. (see video)

https://github.com/buchen/portfolio/assets/45203494/57959598-86c3-4516-b261-c890610e5f31

buchen:
> If the security is decoupled from Portfolio Report, it cannot be recoupled.
> Maybe it has to be this way, but it is annoying that you have to delete the security with all transactions to reconnect this to Portfolio Report.

That is new to me. Why should one has to delete the transactions? The UI is "just" setting or removing the onlineId.

In general, I agree that the implementation is cumbersome and not intuitive. It is possible to decouple by not to re-couple in the edit dialog because the dialog needs some serious refactoring (something for a different pull request).

> A note about the ticker symbol in the AbstractQuoteProviderPages.
> There are some things that don't work here.

Yeah. That's what I mean with serious refactoring... 😉 

* Each quote feed should define - in the API - which information it needs to retrieve historical prices (some require a ticker symbol, some require the URL, some custom attribute such as Json path) -- then the UI is rendered accordingly
* Base on the required information, the UI can also cache the response (if any of the parameters change, re-do the query). Caching is important because it eats up precious API limits just be checking which market to use
* Each quote feed should define whether it requires an API key (any maybe even allow to configure the API right in the configuration dialog)

I think it would be overkill to have the API generically define all required parameters (and generate a UI out of this). I think it would be okay to have a generic implementation for ticker and URL-based providers, and have custom UI for the rest. But today's if/else code is a nightmare. Alas, not for this pull request. Rome wasn't fixed in one day either...

(let me have a look at this PR now)

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/LeewayQuoteFeed.java
```diff
@@ -0,0 +1,249 @@
...
+                if (response.json != null)
+                    cache.put(securityTickerSymbol, response);
+            }
+
+            if (collectRawResponse)
+                data.addResponse(webaccess.getURL(), response.json);
+
+            JSONObject json = (JSONObject) JSONValue.parse(response.json);
+
+            JSONArray pricesJson = (JSONArray) json.get("prices"); //$NON-NLS-1$
```
buchen, line 162: I do not understand the reason for this piece of code:
* First, I think it is misleading to the user to change the response. The response is an array, but we would display it as an object with the property "prices" that has the array. It is not what the server response.
* Second, the test "Response.json != null" does not make sense - it is always not null because we set it one line before
* Third, when parsing, you again extract the array out of the json object.
* Fourth, it looks like we parse the JSON twice - one in line 146, and then again in line 160.
Why not use the array directly?

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/exchange-labels.properties
```diff
@@ -8,6 +8,89 @@
...
+leeway.KO     = Korea Stock Exchange -> KRW
+leeway.KQ     = KOSDAQ -> KRW
+leeway.LIM    = Bolsa de Valores de Lima -> PEN
+leeway.LS     = Euronext Lisbon -> EUR
+leeway.LSE    = NASDAQ Exchange -> USD
+leeway.LU     = Luxembourg Stock Exchange -> EUR
+leeway.LUSE   = Lusaka Stock Exchange -> ZMW
+leeway.MC     = Madrid Exchange -> CAD
+leeway.MCX    = MICEX Moscow Russia -> RUB
+leeway.MONEY  = Money Market Virtual Exchange -> Unknown
```
buchen, line 52: I propose to remove the "--> Unknown" label.
Just looking at it it reads like the Money Market Virtual Exchange is unknown - (but it can handle multiple currencies)

And then: I am not sure that it is required that all securities are traded at one stock exchange are traded in the same currency.

A quick search brought up this example:
> The Singapore Exchange lists equities denominated in different currencies, including Singapore dollars (SGD), US dollars (USD), Hong Kong dollars (HKD), Renminbi (CNY), Japanese Yen (JPY), Euros (EUR), Australian dollars (AUD), Pound sterling (GBP).
https://economics.stackexchange.com/questions/5399/does-each-stock-exchange-only-trade-in-one-currency


Also:
> leeway.MC     = Madrid Exchange -> CAD
Madrid is trading in Canadian Dollars?

> leeway.NASDAQ = London Exchange -> GBP
> leeway.LSE    = NASDAQ Exchange -> USD
Let's also check the stock exchange labels - I would think London is LSX https://en.wikipedia.org/wiki/London_Stock_Exchange and Nasdaq is NASDAQ

Nirus2000:
Hello @buchen

Thanks first of all for all the responses.

I'll go through the points in order:

1. https://github.com/buchen/portfolio/pull/3343#issuecomment-1546849901
   - I communicated a mistake there. It is possible to reconnect to Portfolio Report in the securities overview and if you paste the link from Portfolio Report. Sorry, that was a bit hidden.
   - I had thought that in the main menu of the security I can link the security again with Portfolio Report and not only somewhat awkwardly outside about the security view. 🤔 

2. https://github.com/buchen/portfolio/pull/3343#discussion_r1193106870
   - Ah! The realization has now reached me. Nice, learned something again. I have modified it and added another test.

3. https://github.com/buchen/portfolio/pull/3343#discussion_r1193108266
   - I have removed the "Unknow".
   - I have removed the currencies from the labels.

Thank you 👍🏻

buchen (commented):
I have changed the code to fit to the comments and merged it into the master branch.

buchen:
For the additional changes, see here: c8e391ef961be09f198c824d5bcc78aed174315a

buchen (approved):

buchen:
Hi @Nirus2000,

the behavior described in the [first comment](https://github.com/buchen/portfolio/pull/3343#pullrequestreview-1421100524) still exists:

Search for DE0007164600 and pick the "SAPGF.US" item:

Bildschirmfoto 2023-05-19 um 21 24 21

The edit dialog shows it okay with the ticker correctly shown:

Bildschirmfoto 2023-05-19 um 21 24 30

But the historical prices on the later tab are not shown because there is no exchange for the suffix "US":
Bildschirmfoto 2023-05-19 um 21 24 44

If I then save the instrument nonetheless, it correctly loads prices:
Bildschirmfoto 2023-05-19 um 21 24 57


My proposal:
* if the security has a ticker symbol that has not a (known) stock exchange, add it to the drop down anyway. I believe this is what I do with Yahoo, because we can never be sure that we have the list of exchanges up to date.

As this is now merged, please create a new PR. And as always: thank you very much!!!

PR #3356, opened 2023-05-15
Modify Raiffeisenbank PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-raiffeisenbank-bankgruppe/12535/117

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
@@ -1,15 +1,16 @@
 package name.abuchen.portfolio.datatransfer.pdf;
 
 import static name.abuchen.portfolio.datatransfer.ExtractorUtils.checkAndSetGrossUnit;
-
 import static name.abuchen.portfolio.util.TextUtil.trim;
 
+import java.math.BigDecimal;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import name.abuchen.portfolio.Messages;
 import name.abuchen.portfolio.datatransfer.ExtrExchangeRate;
+import name.abuchen.portfolio.datatransfer.ExtractorUtils;
 import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Block;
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -699,83 +771,124 @@ private void addAccountStatementTransactions()
 
     private <T extends Transaction<?>> void addTaxesSectionsTransaction(T transaction, DocumentType type)
     {
+        // If we have a gross reinvestment,
+        // we set a flag and don't book tax below.
+        transaction
+                .section("n").optional()
+                .match("^Zu (?<n>Lasten) .* \\-[\\.,\\d]+ [\\w]{3}.*$")
+                .match("^Belastung KESt bei aussch.ttungsgleichen Erträgen .*$")
+                .assign((t, v) -> type.getCurrentContext().put("noTax", "X"));
```
buchen, line 780: BTW, the ```DocumentContext``` also supports boolean directly.

You could say ```type.getCurrentContext().putBoolean("noTax", true)``` and use ```getBoolean("noTax")``` which resolves to true only if there is a boolean true in the map.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -677,8 +749,8 @@ private void addAccountStatementTransactions()
 
                     if (context.get("nr").compareTo("01") == 0  && Integer.parseInt(v.get("month")) < 3)
                     {
-                        Integer year = Integer.parseInt(context.get("year")) + 1;
-                        t.setDateTime(asDate(v.get("day") + "." + v.get("month") + "." + year.toString()));
+                        int year = Integer.parseInt(context.get("year")) + 1;
+                        t.setDateTime(asDate(v.get("day") + "." + v.get("month") + "." + Integer.toString(year)));
```
buchen, line 753: If you already have day, month, and year - there is no need to convert them to a string and parse them again:

```
LocalDate.of(year, month, day).atMidnight()
```

buchen:
I'll merge the PR anyway - it is not wrong, it just could be a little bit more efficient.

PR #3357, opened 2023-05-15
Improve IB-Flex-Query Importer to set better note

Nirus2000:
https://github.com/buchen/portfolio/pull/3294#discussion_r1193175946


pfalcon:
@Nirus2000: Thanks again for looking into this and taking effort to make substantial changes! But fairly speaking, that's not what I called for - parsing the data in the dividend's note. Instead, I'd rather see the ` if (!transaction.getType().equals(AccountTransaction.Type.DIVIDENDS))` added in "https://github.com/buchen/portfolio/pull/3294/files#diff-7aa193362f68c7e51a5624dfc2481ee1179984f8449df3994a387c171a382880R279 removed, to bring the situation to how it was before - i.e. record verbatim IBFlex "description" as PP's "note", regardless of the type of transaction. That's simple and robust.

Parsing the description is of course nice feat, but I'm afraid its format can change any time, and then it doesn't work as expected, or whole importer doesn't work at all.


Additional thanks for acting on suggestion to append tradeID/transactionID to the end of the note (instead of the beginning), I guess that's positive change.

Thanks!

Nirus2000:
I am closing the pull-request with no changes to the current state. 
You can edit, delete or add information to the notes at any time.
I offered you a solution, which you rejected.

"But fairly speaking, that's not what I called for" - I'm not your puppet.... and you are not the only user of PP, but so far the only one who has a very "me"-related opinion.

PR #3364, opened 2023-05-20
Improve PWP Leeway to set unknow exchange markets

Nirus2000:
Add ExchangeLabel "unknowStockExchange"
Add exchange market if the exchange is not included in the ExchangeLabels list
Add leeway.US -> "OTC Markets Group"
Fix NPE, when creating a new empty security the ticker symbol is searched.
Description of the limits adjusted in the setting

---

If the exchange market is not present in the ExchangeLabels list, then we insert it. If we go through the API and make a query of the unknown exchange market, then we get a response back in the form of an array, but we don't get any results via the new "code".

I have emailed Leeway again about this. The answer was as expected. The exchanges in US are very inconsistent and are therefore summarized under the sub-abbreviation US. (Group NASDAQ, NYSE, OTC, PINK)

Therefore, US should always be taken. (Label name "leeway.US = "OTC Markets Group")

---

Reminder:
I tried to implement the currency from Portfolio Report into the new column "currency", but unfortunately I didn't succeed. 
Could you have a look at it?


Nirus2000:
I have emailed Leeway again about this. The answer was as expected. The exchanges in US are very inconsistent and are therefore summarized under the sub-abbreviation US.

Therefore, US should always be taken.

buchen (commented):
merged

PR #3366, opened 2023-05-22
Modify DEKA PDF-Importer to support new transactions

Nirus2000:
Add new transactions
Fix dividend share calculation
Remove `PDFMsgMissingShares` in ZN tranlations... not exists


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DekaBankPDFExtractor.java
```diff
@@ -924,6 +929,29 @@ else if (currentSecurityItem != null && currentSecurityItem == securityItem)
...
+                                                t.setType(PortfolioTransaction.Type.SELL);
+
+                                            BigDecimal amountPerShare = BigDecimal.valueOf(asAmount(v.get("amountPerShare")));
+                                            BigDecimal amount = BigDecimal.valueOf(asAmount(v.get("amount")));
+
+                                            int sharesPrecision = Values.Share.precision() * 2;
+                                            BigDecimal shares = amount.divide(amountPerShare, sharesPrecision, RoundingMode.HALF_UP);
+
+                                            t.setNote(Messages.PDFLabelShares + ": " + v.get("shares") + " \u2259 "
+                                                            + shares.setScale(8, RoundingMode.HALF_UP));
```
buchen, line 952: This code does not change any values on the imported transactions (besides adding a comment to the transaction). Shouldn't it call t.setShares()?

And then the code first rounds the division to 16 decimal places only to immediately round it to 8 places. You can skip the first rounding then.

And as long it is just a note, I do not have a problem with the hard-coded scale of 8. But if we use this value somewhere, it should use the precision from the values.

name.abuchen.portfolio/src/name/abuchen/portfolio/messages.properties
```diff
@@ -574,6 +574,8 @@ MsgUnsupportedVersionClientFiled = This file was created with a newer version of
 
 MsgXMLFormatInvalid = Invalid XML Format: {0}
 
+PDFLabelShares = Shares
```
buchen, line 577: There is already a label "shares" - ```CSVColumn_Shares```. I propose instead of duplicating the translation terms, just use the existing label. It is easier for the translators than checking the same thing over again.

(I understand there are duplicate labels between the two packages "portfolio" and "portfolio.ui". It is not so easy to deduplicate them therefore I accepted that. But within on package, let's not add duplications).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DekaBankPDFExtractor.java
Nirus2000: I understand.
Done.
A note is no longer set, but the shares are calculated and set.

name.abuchen.portfolio/src/name/abuchen/portfolio/messages.properties
Nirus2000: Also finished.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DekaBankPDFExtractor.java
buchen: I noticed there is no test case that is going into that branch of code to "fix" the number of shares (I wanted to change the BigInteger -> double -> long conversion because why introduce the low precision double in the middle).

Do you have an example? Or we build an example to safeguard this piece of code?

(I'll merge this change anyway because it makes sense to get the updates to DEKA)

Nirus2000: Yes, sure -> [here](https://github.com/buchen/portfolio/blob/e716242a51e0f7126a4f989476bc4b5cc1d3c0eb/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dekabank/DekaBankPDFExtractorTest.java#L20892:L20905)

PR #3368, opened 2023-05-22
Enhancement of the PDF importer by cleaning up the source code

Nirus2000:
I noticed that we already have 61 PDF importers... OMG. 
No tests have been changed!

**Cleanup of importers:**
- removed unnecessary spaces
- remove the "equal" checks to "getCurrentContext()" and replaced by a primitive data type boolean
- replace various variables with primitive data types
- simplification of various pattern matches (e. g. isJointAccount)
- avoid Object.equals() or Object.equalsIgnoreCase() on null objects
- remove unnecessary $NON-NLS tags (all use @SuppressWarnings("nls"))


Nirus2000:
It worked out great, as we discussed on the phone.
It's going in the right direction.

PR #3370, opened 2023-05-24
Fix transaction dialog with negativ quote

### LINKED ISSUE #3365: Inbound delivery
Would it be possible to allow a free inbound delivery? It would be great for my scrip dividends and the way the tax authorities want me to account for the new shares.
I'm currently accounting 0,01€ but it souldn't be like that and I'm sure that the comunity would mind to allow 0€ inbounds deliveries. It's also a minor change and it would help many other users.
Thanks!!!

Nirus2000:
Fixes #3365

See comment
https://github.com/buchen/portfolio/issues/3365#issuecomment-1557594261

https://github.com/buchen/portfolio/assets/45203494/dc145614-8645-4612-bfa4-40ccf8bbe76a

I have not found a solution for this error case.
grafik

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModelTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModelTest.java
@@ -1,0 +1,159 @@
+package name.abuchen.portfolio.ui.dialogs.transactions;
+
+import static org.junit.Assert.assertEquals;
+
+import java.math.BigDecimal;
+import java.text.MessageFormat;
+
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.junit.Before;
+import org.junit.Test;
+
+import name.abuchen.portfolio.model.Client;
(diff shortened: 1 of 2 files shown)


stoeggich:
I don't think the fix is ​​optimal. If someone is active with options, you could not post more negative prices than in 2020 with crude oil. A truth might be better as the case is very rare

stoeggich:
Negative courses are possible, so the possibility should be retained that you can enter them. It is probably mostly a typo, so asking if the input is correct might be an option.

But the Performance cannot be calculated for this position

Nirus2000:
I have never seen a negative price on securities.
However, the situation in April 2020 with crude oil cannot be transferred to stocks. 
What happens is a total loss or insolvency of the company. See [here](https://depotstudent.de/kann-eine-aktie-ins-minus-fallen/)

buchen:
> I have never seen a negative price on securities.

I believe what @stoeggich wants to say if users are using PP to model options, then it might require to have a negative asset price.

I am hesitant here. I know PP does not officially support options, but I also know that some users have found a good way to track them anyway (I live with the glitches here and there). I do not want to just shut it down.

Maybe we can have an easier transition - by default check the values, but if experimental features are activated, we are more lenient. This way we can find out the feedback from the community. What do you think?

PR #3373, opened 2023-05-26
Add TwelveData as new rest api quote provider

Nirus2000:
Add TwelveData as new rest api quote provider

Add MIC to exchange-labels.properties (market identification codes -> ISO-10383). There are currently over 2000 MIC in the list. I have listed some, but they can all be added at [https://www.iso20022.org/market-identifier-codes](https://www.iso20022.org/market-identifier-codes) if necessary.

Change of the "type" translation of Leeway, that the found types of securities are translated collectively. The current values I have from the Yahoo, Portfolio Report, Leeway, Twelve Data, Coingecko ... compiled. These could now be implemented in the other quote providers. 
If the pull-request is ok for you with the "type", then I would create another pull-request in which the other quote providers are integrated. (I would not care if it is you or me)

All other information in the @ see, @ implNote and @ apiNote .


Nirus2000:
@buchen 
ist dieser nicht schon im master?!

PR #3377, opened 2023-05-31
Improve DEKA PDF-Importer to simplified fallback calculation of shares

Nirus2000:
Improve DEKA PDF-Importer to simplified fallback calculation of shares
Example: [here](https://github.com/buchen/portfolio/blob/e716242a51e0f7126a4f989476bc4b5cc1d3c0eb/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dekabank/DekaBankPDFExtractorTest.java#L20892:L20905)


buchen:
Grrr. My bad. I did not realize that the code fragment is twice in the importer. One fragment is executed during tests, the other not. And I picked the one that isn't. Thanks for noticing and fixing the other spot :-)


This was the Jacoco output:
Bildschirmfoto 2023-05-31 um 10 45 49

buchen:
Grrr. My bad. I did not realize that the code fragment is twice in the importer. One fragment is executed during tests, the other not. And I picked the one that isn't. Thanks for noticing and fixing the other spot :-)


This was the Jacoco output:
Bildschirmfoto 2023-05-31 um 10 45 49

buchen:
merged

PR #3378, opened 2023-05-31
Stake pdf loader

flywire:
https://hellostake.com/au

Note:
1. Security uses ticker code with `.ASX` extension
2. No security name
3. `A$` is `AUD` currency
4. Summary at top of file can be ignored

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/stake/Stake-Buy-01.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/stake/Stake-Buy-01.txt
@@ -1,0 +1,34 @@
+```
+PDFBox Version: 1.8.17
+Portfolio Performance Version: 0.62.1
+-----------------------------------------
+BUY CONFIRMATION
+SUMMARY
+JOHN DOE
+TICKER FLT.ASX
+1 Long Road
+QUANTITY
+SYDNEY, NSW 2000 512
+EFFECTIVE PRICE $20.43
+The following securities have been purchased on your behalf
+QUANTITY 512
+EFFECTIVE PRICE $20.43 TICKER FLT.ASX
+VALUE A$10,455.04 EXECUTION DATE 27-05-2022
+ORDER TYPE LIMIT HIN 12345678
+CONFIRMATION NUMBER 0000001 PID 3556
+SETTLEMENT DATE 01-06-2022 ACCOUNT NUMBER 123456
+BROKERAGE & GST A$3.00 SIDE BUY
+ORDER DATE 27-05-2022
+AMOUNT DUE & PAYABLE A$10,458.04 Funds have already been deducted from your buying power. No action
+is required by you to facilitate settlement.
+This trade was executed and cleared by Stake’s This confirmation tax document is issued by Stakeshop Pty Ltd and is issued subject to directiongs,
+executing broker Finclear Pty Ltd decisions and requirements of the Market Operators, the ASIC Market Integrity Rules, the Market
+Operating Rules and where relevant, the Settlement Rules and the correction of any errors and/or
+ABN 63 607 164 714 as a participant of ASX and CHI-X. This confirmation
+is issued subject to the directions, decisions and requirements of the omissions.
+market operator for the market on which this transaction was executed. Stakeshop Pty Ltd, trading as Stake, ACN 610105505, is an authorised representative (Authorised
+You agree that the trade confirmation made available to you via your Stake Representative No. 1241398) of Sanlam Private Wealth Pty Ltd (Australian Financial Services
+account are deemed as received. For more information, please refer to
+Terms and Conditions Licence No. 337927) and Airwallex Pty Ltd (Australian Financial Services Licence No. 487221).Stake’s
+
+```
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/stake/Stake-Buy-02.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/stake/Stake-Buy-02.txt
@@ -1,0 +1,34 @@
+```
+PDFBox Version: 1.8.17
+Portfolio Performance Version: 0.62.1
+-----------------------------------------
+BUY CONFIRMATION
+SUMMARY
+JOHN DOE
+TICKER LLL.ASX
+1 Long Road
+QUANTITY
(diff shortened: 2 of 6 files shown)


buchen:
@flywire  I think the commit is missing the actual PDF Importer (StakePDFExtractor)

Nirus2000:
@flywire 
please take a look at this [post](https://github.com/buchen/portfolio/pull/3354#issuecomment-1547193772) before you go any further. So change the test cases to the newer version and apply the Contributing standards.

flywire:
@Nirus2000 How can I get a 3 character currency from a single character as shown above?

> apply the Contributing standards

Already using them. They need a mention of `//` <-> `/` conversion in https://regex101.com/
> change the test cases to the newer version

I'm not clear on that currently, I'll have to come back to it.

buchen:
> How can I get a 3 character currency from a single character as shown above?

@flywire 

Although PP is not really using this at the moment, the ```CurrencyUnit``` class knows about the currency symbols - including A$.

https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies.properties#L44

Feel free to add an appropriate static lookup method to CurrencyUnit (or check in the importer-specific code)

buchen:
> change the test cases to the newer version

We added a couple of custom Hamcrest matchers to make the tests more concise.

See here for example:
https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dekabank/DekaBankPDFExtractorTest.java#L21652-L21655

flywire:
> add an appropriate static lookup method to CurrencyUnit (or check in the importer-specific code

I think you misunderstand my contribution, I'm not writing java code, just pattern patching with the docs and existing code to adapt it to a new broker. I know what a static method is but I don't write them. If you could point me to a good code example I could likely work it in.

If I can get this broker done I'll add another one.

buchen:
> I'm not writing java code, just pattern patching with the docs and existing code to adapt it to a new broker.

As you are not writing code, the code looks good ;-) Let me have a look at it the next days

buchen:
@flywire 

I adopted the code to make it work with your examples.

However, I have couple of questions we can discuss here in this pull request:

First about the currency of the investment vehicle.
```
QUANTITY 512
EFFECTIVE PRICE $20.43 TICKER FLT.ASX
VALUE A$10,455.04 EXECUTION DATE 27-05-2022
ORDER TYPE LIMIT HIN 12345678
```
Interestingly it says "$" as effective price and "A$". Is the vehicle traded in USD? I do not think because the exchange rate would make this sum up. So I assume that it is just missing the "A". For now, I pick up the currency of the investment vehicle also from the "VALUE" line. Do you think that is okay?

Interestingly the numbers also do not add up. If it is 512 pieces at 20.43, I'd expect a different value:
512 * 20.43 = 10,460.16  != 10,455.04
For now I ignore the share price (PP will calculate it by dividing the given total of  10,455.04)
Do you have an idea where the difference could come from?

```
"^EFFECTIVE PRICE \\p{Sc}[\\.,\\d]+ TICKER (?<tickerSymbol>[\\w]{3,4})\\.ASX$"
```
Here you are only matching the ticker, but not the exchange suffix (here: ASX). This value is used to find the existing investment vehicle in the existing PP file. Because the ticker is also used to retrieve historical prices, I would assume the configured ticker actually includes the exchange suffix. Therefore my gut feeling is: should the ticker be matched so that it incudes the exchange?

flywire:
> As you are not writing code, the code looks good ;-)

Credit to @Nirus2000 and @buchen for ongoing improvements to the PDF importer documentation and code comments to support hacking by non-java programmers

---

> First about the currency of the investment vehicle.

Effective price does not display currency, just currency sign. The currency is `AUD` in this example.

>  the numbers also do not add up. If it is 512 pieces at 20.43

Hmm, misplaced keystroke changed one and I made the other wrong when I went to fix it, original is `$20.42`. btw, I notice another contract note effective price has `$9.9999` format, dollar to 4 decimal places.

> Here you are only matching the ticker, but not the exchange suffix (here: ASX).

This is something I don't understand. I use Yahoo as the quote provider and it uses `.AX` for the `.ASX` exchange. Initially, I enter a security without the exchange suffix, and then when the Quote Feed Exchange is selected as say `FLT.AX (Australian Stock Exchange (ASX))` the `.AX` is appended to the `symbol`.

This becomes a problem entering dividends. `Date,Ticker Symbol,Value,Cash Account` does not recognise the security based on `Ticker Symbol` (with or without the exchange extension), it must be imported with the `Security Name` field. This is difficult to match character perfectly, especially when the dividend payment lists the ticker symbol (without the extension) and not the security name. Essentially, I need to export PP securities, lookup the security name from ticker symbol given in dividend payment then add it to the csv import file. Crazy stuff!

buchen:
> This is something I don't understand. I use Yahoo as the quote provider and it uses ```.AX``` for the ```.ASX``` exchange.

Yeah, I guess this is called "technical debt". Years back I did not realize it needs multiple fields.

But what we can do is this: when looking for a security, we first try to match it by the full entry of the "tickerSymbol". If that does not work, we try to match it by the prefix (strip away the exchange label).

In your case that would mean:
* the security is configured as XYZ.AX (to get prices from Yahoo)
* the PDF document contains XYZ.ASX, we look for the XZY and find the security above

> This becomes a problem entering dividends

How do you enter dividends? Also via PDF import? Or via CSV?

buchen:
> But what we can do is this: when looking for a security, we first try to match it by the full entry of the "tickerSymbol". If that does not work, we try to match it by the prefix (strip away the exchange label).

I realize that @Nirus2000 has already implemented this feature.
[I have added a test that shows that the importer matches an existing security with the ticker FLT.AX to the Buy01.txt import](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/stake/StakePDFExtractorTest.java#L67-L71).

@flywire Which leaves the question why this matching logic does not work for your dividend import workflow. Via PDF? Via CSV?

flywire:
My comments above were not accurate and have been edited. Moved the discussion about Loading Payments by CSV to #3380.

What is required to close this issue?

buchen:
> My comments above were not accurate and have been edited. Moved the discussion about Loading Payments by CSV to https://github.com/buchen/portfolio/issues/3380.

Great. Let me have a look at the other issue then.


> What is required to close this issue?

Nothing. The code is merged already. I just kept it open to not loose the conversation if it is moved to the "closed" section.

PR #3379, opened 2023-05-31
Fixes wrong "benchmark" suffix for non-benchmark items

OnkelDok:
Fixes wrong "benchmark" suffix for non-benchmark items

see https://forum.portfolio-performance.info/t/auswahlmenue-der-datenreihen/2664/4?u=onkeldok


buchen:
Proper formatting would have helped to avoid that one...

PR #3385, opened 2023-06-04
Improve Stake PDF-Importer

Nirus2000:


buchen:
Hi @Nirus2000,

I think we have to compare notes on the default formatter here. The file was formatted according to project standards. And eventually I want to get back to everything formatted that way. I wasn't always consequent with all contributions in the past (to not demotivate contributors on details). But this is a new file, no history, so no reason to have it formatted differently. I usually have the "save action" activated that "formats all changed lines of code" automatically.

Which formatter are you using?

Bildschirmfoto 2023-06-04 um 10 46 03


Andreas.

PR #3423, opened 2023-06-24
Add MACD as an indicator

daveln:
Add MACD as an indicator

Also requested in
https://forum.portfolio-performance.info/t/chartanalyse-charttools-und-indikatoren/744/7 and https://forum.portfolio-performance.info/t/chart-analyses-in-stock-graph-like-rsi-or-macd-or-others/21484


daveln:
@buchen 
Potential issues:
* I could modify TimelineChart.getOrCreateAxis to accept only SecuritiesChart.ChartDetails but I'm not decided if a defined type is required for the key. The key could be a 'new Object()' and it would still work. For now I think it is ok to not define the key type.
* The color is just a few ints that were already used as color consts in a unique combination. I don't know if you have a preference to change that to something else.
* No translations other than English and Dutch.
* The test is modified from SimpleMovingAverageTest. I could fold testSufficientPriceDataPass and testSufficientPriceDataStartDate into one test. Eh.

Thoughts?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,112 @@
...
+        List<LocalDate> datesMacd = new ArrayList<>();
+        List<Double> valuesMacd = new ArrayList<>();
+        List<Double> valuesSignal = new ArrayList<>();
+
+        // seed the running EMAs with the first value
+        double minuendEma = prices.get(index).getValue() / Values.Quote.divider();
+        double subtrahendEma = minuendEma;
+        double signalEma = 0;
+
+        for (int i = 0; i < prices.size(); i++) // NOSONAR
```
Nirus2000, line 82: Change to... 
for (SecurityPrice price : prices) {
    LocalDate date = price.getDate();
    // ...
}

daveln: Fair. Changed, squashed, force pushed.

Nirus2000 (changes requested):

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private void calculateMacd()
+    {
+        if (security == null)
+            return;
+
+        List<SecurityPrice> prices = security.getPricesIncludingLatest();
+        if (prices == null)
+            return;
```
Nirus2000, line 64:
```suggestion
        if (prices == null || prices.isEmpty())
            return;
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,121 @@
...
+    private double macdSubtrahendSmoothingFactor;
+    private double signalLineSmoothingFactor;
+    private Security security;
+    private ChartInterval interval;
+    private ChartLineSeriesAxes macd;
+    private ChartLineSeriesAxes signalLine;
+
+    public MovingAverageConvergenceDivergence(Security security, ChartInterval interval)
+    {
+        this.security = security;
```
Nirus2000, line 36:
```suggestion
        this.security = Objects.requireNonNull(security);
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public ChartLineSeriesAxes getSignalLine()
+    {
+        return signalLine;
+    }
+
+    private void calculateMacd()
+    {
+        if (security == null)
+            return;
```
Nirus2000, line 60:
```suggestion
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,121 @@
...
+
+        int index = Collections.binarySearch(prices, new SecurityPrice(interval.getStart(), 0),
+                        new SecurityPrice.ByDate());
+
+        if (index < 0)
+            index = -index - 1;
+
+        if (index >= prices.size())
+            return;
```
Nirus2000, line 73:
```suggestion
        index = Math.max(0, index);
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
```diff
@@ -0,0 +1,121 @@
...
+            if (date.isBefore(interval.getStart()))
+                continue;
+
+            valuesMacd.add(round(macdValue, 2));
+            valuesSignal.add(round(signalEma, 2));
+            datesMacd.add(date);
+        }
+
+        Date[] timelineDates = TimelineChart.toJavaUtilDate(datesMacd.toArray(new LocalDate[0]));
+        macd.setDates(timelineDates);
```
Nirus2000, line 108:
```suggestion
        macd.setDates(TimelineChart.toJavaUtilDate(datesMacd.toArray(new LocalDate[0])));
        signalLine.setDates(macd.getDates());
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
daveln: Fair. I've determined that security.getPricesIncludingLatest() can never return null. I'm dropping the null check, and replacing with your suggested prices.isEmpty().

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
daveln: Changing as suggested.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
daveln: Changing as suggested.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
daveln: After changing the calculation for correctness, this code was changed; MACD and MACDSignal have different start dates. In future there will be a xlsx file with the full MACD and MACDSignal calculation.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/MovingAverageConvergenceDivergence.java
daveln: Hmm. In the previous code this index was only used a stop condition (if the interval was in the future of the prices list). I've refactored this and made it more explicit what the intent is.

daveln:
* Added formatting to the MACD/MACDSignal values in the SecuritiesChart/TimelineChartToolTip.
* Calculation revisited after noticing that the start of the series was deviating from correct values. (The end of the series was correct.)
* Test is completely reworked. The test now has full coverage of MovingAverageConvergenceDivergence class.
* Added an excel file with matching csv that contains the suitable calculations for MACD, RSI, EMA, SMA for a security.
* It is used in the test to validate calculation correctness for MACD.
* Processed review comments.

buchen:
> I could modify TimelineChart.getOrCreateAxis to accept only SecuritiesChart.ChartDetails but I'm not decided if a defined type is required for the key. The key could be a 'new Object()' and it would still work. For now I think it is ok to not define the key type.

I think that is okay. The ChartDetails key should not leak into TimelineChart (which is used in other locations as well).

Anyway, the SecurityChart class has grown over time as more and more things have been bolted to it. At one point, we'll have to refactor this class. But for MACD it is okay to add it the existing way.

I'll add some additional translations and will merge it. Thanks.

PR #3424, opened 2023-06-25
Added fees to account transfer dialog

RomanLangrehr:
Let the user add fees for account transfers:
image


buchen:
Does it really have no impact on client filters, reports on fees, etc? I haven't had the time to look into the details, but I pretty much expect that the newly introduced fees require changes in the code that reports, filters, and selects on fees.

Nirus2000:
I think that definitely has an impact on the filters and so on. (See https://github.com/buchen/portfolio/pull/3272)

Furthermore, we had also talked in a meeting that we wanted to include rebookings via importer with this problem. 
- See FlatEx foreign exchange business as example
  - https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/191
  - https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/192

RomanLangrehr:
@buchen @Nirus2000 I missed that I have to make changes to the `ClientPerformanceSnapshot ` to take into account the fees in transactions (for the calculation of fees and for the calculation of foreign currency gains/losses). I fixed that now. But I don't think any of the client filters need adaption (as far as I know there is no filter working based on fees).

I will add more unit tests (for the `AccountTransferModel ` and the `ClientPerformanceSnapshot `) when I have time

The motivation for me to add this feature was also to import automtically forex transactions in the future (for example from Interactive Brokers) and they typically include fees.

Nirus2000:
Hello @RomanLangrehr 

> The motivation for me to add this feature was also to import automtically forex transactions in the future (for example from Interactive Brokers) and they typically include fees.

Yes, that's right.
In the IB-Flex-Query Importer there are already test files with these transactions, which could not be processed yet, because this function does not exist yet. (Example [testIBFlexStatementFile19.xml](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/testIBFlexStatementFile19.xml))
So... could you see if you can integrate this part or if we should implement it in a separate PR?

chirlu:
I’m very suspicious of this introducing unintended side effects.

Furthermore, UI-wise, how would the user distinguish between the two “fee” fields? If the accounts are of different currencies, OK, you can guess it from the currency; but for transfers without currency exchange?

Nirus2000:
The feature itself — adding fees to account transfer transactions — is useful and relevant, especially for automated forex import (e.g. Interactive Brokers). However, the PR stalled after unresolved review concerns:

Correctness: @buchen raised the question of whether the new fees are properly reflected in client filters, fee reports, and performance calculations. This was not conclusively addressed.
UI clarity: @chirlu pointed out that two unlabeled fee fields in same-currency transfers are indistinguishable to the user.
Promised test coverage for AccountTransferModel and ClientPerformanceSnapshot was never added.

If someone wants to revive this idea, please open a new pull request rather than reopening this one. The new PR should address all three points above before submission — particularly the impact on filters and fee reporting, as that was the maintainer's primary concern. The discussion here provides a solid foundation to build on.

PR #3425, opened 2023-06-25
taxonomy/AbstractNodeTreeViewer: Add Symbol column

pfalcon:
Allows to see ticker symbols in taxonomy views, whereas previously only security name was available, while arguably some users are more familiar with tickers than names.


pfalcon:
Fixes https://github.com/buchen/portfolio/issues/3292

PR #3427, opened 2023-06-28
Activate measurement tool in regular mode

Nirus2000:
Please add another description e.g. video as a new feature in the next update.

@OnkelDok
Maybe we can now tackle the crosshair as the next tool ?! :grinning:


OnkelDok:
@Nirus2000  I have the feeling you really need the crosshair?
Maybe the biggest work for this is to find a suitable location for the button with which the crosshair mode can be enabled.
Do you think of a seconds button next to the measurement-tool-button?

Nirus2000:
@OnkelDok 
Hehehe... yes, you started with the Crosshair and I still like it. I would also just put a button to activate it next to the measure tool.

PR #3428, opened 2023-06-29
SecurityDetailsViewer: Use scrollable container

pfalcon:
Allow to scroll vertically contents of security details pane (as shown side by side with security chart), to access data closer to its bottom. For example, user-defined taxonomies are shown there and may be not visible otherwise.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
@@ -12,6 +12,9 @@ import java.util.List;
 import org.eclipse.jface.layout.GridDataFactory;
 import org.eclipse.jface.layout.GridLayoutFactory;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.ScrolledComposite;
+import org.eclipse.swt.events.ControlAdapter;
+import org.eclipse.swt.events.ControlEvent;
 import org.eclipse.swt.layout.FormAttachment;
 import org.eclipse.swt.layout.FormData;
 import org.eclipse.swt.layout.FormLayout;
@@ -353,7 +356,7 @@ public class SecurityDetailsViewer
         }
     }
 
-    private Composite container;
+    private ScrolledComposite container;
 
     private List<SecurityFacet> children = new ArrayList<>();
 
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
```diff
@@ -364,13 +367,22 @@ public SecurityDetailsViewer(Composite parent, int style, Client client)
...
+
+        Composite container1 = new Composite(container, style);
+        container.setContent(container1);
+        container.setExpandHorizontal(true);
+        container.setExpandVertical(true);
+        // Needed to avoid artifact (at least on Linux) when scrollbar appears
+        // and covers right edge (last digits of quotes, etc.) of the content
+        // pane. With setAlwaysShowScrollBars(true) that issue doesn't happen.
+        container.setAlwaysShowScrollBars(true);
+        container1.setBackground(Colors.WHITE);
```
Nirus2000, line 380: I am unsure if this will work with PP's Dark or Withe themes.

buchen:
> ```container.setAlwaysShowScrollBars(true);```

This line is what I am wondering about.

At least on macOS, the change is okay. On macOS, the scrollbar is anyway hidden unless the user is actually scrolling. Therefore this change looks okay (and enables scrolling).

@Nirus2000 - can you check how it looks on Windows? If it is okay, I would merge.


(One note about the dark mode. When in dark mode, the colors are anyway overwritten. So setting the color to white has no effect when in dark mode but apparently is needed for light mode)

pfalcon:
> This line is what I am wondering about.

As the code comment above it says, it took quite a while to arrive at that call as a workaround for artifact (seen at least on Linux) when the pane *always* appeared with the size which hides 10-20 pix of its right edge, where the last digits of quotes are located, etc. I found that quite annoying and tried various ScrolledComposite options, until found that's what worked.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
pfalcon: Tested now on Linux - everything is ok with the Dark Theme. Also, if you look at the patch, I don't really change much, there used to be `container.setBackground(Colors.WHITE);`, and I just changed var name to apply to the same logical container.

> and if "container1" is a good variable name.

I'm all for descriptive var names, but `innerContainer` is a bit too long IMHO, and stuff like `inContainer`, `innerCntnr` is inconsistent/confusing.

buchen:
> As the code comment above it says, it took quite a while to arrive at that call as a workaround for artifact (seen at least on Linux) 

Understood. I should have formulated it better: I understand it is a good workaround for Linux, but I am wondering how it looks on Windows (which are 80% of the users).

I wasn't able to put in today's release (which will take some more time because I am on the road and upload speed veeeery slow). But I will be able to check on my windows machine next week

buchen:
Pinging @Nirus2000 - can you test how it looks on Windows?

Nirus2000:
i can't get this PR to run.
grafik

Pinging @OnkelDok 
Could you check this one?

Nirus2000 (approved):
For me this looks in Windows10 i. O. looks.
@OnkelDok can you have a look in Windows11?

buchen:
Win10 and Win11 will not differ much. Thanks for checking. Will merge.

OnkelDok:
> @OnkelDok can you have a look in Windows11?

I don't have Win11. I use Win10.

Issue #3444, opened 2023-07-15
Add security operation ignores portfolio's base currency, hardcodes EUR by default


pfalcon:
**To Reproduce**
1. Use File -> New to create a new portfolio file
2. Choose USD as the base currency
3. Use "New instrument..." dialog to add new security
4. Search for "AMZN"
5. Choose Yahoo Finance as the source feed.
6. After clicking "Apply" in search result list, security property dialog open and the currency set to EUR.

**Expected behavior**

As the portfolio base currency is set to USD, that what should be used by default for new securities (unless there're strong reasons to do otherwise, e.g. is source indicates that security's currency is different one. In this case we know that AMZN is quoted in USD.)

Version 0.64.4 (git master)

pfalcon:
@buchen , I would suggest it's pretty important to fix this issue, because I may imagine it contributes to a perception that PortfolioPerformance is "intended for European investors" and "less suitable for US investors". (I read such characterizations more than once "on the internets".  Which is of course not true, but rather a subjective perception grounded in cases like above.)

The causes of this issue are apparently:
1. This line: https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java#L50, i.e. hardcoding EUR by default in a Security.
2. And then, creating Security without explicitly setting currency in many places in the codebase.

@buchen, only you, as the author, would know the right way to fix this. E.g., the "ultimate" way would be to remove the default assignment above, disable default (empty) constructor, and make sure that any remaining constructor accepts currency explicitly. Then, fix the codebase to adhere. Yes, that would be a bit "revolutionary" (a lot of changes) way, so again, only you would know if it's worth it.

pfalcon:
For a user who look to just work around the issue on their side, a low-effort way would be to change currency in Security.java. But they would be greeted by failing tests, because again, assumption of EUR is implicitly hardcoded in many places. I would like to contribute at least to resolving such cases: https://github.com/buchen/portfolio/pull/3445 . (But without resolving the root cause, such cases would reappear).

buchen:
@pfalcon Agree. If the currency is not known otherwise (because it is imported via PDF or because it is know from the market), then it should be set to the base currency of the file.

[...]

PR #3445, opened 2023-07-15
tests: Create test securities with explicit currency set

pfalcon:
Don't rely on a particular curreny to be "hardcoded" in a generic security create for a test. These cases where identified by setting Security.currencyCode to a different currency and seeing what tests break.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
@@ -108,7 +108,7 @@ public class CSVAccountTransactionExtractorTest
     public void testDividendTransaction()
     {
         Client client = new Client();
-        Security security = new Security();
+        Security security = new Security("", "EUR");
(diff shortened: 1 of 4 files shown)


pfalcon:
First step towards resolving https://github.com/buchen/portfolio/issues/3444

Nirus2000:
These changes are nonsense.

buchen:
@pfalcon

have a look at #3444

My proposal:
* do not change all tests - annotate the constructor as visibleForTesting only
* update all usages of the default constructor in production code and make sure it defaults to the base currency (and not EUR)
* with the change, I think I have updated the relevant places (Yahoo Finance and CoinGecko seemed most prominent culprits)

closing this pull request

pfalcon:
@buchen :

> do not change all tests 

Sure, I'm all for avoiding sweeping changes across the codebase, whenever they reasonably can be avoided. I hope the same logic would apply to https://github.com/buchen/portfolio/pull/3355, that for sure looks like a scary change across the entire codebase, but not giving much value (I'm about the part which renames all variables/methods from "Ticker" to "TickerSymbol", UI-level terminology definitely could be improved).

buchen:
> sure looks like a scary change across the entire codebase, but not giving much value

Agreed.

Issue #3452, opened 2023-07-19
Security is added to watchlist of other (not active) open file when new security is added/created


OnkelDok:
### Describe the bug
Reported here: https://forum.portfolio-performance.info/t/interferenz-zwischen-zwei-gleichzeitig-geoffneten-dateien/24516

### To Reproduce
Steps to reproduce the behaviour:
1. Have two opened files A and B, where B contains a watchlist.
2. Select watchlist in B and switch back to file A.
3. Create new security via main menu or +-Button
4. Security is added to file A (as expected)
5. **But** also in watchlist of file B the security was inserted (not expected)

### Expected behavior
Only new security in file A. 

### Desktop (please complete the following information):
 - OS: Windows
 - Version 0.64.4 (Juli 2023)

buchen:
fixed

PR #3453, opened 2023-07-19
fix for #3452: add check in SECURITY_CREATED subscriber method to determine if security was created in current file/client or not

OnkelDok:
With this the issue in  #3452 is fixed. ~~But additionally the functionality  is lost that when on the **active** file a watchlist is selected when creating a new security, then the new security is not automatically inserted to this watchlist.~~ *Update: no, it's not. See https://github.com/buchen/portfolio/pull/3453#issuecomment-1645712619*

dbeda8c60a4a0de7871fb34e02333009ddb5554b was just the first solution that comes in my mind.
Maybe better solutions would be (to preserve the automatic insertion into watchlist):
* Fire the event ``SECURITY_CREATED`` only in the active file (don't know if this kind of filtering is possible)
* Or detect in the event handler ``onSecurityCreated`` [method](https://github.com/buchen/portfolio/blob/044c24f049189aadfa413ea77b6765578a55ec71/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java#L384) if the file is active or not
    * maybe there the ``MenuHelper.getActiveClientInput`` is accessible
    * or when firing the event the current client will also passed (additionally to the new security) then in the ``onSecurityCreated`` method we can check if this passed client is the client of the current file (itself)

*UPDATE: I have chosen the last listed option.*

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/NewDomainElementHandler.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/NewDomainElementHandler.java
@@ -60,6 +60,7 @@ import name.abuchen.portfolio.ui.util.Colors;
 import name.abuchen.portfolio.ui.util.FormDataFactory;
 import name.abuchen.portfolio.ui.wizards.security.EditSecurityDialog;
 import name.abuchen.portfolio.ui.wizards.security.SearchSecurityWizardDialog;
+import name.abuchen.portfolio.util.Pair;
 import name.abuchen.portfolio.util.TradeCalendarManager;
 
 public class NewDomainElementHandler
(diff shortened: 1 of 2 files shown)


OnkelDok:
With 39dea9cbf0c5ba7b0dde9682530458a507ac32e9 I restored adding to watchlist and inserted a check in the event handler ``onSecurityCreated`` to determine if the event is called from current client or not.
So with this only the bug #3452 should be fixed without losing any functionality.

buchen:
Thanks @OnkelDok for proposing and implementing a solution.

I will do a slight change: instead of using the generic ```Pair```, I will create a specific event class. I think the use case is important enough to take the effort to create a class. I am not dogmatic on the use of Pair (some argue it is a design issue to have such a class at all), here it will help our future selves to better understand the code.

I will publish a new version today because it can corrupt files. The second file has the security in the watchlist, but not in the list of securities.

buchen:
I also realized that there was already a [SecurityChangedEvent](name.abuchen.portfolio/src/name/abuchen/portfolio/events/SecurityChangeEvent.java). I totally forgot... That change event was already passing the client along. I updated the code to use the same pattern.

buchen:
@OnkelDok if have time to test it, then I would appreciate your feedback before publishing a new version

OnkelDok:
> @OnkelDok if have time to test it, then I would appreciate your feedback before publishing a new version

@buchen: It's working as expected.

PR #3459, opened 2023-07-22
remove wrongly added securities (due to #3452) at startup

OnkelDok:
I wonder if it would be necessary to automatically delete wrongly added securities in watchlist due to #3452 on startup. What do you think?

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
@@ -828,6 +828,10 @@ public class ClientFactory
 
                 // remove obsolete MARKET properties
                 removeMarketSecurityProperty(client);
+            case 57:
+                // remove securities in watchlists which are not present in "all
+                // securities", see #3452
+                removeWronglyAddedSecurities(client);
 
                 client.setVersion(Client.CURRENT_VERSION);
                 break;
@@ -836,6 +840,16 @@ public class ClientFactory
             default:
                 break;
         }
+    }
+
+    private static void removeWronglyAddedSecurities(Client client)
+    {
+        client.getWatchlists().forEach(w -> new ArrayList<>(w.getSecurities()).forEach(s -> {
+            if (!client.getSecurities().contains(s))
+            {
+                w.getSecurities().remove(s);
+            }
+        }));
     }
 
     private static void fixAssetClassTypes(Client client)
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Client.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Client.java
@@ -29,7 +29,7 @@ public class Client
         String WATCHLISTS = "watchlists"; //$NON-NLS-1$
     }
 
-    public static final int CURRENT_VERSION = 57;
+    public static final int CURRENT_VERSION = 58;
     public static final int VERSION_WITH_CURRENCY_SUPPORT = 29;
     public static final int VERSION_WITH_UNIQUE_FILTER_KEY = 57;
 


buchen:
@OnkelDok that is a good idea. I was hoping that the [FixOrphanedSecurtiesCheck](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/checks/impl/FixOrphanedSecurtiesCheck.java) does check for watchlists, but it does not (it is not the first time that some change create dangling securities).

I will merge your pull request - but will a slight change: we will delete the security only if it has no transactions. If there have been transactions created, we add it to the list of securities (and the user can decide)

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
```diff
@@ -838,6 +842,16 @@ else if (flags.contains(SaveFlag.COMPRESSED))
         }
     }
 
+    private static void removeWronglyAddedSecurities(Client client)
+    {
+        client.getWatchlists().forEach(w -> new ArrayList<>(w.getSecurities()).forEach(s -> {
+            if (!client.getSecurities().contains(s))
```
Nirus2000, line 848: I'm not sure if "contains" should be replaced by "equals"?! 

maybe with "contains"...
```
List<Watchlist> watchlists = client.getWatchlists();
List<Security> clientSecurities = client.getSecurities();

for (Watchlist watchlist : watchlists) {
    List<Security> watchlistSecurities = new ArrayList<>(watchlist.getSecurities());
    for (Security security : watchlistSecurities) {
        if (!clientSecurities.contains(security)) {
            watchlist.getSecurities().remove(security);
        }
    }
}
```

buchen: Isn't this saying the same thing - just with more code? I am not sure I understand your comment

buchen:
merged

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
Nirus2000: Die iterierung in der Watchlist hat Vorrang...?!

buchen: Ah okay, I believe this ```w -> new ArrayList<>(w.getSecurities()``` also copies the securities to be able to modify the underlying list.

OnkelDok: Yes, I think it's the same just with ``for``-statement instead of ``.forEach``. The only difference I can see is that @Nirus2000 made only one single call to ``client.getSecurities()``. In my proposal this fcuntion is called N times, were N is the sum of all securities in all the watchlists.

Nirus2000: Allerdings sollten sichergestellt werden, dass die Security-Klasse eine korrekte equals()- und UUID-Methode implementiert ist, damit "contains()" das gewünschten Ergebnisse liefert. 
Wenn das nicht der Fall ist, müssen die contains()-Logik anpassen werden, um die Wertpapiere besser zu vergleichen.
Das habe ich jetzt nicht prüfen können... war von mir nur ein Querschuss... I dont know

Nirus2000: > Yes, I think it's the same just with `for`-statement instead of `.forEach`. The only difference I can see is that @Nirus2000 made only one single call to `client.getSecurities()`. In my proposal this fcuntion is called N times, were N is the sum of all securities in all the watchlists.

Ja, genau das ist der Unterschied, welche aber wichtig ist.

Nirus2000: Egal... wird schon schief gehen... :rofl::rofl::rofl::rofl::rofl::rofl::rofl::rofl::rofl:

buchen: ```Security``` never implemented ```equals``` as we want to work with object identity. I would think Client#getSecurities will be inlined by the JIT (it will run once per file - I do not think it will be a problem).

Anyway, I tested it with a sample, it works as far as I can tell, the release is already for notarization at Apple. If it fails, it will be one me. :-)

Thanks @OnkelDok for the pull request, thanks @Nirus2000 for looking over it

OnkelDok: I tested it with corrupt file (watchlist that contains wrongly added securities and also valid securities) and only the wrongly added securities were deleted. For this functionality I did not tried to optimize the performance, because this function normally is only called once per file and the ``client.getSecurities()`` is not that expensive operation.

PR #3463, opened 2023-07-24
Fix sBroker PDF-Importer and add improve PDFParser

Nirus2000:
Add new PDFParser that mustNotInclude pattern in a document

Some fix in sbroker
Fix wrong fee and feeRefund (detect bei sign)
Fix missing "BargAuszDebitFremdGA" transaction
Improve regEx in ExtrExchangeRate section

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
@@ -26,19 +26,21 @@ import name.abuchen.portfolio.model.TypedMap;
     /* package */ static class DocumentType
     {
         private List<Pattern> mustInclude = new ArrayList<>();
+        private List<Pattern> mustNotInclude = new ArrayList<>();
 
         private List<Block> blocks = new ArrayList<>();
         private DocumentContext context = new DocumentContext();
         private BiConsumer<DocumentContext, String[]> contextProvider;
 
-        public DocumentType(List<Pattern> mustInclude)
-        {
-            this.mustInclude.addAll(mustInclude);
-        }
-
         public DocumentType(String mustInclude)
         {
-            this(mustInclude, null);
+            this.mustInclude.add(Pattern.compile(mustInclude));
+        }
+
+        public DocumentType(String mustInclude, String mustNotInclude)
+        {
+            this.mustInclude.add(Pattern.compile(mustInclude));
+            this.mustNotInclude.add(Pattern.compile(mustNotInclude));
         }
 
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
```diff
@@ -26,19 +26,24 @@
...
 
-        public DocumentType(String mustInclude)
+        public DocumentType(String mustInclude, String mustNotInclude)
         {
-            this(mustInclude, null);
+            this.mustInclude.add(Pattern.compile(mustInclude));
+            if (mustNotInclude != null)
+            {
+                this.mustNotInclude.add(Pattern.compile(mustNotInclude));
+            }
```
buchen, line 46: I propose to remove this null check.
If one does not have an exclude pattern ("mustNotInclude") then do not use this constructor.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/PDFParserTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+import org.junit.Test;
+
+import name.abuchen.portfolio.datatransfer.DocumentContext;
+
+@SuppressWarnings("nls")
+public class PDFParserTest
+{
+    @Test
+    public void testMatches_Type1()
```
buchen, line 16: I do not understand what is meant by "type1" through "type5" - can you shed some light?

Also, the test "type1" and "type2" tests look pretty identical to me - the comment is different, but I do not see a difference in the code. They all put a strings in a map and retrieve them again. Am I overlooking something?

To be honest, I would not bother writing specific tests for the PDFDocument in isolation. **All** PDF tests are testing this functionally implicitly. I cannot image a situation where all PDF tests are green, but this test is failing. And the PDF tests are better because they test on the level of business functionality.

Nirus2000: done... i delete this file

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
Nirus2000: done... i remove this null check

buchen:
merged.

I have made a couple of minor changes - see the commit message.
I also rephrased the commit message to make clear what changed (instead of that something changed).

PR #3466, opened 2023-07-26
Introducing chart tools manager including new crosshair tool

OnkelDok:
Here my suggestion for a chart tool manager which provides the existing measurement tool and a new crosshair tool.
If you have any ideas or suggestion for different and maybe better implementation of such a chart tool management (``ChartToolsManager``), then please share your thoughts.
~~One thing I want to change the next days is the size of the crosshair in the button image. I think it can be a little bit bigger.~~ **done**

https://forum.portfolio-performance.info/t/fadenkreuz-in-charts/7335

And here a short video:

https://github.com/buchen/portfolio/assets/90478568/f7902541-d070-4b03-95d1-e53514b2ccd9

Good night!


Nirus2000:
Another masterpiece... 💯 
If you have time, would you email me?

OnkelDok:
@Nirus2000: how can I contact you? I cannot find any way to send you private message in the PP forum or at github. But Andreas sent me one some days ago in the forum. Did I miss something?
Are you able to send me private messages in the PP forum?

Btw: I enlarged the images for the crosshair buttons.

Nirus2000:
Write on... webmaster ääääätttt nirus-online.de

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
```diff
@@ -0,0 +1,523 @@
...
+
+        private void drawCrosshairValueSecondAxisTextbox(PaintEvent e, Point p)
+        {
+            IAxis axis = chart.getAxisSet().getYAxis(2);
+            if (!axis.getTick().isVisible())
+                return;
+
+            double value = axis.getDataCoordinate(p.y);
+
+            String yText = new DecimalFormat("+#.##%;-#.##%").format(value); //$NON-NLS-1$
```
Nirus2000, line 337: Maybe it is better if we get formattings from here (analogous to L487)
```suggestion
            String yText = Values.PercentWithSign.format(value);
```

Nirus2000 (approved):

Nirus2000:
Maybe it is better to change the button labels according to the Contributing rules.

BtnLabel... --> [see here](https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#label-descriptions)

```
Messages.LabelCrosshair --> Messages.BtnLabelCrosshair
Messages.LabelMeasureDistance--> Messages.BtnLabelMeasureDistance
```

buchen:
About:

> Messages.LabelCrosshair --> Messages.BtnLabelCrosshair

I wasn't aware of this "guideline" in the contributing readme. I will change it.
While there are some labels that start with "BtnLabel", let's not add more.
Often the label can be used in various places - buttons, tooltips, columns, window titles - and then the "BtnLabel" ist just restricting the usage too much.

In general, my thinking is:
* "Label" for plain labels, "Msg" for longer messages & texts, "Column" for column header
* If an domain area is big enough (and specific enough) it might warrant its own domain-specific prefix (say CSV, PDF, Preferences, Chart, ...)
* When in Rome, do as the Romans do - if there is a domain-specific prefix use it, otherwise do not introduce new domains
 

@OnkelDok  - do not changes anything right away - I'll need some more time to look at the code

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
```diff
@@ -0,0 +1,523 @@
...
+import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.ui.Images;
+import name.abuchen.portfolio.ui.Messages;
+import name.abuchen.portfolio.ui.util.Colors;
+import name.abuchen.portfolio.ui.util.SimpleAction;
+import name.abuchen.portfolio.util.Pair;
+
+public class ChartToolsManager
+{
+    private abstract static class ChartToolBase
```
buchen, line 32: Why not define an interface (instead of an abstract class)?
And we do not need the default implementations - only ```onMouseMove``` is not implemented by all tools


And about the name: maybe we can shorten it to ```ChartTool```?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
```diff
@@ -0,0 +1,523 @@
...
+    public enum ChartTools
+    {
+        NONE("", c -> null, null, null), // //$NON-NLS-1$
+        CROSSHAIR(Messages.BtnLabelCrosshair, p -> new CrosshairTool(p.getLeft(), p.getRight()), Images.CROSSHAIR_ON, 
+                        Images.CROSSHAIR_OFF), //
+        MEASUREMENT(Messages.BtnLabelMeasureDistance, p -> new MeasurementTool(p.getLeft(), p.getRight()),
+                        Images.MEASUREMENT_ON, Images.MEASUREMENT_OFF);
+
+        private String actionText;
+        private Function<Pair<TimelineChart, Color>, ChartToolBase> createToolFactory;
```
buchen, line 62: Let's not introduce more usages for the generic ```Pair``` class.
You can use Java's ```BiFunction``` here:

```
private BiFunction<TimelineChart, Color, ChartToolBase> factory;
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
```diff
@@ -0,0 +1,523 @@
...
+
+            // value for horizontal axis
+            drawCrosshairDateTextbox(e, p);
+
+            drawCrosshairValueTextbox(e, p);
+            drawCrosshairValueSecondAxisTextbox(e, p);
+        }
+    }
+
+    private static class MeasurementTool extends ChartToolBase
```
buchen, line 391: You moved the MeasurementTool from a separate file into a member class.
I also tend to separate classes into smaller pieces rather later (there are some big classes 🫣), but I think the measurement and cross hair tools can have their own files - for readability and maintenance reasons.


If you feel like building more tools, we could also add a "tools" packages below the "chart" package.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
```diff
@@ -0,0 +1,523 @@
...
+        private double getYValue(Point p)
+        {
+            return chart.getAxisSet().getYAxis(0).getDataCoordinate(p.y);
+        }
+
+        private void drawCrosshairValueTextbox(PaintEvent e, Point p)
+        {
+            double value = getYValue(p);
+
+            String yText = new DecimalFormat(Values.Quote.pattern()).format(value);
```
buchen, line 300: We should use the pattern that is currently configured for the given chart:

```
chart.getToolTip().getDefaultValueFormat()
```

This is how it currently looks in the performance chart.

Bildschirmfoto 2023-07-31 um 12 51 15

buchen (changes requested):
Thanks @OnkelDok,

looks good and works good (I tested in on my machine - there it is smooth and fast).

I added some small comments and would appreciate if you can update your pull request.

Sorry for the back and forth on the naming of the labels. I will update the contributing guide.


Andreas.

OnkelDok:
@buchen, thank you for your feedback. I understood them all, but I will answer them and make changes when I have more time.
Maybe this evening, or tomorrow.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
```diff
@@ -786,7 +787,7 @@ public class Messages extends NLS
     public static String LabelKeyIndicators;
     public static String LabelMaxDrawdown;
     public static String LabelMaxDrawdownDuration;
-    public static String LabelMeasureDistance;
+    public static String BtnLabelMeasureDistance;
```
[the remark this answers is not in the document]
OnkelDok: I will revert this to the original field names.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
OnkelDok: Ah yes, don't know why I didn't found the ``BiFunction``. It's much better.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
OnkelDok: Totally agree.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
OnkelDok: Yes, I only tested the security and statement of assets charts. I will take note of it.

buchen:
Merged 👍 

Thanks @OnkelDok for the contribution. I love how we now have the tools manager handling all tools and at the same time have short and concise classes for the tools itself. That looks very maintainable.

PR #3469, opened 2023-07-30
Fixing CONTRIBUTING.md

DavesGithub90:
Removed '.*' at last Type, because it leads to an error when adding in eclipse.


DavesGithub90:
Removing '.*' at end of last Type, because it leads to an error in eclipse

buchen:
thanks for the fix

PR #3472, opened 2023-08-01
Added PDF Extractor for Info Reports from BISON app

buchen:
* Matches crypto currencies based on ticker or name
* Creates new crypto currencies with CoinGecko quote feed


Nirus2000:
i think it looks good for me... @buchen

PR #3473, opened 2023-08-02
Fix AppStream metainfo translations

roshanshariff:
This commit makes some changes to the AppStream metainfo file, mostly regarding translation. The translated summary, description, and release notes are correctly displayed in Gnome Software according to the selected locale, but are unfortunately ignored on the Flathub website (flathub/website#88).

**IMPORTANT**: @buchen, your code for generating release notes will need to be modified.

* Remove the xml:lang="en" attribute from the English language release notes
The release description has to have at least one child without an xml:lang attribute; otherwise Flatpak removes the entire description from the exported metadata file.

* Add German translation for summary and description tags

* Change vcs-browser and bugtracker URLs to use portfolio-performance organization. The summary is machine translated, so please check it.

* Add "contribute" and "FAQ" URLs

### DIFF
--- a/portfolio-product/metainfo/info.portfolio_performance.PortfolioPerformance.metainfo.xml
+++ b/portfolio-product/metainfo/info.portfolio_performance.PortfolioPerformance.metainfo.xml
@@ -3,17 +3,28 @@
   <id>info.portfolio_performance.PortfolioPerformance</id>
   <name>Portfolio Performance</name>
   <summary>Track the performance of an investment portfolio</summary>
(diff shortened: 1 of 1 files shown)


buchen:
> your code for generating release notes will need to be modified.

It was generated by ChatGTP. ChatGTP duly updated the code 😆

roshanshariff:
Oops, I copied that text from the PP homepage and didn't notice the soft-hyphens. I've pushed a new commit removing them.

The machine-translated text is just the summary: "Verfolgen Sie die Performance eines Anlageportfolios" for "Track the performance of an investment portfolio".

buchen:
merged (slight adoption of the translated text)

PR #3474, opened 2023-08-02
Fix quirion dividenden in usd

christen90:
fixed quirion Erträgnisabrechnung to support also currency USD. Shall work also with taxes, but no test case available yet.

I tried to rebase onto your master, but struggled. Could you please extend your contributing guidelines to make this proccess easier to understand?


buchen:
> I tried to rebase onto your master, but struggled. Could you please extend your contributing guidelines to make this proccess easier to understand?

I do not think the process is specific to PP but rather it is plain git operations. Personally, I try never to merge the origin/master into my local feature branch but instead rebase the feature branch on top of origin/master. This way I have to adopt my changes to the changes in the main branch (and not the other way around).

What I did now was do a mixed reset to origin/master. That left your changes as unstaged changes in the working directory. Those are just the 3 files (2 modified, 1 added). I was able to create new commit with you as the author.

Let me try if I can force push this into this commit. Or whether I have to create a new pull request (so that @Nirus2000 also can have a look)

buchen:
Replaced by #3475

PR #3475, opened 2023-08-03
Fix Quirion dividend import in USD

buchen:
Issue: #3474

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/quirinbankag/Dividende04.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/quirinbankag/Dividende04.txt
@@ -1,0 +1,72 @@
+```
+PDFBox Version: 1.8.17
+Portfolio Performance Version: 0.64.4
+-----------------------------------------
+Erträgnisabrechnung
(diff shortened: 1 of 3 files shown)


buchen:
@christen90 - apparently, I had to create a new pull request. Please have a look at your change - I might have mangled up something... 🙄

Nirus2000:
I don't understand this pull request at all.
What is changed here and why are useless parts inserted?

buchen:
I let @christen90 comment on the details.

But what I can see is that the "Dividende04.txt" example **cannot** be imported without the changes to the extractor.

```
Failures:
  QuirinPrivatbankAGPDFExtractorTest.testDividende04:592 Buchungswährung USD passt nicht zu Kontowährung EUR
26.07.2023 Dividende USD 1,57 iShsIII-MSCI EM Sm.Cap U.ETF Registered Shares o.N. Dividende04.txt
Expected: is <OK>
     but: was <ERROR>
```

My reading of the change is: like many other sections that have different format (there are already many Format01 and Format02 methods), there apparently is also a differences is the formatting of the "taxes sections".


> What is changed here and why are useless parts inserted?

Can you specify what you mean by "useless parts"?

buchen:
Let's continue here: #3482

PR #3486, opened 2023-08-05
Improve ATH dashboard widget to support all dataseries

ma4nn:
This small PR improves the All-Time-High (ATH) dashboard widget to support all kind of data series (e.g. whole portfolio, benchmarks, securities).

In addition it slightly adapts translations and moves the dashboard menu entry into subcategory "Performance" to better reflect its purpose.

https://forum.portfolio-performance.info/t/all-time-high-ath-des-ganzen-portfolios/24712


ma4nn:
Totally forgot that there is already [another PR](https://github.com/portfolio-performance/portfolio/pull/3037) from @RomanLangrehr with a similar change 🤦 
Maybe this PR has the benefit that the existing ATH widget was extended and not 2 separate widgets (with duplicated code) are required.

buchen (changes requested):
Thanks for the pull request.

In general, I like the idea to extend the ATH widget.

But I do not think using the PerformanceIndex instead of security prices works:
* the performance index fills in values to complete the data series
* benchmark series are calculated relative to the underlying series
* using the performance index also changes the semantics slightly - it includes dividends (if it is not a benchmark)

buchen:
I read @RomanLangrehr change in #3037 correct, then his pr does ATH on the performance (accumulated percentage).


Thinking about the requirements:

One issue I see with the current code (=before your pull request) is the data series selection: The algorithm could easily work with any security (because it is using the security prices), but piggy-backs on the data series selection for instruments (and because one cannot purchase an index such as the DAX, it currently does not show up). Maybe this is the first thing to fix.

The current change I find confusing from a user point of view:
* I could add an instrument twice - say "BASF" and "BASF (Benchmark)". If I never held BASF, then the ATH will be 0 while the BASF (Benchmark) ATH will be non zero.
* At least my gut feeling is, that users will not expect to include dividend payments in the ATH calculation. If the user then compare PP values to some other websites, the user will see different indicators
* Furthermore, if the user has transactions during the interval, it would change the ATH.
* Additionally, the PR is changing the behavior: Today, the code treats securities as benchmarks. But because it piggy-backs on the data series selection for invested securities, the behavior would change if we allow to select both.

Then: Let's say we additionally add composite data series (Gesamtportfolio, taxonomy-based data series) but keep the security ATH as is (=against the security price). This difference is also not trivial to understand: let's say I have a portfolio with only one instrument. Then the ATH for this portfolio is not identical to the ATH for the instrument itself - the former includes dividends, the latter does not.


I would love to read your reaction to my thoughts.

I think more about it, but I could see the following option:
* Extend the existing widget to include all securities (including indices)
* Add a new widget which shows the ATH for existing investments - but only on Gesamtportfolio, filters, and taxonomies. The description must make clear that it is the total valuation which is looked at

ma4nn:
@buchen thank you for your detailed comments on this PR, very much appreciated and again what learned. 

You're right, probably I was too quick switching the implementation to the `PerformanceIndex` while not being fully aware of all the consequences and testing all the cases:grimacing: 

> I read @RomanLangrehr change in #3037 correct, then his pr does ATH on the performance (accumulated percentage).

Ah yes, right :+1: Personally I find it more interesting to see the distance from the total, e.g. for Gesamtportolio.

> The algorithm could easily work with any security (because it is using the security prices), but piggy-backs on the data series selection for instruments (and because one cannot purchase an index such as the DAX, it currently does not show up).

Hm perhaps I am having problems with the PP domain language here: so with "instruments" you mean "invested securities" whereas indices are "non-investable securities"? 
In my test cases all securities (invested and non-invested) can be chosen in the current ATH widget already (including e.g. `^GDAXI`), with the only exception being indices like "HVPI". Or did I misunderstand something? Can you give me an example how to create the index that is then not choosable for the ATH widget?


> Then: Let's say we additionally add composite data series (Gesamtportfolio, taxonomy-based data series) but keep the security ATH as is (=against the security price). This difference is also not trivial to understand: let's say I have a portfolio with only one instrument. Then the ATH for this portfolio is not identical to the ATH for the instrument itself - the former includes dividends, the latter does not.

Good question. From my perspective it makes no sense to calculate e.g. the Gesamtportolio ATH excluding dividends. But yes, this must then be made more clear in the description somewhere.


> Extend the existing widget to include all securities (including indices)

So basically the contents of the "choose benchmark series" dialog? Not sure if there is really a need to have indices as well there. Of course we could implement that, but I haven't noticed this as a request in the forum or by myself.
My idea in this PR to include the benchmark series was to be able to easily get both the ATH of a security (which I errorneously thought to get by using the security benchmark with the `PerformanceIndex`) and the calculated individual performance ATH of the security (considering purchases, dividends, etc.) - but of course the latter may not be that much of interest. 

> Add a new widget which shows the ATH for existing investments - but only on Gesamtportfolio, filters, and taxonomies. The description must make clear that it is the total valuation which is looked at

This is the more important feature I think. From my usability perspective I would love to only have 1 ATH widget which does cover both cases (because the use case is "distance to ATH" no matter of the underlying data series), but I also understand that it might be too confusing or requiring too much change in the data series selection.

* * *

What do you think about modifying this PR to:
1. Remove the benchmark data series
2. Continue to use the existing logic for security prices and additionally use the `PerformanceIndex` only for Gesamtportfolio/filters/taxonomies ATH calculations within the same widget
3. Add the possibility to the data series dialog to have an introductionary help text that explains the difference between choosing a security (price) and Gesamtportfolio/filters/taxonomies (total valuation) in this case or as an alternative add "(price)" or "(total)" after each element in the tree viewer (same as for pretax currently)


But of course I'm also fine with implementing a separate widget as you suggested if you like to.

uraimo:
Hi, thanks for PortfolioPerformance, great and stable app. I see that there are conflicts now, but since being able to see the ATH delta for the whole portfolio would be an interesting metric and currently it just displays 0%, I would greatly appreciate if this could be considered for inclusion in one of the future releases.

PR #3490, opened 2023-08-06
Fixes update issue in PaymentsView when filter is set

OnkelDok:
See issue report in forum: https://forum.portfolio-performance.info/t/zahlungen-ubersicht-aktualisiert-sich-nicht/24707

The error cause in current PP version is:
When a filter is set, the ``filteredClient`` in ``PaymentsViewModel`` is a readonly-instance which contains a state/snapshot of the account data from the moment in time when filter was set or view was initialized. Then, when the user modifies the account within this view (e.g. create new interest booking) then this new entry is added to the original client, but the ``ReadOnlyAccount`` used for the re-calculation is out-of-date: the new entry is missing in this instance.

So the current solution to fix this is to re-create and set the ``filteredClient`` before the recalculation when ``notifyModelUpdated`` is called.


buchen:
Thanks. Looks good. Back at my desk tomorrow

buchen:
merged

PR #3491, opened 2023-08-06
Fixes initial filter was not used in AllTransactionsView

OnkelDok:
When opening the ``AllTransactionView`` the last saved/checked filter was not used.
The user saw a mismatch between the selected filter (e.g. only one account) and the displayed data (entire portfolio).


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AllTransactionsView.java
```diff
@@ -223,11 +220,22 @@ public FilterDropDown(IPreferenceStore preferenceStore)
             // injected into the view immediately
```
buchen, line 220: I'll also remove this comment - because it is obvisously wrong.
The listener is not called - that's why the filter is selected, but not applied.

I checked the other usages of the trackSelectedFilterConfiguraitonKey.
They explicitly setup the filtering. Therefore, I agree with your change to not change the trackSelectedFilterConfigurationKey method

buchen:
merged

PR #3498, opened 2023-08-11
Display WKN in MasterData

OnkelDok:
https://forum.portfolio-performance.info/t/wkn-im-kursdiagramm/24801

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityDetailsViewer.java
@@ -78,6 +78,7 @@ public class SecurityDetailsViewer
         private Label valueName;
         private Label valueISIN;
         private Label valueTickerSymbol;
+        private Label valueWKN;
 
         @Override
         Control createViewControl(Composite parent)
@@ -89,6 +90,7 @@ public class SecurityDetailsViewer
             valueName = new Label(composite, SWT.NONE);
             valueISIN = new Label(composite, SWT.NONE);
             valueTickerSymbol = new Label(composite, SWT.NONE);
+            valueWKN = new Label(composite, SWT.NONE);
 
             // layout
 
@@ -109,6 +111,7 @@ public class SecurityDetailsViewer
 
             below(valueName, valueISIN);
             below(valueISIN, valueTickerSymbol);
+            below(valueTickerSymbol, valueWKN);
 
             return composite;
         }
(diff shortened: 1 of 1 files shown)


pfalcon:
Just a few words on devil's advocate side: local adhoc id's like WKN, CUSIP are known to be subsumed into ISIN. Even on the screenshot from the thread you link to, one can see: WKN: 110487, ISIN: DE0001104867. So, arguably not much new information is shown if adding WKN to the existing ISIN display (though WKN is not a perfect substring of ISIN :-( ). On the other hand, it takes a precious line at the top of that pane (line useless to anyone outside .de), requiring scrolling to see more info (at least now it's possible to scroll it after https://github.com/portfolio-performance/portfolio/pull/3428).

On the other hand, I find it limiting that info pane shows taxonomies, a one kind of user-defined attribute for securities, and doesn't show other user-defined attributes. IMHO, what would be useful is to add a kind of "security card" to open on double-click with full-size chart, and as much as possible other info, optimized for viewing. (All/most info is available in security edit dialog, but that's optimized for editing.)

OnkelDok:
I just checked some securities: not every ISIN contains the WKN (MSFT, APC).
Maybe this applies only to german stocks (e.g. "Allianz, 840400)"?

Some months ago, @Nirus2000 made some good suggestion for some optimization of the security information pane by moving the MasterData from side panel into the chart title: https://github.com/portfolio-performance/portfolio/pull/3004#issuecomment-1400675368
Maybe in the future this could be a good goal.

If you miss information in the pane, then we can discuss this. Please feel free to create a separate ticket for this to have a better place for a discussion. 
This PR is just a suggestion to add missing information with really low effort. Bigger changes should be discussed in separate ticket or in forum where more people can participate.

Nirus2000:
Hello @OnkelDok 
I see this just like you do. Your changes are good and meaningful for now. 👍🏻 

Following up on my comment #3004, we could start by opening an ISSUE and adding it to our TODO list. In a brief phone call, I already spoke with @buchen about opening a project with the top 5 themes, where we can address projects like this one. I would place this under the top 3. I've mentioned the explanations for this before, but we should further discuss the content and milestones. Very good.
Could you please open an ISSUE for this, which we could then integrate into the project list? I think that's where the initial thoughts could develop. Additionally, we could schedule a small team meeting later to exchange thoughts again.

Thank you again... ;-)

@pfalcon
And once again, an unnecessary comment on your part to a pull request that is in no way related. 
A column "WKN" has been added to an existing table. The commit has nothing to do with the security chart and its info card.
Please use the ISSUE system for your thoughts. 👍🏻 Thanks!

buchen:
Thanks for the implementation. Will merge.

I have created #3503 to discuss and track the evolution of the security details viewer.

PR #3504, opened 2023-08-14
Fix small typo in AppStream metadata

roshanshariff:
The "description" tag for the 0.65.0 release is misspelled "decription".

Also add an introductory paragraph to the app description to make appstream-util validate happy.


buchen:
grr. Sorry. Merged. The new binary should include the update upstream metadata (even though the tag seems to look like it does not)

Issue #2216, opened 2021-04-14
anybody interested in PP with a database?


MichaelMeier99:
**Is your feature request related to a problem? Please describe.**
PP is wonderful.
Never the less it is slow in opening and the data take much space

**Describe the solution you'd like**
It would be a boost for performance and power to compact everything into a Firebird database.
The data would load on scratch.
The options to get information about our portfolio boost to limitless (SQL).

**Additional context**
The time and code this complete change of data-storage is enourmous. I cannot handle it alone. And it will take a handful of beta-testers until we can dare to offer the code to our public.
Is there anybody out there, who likes the project as well?

What I can give: 
- a well tested table set specializing in exchange-rates (date, open, high, low, close, volume,....) in Firebird
- SQL codes for saving new data into Firebird
- SQL codes for grabbing information from Firebird

What I do not know how to at the moment
- interface to the rate-grabbing tools into the method, which will perform the SQL-write transactions (in other words: yahoo into the method which will write the data to Firebird)
- interface after having read the information (eg. into text oder figures) to import it into the PP environment

So please let me know 2 things for now:
1) Do you like the idea?
2) Are you willing and able to work with me?

RomanLangrehr:
I really like the idea. I also think it would be useful to store historical open, high, low, close, volume for the (at the moment, PP only stores the close quote. High, low and volume are only stored for the most recent date and open is not stored/fetched) at all. They could be used to calculate, for example, maximum drawdown more accurately, and you could display the trading volumes (and candlestick charts instead of normal charts) in the chart tab.

However, PP stores a lot of different information in the xml file (GUI settings, transactions etc.) and thus I think "compact everything" in a SQL database would require a quite complex database scheme and might be less flexible (and a lot of work). So I think, at least for a starter, we should only put the historical quotes in a database.

Pnda87:
I also like the idea to separate the data into database tables: quotes, securities, transactions, ....
GUI information for PP could stay in the xml file

debegr92:
There was also a comment in a Telegram chat, that it would be great to separate quotes and transactions. With this model, all securities with historical (OHLC+V) values can be stored very efficient in a SQLite file for example.
Helpful for investors with more than one account (kids etc.) but the same stocks.

MichaelMeier99:
Thank you for the comments!

"Candlestick", Drawdowns, Runups... These are old wishes of mine.
[...]

PR #3510, opened 2023-08-15
Store time periods in xml file not in settings file, fixes #2952

### LINKED ISSUE #2952: Store time periods in xml file not in settings file
I'm using PP on different machines with very different screen resolutions (4K and FHD). And I'm heavily relying on the time periods for the analysis. Now there are two equally bad options: 

- Save the settings file beside the xml: I can sync the time periods but I'm also syncing the sizes of different views (like sidebar, graph, etc.) 
- Don't save the settings file: The opposite from above. I don't sync the view sizes and don't sync the time periods either. 

In my opinion the time periods are xml-file specific and not machine specific. In addition the time periods are used for some widgets in the performance overview. Moving the xml file without the settings file breaks the performance dashboard. Therefor the time periods should be saved in the xml file.

OnkelDok:
I am working on #2952.
Closes #2952

This is not the final state, therefore I set the PR to draft until I think it's ready for a full review. But I need a place to ask my questions and I don't want to use the issue for this.

I have following questions to continue my work:
* PP currently stores ``TIME_PERIODS`` in the settings file and ``SELECTED_TIME_PERIOD`` in the file ``copy.e4xmi`` in the workspace.
**Question: is the goal for this task to migrate both settings into the XML? Or only one of them?**
* For migrating the ``SELECTED_TIME_PERIOD`` we need the ``MPart`` in the migration logic. To get this I had to change the instance creation order of ``ClientInput`` and the ``MPart``. Here I don't know if this new order is allowed or if the old order (``MPart`` is created after the ``ClientInput``) is important.
* In general: how can I link this PR and the issue to see the PR as linked to the item in our project? Using ``fixes #2952`` in commit message did not make it. And linking it manually using the side menu is not possible. I cannot click on ``Development``.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import javax.inject.Inject;
 
@@ -65,6 +66,7 @@ import name.abuchen.portfolio.ui.wizards.client.ClientMigrationDialog;
 
 public class ClientInput
 {
+
     public static final String DEFAULT_RELATIVE_BACKUP_FOLDER = "backups"; //$NON-NLS-1$
 
     private String label;
@@ -75,7 +77,7 @@ public class ClientInput
 
     private PreferenceStore preferenceStore = new PreferenceStore();
     private ExchangeRateProviderFactory exchangeRateProviderFacory;
-    private ArrayList<ReportingPeriod> reportingPeriods; // NOSONAR mutable
+    private ReportingPeriods reportingPeriods;
 
     private boolean isDirty = false;
     private List<Job> regularJobs = new ArrayList<>();
@@ -213,7 +215,6 @@ public class ClientInput
                 if (preferences.getBoolean(UIConstants.Preferences.CREATE_BACKUP_BEFORE_SAVING, true))
                     createBackup(clientFile, "backup"); //$NON-NLS-1$
 
-                storeReportingPeriods();
                 ClientFactory.save(client, clientFile);
                 storePreferences(false);
 
@@ -253,7 +254,6 @@ public class ClientInput
         BusyIndicator.showWhile(shell.getDisplay(), () -> {
             try
             {
-                storeReportingPeriods();
                 ClientFactory.saveAs(client, clientFile, pwd, flags);
                 storePreferences(true);
 
@@ -289,7 +289,6 @@ public class ClientInput
         BusyIndicator.showWhile(shell.getDisplay(), () -> {
             try
             {
-                storeReportingPeriods();
                 ClientFactory.exportAs(client, localFile, null, flags);
             }
             catch (IOException e)
@@ -354,7 +353,6 @@ public class ClientInput
             String backupName = constructFilename(clientFile, "autosave"); //$NON-NLS-1$
             File autosaveFile = clientFile.toPath().resolveSibling(backupName).toFile();
 
-            storeReportingPeriods();
             try
             {
                 ClientFactory.save(client, autosaveFile);
@@ -533,18 +531,25 @@ public class ClientInput
         }
     }
 
-    public ArrayList<ReportingPeriod> getReportingPeriods()
+    public ReportingPeriods getReportingPeriods()
     {
         if (reportingPeriods != null)
             return reportingPeriods;
 
+        reportingPeriods = new ReportingPeriods(this, loadReportingPeriods());
+        return reportingPeriods;
+    }
+
+    /* package */ ArrayList<ReportingPeriod> loadReportingPeriods()
+    {
         ArrayList<ReportingPeriod> answer = (ArrayList<ReportingPeriod>) client.getSettings()
                         .getConfigurationSet(WellKnownConfigurationSets.REPORTING_PERIODS) //
                         .getConfigurations().map(c -> reportingPeriodStringToReportingPeriod(c.getData())) //
                         .filter(java.util.Optional::isPresent) //
                         .map(java.util.Optional::get).collect(toMutableList());
 
-        // if periods not in client file, load them from settings file (legacy)
+        // if periods not in client file, load them from settings file
+        // (legacy)
         boolean storeBack = false;
         if (answer.isEmpty())
         {
@@ -557,16 +562,14 @@ public class ClientInput
             answer = defaultReportingPeriods();
         }
 
-        reportingPeriods = answer;
-
         // immediately store periods in case they were loaded from legacy
         if (storeBack)
         {
-            storeReportingPeriods();
+            storeReportingPeriods(answer.stream());
             touch();
         }
 
-        return reportingPeriods;
+        return answer;
     }
 
     private static ArrayList<ReportingPeriod> defaultReportingPeriods()
@@ -601,14 +604,12 @@ public class ClientInput
                         .map(java.util.Optional::get).collect(toMutableList());
     }
 
-    private void storeReportingPeriods()
+    /* package */ void storeReportingPeriods(Stream<ReportingPeriod> periods)
     {
-        if (reportingPeriods == null)
-            return;
-
         var set = client.getSettings().getConfigurationSet(WellKnownConfigurationSets.REPORTING_PERIODS);
         set.clear();
-        reportingPeriods.forEach(rp -> set.add(new Configuration("", "", rp.getCode()))); //$NON-NLS-1$ //$NON-NLS-2$
+        // null values to not generate XML tags for uuid and name
+        periods.forEach(rp -> set.add(new Configuration(null, null, rp.getCode())));
     }
 
     @Inject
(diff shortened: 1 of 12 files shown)


buchen:
Hi @OnkelDok,

sorry for replying only now - we were in the mountains with only my mobile - things like this PR need my development machine.

> This is not the final state [...] But I need a place to ask my questions and I don't want to use the issue for this.

I love it. That makes it easy to discuss on source code 😄 


> PP currently stores TIME_PERIODS in the settings file and SELECTED_TIME_PERIOD in the file copy.e4xmi in the workspace.
>
> Question: is the goal for this task to migrate both settings into the XML? Or only one of them?

The selected time period is stored with the part, because the user can have multiple parts (= windows) open. Then one window can show 2022 and the other 2023 (or whatever time periods to compare). If we have only one property, then a) all windows would have the same reporting period after opening and b) and we would have to dynamically update _all_ windows if the selected time period changes (essentially preventing the user to compare data).

I am open for suggestions, but my initial reaction is:
* migrate the set of configured time periods into the client file (XML)
* keep the selected time period with the persisted data of the part (= allow time period per window)

If the file is opened on a new computer, then the set of periods is available (specifically configured time periods) but the selection does not change. I think it should be reasonable easy to then pick the time period the users wants.


> in the file copy.e4xmi

Some background here: the Eclipse framework is storing the "state" of the application component tree in the e4xmi file. Usually, this is loaded by the framework by default. However, when doing an online update of PP, we have to reset the e4xmi file because the update might introduce new menu items, keyboard shortcuts, etc. When resetting, then also the window layout and the currently opened files are lost. Therefore I always save a copy when closing PP. When PP starts, I check if a new version is detected. If so, I try to merge the part layout into the new e4xmi file before rendering the setup.

Again: my gut feeling is not to change this process for the reasons outline above (allow different reporting periods per window). I am open for ideas.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
```diff
@@ -86,14 +89,17 @@ public class ClientInput
...
     private IEclipseContext context;
 
+    private MPart part;
+
     @Inject
     @Preference
     private IEclipsePreferences preferences;
 
-    /* protected */ ClientInput(String label, File clientFile)
+    /* protected */ ClientInput(String label, File clientFile, MPart part)
```
buchen, line 98: Keep in mind that there can be many parts for one client input.
A part represents a window. There can be many windows open, but the file is only opened once.
Therefore I think the part should not be added to the ClientInput.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
```diff
@@ -529,17 +535,88 @@ private File getPreferenceStoreFile(File file) throws IOException
...
+        for (int ii = 1; ii <= 3; ii++)
+            answer.add(new ReportingPeriod.LastX(ii, 0));
+        return answer;
+    }
+
+    private static String reportingPeriodsToConfigurationString(List<ReportingPeriod> periods)
+    {
+        StringBuilder buf = new StringBuilder();
+        for (ReportingPeriod p : periods)
+            buf.append(p.getCode()).append(';');
```
buchen, line 609: As we use the configuration sets for storing the data, I wonder if we do not need to concatenate the reporting periods anymore. We could have a configuration set "REPORTING_PERIODS". Each configuration is one reporting period. I understand that we do not really need all the different property (uuid, name, data), but then let's leave them empty.

What do you think?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/TimePeriodMigration.java
```diff
@@ -0,0 +1,36 @@
+package name.abuchen.portfolio.ui.editor;
+
+import org.eclipse.e4.ui.model.application.ui.basic.MPart;
+import org.eclipse.jface.preference.PreferenceStore;
+
+import name.abuchen.portfolio.model.Client;
+
+class TimePeriodMigration
```
buchen, line 8: We can do the migration here. Maybe we can also migrate lazily in the ClientInput.

The logic in ClientInput could be:
* if the new configuration set exists, use it
* otherwise use the reporting time periods from perference
* otherwise create the default reporting periods

Then we do not need to a create a new client version. It is just in the getReportingPeriod method in ClientInput


(This was not possible with the client filter, because with the client filter we introduce new UUIDs as a replacement for the auto-generated identifier. Plus: the client filters were used as data series in the chart. But with the reporting period we already have identifier (the code) and we do not have to migrate any configuration in the XML file itself).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
OnkelDok: Agree. As written in https://github.com/portfolio-performance/portfolio/pull/3510#issuecomment-1686881377. When there can be multiple parts, then the current migration logic does not make any sense.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
OnkelDok: I also thought about this, but decied to let this as it is . Because current serialization seems light a lightwight solution.
But if its okay to use separate ``Configuration``s for each period, then I can change this.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/TimePeriodMigration.java
OnkelDok: I understand what you mean. This "Lazy"-Migration is better. Thx for the idea.

buchen:
> You mean the user can create multiple windows for a single XML, each with different reporting period?
> Crazy, I did not knew that. How to do it?

Right click on window title. One can drag and drop the parts within the window to see them next to each other.

Bildschirmfoto 2023-08-22 um 09 45 40

buchen:
@OnkelDok - we can remove the "draft" status, right? I haven't looked into the details yet, but from your last commit message (closes ....) I assume that you see the change as ready?

OnkelDok:
> @OnkelDok - we can remove the "draft" status, right? I haven't looked into the details yet, but from your last commit message (closes ....) I assume that you see the change as ready?

@buchen: no, it's not finished. We need some small additional changes. But I have not found time the last days to push it forward. I've inserted a TODO comment.
Currently the ``storeReportingPeriods`` is called after the client file was saved. So when we keep this lazy migration, then we have to store the reporting periods always. When we would save the settings only when they were changed (time period added, removed or reordered) then with this lazy migration the settings will only arrive in the client file when reporting period lists is changed.
I cannot say if I'll find time for this today.

buchen:
> Currently the storeReportingPeriods is called after the client file was saved. So when we keep this lazy migration, then we have to store the reporting periods always. When we would save the settings only when they were changed (time period added, removed or reordered) then with this lazy migration the settings will only arrive in the client file when reporting period lists is changed.

Ah, Okay, I think I understand. My gut feeling
* We can mark the client file as "dirty" on the first (lazy) migration
* Later, we should make the client file only dirty (= needs saving) if the reporting periods actually changed

buchen:
> I cannot say if I'll find time for this today.

No worries. It's all on our spare time

OnkelDok:
Now I think its ready for review.
I added storing the periods into when they were loaded from legacy string. And also added calling ``client::touch`` when list of reporting periods were changed (in ``ReportingPeriodDropDown``).

buchen:
> To implement the bigger change I planned to create a container ReportingPeriods for the periods-list and pass this container to all other modules which are using the periods (instead of passing the ArrayList).

I like that idea. I think it make sense to a have a container ReportingPeriods. I think it does not have to implement the List interface (as you did).

Inside this container, we could track the reporting periods with an object that keeps both - the reporting period, and the configuration. Then we always know if which underlying record we have to update.

So riffing on your idea:

```java
public class ReportingPeriods
{
    private record Item(ReportingPeriod period, Configuration config) {}
    
    private Client client;
    private List<Item> items;

    private ConfigurationSet set;

    public ReportingPeriods(Client client)
    {
        this.client = client;
        this.items =  // initial loaded from configuration set
                        
        this.set = // well known set
    }

    public void add(ReportingPeriod p) {}

    public Stream<ReportingPeriod> stream() { }

}
```

But thinking about it: I think your solution is simpler. If something changes, then just override the whole configuration set. It will be fast enough anyway. Managing the individual Configuration items is unnecessary complex.

 ---> go ahead and implement your proposal


> So I decided to go the "easy" way as first suggestion.

Completely agree. I often start coding the easy solution first - then it creates a feeling if that will work. If not, only then invest time and energy in a more complicated solution. On Github it is harder, because people tend not to use draft pull requests that much.

buchen:
Thanks @OnkelDok for the contribution.
It now looks very nice - with minimal code.

I have changed one smaller thing: I moved the store / load methods for the periods into the ```ReportingPeriods``` class. The class was the only one using them anyway. It felt they are more encapsulated there.

pfalcon:
@OnkelDok, @buchen : FWIW, I see that these time period setting are stored somewhat inconsistently wrt all other settings:
```
      <entry>
        <string>reporting-periods</string>
        <config-set>
          <configurations>
            <config>
              <data>L1Y0</data>
            </config>
            <config>
              <data>L2Y0</data>
            </config>
```
As can be seen, only `<data>` element is defined, whereas all other settings also have:
```
            <config>
              <uuid>HoldingsPieChartView</uuid>
              <name></name>
              <data>a24e513e-0f43-46e0-8e3b-032142325</data>
            </config>
```
I.e. there's always `uuid` and `name` elements, even if it's not really uuid and name is empty.

Such discrepancies are easily visible when mapping XML content onto a DB model https://github.com/portfolio-performance/portfolio/issues/2216#issuecomment-1703819271 .

Probably not a big deal, but raises (known) questions of XML model consistency and how easy users can access data in it.

PR #3515, opened 2023-08-20
Feature - Deactivation of notes in the importer and store values

### LINKED ISSUE #3500: Deactivation of notes in the importers
Add a checkbox in settings to disable notes extracted by PDF and IB-Flex import.
With the CSV Importer, the user can decide for himself by configuring the splitter. therefore, the feature is not needed here.

**Default value is:** Notes are set by importer.

_Perhaps this deactivation should be made selectable on every importer. 
PDF importer: deactivation of notes
IB-Flex-Importer: deactivation of the notes_

Nirus2000:
Issue #3500 
Closes #3500

~~Allows the user to permanently disable the extraction of notes in the PDF and IB-Flex Query Importer.~~
Allows the user to permanently disable the extraction of notes in the importer wizzard.

Hello @buchen ,
I created a first step for this little feature.
However, I have a problem where I can best implement the function appropriately in the PDF importer and in the IB-Flex importer.


Nirus2000:
if i implement the function in the AbstractPDFExtractor.java, i get a lot of errors. 🦧🦧🦧🦧

```
            for (Item item : items)
            {
                Annotated subject = item.getSubject();

                if (subject instanceof Transaction tx)
                    tx.setSource(filename);
                else if (subject instanceof CrossEntry entry)
                    entry.setSource(filename);
                else if (subject.getNote() == null || trim(subject.getNote()).length() == 0)
                    item.getSubject().setNote(filename);
                else
                    item.getSubject().setNote(trim(item.getSubject().getNote()).concat(" | ").concat(filename)); //$NON-NLS-1$
            }
```

change to this
```
            for (Item item : items)
            {
                Annotated subject = item.getSubject();

                if (subject instanceof Transaction tx)
                    tx.setSource(filename);
                else if (subject instanceof CrossEntry entry)
                    entry.setSource(filename);

                // Automatic import of notes disabled
                if (!PortfolioPlugin.getDefault().getPreferenceStore()
                                .getBoolean(UIConstants.Preferences.PDF_IMPORTER_NOTES))
                {
                    item.getSubject().setNote(null); //$NON-NLS-1$
                }
            }
```

buchen (changes requested):
See my comment on the ImportController above

Nirus2000:
Hello @buchen
Absolutely... this is a much better implementation.
I have implemented the changes.

Is it still necessary to save the variable now?

Regards
Alex

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
```diff
@@ -1516,6 +1516,8 @@ LabelRemoveDividends = Generate an additional withdrawal for dividend transactio
 
 LabelRemoveLogo = Remove logo
 
+LabelRemoveNotesInTransactions = Set notes (if available)
+
```
buchen, line 1520: This is flipping the meaning - the variable is called "remove notes", the text says "do not remove" notes. It works because at some later point the wording is flipped back.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/actions/InsertAction.java
```diff
@@ -44,6 +45,11 @@ public void setInvestmentPlanItem(boolean flag)
         this.investmentPlanItem = flag;
     }
 
+    public void setRemoveNotesInTransactions(boolean flag)
+    {
+        this.addNotesInTransactions = flag;
+    }
+
```
buchen, line 52: This is the other place where the meaning is flipped: the method says it configures whether to remove notes, but the internal flag then is then whether to add notes.

buchen (commented):
Hi Nirus,

I have picked up the change and did some changes (therefore it turned out to be a new commit - I put both of us as co-authors).

The major changes are:
* I remove the separate preference page. Instead, the last selection of the user in the dialog is remembered. No need to go to a different page.
* The preferences are removed *per* importer. Maybe one wants to import notes for bank X but not bank Y. Same to convert deliveries.
* No need to pass the flag to the InsertAction - instead the notes are simply removed from the selected entries before.

buchen:
changed, rebased, and merged

PR #3517, opened 2023-08-22
Use crosshair cursor for chart tool

ma4nn:
This little PR suggests to use the crosshair cursor for the new chart tools in order to better recognize that the chart tools are active and for easier positioning of the mouse cursor.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/ChartToolsManager.java
@@ -13,6 +13,7 @@ import org.eclipse.jface.action.ToolBarManager;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.PaintEvent;
 import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.Cursor;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Event;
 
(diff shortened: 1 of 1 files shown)


OnkelDok (approved):
Cool improvement.

This is how it looks like:

https://github.com/portfolio-performance/portfolio/assets/90478568/a8505167-3ef0-4ba9-9723-15a470c8b751

pfalcon:
I guess it depends on your desktop environment and specific theme whether this is improvement or not. With @OnkelDok's video, with a simple "hairline" crosshair cursor, it definitely looks useful. But here on Linux, with MATE desktop, crosshair cursor is pretty big and actually has a shape of "aim target" with white outline around it, so it's actually pretty hard to see where you position it, as it all obscures context around your point, which is of course important for positioning. Simple pointer cursor worker better here...

OnkelDok:
@pfalcon can you provide a screenshot to get an idea how it looks?

pfalcon:
> @pfalcon can you provide a screenshot to get an idea how it looks?

Thanks for paying attention to this. Producing a screenshot however turned out a bit of chore, as standard Gnome/MATE tool didn't capture the custom crosshair cursor, and instead the screenshot had standard pointer cursor. So, I had to revert to making a photo with a phone:



All in all, I don't know what to make out of it. So, it's mostly FYI that sometimes even obviously simple improvements can be not such for some people...

pfalcon:
@OnkelDok : 

> @pfalcon can you provide a screenshot to get an idea how it looks?

Thanks for paying attention to this. Producing a screenshot however turned out a bit of chore, as standard Gnome/MATE tool didn't capture the custom crosshair cursor, and instead the screenshot had standard pointer cursor. So, I had to revert to making a photo with a phone:



All in all, I don't know what to make out of it. So, it's mostly FYI that sometimes even obviously simple improvements can be not such for some people...

ma4nn:
@pfalcon interesting issue thanks for raising 👍 

But isn't this more a problem of your used theme/[mouse cursor styles](https://wiki.ubuntuusers.de/Mauszeiger/)? Imho the crosshair cursor should be meant for exactly those use cases, e.g. [stock3](https://terminal.stock3.com/#Index) is also using it. Also painting tools like GIMP make use of the crosshair cursor, so you might have difficulties in those tools too?

Do other users see similar issues?

PR #3532, opened 2023-09-01
Stretch text fields for ISIN, Ticker and WKN in SecurityMasterDataPage

OnkelDok:
https://forum.portfolio-performance.info/t/eingabefeld-symbol-in-der-eingabemaske-verlangern/24916

EDIT:
After initial commit I realized that with my change of the ISIN textinput I removed the validation of the ISIN value.
@buchen: If you generally would accept the stretching, then I will find a way to have both: stretch and validation. Just let me know.


buchen:
Hi @OnkelDok,

thanks for picking this up - it is one of those "paper cuts". It is super annoying that the field is not very long - but it is also super easy to fix (as opposed to "just" support bonds).

Personally, I did not like the stretching that much. It fills up the dialog unnecessary with a field which typically contains 10 characters at most. I now increased the width to 30 characters. That should be more than enough.

Bildschirmfoto 2023-09-02 um 08 28 16


Andreas.

buchen:
(I put the two of us as authors of course)

PR #3545, opened 2023-09-11
Modify DKB PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-dkb/4449/100

DKB Bank has redesigned its PDF documents as of September 2023.

We therefore separate the functions with
   XXX_Format01
and
   XXX_Format02

To close the chapter of the old document formats cleanly, the source code is formatted according to the standard Eclipse formatting.


buchen:
> DKB Bank has redesigned its PDF documents as of September 2023.

Too bad.

Just a note for the future: if there are more differences, maybe it is easier to have 2 extractors: one for "old" and one for "current" documents. Then each files remains smaller and easier to maintain. Of course, the tests should run both importers and check that still the right one is chosen to import the files.

PR #3548, opened 2023-09-13
[PDF] [Deutsche Bank] Fix detection of currency for account statement

buchen:
The latest account statements cannot be imported because the currency was not detected. Apparently the latest document (can) introduce a line break between account number and final balance.

With this change, the currency is retrieved from the header of the document. However, to enable this we need to check for two lines.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
@@ -2299,4 +2299,38 @@ public class DeutscheBankPDFExtractorTest
                         "Deutsche Bank Privat- und Geschäftskunden AG", "GiroKontoauszug05.txt");
         assertEquals(expectedErrorMessage, firstError.getMessage());
     }
+    
+    @Test
+    public void testGiroKontoauszug06()
+    {
+        DeutscheBankPDFExtractor extractor = new DeutscheBankPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        List<Item> results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "GiroKontoauszug06.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(results.size(), is(3));
+
+        // check dividends transaction
+        assertThat(results, hasItem(removal( //
+                        hasDate("2023-08-14"), hasShares(0), //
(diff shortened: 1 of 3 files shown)


buchen:
Hi @Nirus2000,

please have a look. This change attempts to fix a NPE when importing the latest account statement. The reason is that the currency is not detected, the transaction is created without currency, that crashes the UI later.

Besides this change, I am thinking about two more (separate) changes.
I would love to get your feedback there.

First, as you can see the following block is imported as a **deposit** but it is clearly a dividend payment.

```
16.08. 16.08. Verwendungszweck/ Kundenreferenz + 33,23
2023 2023 ZINSEN/DIVIDENDEN/ERTRAEGE FIL/DEPOT-NR:
218/338781300IS.E.R.G.G.1.5-2.5Y U.ETF DE
INH.ANT.E UR(D.)
```

What is your proposal how to address this?

A) Do not import that transaction at all (and thereby rely on the fact that the user is separately importing the document with the dividend payment)?
   * Pro: if the user imports all documents, then he has no work to remove the wrong transactions
   * Contra: The document contains three entries, but only one is created - it might be confusing why 2 are missing

B) Create a dividend transaction - but with reduced information (it will be hard to match to an existing security as there is no ISIN, etc.)?
   * Pro: If the user were only to import the account statement, at least the dividend is there for matching
   * Contra: Every time I have a manual action to not import this transaction - which is annoying

How do you do this in other importer?




Second, I think we should have a central check that no transaction is created with a "null" currency. Because that crashes the UI. I'll have a look where that check could make sense. The challenge is that the rendering requires the currency to be set.

stoeggich:
sorry for the mix. But many other imports are handled in such a way that no books are imported twice. In the last or penultimate update, this was also adjusted for stock purchases, for example [https://github.com/portfolio-performance/portfolio/pull/3537](url)

Nirus2000:
Hello @buchen
exactly as @stoeggich has written.

We don't extract transactions from bank statements, only in and out deliveries. This is true for all importers. 
The exception is a depot statement.

For the transaction of securities/dividends. We cannot determine fees or taxes via the bank statement. It also does not work this way for foreign currencies of securities.

Regards
Alex

buchen:
> Please create a PDF debug and post it in the forum or create an issue.

Haha... it is already part of the pull request.


Sorry for mixing up two discussions: **this** pull request is about fixing the currency detection. While doing this, I noticed the behavior to import dividend transactions as deposit transactions (I see you created a separate pull request for the skipping).

buchen:
Okay, I see in the other pull request you copied the debug output and also changed the detection. This duplicates the pull requests.

I wasn't quite sure if it is reliable enough to go match the IBAN. But I trust your experience here. It certainly makes it easier

PR #3550, opened 2023-09-13
[PDF] [Deutsche Bank] Prevent transactions from statement if no currency is found

### LINKED ISSUE #3549: Prevent faulty transaction without currency to crash Import UI with NPE
If an importer, for whatever reason, cannot extract the currency, then currently the UI crashes.

See discussion in #3548

buchen:
The transaction year and transactions currency is parsed from the document header into the context. Until now, the information is not checked if it is actually found and leads to faulty transactions.

Closes #3549


buchen:
This is how it is shown to the user:

Bildschirmfoto 2023-09-13 um 20 45 30

If the proposal makes sense, I can also to the translation (or we leave it in German because it is about German documents)

buchen:
Obsolete with #3556

PR #3554, opened 2023-09-14
PDF parser: add required attributes from the document context to section

buchen:


buchen:
As discussed, this change allows to "require" document context attributes.

It really does not reduce the number of lines of code in the extractor - but it fails with better error messages if a property such as currency is actually missing.

What do you think?
Merge? Or not worth it?

buchen:
> Is it still possible that documentContext could also be optional?

If it is required, we could add a

```
.optionalDocumentContext("currency")
```

I was trying to avoid it, because it makes the code harder to understand and maintain.

Why? The attributes of the sections are always required. The flag "optional" is an all-or-nothing flag on the whole section. Now we would add an optional flag no the attributes level - but only on some attributes. From a semantic point of view it becomes kraut and rüben.

Plus: the added value of defining an optional document attributes is very small. For required document attributes we fail if they do not exist - but obviously we do not check this for optional attributes. So it boils down to writing ```v.get("X")``` instead of ```type.getCurrentContext().get("X")``` which is a difference of only 23 characters.

Nirus2000:
Okay, i understand.
Let's start...

PR #3555, opened 2023-09-14
Revision ComDirect PDF-Importer

### LINKED ISSUE #1131: java.lang.UnsupportedOperationException, wenn Steuermitteilung durch PDF Import 0,00€
image

Wenn ich eine Steuermitteilung von comdirect importiere, die aber 0,0€ umfasst, wird diese so eingetragen. Wenn ich aber anschließend auf das Wertpapier klicke, kommt es zum angezeigten Fehler. Öffne ich die Steuermitteilung manuell, bekomme ich auch die Meldung, dass 0,00€ kein valider Wert ist.

### LINKED ISSUE #3313: Comdirect PDF Importer: use of manually changed test documents; use of tax documents
Keep track of changes needed for the Comdirect Importer:

@Nirus2000 commented in #3305

>  2. the problem is here also that in the ComDirect still some "manipulated" PDF debugs are and the tests say "Ok", but are not actually.
>
> 3. also tax documents were misused for dividend transactions or sales.
>
> 4. also withholding taxes appear in the regular dividend transactions, but no longer in the tax treatment.

### LINKED ISSUE #3553: Revision of ComDirect Importer
ComDirect provides two documents for the transaction.
The security transaction and the taxes treatment.
Both documents are provided as one PDF or as two PDFs.

The security transaction includes the fees, but not the correct taxes
and the taxes treatment includes all taxes (including withholding tax),
but not all fees.

Therefore, we use the documents based on their function and merge both documents,
if possible, as one transaction.

Always import the securities transaction and the taxes treatment for a correct transaction.
Due to rounding differences, the correct gross amount is not always shown in the securities transaction.

Analogous to the Targo Importer, modify the Comdirect Importer so that the postprocessing() is changed for dividends or sales and the taxes are also posted (make one out of two transactions).

Assign incorrectly used documents correctly.
Tax notices should also be only tax notices and not also misused for dividend transactions.

Nirus2000:
Closes #3553
Closes #3313
Closes #1131

https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/229
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/236
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/242
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/254
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/270
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/276
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/283

ComDirect provides two documents for the transaction. The security transaction and the taxes treatment.
Both documents are provided as one PDF or as two PDFs.

The security transaction includes the fees, but not the correct taxes and the taxes treatment includes all taxes (including withholding tax), but not all fees.

Therefore, we use the documents based on their function and merge both documents, if possible, as one transaction.

Always import the securities transaction and the taxes treatment for a correct transaction. Due to rounding differences, the correct gross amount is not always shown in the securities transaction.

@buchen
Many work... so plz check the postProcessing if you find any problems.

Maybe it's possible to leave a note about this in the next changelog.

---

A special thanks goes to those who sent me more than 400 PDF debugs.

**Forum user:** 

- [jensinho](https://forum.portfolio-performance.info/u/jensinho)
- [DonQuijote](https://forum.portfolio-performance.info/u/donquijote)
- Manuel ... i dont know the user name in PP-Forum
- Max ... i dont know the user name in PP-Forum
Maybe you find by email adress the user in the PP-Forum userlist


Nirus2000:
Hello @buchen 
One point occurred to me today, where I still have no solution in mind.

**Scenario:**
When importing on the **same day** with same security...
- 1x dividend transaction without taxes treatment
- 1x tax treatment of sale
...is imported.

I think when creating the maps in postProcessing, it still needs to be identified somehow if the taxes treatment is a purchase/sale or a taxes treatment of a dividend.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1155,170 +1728,332 @@ else if (v.get("note1").matches("^(?i:Visa\\-Kartenabre.*)$"))
...
+                if (transactions.size() == 1)
+                {
+                    if (transactions.get(0).getSubject() instanceof BuySellEntry portfolioTransaction)
                     {
-                        a2 = (AccountTransaction) similarTransactions.get(1).getSubject();
+                        BuySellEntry sellTransaction = (BuySellEntry) transactions.get(0).getSubject();
+
+                        // Set the time to the transaction date if available
+                        if (sellTransaction.getPortfolioTransaction().getSecurity().getAttributes()
+                                        .get(new AttributeType(ATTRIBUTE_DATE_TIME)) != null)
```
buchen, line 1914: I understand the reasoning: the sale document contains a time, the tax document does not.

However, storing the date time as attribute does not work for multiple reasons: first: what happens if there are multiple sale transactions for one security - which date/time is winning? Second: the attributes are polluting the XML.

Maybe a better way is to create the grouping not on LocalDateTime but on LocalDate. In the process, it is "just" stripping the time. But the transaction keeps the original time. Something along these lines:

```java
// Group dividend and taxes transactions together and group by date and security
Map<LocalDate, Map<Security, List<Item>>> dividendTaxesTransactions;
[...]
dividendTaxesTransactions = Stream.concat(dividendTransactionList.stream(), taxesTransactionList.stream())
   .collect(Collectors.groupingBy(item -> item.getDate().toLocalDate(), Collectors.groupingBy(Item::getSecurity)));
```

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1155,170 +1728,332 @@ else if (v.get("note1").matches("^(?i:Visa\\-Kartenabre.*)$"))
...
+        Iterator<Item> iter = items.iterator();
+        while (iter.hasNext())
         {
-            Annotated s = txi.getSubject();
-            if (s instanceof AccountTransaction a)
+            Object o = iter.next().getSubject();
+            if (o instanceof AccountTransaction a)
             {
-                return AccountTransaction.Type.DIVIDENDS.equals(a.getType());
+                if (ATTRIBUTE_TO_BE_DELETED.equals(a.getNote()))
```
buchen, line 2051: While the deletion logic is not wrong, I wonder if we can make it easier by using a Set:

```java
Set<AccountTransaction> toBeDeleted = new HashSet<>();

...

toBeDeleted.add(taxesTransaction)

...

Iterator<Item> iter = items.iterator();
while (iter.hasNext())
{
    Object o = iter.next().getSubject();
    if (toBeDeleted.contains(o))
        iter.remove();
}
```

(or, if you kept the item around, then ```items.remove(taxesTransactionItem)```)

buchen (changes requested):
Hi @Nirus2000,

what a change. I started digging through it - here are my comments on the postProcessing method.

The main point: we should not use attributes of the security. First, there can be one security for many transactions (which one wins?). Second, it pollutes the XML document.

Bildschirmfoto 2023-09-23 um 08 26 10

I have tried to make suggestions how to store the data differently.
The Item has a generic ```setData``` and ```getData``` which is currently not used at all.
If needed, we could extend it to (typed) map.

Finally, maybe it could enhance the code if the post processing of sale and dividend transactions are two different methods. The way I read the post processing, sale are dividend processing is completely separated - but the code is kind of mixed: prepare list S, prepare list D, process S, process D.

buchen:
> I think when creating the maps (concat() -> taxes treatment with sell or dividend) in postProcessing,
> it still needs to be identified somehow if the taxes treatment is a purchase/sale or a taxes treatment of a dividend.

I guess it depends on the "somehow".
If the document contains some kind of transaction identifier, then this could be stored in the Item and used to match.
But if the security + the date is the only identifier, then I do not see an way how we should know for which one the tax statement is.

Nirus2000:
Hallo @buchen

Thank you for the review.

I have implemented these points.
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334934809
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334934312
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334933651

If you have time, let's do a TeamViewer meeting on this so we could work out a solution together here.
- [ ] https://github.com/portfolio-performance/portfolio/pull/3555#pullrequestreview-1640794012
- [ ] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334933362

Regards
Alex

Nirus2000:
Hallo @buchen

Thank you for the review.

I have implemented these points.
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334934809
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334934312
- [x] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334933651

I need ATTRIBUTE_GROSS_TAXES_TREATMENT and ATTRIBUTE_EXCHANGE_RATE as attribute to read them from transaction A and process them in transaction B if necessary.
If you have time, let's do a TeamViewer meeting on this so we could work out a solution together here.
- [ ] https://github.com/portfolio-performance/portfolio/pull/3555#pullrequestreview-1640794012
- [ ] https://github.com/portfolio-performance/portfolio/pull/3555#discussion_r1334933362

Unfortunately, converting/rounding is now giving me a problem in `testDividend14WithTaxTreatmentOfDividende14()`
-   _Imported gross value (EUR 13.69) and calculated gross value (EUR 13.70) do not match_ 

and `ExtractorUtils.fixGrossValueA();` cannot be applied in postProcessing.

Regards
Alex

buchen:
> I need ATTRIBUTE_GROSS_TAXES_TREATMENT and ATTRIBUTE_EXCHANGE_RATE as attribute to read them from transaction A and process them in transaction B if necessary.

My understanding is you need both in the ```postProcessing``` method. In that method, you also  have the ```Item``` at hand and therefore also the ```Item#getData``` method (you'd need to create the maps wither with the item - or maybe create a record (item, transaction) - we can do  team viewer later)

buchen:
merged

PR #3556, opened 2023-09-14
[DRAFT] Parse document properties with document sections

buchen:
* Allow for option/mandatory properties
* Fail if the mandatory properties do not exist


buchen:
@Nirus2000  have a look at the deutsche bank importer - this is how it could look like

(the pr needs some more polishing, but I would like your feedback first)

PR #3564, opened 2023-09-19
nullcheck of taxonomy in EarningsByTaxonomyChartWidget

OnkelDok:
Maybe reason for https://forum.portfolio-performance.info/t/fehlermeldung-java-lang-illegal-argumentexeption-bereich-ungultig/25118
But so far I could not reproduce the error exactly like it was reported in the initial post. But when I tried to create an new file (no taxonomies) and create a ``EarningsByTaxonomyChartWidget`` then I got NPE and this is fixed by this PR.


buchen:
Thanks for the fix.

The message in the forum is difficult to debug because it is missing any relevant stack trace (I have no idea how to get rid of the "AccessibleObject" stack trace on Linux).

PR #3571, opened 2023-09-24
Show message when security with same symbol already exists (in SearchSecurityWizardPage)

OnkelDok:
To let the user know why the button is disabled.

grafik

See: https://forum.portfolio-performance.info/t/uebernehmen-ausgegraut-in-ergebnisliste-der-wertpapiersuche/9526/32


buchen:
Thanks for the contribution. I just took some time for me to test it locally. sorry.

PR #3577, opened 2023-10-02
Feature rework comdirect

buchen:


buchen:
Hi @Nirus,

please have a look at this commit: https://github.com/portfolio-performance/portfolio/pull/3577/commits/4c23cefdda561032be738637129aa599e3e33763

It is a commit on top of your pull request #3555.

It addresses the issue that the exchange rate is stored permanently in the security. For one, this is polluting the XML. Furthermore, the current algorithm cannot handle if one import contains multiple such documents (as they override the exchange rate).

This [change](https://github.com/portfolio-performance/portfolio/pull/3577/commits/4c23cefdda561032be738637129aa599e3e33763) does the following:
* it removes the handling of the exchange rate from parsing the transactions
* instead it uses the postProcessing method to retrieve the exchange rate from the sale and attach it to the taxes

The only change in the test is that "failed" transactions (the ones that will never be imported), do not have a grossValue of 0 anymore.

As you can see, I added a method #matchSaleAndTaxTransactions which creates a list of pairs of sale and tax transaction. The same list could later be used to merge the sale and tax where applicable. I changed it so that I have directly already typed objects.

Nirus2000:
Hallo @buchen 
Yesterday I had to sit down at [Thalia](https://www.thalia.de/) with a coffee... 
Well, the weather was nice and I drank a coffee to understand the "Warum & Wieso" and how does that work?. 
:exploding_head: :exploding_head: :exploding_head: 

I modified your source code a bit so that I (sorry) understood it better when reading and added it to my pull request.
I added you as a co-author :zany_face:
https://github.com/portfolio-performance/portfolio/pull/3555/commits/698d6f36d09b22c7631abfccf0a55a699a1b3a84

As a side note:
If I now, for example, For example, if I looked at the Targobank importer, then I would have assumed that I could also use this method to solve the problem with the pay_day - but that's not the case, is it?

Thank you
Alex

buchen:
> If I now, for example, For example, if I looked at the Targobank importer, then I would have assumed that I could also use this method to solve the problem with the pay_day - but that's not the case, is it?

Yeah. I justed looked at the Targobank importer. The post processing looks very similar

buchen:
Closing because it is now included in your pull request

PR #3578, opened 2023-10-04
Use collected sale/tax transactions to merge taxes into sale

buchen:
Issue: #3555

@Nirus2000 - this pr uses the list of paired tax and sale transactions also for the merging. I think it could simplify the code: first, the items are already cases. Second, the hash map is not needed to keep the deleted items - we can delete them right away.


Nirus2000 (approved):
See pull request #3584

buchen:
merging - as you cherry picked it into your branch, I believe you see this as okay :-)

PR #3584, opened 2023-10-07
Improve postProcessing in ComDirect PDF Importer after #3578

Nirus2000:
Add and merge https://github.com/portfolio-performance/portfolio/pull/3578

Refactors and optimizes the source

TransactionPair replaces SaleTaxPair for matched transaction pairs

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
@@ -10,17 +10,14 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.time.LocalDate;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import name.abuchen.portfolio.Messages;
 import name.abuchen.portfolio.datatransfer.ExtrExchangeRate;
@@ -50,13 +47,13 @@ import name.abuchen.portfolio.util.Pair;
  *
  *           Therefore, we use the documents based on their function and merge both documents, if possible, as one transaction.
  *           {@code
- *              matchSaleAndTaxTransactions(List<Item> saleTransactionList,List<Item> taxesTransactionList)
+ *              matchTransactionPair(List<Item> transactionList,List<Item> taxesTreatmentList)
  *           }
  *
- *           The separate taxes statement does only contain taxes in the account currency.
+ *           The separate taxes treatment does only contain taxes in the account currency.
  *           However, if the security currency differs, we need to provide the currency conversion.
  *           {@code
- *              fixMissingCurrencyConversionForTaxesTransactions(Collection<SaleTaxPair>)
+ *              fixMissingCurrencyConversionForSaleTaxesTransactions(Collection<TransactionTaxesPair>)
  *           }
  *
  *           Always import the securities transaction and the taxes treatment for a correct transaction.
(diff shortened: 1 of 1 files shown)


Nirus2000:
Hello @buchen 
I could imagine that we could move the matchTransactionPair() and concat() function to the ExtractorUtils().
The applications would be, Targobank, Degiro and ib-flex.
Of course, the importers would still need to be customized and refactored.

Regards
Alex

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -76,7 +73,7 @@
 @SuppressWarnings("nls")
 public class ComdirectPDFExtractor extends AbstractPDFExtractor
 {
-    private static record SaleTaxPair(Item sale, Item tax)
+    private static record TransactionPair(Item transaction, Item tax)
```
buchen, line 76: There is already a class named [TransactionPair](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/TransactionPair.java). Therefore I suggest to name this one differently to avoid confusion.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1809,249 +1822,191 @@ public List<Item> postProcessing(List<Item> items)
...
+        // Match transactions (sale + taxes treatment)
+        var saleTaxPairs = matchTransactionPair(saleTransactionList, taxesTreatmentList);
+
+        // Match transactions (dividend + taxes treatment)
+        var dividendTaxPairs = matchTransactionPair(dividendTransactionList, taxesTreatmentList);
+
+        // Fix missing currency conversion for sale tax transactions
+        fixMissingCurrencyConversionForSaleTaxesTransactions(saleTaxPairs);
+
+        // @formatter:off
```
buchen, line 1834: I understand that you need to switch off the formatter for text copied from the PDF documents - every space matters. I am not sure why it is needed for plain text comments - doesn't it just add two additional lines?

buchen:
> I could imagine that we could move the matchTransactionPair() and concat() function to the ExtractorUtils().

If needed, then it makes sense to have a separate class for this purpose - let's not bloat ExtractorUtils unnecessarily.

I think it is your call - I have no gut feeling how similar the processing is, if some banks work differently, etc.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
Nirus2000: I didn't even notice that anymore, but you're right. i reduced the comments.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
Nirus2000: done... renamed

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1809,249 +1822,177 @@ public List<Item> postProcessing(List<Item> items)
...
-                        }
+                saleTransaction.setNote(concat(saleTransaction.getNote(), taxesTransaction.getNote()));
 
-                        // Subtract the taxes from the taxes transaction from the total amount
-                        dividendTransaction.setMonetaryAmount(dividendTransaction.getMonetaryAmount()
-                                        .subtract(taxesTransaction.getMonetaryAmount()));
-
-                        // Add taxes as tax unit
-                        dividendTransaction.addUnit(new Unit(Unit.Type.TAX, taxesTransaction.getMonetaryAmount()));
+                items.removeIf(item -> item == pair.tax());
```
buchen, line 1856: Isn't that just removing one item from the list?

So just:
```
items.remove(pair.tax());
```

RemoveIf iterates over the full list to check the condition which will be true only once - for the tax item

buchen:
merged

PR #3592, opened 2023-10-11
Ignore inactive securities when searching for a matching identifier

rcjan:
This is a small step towards improving support of "duplicate" securities (e.g. same ISIN), or rather, making it work again.

Currently, importing PDFs is basically impossible if the security ISIN/WKN exists twice. This has been acknowledged as a shortcoming in these forum threads:

https://forum.portfolio-performance.info/t/csv-import-of-portfolio-transaction-does-not-work-if-isin-is-not-unique/7386
https://forum.portfolio-performance.info/t/fehler-isin-existiert-mehrfach-in-wertpapierliste-beim-pdf-import/11087/25

To quote @buchen:

> Anyway: PP has a very simple logic. When importing data from CSV, I tries to match the information to the existing securities configured in the file. If the identifier (ISIN, WKN, ticker) points to two securities, then PP does not know which one to use and fails. **That might not be the most intelligent strategy** but it is then implemented. PP could allow the user to select the security in the dialog after parsing the CSV file. I never came around to implementing it, so if you want to give it a go, then I can point you to the place in the code

This change addresses the problem by at least allowing inactive securities and one active security to exist without blocking the PDF import. Inactive securities are ignored when searching for a security with matching identifier(s). Thus, the PDF import will match the single active security and allow importing again. This should solve most real-world scenarios for using multiple identical securities (as the "old" securities are likely already inactive or at least easily could be deactivated). It would also allow a workaround for users even in scenarios with identical active securities if they simply temporarily deactivate one of them for the PDF import and then activate it again.

Further improvements to the described problem could be:

1) During PDF import for a specific securities account, in the case of duplicates choose the security that is already an asset in this account, if exactly one matching asset exists in the account.

2) In any case where, even with the better matching described in 1) and in this change, more than one security matches, show a dialog allowing the user to choose the right security. This should be a rare scenario though because it could only happen if you trade two "identical" securities in the same account but somehow logically separated.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/SecurityCache.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/SecurityCache.java
@@ -32,18 +32,20 @@ public class SecurityCache
     {
         this.client = client;
 
-        this.localMaps.add(client.getSecurities().stream().filter(s -> s.getIsin() != null && !s.getIsin().isEmpty())
+        this.localMaps.add(client.getSecurities().stream()
+                        .filter(s -> s.getIsin() != null && !s.getIsin().isEmpty() && !s.isRetired())
                         .collect(Collectors.toMap(Security::getIsin, s -> s, (l, r) -> DUPLICATE_SECURITY_MARKER)));
 
         this.localMaps.add(client.getSecurities().stream()
-                        .filter(s -> s.getTickerSymbol() != null && !s.getTickerSymbol().isEmpty())
+                        .filter(s -> s.getTickerSymbol() != null && !s.getTickerSymbol().isEmpty() && !s.isRetired())
                         .collect(Collectors.toMap(Security::getTickerSymbolWithoutStockMarket, s -> s,
                                         (l, r) -> DUPLICATE_SECURITY_MARKER)));
 
-        this.localMaps.add(client.getSecurities().stream().filter(s -> s.getWkn() != null && !s.getWkn().isEmpty())
+        this.localMaps.add(client.getSecurities().stream()
+                        .filter(s -> s.getWkn() != null && !s.getWkn().isEmpty() && !s.isRetired())
                         .collect(Collectors.toMap(Security::getWkn, s -> s, (l, r) -> DUPLICATE_SECURITY_MARKER)));
-
-        this.localMaps.add(client.getSecurities().stream().filter(s -> s.getName() != null && !s.getName().isEmpty())
+        this.localMaps.add(client.getSecurities().stream()
+                        .filter(s -> s.getName() != null && !s.getName().isEmpty() && !s.isRetired())
                         .collect(Collectors.toMap(Security::getName, s -> s, (l, r) -> DUPLICATE_SECURITY_MARKER)));
 
     }


buchen:
I like the idea. However, I propose *not* to filter all inactive securities. Instead, I think the mergeFunction could check if one of the securities is retired. If so, it takes the other one. If both are retired or both are active, the behaviour does not change.

The one edge case I see there: if there are three instruments with the same identifier and two of them are inactive. Then the merge function might mark this instrument as not importable.

What do you think?

rcjan:
Thanks for taking a look. I must admit I fail to see the advantage of your approach. In which scenarios is it better? As you mention, not ignoring all inactive securities can result in situations where it's still not possible to import/select the right security. In fact, I have such circumstances in my own portfolio. The "same" security exists three times, two of them being inactive.

buchen:
Today, transactions can be imported if the identifier (ISIN, WKN, ticker) exists only once. That applies regardless whether the security is active or not.

With your change, only transactions for **active** securities can be imported - even if a retired security has a unique identifier (ISIN, WKN, ticker). My argument is: why make this restriction of only importing active instruments. **Only** if there are multiple securities with the same identifier, we use the "retired" flag to decide which one to use.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/SecurityCache.java
```diff
@@ -32,18 +32,20 @@ public SecurityCache(Client client)
     {
         this.client = client;
 
-        this.localMaps.add(client.getSecurities().stream().filter(s -> s.getIsin() != null && !s.getIsin().isEmpty())
+        this.localMaps.add(client.getSecurities().stream()
+                        .filter(s -> s.getIsin() != null && !s.getIsin().isEmpty() && !s.isRetired())
                         .collect(Collectors.toMap(Security::getIsin, s -> s, (l, r) -> DUPLICATE_SECURITY_MARKER)));
```
buchen, line 37: Instead of the existing merge function ```(l, r) -> DUPLICATE_SECURITY_MARKER ```, we could check for the retired flag there.

```

(l, r) -> {
   if (l.isRetired() ^ r.isRetired()) {
     return !l.isRetired() ? l : r;
  }
  else return DUPLICATE_SECURITY_MARKER;
}

```

Maybe we would also have to check whether l or r is a DUPLICATE_SECURITY_MARKER. 

(I am not at my laptop so it is just jotted down to share my thinking)

rcjan:
> With your change, only transactions for **active** securities can be imported - even if a retired security has a unique identifier (ISIN, WKN, ticker).

You are right, I did not consider this. I can understand the reasoning, even though I don't agree with it. By definition, a security would not be inactive if you still need to import PDFs for it. Should it really occur because someone wants to amend/improve data for a deactivated security, they could still temporarily activate the security again. Personally, I think this situation is rarer than the one with more than one inactive "identical" security.

Obviously, unless you actually want to offer a choice dialog after all, only one of the scenarios can be solved fully.

I would argue in favor of my variant (ignoring all inactive) mainly because of this: With my approach, the edge case that is not solved can be "solved" with a workaround by the user (reactivating the inactive security). With yours, the edge case is not solvable for a user at all, except maybe by removing identifiers from existing securities temporarily (which is quite a hassle compared to just setting it inactive/active).

buchen:
Hey @rcjan, I do think we are very close in our understanding. What if want: If there are multiple instruments with the same identifier, but only one is them is active (= not retired), we want to allow the import to this active instrument.

I like the idea of using the retired flag. The difference is that I am proposing to check the retired flag **only if** there is a conflict (= multiple instruments with the same identifier) while this change proposes to ignore all inactive instruments. Anyway, next week I am back from vacation and will open my IDE to give it a go.

rcjan:
Ah, I see. Yes, a two-step process would work, with the default logic first, and if there are conflicts, try again but this time ignoring all inactive securities. Looking forward to that. I currently don't have a working project setup so cannot give it a try myself.

buchen:
Hi @rcjan,

I have added a new commit with the updated implementation as discussed. I have put both of us as authors as it was really (fruitful) joint discussion to come to a good solution. Thanks for your contribution.

Andreas.

rcjan:
Awesome, thank you. Looking forward to this change. Finally I can import my PDFs again without using a local build 😆

PR #3600, opened 2023-10-22
Enhance trade activity widget - allows to filter by specific transaction types

Pnda87:


buchen:
rebased and merged

I did a slight change by using a switch statement as I thought it would improve readability

PR #3603, opened 2023-10-23
Reset error message in SearchSecurityWizardPage

OnkelDok:
I've found another small thing in the ``SearchSecurityWizardPage``.
behaviour before the PR:

https://github.com/portfolio-performance/portfolio/assets/90478568/1fd273de-d134-4051-b15d-243fa8ffa9d1

with this PR:

https://github.com/portfolio-performance/portfolio/assets/90478568/25db3ea9-c077-4c3c-b242-ba378facf42d


buchen:
Thanks. Will merge.

As a side note: maybe it does not make sense to run a search request at all if the user entered no query. But then again I think that is not the normal scenario.

PR #3610, opened 2023-10-26
Adds the option for non cash-effective dividend payments

Pnda87:
This commit allows to mark dividends as non cash-effective. It can be used, e.g., for spin-offs or stock dividends, where the paid amount is directly reinvested in a stock and not paid in cash. The marked dividends can be excluded in the payment view to keep a clean view of the dividend development. All non cash-effective dividends can be filtered and listed in the transaction view.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1305,6 +1305,10 @@ public class Messages extends NLS
     public static String OptionDateIsInTheFuture;
     public static String OptionDateIsInThePast;
     public static String YearlyPerformanceHeatmapToolTip;
+    public static String LabelNonCashEffective;
+    public static String MenuNonCashEffective;
+    public static String MsgInfoNonCashEffective;
+    public static String TransactionFilterNonCashEffectiveDividend;
     static
     {
         // initialize resource bundle
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
@@ -28,6 +28,7 @@ import org.eclipse.jface.databinding.swt.typed.WidgetProperties;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.graphics.Image;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
@@ -46,12 +47,14 @@ import name.abuchen.portfolio.money.Values;
 import name.abuchen.portfolio.snapshot.ClientSnapshot;
 import name.abuchen.portfolio.snapshot.PortfolioSnapshot;
 import name.abuchen.portfolio.snapshot.SecurityPosition;
+import name.abuchen.portfolio.ui.Images;
 import name.abuchen.portfolio.ui.Messages;
 import name.abuchen.portfolio.ui.UIConstants;
 import name.abuchen.portfolio.ui.dialogs.transactions.AccountTransactionModel.Properties;
 import name.abuchen.portfolio.ui.util.FormDataFactory;
 import name.abuchen.portfolio.ui.util.LabelOnly;
 import name.abuchen.portfolio.ui.util.SWTHelper;
+import name.abuchen.portfolio.ui.util.swt.ControlDecoration;
 
 public class AccountTransactionDialog extends AbstractTransactionDialog // NOSONAR
 {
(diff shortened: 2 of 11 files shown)


buchen:
Hi @Pnda87,

thanks for the pull request. I fully understand your motivation. Personally, I think it would be valuable to be able to maintain a securities account **without** a cash account. That would mean having "non-cash effective" dividend, tax, and fee transactions. If we had those, then one could also build "composite transactions" for stock dividends, etc. And, as a benefit, we actually could map dividend transactions better to the portfolio (today multiple portfolios could have the same reference account).

So far my thinking was to allow a limited set of ```AccountTransaction```s also on the ```Portfolio``` and adapt the performance calculation accordingly. This includes the dialog to create and edit transactions, the ```ClientSnaphot``` and ```ClientIndex``` classes with the main calculations, the client filter, etc. As you can guess that is a very big change and the effort is the reason I never got to implementing it.

Your proposal is limited in scope - by introducing the "nonCashEffective" flag on the account transaction. I see two challenges:
* First, logically it is hard for the user to understand that he still needs a cash account even though the transaction is not cash effective. This muddles the UI.
* Second, I believe once we adopt the implementation to honor the new flag, the effort will not be significantly smaller than for the solution outlined above. In the current pull request, the balance does not reflect non cash effective dividends, the (filtered) performance is calculated as if it were cash effective, internal rate of return is not including it, etc.

For these reasons, I am at the moment not inclined to introduce a "non cash effective" flag on the transaction. However, I leave this PR open for the time being to continue our discussion and hear your arguments

AndreasHoffmann2:
May I ask if it might be a viable solution to add a new type of `PortfolioTransaction` ("SpinOff") for this instead of continuing to store spin-off as dividends?
The new type would behave very similar to "Delivery", since this is what it actually is: Some shares pop up in the portfolio without any cashflow occurring. Gui-fields would be the same as "Delivery (inbound)", but with the addition of:

- SourceSecurity
- SourceShares
- x shares for y shares

This way, the spin-off is stored as what it actually is: a spin-off.

For the statistics, I am not sure how to display this for the SourceSecurity.
image

The absolute performance would need to benefit from the spin-off of course. But in which column should the raw benefit of the spin-off be displayed? The best would be IMO to find an existing column where this could be added. If this does not exist, a new column would be required. And there would probably be some more placed, where this needs to be included.

I am be willing to invest the effort of further specifying and implementing this, but I would love to have some short feedback if this would have any chance of being merged.

Thank you!

Pnda87:
I like the proposal @AndreasHoffmann2 !
In terms of source security column: I sugget to show the perf. contribution in the dividend column, as a spin off is basically a different kind of distribution to shareholders, similar to dividends.
looking forward to get this implemented at some point in the future. I think PP really has a significant gap there atm. Whats your opinion on this @buchen ?

buchen:
First, I agree that PP needs more transaction types.

Let's step back for a moment. My understanding is that spin-off is one out of many *corporate actions*. (That does not mean we can implement everything right away, but to set the broader scene. We have to decide whether to create types for each of those corporate actions, or have one type with the related class holding the specific data).

```
Corporate Action
 ├─ Cash Distribution
 │   └─ Dividend
 └─ Securities Distribution
     ├─ Stock Dividend
     ├─ Spin-Off
     ├─ Bonus Issue
     └─ Rights Distribution
```

And then one more thought: in #5453, we discuss "instruments quoted percentage-based" which are bonds after all. The corporate actions there look different. Therefore, I think we should also think about a mechanism the separate them to type.

```
Fixed Income Corporate Action
 ├─ Income
 │   ├─ Coupon Payment
 │   ├─ PIK Interest
 │   └─ Defaulted Interest
 ├─ Redemption
 │   ├─ Maturity
 │   ├─ Partial Redemption
 │   ├─ Call
 │   └─ Put
 └─ Structural / Credit
     ├─ Conversion
     ├─ Exchange
     ├─ Restructuring
     └─ Default
```

And then one note about the implementation specifically for "Spin-Off". As this is a corporate action that touches two instruments, it could make sense to handle it similar to a transaction that touches two accounts (like purchase, sale, transfer). Similar I think a "rights Issue" can have a cash flow which then would include an account. For purchases, sales, transfers, I use an "Entry" implementation which "manages" **two** transactions.

Background: let's say one has a FIFO calculation that needs to quickly collect all transactions relevant for an instrument. Right now it looks for transaction->security. It would require many places to change if we had to change that to transaction->type(=corporate action)->subtype(=spin-off)->sourceSecurity as well.


@AndreasHoffmann2 asks:

> But in which column should the raw benefit of the spin-off be displayed?

Very concretly, I think we should not add it to "Sum of dividends" column.
Maybe we add one column "Total Distribution" which includes Cash and Securities Distribution.
And add one column "Securities Distribution" which holds the remaining corporate actions.
I do not think we it is needed to have columns for "Stock Dividend", "Spin-Off", etc.

> And there would probably be some more places, where this needs to be included.

Yes. We would need to change the performance calculation (ClientIndex), the filtering (ClientFilter) as well.

If you want to give it a go, as first step could be a draft pull request that outlines the extensions to the PortfolioTransaction for the new type.

buchen:
I created #5466 to continue the conversation.

I will close this pull request, because the given implementation (a flag whether something is cash effective) is not the way to pursue corporate actions.

PR #3611, opened 2023-10-27
Crosshair & measurement tool - round price to 2 digits

Pnda87:


Pnda87:
the apparent precision is useless for stocks, but I realized it makes sense to keep it for cryptos, which can be tracked too.

PR #3614, opened 2023-10-29
Improve some colors for dark theme

Nirus2000:
https://forum.portfolio-performance.info/t/dark-mode-fur-pp/3739/84

The screenshot does not show this quite as well, but it is significantly better than before. 
Left new <--> right old

(Click picture to expand)
light_dark_theme

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -347,17 +347,18 @@ public class SecuritiesChart
 
     private Color colorQuote;
 
-    private static final Color colorEventPurchase = Colors.getColor(26, 173, 33);
-    private static final Color colorEventSale = Colors.getColor(232, 51, 69);
-    private static final Color colorEventDividend = Colors.getColor(128, 0, 128);
+    private static final Color colorEventPurchase = Colors.getColor(26, 173, 33); // #1AAD21
+    private static final Color colorEventSale = Colors.getColor(255, 43, 48); // #FF2B30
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -347,17 +347,18 @@ public static ChartRange createFor(List<SecurityPrice> prices, ChartInterval cha
 
     private Color colorQuote;
 
-    private static final Color colorEventPurchase = Colors.getColor(26, 173, 33);
-    private static final Color colorEventSale = Colors.getColor(232, 51, 69);
-    private static final Color colorEventDividend = Colors.getColor(128, 0, 128);
+    private static final Color colorEventPurchase = Colors.getColor(26, 173, 33); // #1AAD21
+    private static final Color colorEventSale = Colors.getColor(255, 43, 48); // #FF2B30
+    private static final Color colorEventDividend = Colors.getColor(191, 62, 255); // #BF3EFF
```
OnkelDok, line 352: In my personal opinion the color is too bright for the dark background and therefore the numbers are harder to read. So slightly darker than your current proposal would be nice. Currently I have not time to test with some colors around.

Nirus2000: Adjusted

buchen:
Thanks @Nirus2000.

Looking at the screenshots, the colors seem to be an improvement. I will merge it.

BTW, if we want to have different colors for the light or dark theme, then it is possible to extend the CSS handler for the SecurityChart. At the moment, only the color for the historical price ("quote-color") is managed via CSS.

https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/plugin.xml#L119-L125

PR #3615, opened 2023-10-29
Reverse #2752 Logos are cut off on the left in OS_WIN

Nirus2000:
Reverse #2752
Logos are no longer cut off on the left under OS_WIN32

https://forum.portfolio-performance.info/t/dark-mode-fur-pp/3739/75

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
@@ -2,7 +2,6 @@ package name.abuchen.portfolio.model;
 
 import java.util.HashMap;
 
-import org.eclipse.core.runtime.Platform;
 import org.eclipse.swt.graphics.Image;
 
 import name.abuchen.portfolio.model.AttributeType.ImageConverter;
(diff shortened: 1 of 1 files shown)


Nirus2000:
Hello @buchen
The Logo...
Logo_Bayer_remove_background
grafik
What surprises me is that the transparent background is not accepted. Do you know why?

Alex

Morpheus1w3:
@Nirus2000 
Transparency and SWT with adjusting the scaling will be tricky as described at https://stackoverflow.com/questions/58171769/how-to-scale-a-transparent-org-eclipse-swt-graphics-image-loaded-from-a-png-ja

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
```diff
@@ -22,12 +22,37 @@ private ImageManager()
...
+     * @see <a href="https://github.com/portfolio-performance/portfolio/pull/2752">#2752</a>
+     *
+     * @param target The Attributable object to retrieve the image for.
+     * @param attr   The AttributeType specifying the image.
+     * @return The retrieved image or null if not found.
+     */
     public Image getImage(Attributable target, AttributeType attr)
     {
-        int xOffset = Platform.OS_WIN32.equals(Platform.getOS()) ? 1 : 0;
+        int xOffset = Platform.OS_WIN32.equals(Platform.getOS()) ? 0 : 0;
```
buchen, line 41: If the x offset is not needed anymore, then we can completely remove this statement as it is always zero.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
```diff
@@ -22,12 +22,37 @@ private ImageManager()
...
 
+    /**
+     * Retrieves an image associated with the given Attributable and AttributeType.
+     * If not found, a default image is returned.
+     *
+     * @implNote For some time, Windows cut off the logos on the left.
+     *           An offset was added with #2752 In the meantime the problem seems to be fixed,
+     *           so we set the offset to 0 again.
+     *
+     * @see <a href="https://github.com/portfolio-performance/portfolio/pull/2752">#2752</a>
```
buchen, line 33: Let's remove this comment. If one wants to have a look at the history, then one can use git. We should have comments only for "strange" workarounds which are in the code, not for the ones we do not need (anymore).


The commit message can include the explanation that it is not needed anymore.

buchen (changes requested):
Thanks for the change.

I have not tested on Windows lately. So it is good news that the x offset of 1 is not needed anymore.

buchen:
Then about the transparency. The SWT framework has its challenges...

Let's move the discussion about transparency to a separate issue: #3619


> I could not fix this error.
> The columns are wider than needed.

@Nirus2000  - can you provide some more details (maybe in a separate issue)?
My understanding is: the columns have whatever width the user has given them (by resizing it via double click). If one double-clicks on the column header (or on the line between two column headers), it should resize the column to the widest value.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
Nirus2000: Done

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ImageManager.java
Nirus2000: Done

buchen:
rebased and merged.

Issue #3616, opened 2023-10-29
Possible improvements to (transaction) CSV importer


pfalcon:
CSV transaction importer with all its column typing and mapping is really smart and neat feature. It still falls short in various real-world cases, this ticket tries to enumerate seemingly low-hanging changes which can improve it:

1. For the "(transaction) type", the mapping is "PortfolioPerformance internal operation" -> "value from CSV column". That's needlessly limiting, because it assumes there's a one-to-one mapping from PP's operation to what may appear in a random statement around the world. But maybe there's stuff like "Buy" vs "Buy to open" vs "Buy to close", or maybe there're different ways of depositing/transferring. So, the mapping should be the opposite, "value from CSV column" -> "PortfolioPerformance internal operation". And there should be support for many-to-one mapping. It may be a challenge (or chore) to support that in UI, but as long as that mapping can be edited in CSV importer JSON config which can be exported/imported, that's already a great improvement.
2. CSV forcibly ignores amount signs, by calling Math.abs(). This seems to be a generic problem with PP, see e.g. https://github.com/portfolio-performance/portfolio/issues/3450. For CSV importer, this enforces unrealistic model when each transaction would be explicitly classified as "buy" and "sell". Instead, a buy transaction would be one with the negative amount and sell - with positive (sign definitely depends on the "side" of the statement, the above is from the point of view of cash account). So, again, CSV should not just throw away sign, but use it instead.
3. There should be a way to specify a quote feed for newly imported securities, that's https://github.com/portfolio-performance/portfolio/issues/3450. Failing way to do that in UI (a chore to implement), a "golden standard" default feed should be set, which is Yahoo Finance.

pfalcon:
> 2. CSV forcibly ignores amount signs

Posted https://github.com/portfolio-performance/portfolio/pull/3617 for this.

flywire:
@Nirus2000 Was the outcome of supporting tax sign implemented or ignored?

aanno:
This would at least be possible, see my PR draft https://github.com/portfolio-performance/portfolio/pull/5445

PR #3617, opened 2023-10-29
CSVAccountTransactionExtractor: Do not blindly ignore amount sign

pfalcon:
CSV importer's existing mental model is that each transaction would be explicitly categorized (via a "type" column) as Buy or Sell. It then just ignores the sign of transaction amount (by calling Math.abs()).

This assumption is not realistic to cover all the cases. Another way is to have a generic "security transaction" type, and the sign of the amount signifies the actual operation, e.g. negative for Buy, positive for Sell.

This change allows to handle such CSV statements, by "inverting" the configured transaction type. For example, if given type column value is mapped to "BUY", but amount is negative, it is turned into "SELL".

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractor.java
@@ -236,6 +236,17 @@ import name.abuchen.portfolio.money.Money;
                 type = amount.isNegative() ? AccountTransaction.Type.REMOVAL : AccountTransaction.Type.DIVIDENDS;
             else
                 type = amount.isNegative() ? AccountTransaction.Type.REMOVAL : AccountTransaction.Type.DEPOSIT;
+        } else {
+            if (amount.isNegative()) {
+                if (type == AccountTransaction.Type.DEPOSIT)
+                    type = AccountTransaction.Type.REMOVAL;
+                else if (type == AccountTransaction.Type.REMOVAL)
+                    type = AccountTransaction.Type.DEPOSIT;
+                else if (type == AccountTransaction.Type.BUY)
+                    type = AccountTransaction.Type.SELL;
+                else if (type == AccountTransaction.Type.SELL)
+                    type = AccountTransaction.Type.BUY;
+            }
         }
         return type;
     }


pfalcon:
This addressed p.2 from https://github.com/portfolio-performance/portfolio/issues/3616

buchen (changes requested):
Hi @pfalcon,

thanks for the pull request. I like the change. It will give some more flexibility.

The build failure is due to code formatting of the new change (curly braces). Can you be so kind and format it with the default formatter in Eclipse (Ctrl-F)?

I think we should add a **test case** to [CSVAccountTransactionExtractorTest](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java) that tests the new behavior. Otherwise it is too easy to accidentally break this new feature with a future commit. Can you be so kind and add one?

Finally, I believe we should have the same logic when importing "only" portfolio transactions: [CSVPortfolioTransactionExtractor.java#L243-L249](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractor.java#L243-L249)

Thanks for taking the time for a contribution to Portfolio Performance.

Andreas.

Nirus2000:
I have my doubts here that this is a good idea....

Several problems will arise here.
1) If the sale value of the shares is less than the fees. (Sale with negative amount)
2.) Hereby cancellations are made possible of purchase and sale and not removed. The transactions are posted or duplicated at the wrong time. (Purchase or sale cancellation from day X).
3.) The same applies to reverse transactions of deliveries and removals.

buchen:
@Nirus2000 writes:
> Several problems will arise here.

Let's recap what this change enables.

Let's say we have a CSV of this format:
```csv
date;type;amount
2023-10-31;Ein-/Auszahlung;2000
2023-11-01;Ein-/Auszahlung;-1000
```

In this case, the type cannot be used to determine the transaction - the sign of the value is needed as well.
The sign of the amount can be used to make it a DEPOSIT or REMOVAL (= withdrawal).
In this case, one can import the CSV as a 2000 EUR deposit and 1000 EUR removal.
The imported transactions does not carry the sign anymore, of course, as all internal values must be positive in the model of Portfolio Performance.

If I understand your concerns, they are about negative values in the internal transactions model of Portfolio Performance. At the moment, I do not see this change allowing the negative values to leak in.

pfalcon:
```
date;type;amount
2023-10-31;Ein-/Auszahlung;2000
2023-11-01;Ein-/Auszahlung;-1000
```

Yes, this exemplifies a case we see here. I understand that it may change behavior for some users of CSV import, so it would depend on benefits it brings vs possible corner cases, and IMHO benefits outweigh. All in all, I posted it for feedback. I'll try to address review comments as time permits. (I otherwise have a whole bunch of changes in my fork, some adhoc, all without tests, but some certainly may be useful to other people, and to make PP a more generic tool (== cover more usecases), which should be a good aim for an open-source project.)

flywire:
> I otherwise have a whole bunch of changes in my fork

"_This PR is hosed by now and should be reworked_." It's unfortunate simple ideas are so hard to propose with Git.

aanno:
This would at least be possible, see my PR draft https://github.com/portfolio-performance/portfolio/pull/5445

Nirus2000:
The change itself was well-received — thanks again to @pfalcon for the contribution. However, the review requests from @buchen (code formatting fix to resolve the build failure, a test case in CSVAccountTransactionExtractorTest, and applying the same sign-inversion logic to CSVPortfolioTransactionExtractor) have not been addressed since the last activity in late 2023.
If anyone wants to pick this up, the approach is sound. A new PR with the three open points resolved should have a good chance of being merged. The relevant discussion and the linked issue #3616 provide all necessary context.

PR #3621, opened 2023-10-31
Improve the dark and light theme and move the values in the css handler

Nirus2000:
Improve the dark and light theme and move the values in the css handler


buchen:
merged. That will give more flexibility to change colors in the future.

If you look at the pull request: I have kept the original formatting of the plugin.xml so that it is clear what has been changed with this commit. I also removed some of the comments. If it is an implementation of the interface, I think it is okay to have the description in the interface. I also removed the attribute descriptions - if it is just straight forward attributes, no one is looking at the description anyway and it just uses up precious space. The other comments I kept.

PR #3628, opened 2023-11-03
fix wrong positioning of tooltip when PP is running on secondary screen, fixes #3622

### LINKED ISSUE #3622: No tooltips on exernal monitor
Most of the diagrams show a tooltip with detailed information if I click the left mouse button on any item. This works fine on my laptop screen but if I move the PP window to an external monitor these tooltips does not show up anymore. Moving the window back to my laptop screen and they work again.

 - OS: macOS (Ventura 13.5.1) 
 - PP Version 0.65.4

OnkelDok:
fixes #3622

Setting y to ``PADDING`` when y (``int y = pt.y - size.y - PADDING;``) was smaller results in tooltip beeing displayed on primary screen when PP was running on secondary screen (vertically arranged, primary bottom, secondary above). Because ``pt.y`` is negative when the mouse is on secondary screen, this condition always set y to ``PADDING`` and the tooltip is always on top of primary screen.


buchen:
merged

PR #3631, opened 2023-11-06
VIAC French import for PDF Extractor, basic tests

denispol:
Adding PDF Importer for WirBank documents (VIAC) in French (only English and German are supported in current release)

PR #3633, opened 2023-11-06
Fixes the errors from #3631

Nirus2000:
Fixes the errors from #3631


buchen:
See comment in #3631 - merged

PR #3635, opened 2023-11-07
Set default colors in SecuritiesChart

OnkelDok:
One first fix to prevent the nullreference exception reported here: https://forum.portfolio-performance.info/t/fehlermeldung-java-lang-illegalargumentexception-argument-cannot-be-null-im-bereich-buchungen/25515
I used the old colors before introdution of CSS: https://github.com/portfolio-performance/portfolio/commit/9f44ee6a75ca7624d75615877d5d520da8ff8e52

But the main cause is not resolved with this: Why the colors from CSS are not set.

And additionally (optional) we can implement some check/fallback when [passing the marker color to the chart](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java#L287): ``e.gc.setForeground(marker.color);``
We have following options when ``marker.color == null``:
* do not paint the line at all (cons: as long as nobody misses the line, the error remains hidden)
* use a fallback color (magenta, yellow, ...)
* no check - let the error occur: disadvantage is that the tool is not usable until the error is fixed

@buchen, what is your opinion?

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -345,18 +345,18 @@ public class SecuritiesChart
         }
     }
 
-    private Color colorQuote;
-    private Color colorQuoteAreaPositive;
-    private Color colorQuoteAreaNegative;
+    private Color colorQuote = Colors.getColor(77, 52, 235); // #4D34EB
+    private Color colorQuoteAreaPositive = Colors.getColor(90, 114, 226); // #5A72E2
+    private Color colorQuoteAreaNegative = Colors.getColor(226, 91, 90); // #E25B5A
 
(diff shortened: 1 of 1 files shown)


buchen:
Thanks.

I think that should be enough in terms of "safeguarding".

I realized that I do something similar already with the other themed colors, for example here:
https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/Colors.java#L19-L29

> But the main cause is not resolved with this: Why the colors from CSS are not set.

I agree. I have no good idea why this is supposed to happen.
I understand it can happen on all platforms (there was one report for Linux, one Windows).
Maybe it is because after the update (sometimes?) the old bundle still stays around and maybe old CSS are read?
I find it hard to troubleshoot without reproducing it - if you, or anybody, has a good idea or approach, I am happy to investigate

PR #3637, opened 2023-11-08
Sanitize header text in heat maps

RalfSchmelter:
The header text for the heap maps is not sanitized to make sure ampersands are not interpreted as mnemonics. This PR fixes this.


buchen:
Vielen Dank Ralf für die Contribution - every detail matters

PR #3644, opened 2023-11-12
TradesTableViewer: Actually show cash account in trade breakdown tooltip

pfalcon:
A tooltip with per-transaction breakdown of a trade, shown when hovering over a trade entry, has a column named "Cash Account". However, what's actually shown there is *portfolio* account. So, either column title is wrong, or value shown is wrong. Showing portfolio account wouldn't make much sense, as a trade always belongs to a single portfolio account, so value shown would be invariantly the same in each row. And user can see portfolio account in trade entry (if corresponding column is enabled), so duplicating it in the tooltip doesn't make much sense. Thus, it seems that the corresponding cash account was indeed intended to be shown, so let's make it so.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/TradesTableViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/TradesTableViewer.java
@@ -174,7 +174,7 @@ public class TradesTableViewer
                     row[5] = pair.getTransaction().getUnitSum(Unit.Type.FEE);
                     row[6] = pair.getTransaction().getUnitSum(Unit.Type.TAX);
                     row[7] = pair.getTransaction().getMonetaryAmount();
-                    row[8] = pair.getOwner();
+                    row[8] = pair.getTransaction().getCrossEntry().getCrossOwner(pair.getTransaction());
 
                     builder.addRow(row);
                 });


pfalcon:
@buchen: Dunno if additional typechecking is required here, up to you. I for one consider things like this concerning: https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractor.java#L140 , disabled it in my fork (patching CSV imported transactions behind user's back from Sell to Outbound Delivery).

chirlu:
> Showing portfolio account wouldn't make much sense, as a trade always belongs to a single portfolio account, so value shown would be invariantly the same in each row.

This is not true. Because of transfers, it is possible that the first buy happened in portfolio A, the second buy in portfolio B, and the sell in portfolio C.

buchen:
I tend to agree that the column is confusing.

First, the label is wrong. It should be the "securities account" (Depot).
Second, because in almost all cases a trade is done in the same account, it will almost always show the same information. @chirlu pointed out the case when it is not different.

I am not so sure if displaying the cash account is more helpful:
* For most users, it will be the same cash account - again displaying identical data
* For inbound/outbound deliveries, there is no cash account - and then nothing is displayed.
* The current information is helpful to understand how trades are constructed if there are transfers between securities account - maybe the cash account conveys the same information, but maybe not (if one  cash account is used for multiple securities accounts)

I am open for counterarguments 😄

PR #3647, opened 2023-11-15
SecurityListView: Use Pattern.DOTALL to properly search in multiline …

pfalcon:
…notes

Security's notes field may easily contain newlines. Newlines are by default not matched by the regex "." operator, causing any notes containing newlines to not be matched by search filter.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityListView.java
@@ -431,7 +431,7 @@ public class SecurityListView extends AbstractFinanceView
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


pfalcon:
It was pretty scary to be hit by this issue. I'd recommend to review all other uses of regexps to ensure DOTALL is used.

Also, seeing that regexps are used for searching, why not add an option for advanced user to search using regexps natively (i.e. skip quoting step)? @OnkelDok, random ping if you might be interested to implement this. In my fork I'll probably just switch to regexps altogether.

Ah, another thing I'll be looking to implement now - do not clear the search box when switching between watchlists. Usecase: see in what watchlists you have a particular ticker. Again probably would warrant a config option, but I'll be just switching to this non-clear behavior unconditionally.

buchen:
> Also, seeing that regexps are used for searching, why not add an option for advanced user to search using regexps natively (i.e. skip quoting step)?

How would we enable this from a UX perspective? I would want to avoid cluttering the UI with functionality that less than maybe 1% will be using.

> Ah, another thing I'll be looking to implement now - do not clear the search box when switching between watchlists. 

Nice idea - I'd love a pull request

PR #3650, opened 2023-11-18
Show closing transactions for trades in the Payments view

chirlu:
In the Payments view, there is a tab showing the individual transactions that are aggregated in the tables and graphs of the other tabs. However, in the mode that considers closed trades, this tab has been intentionally empty.

There are reasons for and against this behaviour: On the one hand, a single trade comprises multiple transactions (one or more opening and one closing), some of which may be outside the time period being considered, and the transactions (unlike for the other view such as fees, taxes) don’t immediately show the profit of the trade. On the other hand, users have been repeatedly confused ([A](https://forum.portfolio-performance.info/t/buchungen-in-trades-und-zahlungen-nicht-enthalten/20370), [B](https://forum.portfolio-performance.info/t/berichte-zahlungen-unter-geschlossene-trades-und-summe-fehlerhaft/25561/3), [C](https://forum.portfolio-performance.info/t/kaufe-verkaufe-unter-einnahmen-ausgaben-als-einlieferung-auslieferung-dargestellt/17252/4)) that there were no transactions listed for trades, and it makes verifying the aggregate results for trades hard.

Change this behaviour and from now on include (only) the closing transaction of each trade. The closing transaction always is within the observed time period; it gives the user a general idea of what has been included, and the date allows looking up the trade in the separate Trades view easily, for full details.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
@@ -318,6 +318,8 @@ public class PaymentsViewModel
 
                     sum.values[index] += value;
                     sum.sum += value;
+
+                    transactions.add(trade.getLastTransaction());
                 }
             }
         }
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
@@ -6,6 +6,7 @@ import java.time.Month;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -280,11 +281,11 @@ public class PaymentsViewModel
         Interval interval = Interval.of(LocalDate.of(startYear - 1, Month.DECEMBER, 31), now);
         Predicate<Transaction> checkIsInInterval = t -> interval.contains(t.getDateTime());
 
+        Set<TransactionPair<?>> transactions = new HashSet<>();
         Map<InvestmentVehicle, Line> vehicle2line = new HashMap<>();
 
         this.sum = new Line(null, false, this.noOfmonths);
         this.sumRetired = new Line(null, useConsolidateRetired, this.noOfmonths);
-        this.transactions = new ArrayList<>();
 
         EnumSet<Mode> processGainTx = EnumSet.of(Mode.TRADES, Mode.ALL);
         if (processGainTx.contains(mode))
@@ -469,6 +470,8 @@ public class PaymentsViewModel
                 }
             }
         }
+        this.transactions = transactions.stream()
+                            .sorted(TransactionPair.BY_DATE).toList();
         this.lines = new ArrayList<>(vehicle2line.values());
     }
 


chirlu:
As an aside: The issue with duplicate transactions is a reminder that the “sum” mode (which sums dividends, interest, taxes, fees, and the profit from closed trades) is somewhat inconsistent. Fees and taxes are included twice – both directly, as fees and taxes, and again by lowering the profit of their trade. But I don’t see any way to fix this, except by excluding trades from the “sum” mode; and “sum” mode is only there to give you a general idea anyway, I suppose.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
```diff
@@ -280,11 +281,11 @@ private void calculate()
         Interval interval = Interval.of(LocalDate.of(startYear - 1, Month.DECEMBER, 31), now);
         Predicate<Transaction> checkIsInInterval = t -> interval.contains(t.getDateTime());
 
+        Set<TransactionPair<?>> transactions = new HashSet<>();
         Map<InvestmentVehicle, Line> vehicle2line = new HashMap<>();
 
         this.sum = new Line(null, false, this.noOfmonths);
         this.sumRetired = new Line(null, useConsolidateRetired, this.noOfmonths);
-        this.transactions = new ArrayList<>();
```
chirlu, line 287: Note, the remainder of this method refers to just `transactions`, which got mapped to the member `this.transactions`; but now the same code refers to the new local variable instead.

Morpheus1w3:
@chirlu 
> The issue with duplicate transactions

I guess something like the below code snippet is needed to get rid of this behaviour:

```
<TransactionPair<?>> transaction = new TransactionPair<>(account, transaction);

transactions.computeIfAbsent(transaction,  s -> transaction);
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewModel.java
buchen: My IDE also hints at this with "Local variables should not shadow class fields" 😉

buchen:
@Morpheus1w3 writes:

> I guess something like the below code snippet is needed to get rid of this behaviour

I believe @chirlu solved it technically with a hash set - which is also fine with me. No transaction is displayed twice.

I understand his point as: by including the category "trade" into the payment view, it is mixing something conceptually different into the calculation. A tax, fee, etc. transaction clearly belongs to one month/quarter/year. But a trade necessarily has multiple dates (at least two: opening and closing trade). And it involves taxes and fees from multiple periods that are/should now be compared to taxes/fees of one period? Anyway... the feature is as it is.

buchen:
@chirlu writes:

> Change this behaviour and from now on include (only) the closing transaction of each trade.

I think that makes sense. I will merge this pull request. But let's use it first to discuss further.

I thinking what makes sense for the **other** transactions of the trade:
* They might fall into the reporting period 
* They might include fractional values (because a transaction is included only partially in the trade)

One way to include them could be
* Create a "FractionalPortfolioTransaction" that holds the original transactions, but has only the fractional values
* Enhance the transactions table to visually differentiate fractional transactions (italics?) and a tooltip with the original one
* Enhance context menu to not edit the fractional transaction but unwrap the original instead
* If multiple fractional transactions of the same transaction add up to 100%, then only include the original trancation

What do you think?
To be honest, I am unsure if it will provide the value to the user (while I worry about code maintenance - it is complex)

Morpheus1w3:
Regarding determine the closing sell transactions, isn't it suitable to add the sourcing transaction during calculations somehow at https://github.com/portfolio-performance/portfolio/blob/a77ff90451fc9079a6a331e5efb1f734c2de626f/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/TradeCollector.java#L196 ?

buchen:
@Morpheus1w3 writes:

> Regarding determine the closing sell transactions, isn't it suitable to add the sourcing transaction during calculations somehow at

I understand your point. The way @chirlu does it, works because the pull request is using Trade#getLastTransaction on a closed trade. In that case, the last transaction is always the closing transaction. But semantically it might make sense to have a ```Optional<Transaction> getClosingTransaction()``` method which encapsulates this logic.

chirlu:
Regarding the other transactions (the opening ones), I don’t think it is worth the trouble to include them. As you say, it would get rather complicated; and in the chronological transactions tab, separate from the other transactions belonging to the trade, I think they might be confusing rather than helpful.

One thing I could imagine might be to add a link (to trade-closing transactions) that leads to the Trades report, scrolling to the relevant trade there; or a tooltip giving information about the trade, such as all transactions. But either way would require quite a bit of effort, and I’m not sure it’s worth it.

> My IDE also hints at this with "Local variables should not shadow class fields" 😉

I don’t think I agree with the IDE (as long as the local variable isn’t unrelated to the field, in which case it probably would make sense). Do you? :)

> semantically it might make sense to have a `Optional<Transaction> getClosingTransaction()` method which encapsulates this logic.

Yes, I suppose I could do that.

PR #3651, opened 2023-11-18
Modify ComDirect PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/319
Improve concat()

---

Hello  @buchen 
I have modified the concat() a bit because I want to remove this function from the PDF importer and move it to a more suitable location. The only question is where. [ExtractorUtils.java](https://github.com/portfolio-performance/portfolio/blob/a77ff90451fc9079a6a331e5efb1f734c2de626f/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java) or [TextUtil.java](https://github.com/portfolio-performance/portfolio/blob/a77ff90451fc9079a6a331e5efb1f734c2de626f/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java)

We could then use this function here
https://github.com/portfolio-performance/portfolio/blob/97e7c08635dcf8364c0c9f86a56defe14464a1dc/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AbstractPDFExtractor.java#L130

and here we could remove the "concat" and make the adjustments in postprocessing analogous to the Comdirect PDF importer.

https://github.com/portfolio-performance/portfolio/blob/97e7c08635dcf8364c0c9f86a56defe14464a1dc/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TargobankPDFExtractor.java

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
@@ -1139,6 +1139,9 @@ public class ComdirectPDFExtractor extends AbstractPDFExtractor
                                                             }
                                                             else
                                                             {
+                                                                // Store in transaction context
+                                                                v.getTransactionContext().put(ATTRIBUTE_GROSS_TAXES_TREATMENT, grossBeforeTaxes);
+
                                                                 t.setMonetaryAmount(deductedTaxes);
                                                             }
                                                         }),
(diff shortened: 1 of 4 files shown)


buchen:
> I have modified the concat() a bit because I want to remove this function from the PDF importer and move it to a more suitable location. The only question is where. [ExtractorUtils.java](https://github.com/portfolio-performance/portfolio/blob/a77ff90451fc9079a6a331e5efb1f734c2de626f/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java) or [TextUtil.java](https://github.com/portfolio-performance/portfolio/blob/a77ff90451fc9079a6a331e5efb1f734c2de626f/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java)

I believe both locations are good. I tend to suggest TextUtil because it is only "plain old string manipulation" and not specific to extractors.

But let's think about the name of the method for a moment. It behaves differently than the usual String#concat or String#join methods:
* Maybe ```concatenate``` (the full verb shows it is something different)?
* Or ```combine``` (hinting that identical values are not simply concatenated)?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1988,14 +2006,29 @@ private void fixMissingCurrencyConversionForSaleTaxesTransactions(Collection<Tra
...
+     * Concatenates two strings with a specified separator.
+     * If both strings are null, the result is null.
+     * If the first string is not null and the second is null, the first string is returned.
+     * If both strings are non-null and equal, only the first string is returned.
+     * Otherwise, the two strings are concatenated with the specified separator.
+     *
+     * @param first     The first string to concatenate.
+     * @param second    The second string to concatenate.
+     * @param separator The separator to use when concatenating non-null strings.
+     * @return The concatenated string or null if both input strings are null.
```
buchen, line 2019: I appreciate the effort to add more documentation.
But I am very, very skeptical of such parameter documentation.
Do I really need the written information that ```String first``` is the "first string to concatenate"?!?
I am expecting the developer to understand and read the method signature.

In fact, personally I first read the method signature. And mostly only read the method signature.
Documentation is not compiled nor tested and can get quickly outdated. And it's not that we are writing libraries here.

PR #3652, opened 2023-11-18
Modify Baader Bank PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-baader-bank-scalable-capital/2057/276 
Simplification of some regular expressions
Return the source code to the standard Eclipse formatting.


buchen:
> Return the source code to the standard Eclipse formatting.

I hope you are not using standard Eclipse formatting but the "Portfolio Performance code formatter" 😄 
In general, it might make it easier to separate the reformatting into a separate commit - otherwise it is very hard to understand what was the content change.

PR #3656, opened 2023-11-19
Improve legend consistency in payment charts builder

Nirus2000:
#### Description:
Improvement of the three chart builder classes PaymentsPerMonthChartBuilder, PaymentsPerQuarterChartBuilder, and PaymentsPerYearChartBuilder.
The problem is that the legend for the diagram is not always displayed correctly. Sometimes ascending and sometimes descending.
https://forum.portfolio-performance.info/t/reihenfolge-der-legende-in-ertragsansicht/25705

#### PaymentsPerMonthChartBuilder and PaymentsPerQuarterChartBuilder:
- Introduced TreeMap<Integer, IBarSeries> (yearToSeries) for automatic sorting of yearly data. 
- Replaced the loop with a more expressive computeIfAbsent method. 
- Adjusted the iteration over the months to consider a variable (monthsToPopulate) for a clearer understanding of the loop's purpose.

#### PaymentsPerYearChartBuilder:
- Implemented TreeMap (yearToSeries) for automatic sorting of yearly data. 
- Replaced the loop with a more expressive computeIfAbsent method. 
- Adjusted variable names.
- Removed unnecessary logic for creating the bar series. 
- Removed the hasNegativeNumber logic as it is no longer needed.

---

Hello @buchen 
I'm not 100% sure, but why was the special function with negative values introduced in the annual bar series?
I have tried this on my system and so far everything is displayed correctly.
I have therefore removed it, otherwise we will have to add it again. But I don't really see the point.
Does anyone else still have the reason in mind? @OnkelDok @chirlu 

Regards
Alex

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsPerMonthChartBuilder.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsPerMonthChartBuilder.java
@@ -3,6 +3,8 @@ package name.abuchen.portfolio.ui.views.payments;
 import java.text.DateFormatSymbols;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
 import java.util.function.Consumer;
 
 import org.eclipse.swt.SWT;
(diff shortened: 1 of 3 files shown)


buchen:
> I'm not 100% sure, but why was the special function with negative values introduced in the annual bar series?
I have tried this on my system and so far everything is displayed correctly.
I have therefore removed it, otherwise we will have to add it again. But I don't really see the point.

The reason is that stacked series must not have negative values.
But we use stacked series to have differently colored bars in the yearly chart.
However, depending on the selection of the user, we can have negative values.
Then we fall back to just one data series (not colored, but at least it is displayed correctly)

buchen:
I believe the cause of the issue is this:
* If stacked series are used (and that is only the case for the year), then the legend is reversed
* Somehow, in flipping back between category and non-category series, it sometimes is correct

See 

https://github.com/eclipse/swtchart/blob/develop/org.eclipse.swtchart/src/org/eclipse/swtchart/internal/Legend.java#L226

SWTChart says:
```
// gather the stacked series reversing the order of stack series
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsPerMonthChartBuilder.java
```diff
@@ -144,25 +146,45 @@ public void configure(Chart chart, Consumer<TabularDataSource> selectionListener
...
+     * representing monthly payment data. Uses a TreeMap to automatically sort
+     * the years in ascending order.
+     */
     @Override
     public void createSeries(Chart chart, PaymentsViewModel model)
     {
         chart.setData(PaymentsViewModel.class.getSimpleName(), model);
 
+        // Use a TreeMap to automatically sort the years
+        Map<Integer, IBarSeries> yearToSeries = new TreeMap<>();
```
buchen, line 160: I do not think the month is ever in the wrong order, because it is never using a stacked series.
Plus: the order is determined by the #createSeries call. The TreeMap does not make a difference here.

buchen:
I think I now understand the behavior.

The reversing of the legend is only done, if it is a stacked series.
If we update the series from a non-stacked series (say there are negative values) to a stacked series (because there are none and we want it colored), then SWTChart does not reverse the order because due to the suspension of the update, the internal data model is not fully setup (in particular series#stackedSeries is null).

I think it could be fixed by not suspending updates, but that makes it slow.
It can be fixed by unsuspending updates twice.

Basically, it is because the updateLayout first sorts, but then updates the stack data.
```java
    public void suspendUpdate(boolean suspend) {
        if (updateSuspended == suspend) {
            return;
        }
        updateSuspended = suspend;

        // make sure that chart is updated
        if (!suspend) {
            updateLayout();                                                                          <--- sorts the data series
            ((SeriesSet) getSeriesSet()).updateStackAndRiserData();   <--- updates the stackSeries
        }
    }
```

buchen:
I have now a) reversed the creation of the data series and b) unsuspended the chart to force an update of the internal data:

9f927c885f19ac34d52c38fe3ac6d0f58a647786

PR #3678, opened 2023-12-18
added Fidelity PDF import of trade confirmations

ggrossbe:
added Fidelity PDF import of trade confirmations (not monthly/quarterly statements)

PR #3684, opened 2023-12-22
added Fidelity PDF import of trade confirmationsFidelity import

ggrossbe:
see #3678 
resynced my forked branch and initial PR got closed

added last commit:

- use locale en_US directly (without overriding as... methods) and currency from text where possible
- in addSummaryStatementBuySellTransaction() there is no dollar sign, so have to hardcode currency to USD
- in addBuySellTransaction() date on brokerage statement is "DEC" but DateFormatter only converts if "Dec": Not a valid date: 'APR/09/2023'


buchen:
Thanks for the contribution.
I will be published with the next version - I plan for one more update this year

PR #3685, opened 2023-12-22
Modify DKB PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-dkb/4449/102

Remove BOM in test files Dividende01.txt, Dividende02.txt and DIvidende11.txt

---
The character code 65279 represents the Unicode code for the Byte Order Mark, which often appears as an invisible character at the beginning of UTF-8 encoded files.

https://github.com/portfolio-performance/portfolio/blob/23a649217b18dc614f07e34d4dc416c49ebd1083/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java#L187

```
    String currentLine = lines[ii];
    int firstCharCode = (int) currentLine.charAt(0); // Charakter-Code des ersten Zeichens

    System.out.println("Character Code of first character in line [" + ii + "]: " + firstCharCode);

    Matcher matcher = startsWith.matcher(currentLine);
    System.out.println("Match gefunden => " + matcher.matches() + " [" + ii + "] -> " + currentLine);

    if (matcher.matches())
        blocks.add(ii);
```

Result:
```
Match gefunden => java.util.regex.Matcher[pattern=^10919 Berlin( Seite 1)?$ region=0,21 lastmatch=] -> ﻿10919 Berlin Seite 1
---------------
Character Code of first character in line [0]: 65279
```

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dkb/Dividende01.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dkb/Dividende01.txt
@@ -1,4 +1,7 @@
-﻿10919 Berlin Seite 1
+﻿PDFBox Version: 1.8.17
+Portfolio Performance Version: 0.66.2
+-----------------------------------------
+10919 Berlin Seite 1
 Depotnummer 500123451
 Kundennummer 51234561
 Max Mustermann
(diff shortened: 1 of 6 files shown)


buchen:
I imagine this character was added due to copying the UTF-8 output. I do not think it is in the string input that we generate out of the PDF.

PR #3694, opened 2023-12-25
Rework Fidelity International PDF-Importer

Nirus2000:
Rework #3684
Improve source code
Modification of ExtractorUtils.java DateTimeFormatter to use the helper method to create case-insensitive DateTimeFormatter instances 
Rename Fidelity to Fidelity International 
PDF importer Improve and rename test files
Improve and rename test cases

Hello @ggrossbe 
I have revised the importer. This will make it easier to customize.
I also fixed the date problem and other small things.
The source code is now structured and the test cases are up to date.
Have a look at this. The results are the same as you already had.
I have only added the CUSIP and, if available, the ISIN.

@buchen 
Please check the helper method in ExtractorUtils.java --> DateTimeFormatter to see if this is OK.
Please only add me as a co-author. The author would then be @ggrossbe .
👍🏻


buchen:
Merged.

I plan to release one more version this year.


> Please check the helper method in ExtractorUtils.java

I took the liberty to shorten the method name - it is not an insensitive method but a case insensitive method. Anyway, I think simply "createFormatter" works as well.

> Please only add me as a co-author. The author would then be @ggrossbe .

Next time let's work on this together in a pull request.

PR #3695, opened 2023-12-26
Add new AJ Bell Securities Limited PDF-Importer

Nirus2000:
Add new AJ Bell Securities Limited PDF-Importer

@buchen 
My tests are ok


buchen:
> My tests are ok

yes. I was too sloppy :-(

PR #3697, opened 2023-12-27
MyDividends24 Export Rework

mfrkn702001:
Changed the UI in PortfolioPerformance at: Online > Upload at myDividends24 to be able to select a specific Portfolio of PortfolioPerformance.


mfrkn702001:
@buchen can you please Respond if this will be uploaded into the Release Version or if I have to adjust anything and maybe when it will be published. I need it for my Job.

Greetings,
Furkan 
in name of MyDividends24

buchen:
Hi @mfrkn702001,

I was taking a couple days off over new year. As this is an open source project run on spare time and not a business, there is no guaranteed response time. I'll try to have a look at.

Andreas.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UploadToMyDividends24Handler.java
```diff
@@ -14,26 +15,46 @@
...
-import name.abuchen.portfolio.ui.dialogs.ListSelectionDialog;
 import name.abuchen.portfolio.ui.editor.ClientInput;
 import name.abuchen.portfolio.ui.jobs.AbstractClientJob;
 import name.abuchen.portfolio.ui.util.swt.ActiveShell;
+import name.abuchen.portfolio.ui.util.viewers.CopyPasteSupport;
 import name.abuchen.portfolio.util.Pair;
 
 public class UploadToMyDividends24Handler
 {
+    record Selection(String stringValue, Portfolio portfolio)
```
buchen, line 54: Naming of "stringValue".

To be honest, it is super confusing to understand what the right identifier of a MY24 depot is.
The MyDividends24Uploader#getPortfolios method retrieves a list of pairs with the index and a name.
Reviewing the code, I think the index is never used later on.
Instead the name is eventually put into this stringValue property and used in the API.

You can see this also from the method signatures: #getPortoflios retrieves a list of (int, String), but #upload only takes a String.

My proposal therefore is:
* Refactor to the getPortfolios method to just return the list of names, i.e., drop the unnecessary index
* Rename this variable to something meaningful. I could imagine MY24Portfolio[Id] because it is about a "remote/external" portfolio. Below you use MyDividendsPortfolio which is also okay.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UploadToMyDividends24Handler.java
```diff
@@ -147,3 +172,147 @@ protected IStatus run(IProgressMonitor monitor)
...
+    {
+        this.myDividendsPortfolios = myDividendsPortfolios;
+    }
+
+    public void setPortfolio(List<Object> portfolios)
+    {
+        this.portfolios = portfolios;
+    }
+
+    public Object getSelectedPPPortfolio()
```
buchen, line 205: Naming: getSelectedPortfolio.
If it is about PP (and this is the source code of PP), then we do not prefix this with PP.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UploadToMyDividends24Handler.java
```diff
@@ -147,3 +172,147 @@ protected IStatus run(IProgressMonitor monitor)
...
+    {
+        return selectedPortfolio;
+    }
+
+    public Pair<Integer, String> getSelectedMyDividendsPortfolio()
```
buchen, line 210: Naming: make it consistent to whatever you choose above (see the first comment about MY24Portfolio).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UploadToMyDividends24Handler.java
```diff
@@ -147,3 +172,147 @@ protected IStatus run(IProgressMonitor monitor)
...
+    private List<Object> portfolios;
+    private List<Pair<Integer, String>> myDividendsPortfolios;
+
+    private Object selectedPortfolio;
+    private Pair<Integer, String> selectedMyDividendsPortfolio;
+
+    private TableViewer portfoliosTableViewer;
+    private TableViewer myDividendsTableViewer;
+
+    private boolean includeTransactions = false;
```
buchen, line 188: What is the use of this boolean?

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MyDividends24Uploader.java
```diff
@@ -50,21 +53,37 @@ public List<Pair<Integer, String>> getPortfolios() throws IOException
...
+        {
+            // Case: one portfolio of PP is selected
+            stream = ppPortfolio.getTransactions().stream();
+        }
+        else
+        {
+            // Case: all portfolios of PP is selected
+            ClientSnapshot snapshot = ClientSnapshot.create(client, converter, LocalDate.now());
+            PortfolioSnapshot portfolio = snapshot.getJointPortfolio();
+            stream = portfolio.getPortfolio().getTransactions().stream();
```
buchen, line 72: I believe this can be simplified to

```java
stream = client.getPortfolios().stream().flatMap(p -> p.getTransactions().stream());
```

Because you are interested only in the transactions (see the code further below).
There is no need to calculate a snapshot then (which also means you do not need the CurrencyConverter here).

And: wording of the comment: "case: all portfolios **are** selected"

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MyDividends24Uploader.java
```diff
@@ -50,21 +53,37 @@ public List<Pair<Integer, String>> getPortfolios() throws IOException
...
                             double buyingprice = item.getGrossValueAmount() / Values.Amount.divider() / quantity;
                             String purchasedate = item.getDateTime().toString();
                             String isin = item.getSecurity().getIsin();
                             String type = item.getType().isPurchase() ? "buy" : "sell"; //$NON-NLS-1$ //$NON-NLS-2$
 
-                            // Creates a JSONObject for each transaction with
+                            // Creates a JSONObject for each transaction
+                            // with
                             // the calculated data.
                             JSONObject jsonObject = new JSONObject();
```
buchen, line 88: I am not sure if you are aware, but

* because you are reading the raw transactions, the transactions can be in any currency. A client can contain multiple accounts in different currencies, and instruments can be in different currencies. Don't you need a currency symbol in the json object then?
* you are also including the transfer transactions (transfer of instruments between accounts). You would loose this information because it is converted to "buy" and "sell" (which then has impact on FIFO calculations). I would think you would want to skip the transfers (if they are between portfolios that are included in the upload).

buchen (changes requested):
Hey, thanks for the contributions. It generally looks good to me. Have a look at the review comments I added.

buchen:
@mfrkn702001 still interested in this change?

mfrkn702001:
@buchen yes I am Still interested. I just had to Focus on there Tasks and Bugfixes with higher importance and could only spare time for this Today. I took care of the Changes as requested by you. 

I did not cover only one Part/Comment of yours. The Currency Part and the transactions for the Stock Changes across Portfolios from e.g. First Portfolio to Second Portfolio, are covered with our API. So we don't have to handle them here. 

But the other Points are covered and I refactored the code as you wished.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MyDividends24Uploader.java
```diff
@@ -47,29 +42,26 @@ public List<Pair<Integer, String>> getPortfolios() throws IOException
...
 
-        // Maps each portfolio to a pair of its index and its name and returns
-        // the list of pairs.
-        return IntStream.range(0, portfolios.size()).mapToObj(i -> new Pair<>(i, (String) portfolios.get(i))).toList();
+        return portfolios;
     }
 
     @SuppressWarnings("unchecked")
-    public void upload(Client client, CurrencyConverter converter, String portfolioId, Portfolio ppPortfolio)
+    public void upload(Client client, CurrencyConverter converter, String MY24PortfolioID, Portfolio portfolio)
```
Nirus2000, line 49: Please use consistent variables and function names.
MyDividends24Uploader vs. selectedMY24Portfolio vs. MyDividends24 vs. MY24PortfolioID ...

Use "myDividends24XXXXX" as an example.

Nirus2000 (approved):

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MyDividends24Uploader.java
mfrkn702001: Hi Nirus2000,
in my last Commit I change the Namings all to myDividends24xxxxx as you requested it. 
I want to mention that I will no longer be available for this project so I want to be this approved. I got fired at my Job, what reasons my unavailability.

buchen:
I have squashed and merged the changes.
I have removed the unnecessary and unrelated changes to class path, Imagio registration, pom.xml.
As you can see, I have added another commit with some more cosmetic changes to remove unneeded variable and improve the naming of variables (e.g. from "stringValue" to "md24Portfolio")


Thanks for your contribution.

mfrkn702001:
@buchen as this is my first Contribution to such an project like PortfolioPerformance dont know if my Changes got accepted and what it means that this pull request is closed.

OnkelDok:
@mfrkn702001, your PR was accepted (https://github.com/portfolio-performance/portfolio/commit/d1a3cef7bf996a9035fd0ef3f1d97a622d0d2d4f + https://github.com/portfolio-performance/portfolio/commit/b5456ae8470143154db12322028b278ac98c17a3). See this comment above: https://github.com/portfolio-performance/portfolio/pull/3697#issuecomment-1912576152

PR #3702, opened 2023-12-30
Add test case for Baader Bank PDF-Importer

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-baader-bank-scalable-capital/2057/297


buchen:
Is this PR missing the actual modification to the Importer?

The commit message says "modify importer to support new transaction" but it looks like it is "just" a new test case.

Nirus2000:
I'm to sloppy... is only add test case

buchen:
I'm half way through the publishing - will merge afterwards then

PR #3729, opened 2024-01-13
Modernize and secure temp file creation

zcarroll4:
This change replaces the usage of [`java.io.File#createTempFile`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/io/File.html#createTempFile(java.lang.String,java.lang.String)) with [`java.nio.file.Files#createTempFile`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/nio/file/Files.html#createTempFile(java.lang.String,java.lang.String,java.nio.file.attribute.FileAttribute...)) which has more secure attributes.

The `java.io.File#createTempFile()` method creates a file that is world-readable and world-writeable, which is almost never necessary. Also, the file created is placed in a predictable directory (e.g., `/tmp`). Having predictable file names, locations, and will lead to many types of vulnerabilities. History has shown that this insecure pattern can lead to [information leakage](https://www.cvedetails.com/cve/CVE-2021-28168/), [privilege escalation](https://www.cvedetails.com/cve/CVE-2021-29428/) and even [code execution](https://www.openwall.com/lists/oss-security/2022/02/25/3).

Our changes look something like this:

```diff
+  import java.nio.file.Files;
   ...
-  File txtFile = File.createTempFile("acme", ".txt");
+  File txtFile = Files.createTempFile("acme", ".txt").toFile();
```

<details>
  <summary>More reading</summary>

  * [https://cwe.mitre.org/data/definitions/378.html](https://cwe.mitre.org/data/definitions/378.html)
  * [https://docs.fluidattacks.com/criteria/vulnerabilities/160/](https://docs.fluidattacks.com/criteria/vulnerabilities/160/)
  * [https://github.com/apache/druid/issues/11130](https://github.com/apache/druid/issues/11130)
  * [https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File](https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File)
  * [https://nvd.nist.gov/vuln/detail/CVE-2022-41954](https://nvd.nist.gov/vuln/detail/CVE-2022-41954)
  * [https://www.cvedetails.com/vulnerability-list/cwe-378/vulnerabilities.html](https://www.cvedetails.com/vulnerability-list/cwe-378/vulnerabilities.html)
</details>

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
@@ -1,5 +1,6 @@
 package name.abuchen.portfolio.datatransfer.ibflex;
 
+import java.nio.file.Files;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -47,7 +48,7 @@ public class IBFlexStatementExtractorTest
 {
     private Extractor.InputFile createTempFile(InputStream input) throws IOException
     {
-        File tempFile = File.createTempFile("IBFlexStatementExtractorTest", null);
+        File tempFile = Files.createTempFile("IBFlexStatementExtractorTest", null).toFile();
         tempFile.deleteOnExit();
         FileOutputStream fos = new FileOutputStream(tempFile);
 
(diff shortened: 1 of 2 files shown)


kimmerin:
While chaging to the use of more recent APIs, I don't see much benefit here in terms of the issue you're trying to "fix".

One change happens within a test case, so you'll have leakage of public data - an attacker could simply look into this git-repository to see the content.

The other change happens when a temporary file is created in the installation directory as part of the installation. The files to be installed are not secret, so data leakage isn't an issue here and an attacke who has delete/write-access can simply exchange the DLLs in the installation instead of intercepting the installation/update process.

Or do I miss something here? Maybe we should add a note to the installation-process that running PP as root is discouraged?

drdavella:
Hi @kimmerin I'm a maintainer of @pixeebot. Thanks so much for the feedback; it's very much appreciated.

In this particular case it seems like this change does not address a major security issue (which is good!). However the purpose of this change is to be proactively defensive and it implements a recommended best practice. One of the benefits of @pixeebot is that it monitors _new_ PRs to your repository as well so it would also detect and correct any new code that introduced the older API.

If you're receptive we'd love to get feedback from OSS maintainers like yourself about the kinds of changes we introduce. Thanks again!

PR #3732, opened 2024-01-13
PDF import for Computershare

jimjaeger:
Pullrequest contains a pdf extractor and testcase for computershare documents .
Unfortunately CS only provides one document type and only purchase entries.
Dividends are only shown on the page but not in documents.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComputersharePDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComputersharePDFExtractor.java
@@ -1,0 +1,172 @@
+/**
+ * 
+ */
+package name.abuchen.portfolio.datatransfer.pdf;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import name.abuchen.portfolio.datatransfer.ExtractorUtils;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Block;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.DocumentType;
+import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Transaction;
+import name.abuchen.portfolio.model.BuySellEntry;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.money.Values;
+
+/**
+ * Extractor for Computershare (https://www.computershare.com/) depot documents.
+ * Supported: * Import transaction purchase entries from Statement documents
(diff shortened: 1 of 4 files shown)


Nirus2000:
Nope... plz reading the contributing rules.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComputersharePDFExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+
+
+        final DocumentType type = new DocumentType(DOCUMENT_IDENTIFIER_STATEMENT_DOC, 
+                        (context, lines) -> {
+                            // currently no parsable security information
+                            // identifier in statement document - needs line
+                            // guessing
+                            final Pattern singleDatePattern = Pattern.compile(PATTERN_STRING_SINGLE_DATE);
+                            final Pattern doubleDatePattern = Pattern.compile(PATTERN_STRING_DOUBLE_DATE);
+                            final Pattern cusipPattern = Pattern.compile(PATTERN_STRING_CUSIP);
```
buchen, line 71: I understand the thinking for extracting the pattern/attributes names into constants.
For reasons of readability, we have decided against it for most of the time.
Take the cusip pattern: the constant is defined and then used once, but now it is defined far away (in terms of lines) so that one cannot glance to see the context.

The thinking was: attributes must be consistent between ```match``` and ```assign``` but not beyond. So we scope it those methods.

Other than that it looks good to me (for now only looked in the browser)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComputersharePDFExtractor.java
```diff
@@ -0,0 +1,172 @@
...
+                                        .documentContext(PATTERN_KEY_TICKERSYMBOL, PATTERN_KEY_WKN)
+                        .match(PATTERN_PURCHASE_ENTRY)//
+                                        .assign((t, v) -> {
+                            v.put(PATTERN_KEY_CURRENCY, "USD");
+                            t.setDate(asDate(v.get(PATTERN_KEY_DATE)));
+                                            t.setCurrencyCode(asCurrencyCode("USD"));
+                            t.setAmount(asAmount(v.get(PATTERN_KEY_AMOUNT)));
+                            t.setShares(asShares(v.get(PATTERN_KEY_SHARES)));
+                                            final Security security = getOrCreateSecurity(v);
+                                            security.setCurrencyCode(asCurrencyCode("USD"));
```
buchen, line 129: This is not allowed.
The security matched by the importer might already exist in the data file of the user.
Therefore we must not change the currency (also considering that the security already has transactions and historical prices attached that assume another currency).

I will merge the importer but will change this before.

buchen:
Merged. With changes.

PR #3751, opened 2024-01-18
Modify sBroker PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-s-broker-sparkasse/5195/92
https://forum.portfolio-performance.info/t/pdf-import-von-s-broker-sparkasse/5195/95

Add some test cases
Update regulare expressions
Remove obsolet sections
Format source to standard eclipse format


buchen:
rebased and merged

PR #3762, opened 2024-01-20
Improve CoinGecko API to set call frequency limit

Nirus2000:
https://forum.portfolio-performance.info/t/ratelimit-der-coingecko-api-berucksichtigen/21808
Change of rateLimit for API calls.

Since we do not count the calls of API requests, it can happen that with many cryptocurrencies, the API is quickly at the limit.
As soon as the call frequency exceeds 0.1583, the 30 calls per minute are over very quickly. With this variant, the price updates run constantly.

The other problem is that if we start PP and then update the prices for around 50 cryptocurrencies and also add 1 or 2 new cryptocurrencies, we also exceed the limit. This is also the problem with counting the calls.

---

### 1st case
Start PP and wait until all prices are automatically updated. (with this PR -> ok)
(do not save the test file --> retry)

### 2nd case
Start PP and while all rates are updated, create additional new cryptocurrencies... approx. 2-4 pieces
Error... :-( API limit reached...
(do not save the test file --> retry)

Both can be tested with the test file.
[TestDepot.zip](https://github.com/portfolio-performance/portfolio/files/13999154/TestDepot.zip)


buchen:
merged

dakkusingh:
running Version: 0.68.1 (Mar 2024)
I still see this in my error log:

```
!ENTRY name.abuchen.portfolio 1 0 2024-03-12 13:47:22.865
!MESSAGE Saving myfile.xml with [XML]

!ENTRY name.abuchen.portfolio.ui 4 0 2024-03-12 13:47:51.990
!MESSAGE CoinGecko rate limit exceeded.
!STACK 0
name.abuchen.portfolio.util.RateLimitExceededException: CoinGecko rate limit exceeded.
	at name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed.getHistoricalQuotes(CoinGeckoQuoteFeed.java:209)
	at name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed.previewHistoricalQuotes(CoinGeckoQuoteFeed.java:131)
	at name.abuchen.portfolio.ui.wizards.security.HistoricalQuoteProviderPage$LoadHistoricalQuotes.run(HistoricalQuoteProviderPage.java:323)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

Issue #3767, opened 2024-01-27
URL for Credit Suisse (VIAC) quotes is not working anymore


KopolJunam:
**Description**
Credit Suisse has stopped replying to the URL that has been used up to now. There is a new URL that replies with a differently formatted file.

**To Reproduce**
Add a security with a CS/VIAC feed with an URL of the form:
https://amfunds.credit-suisse.com/ch/de/institutional/fund/history/CH0017844686?currency=CHF

The quotes will not be updated anymore and in the log you will get messages like:
`
java.io.IOException: https://amfunds.credit-suisse.com/ch/de/institutional/fund/history/CH0017844686?currency=CHF

Connect to https://amfunds.credit-suisse.com:443 [amfunds.credit-suisse.com/34.241.70.217, amfunds.credit-suisse.com/34.246.166.124] failed: Connect timed out
`

**Expected behavior**
The feed should provide quotes based on the new address and file format.

KopolJunam:
I will start working on this now.

KopolJunam:
I have finished work basically and will prepare a PR in the next few days. Maybe even this evening if I have time.

PR #3771, opened 2024-01-31
CS Quote Feed uses new URL and format (#3767)

KopolJunam:
[Implements what is described in issue 3767](https://github.com/portfolio-performance/portfolio/issues/3767)


buchen:
squashed, rebased, and merged.

I plan to publish an update later today.

PR #3773, opened 2024-02-01
Use getTextColor for the Limit Exceeded Widget and use black text color for Color.GREEN backgrounds

kimmerin:
Hi,

I've started using limits and with the first entries exceeding them, I had a hard time reading the quotes:

grafik

Pure green should have a black text color so I took the liberty to change the widget to actually set a forground color and use getTextColor for calulation. Because that had no effect due to the high threshold of 0.6 I've adjusted it to use 0.55 which "covers" pure green while keeping the threshold as close to the original one.

BTW:

        private Color redBackground = Colors.GREEN;
        private Color greenBackground = Colors.RED;

That looks wrong, so I've swapped the values.


kimmerin:
Hm, something went wrong, I'll close this because there are now a lot of changes that are not mine

PR #3774, opened 2024-02-01
Use getTextColor for the Limit Exceeded Widget and use black text color for Color.GREEN backgrounds (second attempt)

kimmerin:
Hi,

I've started using limits and with the first entries exceeding them, I had a hard time reading the quotes:

301603821-a72d2ca3-3858-4ae7-9512-7c4a1cf4bafe

Pure green should have a black text color so I took the liberty to change the widget to actually set a forground color and use getTextColor for calulation. Because that had no effect due to the high threshold of 0.6 I've adjusted it to use 0.55 which "covers" pure green while keeping the threshold as close to the original one.

BTW:

    private Color redBackground = Colors.GREEN;
    private Color greenBackground = Colors.RED;

That looks wrong, so I've swapped the values.


buchen:
I will merge this. It makes sense to calculate the foreground based on background color.

BTW, @chirlu did build the option to have custom colors per attribute. So you could also change the green there.

Bildschirmfoto 2024-02-01 um 20 49 42

buchen:
merged

buchen:
@kimmerin  I had to remove the test case - for whatever reasons, it does not run on the Github actions build machines initializing GTK.

PR #3789, opened 2024-02-11
Modify Swissquote Bank PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-swissquote-bank/15835/8


Nirus2000:
Hello @buchen 
I think we need a function analogous to the check by `documentContext("currency")` to move the `PeriodicHelper` or similar into the extractutils.
For example, `baseCurrency` and `year` should be found. (see [Postbank PDF Importer](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostbankPDFExtractor.java#L534:L554))

What did you think?

buchen:
> I think we need a function analogous to the check by documentContext("currency") to move the PeriodicHelper or similar into the extractutils.

If there is significant re-use, then why not.
Do you have an idea how you want to do it?

The current code around the ```PeriodicHelper``` is short and easily understandable. So one could make the ```PeriodicItem``` an abstract class with only the line number and each importer is extending it with whatever fields it needs (currency, year, date, whatever).

I looked at the change for the ```documentContext("currency")```. Supporting it here is much more tricky. Maybe we could allow each importer to define their own record class (containing date, currency, whatever is needed) and parse a pseudo block that collects those information in the range of the block (start - end line number). That sounds like a couple hours...


> The other point is that I can only check one currency via check in the tests, but not EUR, USD, GPB, CHF

Can't we just add the other currencies (of course, where needed, I don't think it makes sense to add this to all tests).

```
new AssertImportActions().check(results, CurrencyUnit.EUR);
new AssertImportActions().check(results, "USD");
```

Nirus2000:
Hello @buchen 
I have opened an issue or extension for this.
Let's bring this PR together and see how we can implement the idea.
Alex

PR #3793, opened 2024-02-15
fixed MsgPreviousStockSplit date and detail order

Hotdogkiro:
Changed note from: 
"Please note: there is stock split of 2022-07-18 recorded on 1:20 for this security. You have to manually convert shares for this transaction."
To:
"Please note: there is stock split of 1:20 recorded on 2022-07-18 for this security. You have to manually convert shares for this transaction."

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/StockSplitWarning.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/StockSplitWarning.java
@@ -20,7 +20,7 @@ public class StockSplitWarning
         if (evt.isPresent())
         {
             SecurityEvent event = evt.get();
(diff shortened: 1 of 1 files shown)


Morpheus1w3:
@Hotdogkiro 
Veto. This is a translation error at https://github.com/portfolio-performance/portfolio/blob/66eb7ba6c177cbd9db2244c4649e9d33ad147cc1/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties#L2067. Prior before changing the code, fix and review the language depending settings of `MessageFormat.format(Messages.MsgPreviousStockSplit` at https://github.com/portfolio-performance/portfolio/tree/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui. Otherwise you're changing the behaviour at all other languages.

buchen:
Thanks @Hotdogkiro also for pointing out. Bolzass also found this when going through the translations. It could have been fixed both ways: switching the order of the properties and fixing the German translation, or fixing the other translations. 🙃

PR #3796, opened 2024-02-18
Update messages.properties

Morpheus1w3:
#3793


buchen:
Thanks.

Bolzass (who is maintaining the Spanish) translation realized that as well and fixed the translations in POEditor. I now synced those changes back to the main branch. That makes this pr obsolete. But thank you very much.

PR #3748, opened 2024-01-17
new PDF Importer for J&T Direktbank

ZfT2:
support  J&T Direktbank Tagesgeldkonto-Auszüge

PR #3804, opened 2024-02-21
Improve PDF Importer for J&T Direktbank

Nirus2000:
Rename test files
Rename test cases
Remove BiConsumer
Remove contextProviderCommon map
Remove interestChargeTransaction for "Storno"
Cancellation is not an interest charge
Add note
Improve source code

---

Hello @ZfT2 and @buchen 
Unfortunately I have to change the source code of pull request #3748 again.

1. I have adapted the source code to the contributing rules.
2. I have renamed the test cases and the test files.
3. I have removed and optimized unnecessary source code.
4. I have added notes.
5. I have removed the " cancellation", which was posted as "interest charge". This is wrong and would lead to incorrect key figures. Instead, I have marked them as "not supported".

Regards Alex


ZfT2:
Hello @Nirus2000 ,

Thank you for optimizing this. But I think, this:

>     5. I have removed the " cancellation", which was posted as "interest charge". This is wrong and would lead to incorrect key figures. Instead, I have marked them as "not supported".

is still corect and if I see right, you kept it also. Because as seen in Kontoauszug 4, it really might happen.

Nirus2000:
> Hello @Nirus2000 ,
> 
> Thank you for optimizing this. But I think, this:
> 
> > ```
> > 5. I have removed the " cancellation", which was posted as "interest charge". This is wrong and would lead to incorrect key figures. Instead, I have marked them as "not supported".
> > ```
> 
> is still corect and if I see right, you kept it also. Because as seen in Kontoauszug 4, it really might happen.

We mark this incorrect booking by the bank as a "cancellation", but the user must remove the transaction manually as we do not support "cancellation".
It is not an interest charge, but a reversal.
If we were to post this as an interest charge, the key figures would be calculated incorrectly.

Issue #3795, opened 2024-02-17
Simplification of the identification of tags (PDF importer)


Nirus2000:
Hello @buchen 
I think we need a function analogous to the check by `documentContext("currency")` to move the `PeriodicHelper` or similar into the extractutils.
For example, `baseCurrency` and `year` should be found. (see [Postbank PDF Importer](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostbankPDFExtractor.java#L534:L554))

The other point is that I can only check one currency via check in the tests, but not EUR, USD, GPB, CHF
`new AssertImportActions().check(results, CurrencyUnit.EUR);`
However, if I only check one currency, it is always displayed incorrectly. We need a list of currencies.

What did you think?

_Originally posted by @Nirus2000 in https://github.com/portfolio-performance/portfolio/issues/3789#issuecomment-1937801138_

PR #3816, opened 2024-02-27
Added range attributes to the PDF parsing

buchen:
Range attributes are only valid within a given block, i.e. a start and and end line number. In order to use range attributes, one has to
* add range blocks to the document type
* add a rangeAttribute() to the section

If the given attribute does not exist, the parsing of the document fails.

Issue: #3795


Nirus2000:
Hello @buchen 
i like this... 👍🏻

Issue #3733, opened 2024-01-14
Stacked chart for Statement of Assets evolution chart


victor-marino:
**Is your feature request related to a problem? Please describe.**
No, it would be a feature request/improvement.

**Describe the solution you'd like**
In the Reports -> Statement of Assets -> Chart screen, it would be very convenient to have the option of showing the data as a **stacked** area chart. It's a perfect way to show the evolution of our "net worth" overtime.

**Describe alternatives you've considered**
Exporting the chart data as CSV and plotting it in Excel as a stacked area chart provides the same effect. But this requires using a third party app and exporting the data every single time we change or update the view. It would be great to have it built into Portfolio Performance, since all the data is already there.

**Additional context**
When looking at the evolution of our assets over time, a great way to visualize this is to use a **stacked** area chart. Compared to the current format, where all the series are superimposed over each other, a stacked option would allow us to nicely see the evolution of our net worth over time. We can currently toggle the "Show Area" option for individual series, but this is only a cosmetic change that doesn't stack the series.

Current chart in Portfolio Performance:
<img width="1256" alt="PP_no_stack" src="https://github.com/portfolio-performance/portfolio/assets/1933443/362c31fd-75af-4ba3-9202-5f2c588a9f67">

Example of the same data plotted as a stacked area chart in Excel:
![excel_stack](https://github.com/portfolio-performance/portfolio/assets/1933443/4e4115f6-e2ec-496f-84ba-ca2c6daf4c17)

Would you consider adding this option?

PR #3819, opened 2024-03-01
Proposition of new Stacked Chart in actual value for Taxonomy

mierin12:
Partially closes https://github.com/portfolio-performance/portfolio/issues/3733
Issue : https://forum.portfolio-performance.info/t/flaechendiagramm-von-portfoliobestandteilen-ueber-zeit/10954

Hello, 

This is a proposition of a new stacked chart graph, with real value in Y axis, in order to complement the existing stacked chart graph in %. It partially closes [3733 ](https://github.com/portfolio-performance/portfolio/issues/3733) as it is done in Taxonomy rather than in Statement of Asset. Several users were interested in the feature as per the forum link.

Given that there is already a stacked chart in %, it seems to me that this could be implemented with minimal changes. 
I basically copied and tried to mimic the existing code of the % stack chart with some small changes:

- `StackedTimeLineChart` vs the new `StackedTimeLineActualValueChart` : the % decimal format for tooltip is removed, which makes the tooltip format, by default, the same as in Statement of Asset charts (I think). 
- `StackedChartViewer` vs the new `StackedChartActualValueViewer` : 
  - removed the % decimal format for Y axis, replaced by the Thousand format. 
  - `getValue` method no longer divides by the total but by 100d : if I understand the format correctly, weigth and taxonomy value are stored in int/long with 2 decimals number.
  - in `rebuildChartSeries ` method : `chart.getAxisSet().adjustRange()` is moved after `chart.suspendUpdate(false)` as `adjustRange`does not work when done after a `chart.suspendUpdate(true)`
 
- Logos: well I did not find a new suitable logo for this new graph on iconmonstr. So currently, I am using the existing stack chart logo for the new actual value stacked chart, and proposing a new icon for the existing stacked chart, which is the same icon but upsidedown. (if unclear, the screenshot below will make it simple) It does match the graphics. I believe it is allowed to modify icon from iconmonstr ? It visually makes sense, however it may slightly change the habit of the users ?

- in both `StackedChartViewer` / `StackedChartActualValueViewer`, : `getModel().fireTaxonomyModelChange(getModel().getVirtualRootNode())` is called in `configMenuAboutToShow`. This allows to keep the graph updated with the state of "Sort by Classification" when switching between the two graphics. 

I do not know what kind of test could be done. I have checked visually on kommer and my own file, on Windows.
Capture d'écran 2024-03-01 221439

Capture d'écran 2024-03-01 220248

The percentage stacked graph is accessible through this icon :
Capture d'écran 2024-03-01 204505


buchen:
First: nice new feature. I like it. Thanks for the contribution.

> I believe it is allowed to modify icon from iconmonstr ? It visually makes sense, however it may slightly change the habit of the users ?

The iconmonstr license says: *"Licensee may adapt or change the Work according to his or her requirements."*. That covers in my understanding also taking the icons and modifying as needed.

I played around with the icons - maybe we can go into this direction - one showing the full area (= percentages) and one not (=actual/absolute values)

Bildschirmfoto 2024-03-04 um 16 58 05


About the code: I appreciate that you did not start by completely refactoring all code. But I think we can do some changes in order to reduce the amount of code duplication. Let me try to suggest some changes.

buchen:
Hey @mierin12,

I have refactored the stacked charts to de-duplicate the code a little bit.
Here is the pull request #3823.
Maybe you can test it and check if it still works as you designed it.

Andreas.

mierin12:
Hi @buchen 
I am glad the feature is appreciated !
I left a comment on pull request #3823, one line was not working. Once corrected, it is working exactly the same way as in this commit ! The % chart is still working. "Sort by Classification" is kept while changing tab and working. "Include Other Classification" is also kept while changing tab and working.

To be noted (applicable to both pull request): I do not quite understand how the min range of `adjustRange` is exactly calculated in the case of such stacked graph. Sometimes it gives a good margin to the lowest data serie, sometimes a lot less (even for high value, as per second picture). I am not sure if this "a bug or a feature" of `adjustRange`. 
Capture d'écran 2024-03-04 204809
Capture d'écran 2024-03-04 205026

(also, is the "Alt" action to sort the labels supposed to be working here ? I can't manage to trigger it, but we are not breaking it here, it is also the case on the existing release).

buchen:
Replaced by #3823

PR #3826, opened 2024-03-06
fix csv export for custom filename

mierin12:
Issue: https://forum.portfolio-performance.info/t/csv-export-von-konto-oder-depotumsatzen-nicht-lesbar/24565

Hello,
in the above forum thread, one csv export issue is mentioned. One user replies that the .csv extension can be lost when the file is renamed. 
Indeed, if we rename the automatic file name before saving it, I am reproducing the issue (on Windows).

If this is not intended, then this small proposition consists in adding a '*.csv' filter extension when exporting csv instead of `*.*`. 
Note: with the filter extension added, maybe `fileName.append(".csv"); //$NON-NLS-1$` might become unnecessary and can be deleted ? I did not notice a difference with or without it when testing this commit. I am using Windows, I cannot see the behaviour for MacOS or Linux (and maybe the issue is not even present outside Windows? ).

Summary with pictures:
Currently the exported file type extension is not set. :
Capture d'écran 2024-03-06 235730
If you rename the automatic filename, then the csv extension is lost.
With the filter extension, you can rename now the filename before saving it:
Capture d'écran 2024-03-07 000341

Capture d'écran 2024-03-07 000240

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/AbstractCSVExporter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/AbstractCSVExporter.java
@@ -22,6 +22,9 @@ public abstract class AbstractCSVExporter
 
     protected abstract Shell getShell();
 
+    private static final String[] FILTER_NAMES = { "CSV (*.csv)" }; //$NON-NLS-1$
+    private static final String[] FILTER_EXTS = { "*.csv" }; //$NON-NLS-1$
+
     protected abstract void writeToFile(File file) throws IOException;
 
     public void export(Object... labels)
@@ -48,6 +51,8 @@ public abstract class AbstractCSVExporter
     {
         FileDialog dialog = new FileDialog(getShell(), SWT.SAVE);
         dialog.setFileName(TextUtil.sanitizeFilename(fileName));
+        dialog.setFilterNames(FILTER_NAMES);
+        dialog.setFilterExtensions(FILTER_EXTS);
         dialog.setOverwrite(true);
         String name = dialog.open();
         if (name == null)


Morpheus1w3:
Well, this does not fix the problem as the user still could remove the suffix. You have to check at here:

https://github.com/portfolio-performance/portfolio/blob/67fee14b6f350fa2ebf905854b11994a36c8377c/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/AbstractCSVExporter.java#L53-L56

buchen:
Thanks @mierin12 for the contribution.

Apparently, this is platform-specific behavior. On Windows, the "File Save Dialog" has the option for the file types. And it will make sure that the file has the appropriate extension. (See [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=24877) which is an old Eclipse bug that proposes to fix the behavior to have the extension added and [here](https://bugs.eclipse.org/bugs/show_bug.cgi?id=332178) a bug that describes that the overwritten check did not used to work properly).

I tested it on macOS and Linux - there it does not make a difference. If the user deletes the extension, then the extension is gone. My thinking is: well, that then what you want. We should not change the file name after the fact because the check for overwriting an existing file will not be applied.

@Morpheus1w3 - which OS are you using?

Morpheus1w3:
@buchen 
I'm using Win10 but I do not have any complaints with the current implementation, because this is a standard Windows behaviour even at commercial apps like Word or Excel. If the user delete the file extension, which application should be assigned?

buchen:
> If the user delete the file extension, which application should be assigned?

I finally managed to install a ARM windows into my virtual machine and was able to test it.

On Windows, if the filter extension is given in the save dialog, then the file extension is always appended. With the change by @mierin12, the essentials forces the ".csv" extension. And the input field for the file name does not contain the extension anymore.

I have added it now to some other save dialogs.

buchen:
And I have to correct myself. It also works on macOS.

There, if I add a different extension to the file name, it will actually ask me which one I want to use:

Bildschirmfoto 2024-03-08 um 11 30 52

buchen:
And Linux behaves similar to Windows, only that the input field actually contains the file extension.

buchen:
merged.

I plan to release a new version later tonight.

mierin12:
> And the input field for the file name does not contain the extension anymore.

Actually on my side, Windows 11,the input field never show the extension in the filename. It seems that in @Morpheus1w3 case, it appears even on Windows. So it may not be only OS dependent ?
This is what I have in 0.67.3: no ".csv" in the input for name. So if want to slightly change this name, it brakes the extension, but unknowingly to the user :
Capture d'écran 2024-03-08 113505

PR #3838, opened 2024-03-12
Fix some label position

mierin12:
Hello PP team,
Very small proposition for a very small 'bug':
In the Custom Event Page, the location of label is not ideal in the case of a wide "EventLabel". This is unvisible in English or Deutsch language, but visible in Spanish and French language.

**Current:**
Capture d'écran 2024-03-13 002145
Capture d'écran 2024-03-13 002333

**After Fix:**
Capture d'écran 2024-03-13 002438

(Please note that I am unable to check the impact of the fix for other language than French, as when launched from Eclipse, the language setting is set by default on Automatic (this is a recent strange behaviour), so French for me. If I change the language, close and start again, I am stuck on automatic.. Only happening from Eclipse, the current official release does not have this issue)

I have also updated the Stock Split event in a similar way (once again, invisible in English & Deutsch setting):
**Before**
Capture d'écran 2024-03-13 002604

**After**
Capture d'écran 2024-03-13 004019

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/events/AddCustomEventPage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/events/AddCustomEventPage.java
@@ -71,7 +71,7 @@ public class AddCustomEventPage extends AbstractWizardPage
         setControl(container);
         container.setLayout(new FormLayout());
 
-        Label labelSecurity = new Label(container, SWT.NONE);
+        Label labelSecurity = new Label(container, SWT.RIGHT);
         labelSecurity.setLayoutData(new FormData());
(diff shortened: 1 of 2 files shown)


Nirus2000:
I just saw that the labels say "Ex-Date"... That is not correct. We are currently working with the value date.
The pull request for the ex-date is still being processed and is waiting for a decision, so @buchen is to correct the label.

Greetings
Alex

buchen:
@Nirus2000 writes

> I just saw that the labels say "Ex-Date"...

The "dividend event" already has an ex-date, but.... the dividend transition does not. But the latter is the relevant one in order to adapt the processing and calculations

buchen:
@mierin12 thanks for the fix. It makes it more consistent. The width was definitely missed.

PR #3843, opened 2024-03-13
Light mode : set selected tab to white

mierin12:
Issue : https://forum.portfolio-performance.info/t/payment-tab-color-in-light-mode/27659

This change of color affects Performance/Calculation tabs + Performance/Payment tabs.

A very naive modification : is that ok to use a fixed color instead of some "#org-eclipse-ui-workbench-" colors here ? `white` is already used as the hover tab color. I cannot see the effect on Mac and Linux.

On my Windows this change from:
Capture d'écran 2024-03-13 231843
to
Capture d'écran 2024-03-13 231918

(Another possible way would be to modify the unselected tab instead with **IN**ACTIVE_UNSELECTED instead of ACTIVE_UNSELECTED
`swt-unselected-tabs-color: '#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_START' '#org-eclipse-ui-workbench-INACTIVE_UNSELECTED_TABS_COLOR_END' 100% 100%;` which produces darker grey unselected tab:
Capture d'écran 2024-03-13 231956

but it is still not super different from the selected in my view, but it allows to keep "#org-eclipse-ui-workbench-" colors for both the selected and unselected tab.)
Also possible to mix and use darker grey for unselected and white for selected.


buchen:
This is how it looks on mac (top: new color, bottom: old color)

Bildschirmfoto 2024-03-19 um 13 02 25

This is how it looks on Linux (top: new, bottom: old)

Bildschirmfoto 2024-03-19 um 13 07 56

Bildschirmfoto 2024-03-19 um 13 05 25

I like the change on linux, but I am not so happy on mac.
So I think the best way is that we override the change on mac to revert back to the original color.

buchen:
merged.

Issue #1506, opened 2020-04-21
Management of free shares allocation: allow null value transaction?


Alain-CYS:
Hi,

I'm trying to track the allocation of free shares (which is different of getting dividend paid in shares rather than in cash).

For example, every 2 years Air Liquide (AI.PA) gives free shares. In 2019, they give 1 free share for 10 shares owned.

To track this in Portfolio Performance, I tried to "buy" the free share at 0€ price but PP doesn't not allow a null value transaction. I could use the "stock split" (11 for 10) but I'm not sure this allocation of free share should be considered a stock split.

For now, the workaround I use is to "buy" the free share at 0,01€ price, which seems to be the minimum allowed by PP for a transaction (so for 2 shares I would use a 0,005€ price for a 0,01€ value transaction). But the real price should be 0€, that's how the broker consider it.

Am I missing the correct way to manage it?
If not, could you allow transaction at null price in PP?

Regards.

buchen:
It is not supported directly unfortunately.

But what you can do is:
- create a purchase for the free shares
- create a dividend payment with the same value

Then the purchase is performance neutral, but the dividend statement (= the shares) is counted towards the performance.

I think setting the purchase with 0,01 Euro is not correct because the shares actually have a value. And that value should be reflected in the performance.

Alain-CYS:
Hi Andreas,

Those "free" shares indeed do have a value but they don't have a cost (at least, not explicitly and not from a tax point of view).

The problem with the dividend/purchase method is that the "purchase price" calculated by PP differs from the one calculated by the broker, which is the one used from a tax point of view.
From what I could see, the broker really considered those free shares at a null cost, i.e. zero purchase price.
But, as far as I can tell, once those new shares have been allocated, it seems the historical quotes have been adjusted to take into account the creation of those "free" shares (like in a stock split). For example, if the historical quote was 100€, with the new shares it has been adjusted to 90,9€ (11 for 10 ratio).

I have another security which offered to pay dividend in shares instead of cash.
In this case, the dividend/purchase method perfectly works and the "purchase price" value coincides with the one calculated by the broker.

cmaoling:
Hi @alain, I might miss something here, but I am using the "Delivery Inbound" record in these cases... Does this also imply some issues down the road?

Alain-CYS:
@cmaoling I just try it but the problem is the same with "delivery inboud": I can't input a null value for "quote" and by consequence for "value of inbound delivery".
[...]

PR #3847, opened 2024-03-14
trade republic document import "SPLIT"

christen90:
adds document "split"

would also be good to fix the gui to enable users also to set value to 0.
#1506

BR Christian

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/Split01.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/Split01.txt
@@ -1,0 +1,24 @@
+```
+PDFBox Version: 1.8.17
+Portfolio Performance Version: 0.68.3.qualifier
+-----------------------------------------
+TRADE REPUBLIC BANK GMBH  KASTANIENALLEE 32  10435 BERLIN
+Vorname Nachname SEITE 1 von 1
+Straße 1 DATUM 04.06.2022
+12345 Ort DEPOT 0000000000
+SPLIT
+ÜBERSICHT
+NR. BUCHUNG WERTPAPIER BETRAG
+1 Einbuchung Amazon.com Inc. 17,4743 Stk.
+Registered Shares DL -,01
+US0231351067
+Amazon.com Inc. Registered Shares DL -,01 in Girosammelverwahrung in Deutschland.
+Diese Abrechnung wird maschinell erstellt und daher nicht unterschrieben.
+Sofern keine Umsatzsteuer ausgewiesen ist, handelt es sich gem. § 4 Nr. 8 UStG um eine umsatzsteuerfreie 
+Leistung.
+Trade Republic Bank GmbH www.traderepublic.com Sitz der Gesellschaft: Düsseldorf Geschäftsführer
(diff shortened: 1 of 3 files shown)


Nirus2000:
Hello @christen90 
I don't think what you're doing here is a good idea. 
A split is not just a posting, but various transactions or the existing securities must be derecognized and posted with the new ones. As we cannot determine the current position of the security in the portfolio using the PDF importer, this variant is not recommended.
These are so-called "special booking types" and should be carried out manually by the user. However, we can mark this transaction as "NonImportable" and set it with the correct "Message" (see other importers).

Regards
Alex

christen90:
Hello Alex,
you are right.

Using the Split function in GUI did the trick of recalculating all buys and fixed the Einstandskurs (and elliminated the drop of value in diagram which came due to split.
Somehow I needed to change the Kurslieferant. Portfolio Report still had the drop in their data. Yahoo Finance delivered recalculated data.

I will look into the marking of NonImportable item.

 BR Christian

christen90:
Hello Alex,
I need support. I don't get the error message set. Is it already implemented to attach a Error Message to a Portfolio Transaction? I only found examples of Account Transactions.

BR Christian

Nirus2000:
Hello @christen90 
little helper... ;-)

```
                        .wrap((t, ctx) -> {
                            TransactionItem item = new TransactionItem(t);

                            if (ctx.getString(FAILURE) != null)
                                item.setFailureMessage(ctx.getString(FAILURE));

                            return item;
                        });
```

And the test like this...

```
    @Test
    public void testBlabla01()
    {
        BlaBlaPDFExtractor extractor = new BlaBlaPDFExtractor(new Client());

        List<Item> results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "BlaBlaBlaBla.txt"),
                        errors);

        assertThat(errors, empty());
        assertThat(countSecurities(results), is(1L));
        assertThat(countBuySell(results), is(0L));
        assertThat(countAccountTransactions(results), is(1L));
        assertThat(results.size(), is(2));
        new AssertImportActions().check(results, CurrencyUnit.EUR);

        // check security
        assertThat(results, hasItem(security( //
                        hasIsin("XXX"), hasWkn("XXX"), hasTicker(null), //
                        hasName("XXX"), //
                        hasCurrencyCode("EUR"))));

        // check unsupported transaction
        assertThat(results, hasItem(withFailureMessage( //
                        Messages.MsgErrorTransactionTypeNotSupported, //
                        dividend( //
                                        hasDate("2000-01-01T00:00"), hasShares(1.234), //
                                        hasSource("BlaBlaBlaBla.txt"), //
                                        hasNote(null), //
                                        hasAmount("EUR", 1.23), hasGrossValue("EUR", 1.23), //
                                        hasForexGrossValue("USD", 1.23), //
                                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00)))));
    }
```

christen90:
Hello Alex,
thanks for the hint. I think it should now work to mark them with error. However it would be more helpful for the user to give a hint that a Split wizard exists.
I tried to add a specific message, but it seems to be required to include all translations :/
Too much effort.

BR Christian

buchen:
> I tried to add a specific message, but it seems to be required to include all translations :/

It does not make sense for each and every special case. But in this case it is a generic message and could be reused, so I took the liberty to create an extra label.

At the moment, I am flipping between Google Translate and DeepL. The latter was my favorite for quite some time, but lately I do like the translations of the former better.

PR #3853, opened 2024-03-18
dynamic factor for calculation of y for dividend markers

OnkelDok:
to prevent dividend markers being drawn to far below the chart for charts with small y-value range
see https://forum.portfolio-performance.info/t/nachkommastellen-von-dividenden-im-kurs-diagramm/26378/28

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -1358,10 +1358,12 @@ public class SecuritiesChart
 
             IAxis yAxis1st = chart.getAxisSet().getYAxis(0);
 
-            // factor converges to 0.84 for large ranges
-            // and to 1 for small ranges to prevent dividend markers being drawn
-            // to far below the chart
-            double factor = 1 - (1 - (Math.max(yAxis1st.getRange().lower, 0) / yAxis1st.getRange().upper)) / 6.25;
+            // Factor should be between 0.84 (for wide range charts) and 1 (for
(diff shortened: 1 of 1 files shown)


Nirus2000:
Hello @OnkelDok 
Quick question... How did you arrive at the value 6.25?
Can you check the whole thing again if you use different font sizes?

I think another Y-axis, which is arranged in the lower 8% area, which would then contain the dividend markers would be a better solution, especially as this would also solve the zoom-in/out problem.
However, I am not sure if SWT accepts another Y-axis.

Would that be an idea?

Regards
Alex

OnkelDok:
> How did you arrive at the value 6.25

Start formula is ``1 - (1-R)/x = 0.84`` where ``R=lower/upper``. And for wide y-ranges (``lower << upper``) R diverges to 1. So we have ``1 - 1/x = 0.84``. Now simple formula conversion results in: ``x = 6.25``

So as a result we have a dynamic factor which is a value between 1 and 0.84. So the dividend marker is placed almost the same as before for charts with wide y-range. And for charts with small y-range the dividend markers are placed closer to the chart.

I've not tested different font sizes, because the change only affects the initial y-value of the dividend markers and not the existing placements for the labels.
If I find time, I can check this.

Maybe there are other solutions with additional y-axis. But that would require more investigation.
So you can see my PR as a quickfix/hotfix for the existing unfavorable scaling for small-range-charts.

pfalcon:
I also reported such an issue about a year ago or so, didn't have a chance to look into it myself. Confirming it fixes display where there were issues for us, and doesn't seem to deteriorate where there weren't. Thanks!

OnkelDok:
> I also reported such an issue about a year ago

In a github issue? Can you link it here for closing it, in case my PR is accepted?
Thank you for testing.

PR #3874, opened 2024-03-24
Modify Trade Republic PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-trade-republic/5107/489

Merge/Remove addLiquidationTransaction()
Merge/Remove addCaptialReductionTransaction()
Merge/Remove addCaptialReductionTransaction()
Remove addDeliveryInOutBoundTransaction()
Remove addTaxStatementTransaction();
Add addNonImportableTransaction()
Add notes
Fix testWertpapierVerkauf08()
Rename test file USQuellensteuer01.txt to Steuerkorrektur01.txt

There were previously some imports that were not solved correctly. For example, a SPIN-OFF was posted as a posting at 0.00. Also inbound/outbound deliveries which have no currency or amount are to be marked as NonImportable. The same applies to fiscal exchanges. Once without tax collection but delivery or derecognition has taken place without the currency of the security and taxes may have been offset. All in all, these are all special cases which we cannot import correctly and clearly.


christen90:
Hi Alex,
why do we import savebacks as buys? I would book them as inbound delivery, as they do not cause costs on depotkonto.

BR Christian

Nirus2000:
> Hi Alex, why do we import savebacks as buys? I would book them as inbound delivery, as they do not cause costs on depotkonto.
> 
> BR Christian

I asked in the forum how this should be booked, but now that you say it, it makes sense. 
Unfortunately I don't know "Saveback".

PR #3891, opened 2024-03-27
Fix Bigbank PDF Importer ignoring cent amounts

MonkeySon:
Bigbank uses the Estonian number format in their account statement transactions, namely a comma as decimal separator.

Before, numbers were parsed by the ExtractorUtils with the de_CH country code, which uses a dot as decimal separator. This has the effect that cent amount are ignored, but the test still passed since it did not include cent amounts.

This pull request:
* Modifies the test case to include cent amounts
* Adjust the ExtractorUtils' country code from de_CH to et_EE (see: https://www.localeplanet.com/icu/et-EE/index.html)


Nirus2000:
Hello @MonkeySon 
It shows how important test cases and clean source code are.
Good job! 👍🏻 

Rebase... merge...

PR #3913, opened 2024-04-04
Add messages to non-descript Exceptions

kimmerin:
Hi,

in past discussions in the forum there often are failures to e.g. load a saved portfolio file where a non-descript InvalidArgumentException is thrown and you can only tell what happened but have no clue what actual data lead to this. So I've gone through the source to look for occurrences and added exception messages that contain parts of the data that lead to the problem with the hope that this might help with future occasions of this type.

### DIFF
--- a/name.abuchen.portfolio.bootstrap/src/name/abuchen/portfolio/bootstrap/CustomSaveHandler.java
+++ b/name.abuchen.portfolio.bootstrap/src/name/abuchen/portfolio/bootstrap/CustomSaveHandler.java
@@ -38,7 +38,7 @@ public class CustomSaveHandler extends PartServiceSaveHandler
     {
         Object clientInput = dirtyPart.getTransientData().get(ModelConstants.EDITOR_INPUT);
         if (clientInput == null)
-            throw new IllegalArgumentException();
+            throw new IllegalArgumentException("no client input found");
 
         EModelService modelService = dirtyPart.getContext().get(EModelService.class);
         MApplication app = dirtyPart.getContext().get(MApplication.class);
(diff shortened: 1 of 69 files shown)


buchen:
Fair. I always thought those are programming errors and need a fix. But, yes, a hint in the message can help people understand what are possible workarounds. I also agree with not translating them: the goal must be that these exceptions never show up anyway.

I am a little unsure about the extensive check for the translated message in the test case. My goal is to never have to change a test case anymore. Of course, this is unrealistic. But it is something to strive for. And labels are something that changes relatively easily.

Give me some time to review and merge.

kimmerin:
> I am a little unsure about the extensive check for the translated message in the test case. My goal is to never have to change a test case anymore.

When writing test cases I'm always trying to get good - if possible 100% - coverage with the effect that any change in the logic of some class (which includes changes in log- or other messages) without adjustment of the corresponding test case leads to a test failure. But in a way it's the same goal as you have: A test case doesn't need to be changed as long as the CuT hasn't changed it's logic.

buchen:
>  coverage with the effect that any change in the logic of some class (which includes changes in log- or other messages) without adjustment of the corresponding test case leads to a test failure.

Agree. I think writing good tests might be even harder than writing the application code. It's an art form ;-) I have seen projects that had great test coverage but any change required also touching existing test cases due to the extensive use of mocking.

That said, I don't think the test is bad, but the message is also just a debug output. Hopefully tomorrow I get to merge this PR.

buchen:
I added the tags that prevent warnings on non-externalized strings as I try to keep the source code without those warnings.

Rebased and merged.

Issue #3896, opened 2024-04-02
CoinGecko free API limited to querying historical data to past 365 day


Nirus2000:
@buchen 
https://forum.portfolio-performance.info/t/kryptowahrung-ckb-nervos-network-kurse-werden-nicht-dargestellt-coingecko/27981/6

Sn1kk3r5:
Just FYI

a workoaround is available. During creation of a Crypto Currency, deactivate the auto download. Create a manual historic quote (Less than 365 days in the past), reactivate auto download.

buchen:
I limit it now to max 365 days. I tested locally and then I can load historical prices.

PR #3917, opened 2024-04-05
Added option to configured Coingecko API key from a paid plan

buchen:
Issue: #3896

As discussed, optionally allow the user to provide an API key that
a) increases the rate limit
b) allow for more historical data

One caveat: the test API key does not work with the pro API endpoints. But just looking at the test key, I do not see an option to understand if is a test or pro API key. As the test API does not provide more data/rate limit, we simply do not support it.


Nirus2000:
Hello @buchen 
thx for review...

I think we should merge these translations. They are the same except that the plural has been formed, which is unnecessary. There is no input with multiple API keys.
```
  PrefTitleAPIKey = API Key
  PrefTitleAPIKeys = API Keys
```
---
In these translations, the semicolon should be replaced by a dot.
```
  PrefDescriptionCoingecko
```

Otherwise the rest should fit... 👍🏻 
Alex

buchen:
> I think we should merge these translations.

The plural form is used for the section (all API Keys), the singular is used to enter one specific API Key.
As the plural form is already available and translated, I am not sure if there is a benefit in deleting the key.
And it is useful for a section that is about multiple API keys.

I added the singular form without the name of the product to be able to use it for more keys in the future. Right now there are "DivvyDiary API Key", and "AlphaVantage API Key", etc. That was unnecessary.

> the semicolon should be replaced by a dot.

Probably the reason is that the English sentence (suggested by ChatGTP) has the semicolon and then it got into the translations. Also in German it is grammatically correct, but I agree it is rarely used. I'll fix it later and update the PR.

buchen:
fixed translations and merged.

Issue #2075, opened 2021-02-12
Request: Ability to search the 'Trades' screen


metal450:
The 'Search' box at the top of the 'All Transactions' screen is indispensable.  It'd be extremely useful to have the same on the 'Trades' screen.

mierin12:
Done since [release 0.68.4](https://github.com/portfolio-performance/portfolio/releases/tag/0.68.4) with https://github.com/portfolio-performance/portfolio/pull/3926

Issue #1597, opened 2020-06-14
Add filter to the Trade view


iwalucas:
Hello, could you add the filtering options on the Trade view?

I use that for tax reporting and usually I have to download and filter it on excel.

My usual filters (which may be a good insight if you think thats a good feature):

Filter by date range (applied on the Close Date)
Filter by the regular filter on the other views under Performance

pegasus87:
I would agree to that and just make another consideration:

Add a filter button to just show the open/closed trades in the list.

debegr92:
That would be a great feature! The CSV export should then only export the current filtered trades.
Maybe we can add filters like in Excel? So that every column can be filtered?
A filter for the holding period would be great.

metal450:
Also: it would be nice to be able to filter by custom sets of accounts (i.e. this makes it possible to filter only taxable accounts, for instance). We can currently do this on the Earnings & Expenses screen.

Sn1kk3r5:
Search and filter partially done

PR #3926, opened 2024-04-10
Add a Search box in Trades view

mierin12:
Issues : https://github.com/portfolio-performance/portfolio/issues/2075

This is a proposition to allow filtering the trades by a security name/ISIN/WKN. It is a copy paste of what is already existing in `SecurityListView`.

Well, I actually did not find a lot of request for it, but it seems relevant I guess ? Trades list can get quite long.

If relevant:

- is a search by security name/ISIN/WKN sufficient ? The search for AllTransaction is a bit different, covering all columns I think, but its code is a bit more complex to understand. So I went the easy way and mimicked `SecurityListView`..
- Do we want a Trades search to apply to more columns ? Securities Account ? Note ? numbered column (holding period, IRR..) ?

Side note :
It makes me notices that the SecurityList, TransactionList and Taxonomy search buttons do not search for the same items. Maybe an hovering label above each search button could add more information about the exact use of each search box ?

On the code itself, I was unsure if `table.getTableViewer().addFilter` should be before or after `update()` in `createBody`. Both worked on my side.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
@@ -2,6 +2,7 @@ package name.abuchen.portfolio.ui.views.trades;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


metal450:
Thanks for this! If it's similar/relatively easy, it would also be very useful to be able to *filter* the trades view :) 🙏  https://github.com/portfolio-performance/portfolio/issues/1597

mierin12:
@metal450 I think this should do the trick https://github.com/portfolio-performance/portfolio/pull/4005

metal450:
Awesome, thanks! Will try it out once it's in the release

PR #3927, opened 2024-04-10
added widget for the dashboard and columns for the information of dividends

kimmerin:
Hi,

in #1669 and in the forum there seems to be demand for information widgets and columns showing informations about upcoming dividends. So here is my shot at it. This adds a widget for the dashboard, showing a list of upcoming dates (ex dividend, payment or both) of securities and the expected amounts. As well I've added columns for various tables (security view, etc.) that shows the date of the next ex/payment date for each security.

I've added new entries to be localized at the end of Messages and the properties-files for german and english to make it easier to spot them for the other languages. They can be resorted at the end of the process, of course.

Hope that helps and cheers, Lothar

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidgetTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidgetTest.java
@@ -1,0 +1,251 @@
+package name.abuchen.portfolio.ui.views.dashboard;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertEquals;
+
+import java.time.LocalDate;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.model.SecurityEvent;
+import name.abuchen.portfolio.model.SecurityEvent.DividendEvent;
+import name.abuchen.portfolio.money.Money;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateEndRange;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateStartRange;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateType;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DividendItem;
+
+public class DividendListWidgetTest
+{
+    private Locale defaultLocale;
+
+    private Security sec1;
+    private Security sec2;
+    private DividendEvent de1_1;
+    private DividendEvent de1_2;
+    private SecurityEvent de1_3;
+    private DividendEvent de1_4;
+    private DividendEvent de2_1;
+    private DividendEvent de2_2;
+    private SecurityEvent de2_3;
+    private DividendEvent de2_4;
+
+    @Before
+    public void setupLocale()
+    {
+        defaultLocale = Locale.getDefault();
+        Locale.setDefault(Locale.GERMANY);
+    }
+
+    @After
+    public void resetLocale()
+    {
+        Locale.setDefault(defaultLocale);
+    }
+
+    @Before
+    public void setupTestData()
+    {
+        sec1 = new Security("Security 1", "EUR");
+        de1_1 = new DividendEvent(LocalDate.of(2024, 4, 9), LocalDate.of(2024, 4, 12), Money.of("EUR", 55), "source");
+        de1_2 = new DividendEvent(LocalDate.of(2024, 7, 14), LocalDate.of(2024, 8, 11), Money.of("EUR", 22), "source");
+        de1_3 = new SecurityEvent(LocalDate.of(2024, 2, 3), SecurityEvent.Type.NOTE, "some note");
+        de1_4 = new DividendEvent(LocalDate.of(2024, 2, 4), LocalDate.of(2024, 3, 5), Money.of("EUR", 33), "source");
+        sec1.addEvent(de1_1);
+        sec1.addEvent(de1_2);
+        sec1.addEvent(de1_3);
+        sec1.addEvent(de1_4);
+
+        sec2 = new Security("Security 2", "EUR");
+        de2_1 = new DividendEvent(LocalDate.of(2024, 4, 5), LocalDate.of(2024, 4, 9), Money.of("USD", 55), "source");
+        de2_2 = new DividendEvent(LocalDate.of(2024, 7, 14), LocalDate.of(2024, 7, 15), Money.of("USD", 122), "source");
+        de2_3 = new SecurityEvent(LocalDate.of(2024, 8, 25), SecurityEvent.Type.STOCK_SPLIT, "some stock split");
+        de2_4 = new DividendEvent(LocalDate.of(2024, 2, 5), LocalDate.of(2024, 3, 5), Money.of("USD", 1333), "source");
+        sec2.addEvent(de2_1);
+        sec2.addEvent(de2_2);
+        sec2.addEvent(de2_3);
+        sec2.addEvent(de2_4);
+
+    }
+
+    @Test
+    public void testDividendItemInstantiation()
+    {
+        DividendItem di;
+        
+        di = new DividendItem(DateType.EX_DIVIDEND_DATE, sec1, de1_1);
+        assertThat(di.getSecurity(), is(sec1));
+        assertThat(di.type, is(DateType.EX_DIVIDEND_DATE));
+        assertThat(di.div, is(de1_1));
+    }
+
+    @Test
+    public void testDateStartRange()
+    {
+        assertEquals("FROM_TODAY,FROM_ONE_WEEK,FROM_ONE_MONTH,FROM_YTD", //
+                        Arrays.stream(DateStartRange.values())
+                        .map(DateStartRange::name).collect(Collectors.joining(",")));
+
+        LocalDate now = LocalDate.of(2024, 4, 8);
+        assertThat(DateStartRange.FROM_TODAY.getDate(now), is(now));
+        assertThat(DateStartRange.FROM_ONE_WEEK.getDate(now), is(LocalDate.of(2024, 4, 1)));
+        assertThat(DateStartRange.FROM_ONE_MONTH.getDate(now), is(LocalDate.of(2024, 3, 8)));
+        assertThat(DateStartRange.FROM_YTD.getDate(now), is(LocalDate.of(2024, 1, 1)));
+    }
+
+    @Test
+    public void testDateEndRange()
+    {
+        assertEquals("UNTIL_EOY,UNTIL_ONE_MONTH,UNTIL_ONE_WEEK,UNTIL_TODAY", //
+                        Arrays.stream(DateEndRange.values()).map(DateEndRange::name).collect(Collectors.joining(",")));
+
+        LocalDate now = LocalDate.of(2024, 4, 8);
+        assertThat(DateEndRange.UNTIL_TODAY.getDate(now), is(now));
+        assertThat(DateEndRange.UNTIL_ONE_WEEK.getDate(now), is(LocalDate.of(2024, 4, 15)));
+        assertThat(DateEndRange.UNTIL_ONE_MONTH.getDate(now), is(LocalDate.of(2024, 5, 8)));
+        assertThat(DateEndRange.UNTIL_EOY.getDate(now), is(LocalDate.of(2024, 12, 31)));
+    }
+
+    @Test
+    public void testDateType()
+    {
+        assertEquals("ALL_DATES,EX_DIVIDEND_DATE,PAYMENT_DATE", //
+                        Arrays.stream(DateType.values()).map(DateType::name).collect(Collectors.joining(",")));
+
+        assertThat(DateType.ALL_DATES.getDate(de1_1), nullValue());
+        assertThat(DateType.EX_DIVIDEND_DATE.getDate(de1_1), is(de1_1.getDate()));
+        assertThat(DateType.PAYMENT_DATE.getDate(de1_1), is(de1_1.getPaymentDate()));
+    }
+
+    @Test
+    public void testGetUpdateTask() throws Exception
+    {
+        LocalDate testnow = LocalDate.now();
+        @SuppressWarnings("deprecation")
+        AbstractSecurityListWidget<DividendItem> widget = new DividendListWidget()
+        {
+            @Override
+            public Supplier<List<DividendItem>> getUpdateTask(LocalDate now)
+            {
+                assertThat(now, is(testnow));
+                return null;
+            }
+        };
+        assertThat(widget.getUpdateTask(), nullValue());
+    }
+
+    @Test
+    public void testGetUpdateTaskNow()
+    {
+        List<DividendItem> list;
+        AtomicReference<DateStartRange> dateStart = new AtomicReference<>(DateStartRange.FROM_TODAY);
+        AtomicReference<DateEndRange> dateEnd = new AtomicReference<>(DateEndRange.UNTIL_TODAY);
+        AtomicReference<DateType> dateType = new AtomicReference<>(DateType.ALL_DATES);
+
+        @SuppressWarnings("deprecation")
+        DividendListWidget widget = new DividendListWidget()
+        {
+            @Override
+            DateStartRange getStartRangeValue()
+            {
+                return dateStart.get();
+            }
+
+            @Override
+            DateEndRange getEndRangeValue()
+            {
+                return dateEnd.get();
+            }
+
+            @Override
+            DateType getDateTypeValue()
+            {
+                return dateType.get();
+            }
+
+            @Override
+            List<Security> getSecurities()
+            {
+                return Arrays.asList(sec1, sec2);
+            }
+        };
+
+        LocalDate now = LocalDate.of(2024, 4, 9);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("2024-04-09 Security 1 EUR 0,55 Ex-Tag\r\n" //
+                        + "Security 2 USD 0,55 Zahltag", getListAsString(list));
+
+        now = LocalDate.of(2024, 4, 8);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("", getListAsString(list));
+
+        dateStart.set(DateStartRange.FROM_YTD);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("2024-02-04 Security 1 EUR 0,33 Ex-Tag\r\n" //
+                        + "2024-02-05 Security 2 USD 13,33 Ex-Tag\r\n" //
+                        + "2024-03-05 Security 1 EUR 0,33 Zahltag\r\n" //
+                        + "Security 2 USD 13,33 Zahltag\r\n" //
+                        + "2024-04-05 Security 2 USD 0,55 Ex-Tag", getListAsString(list));
+
+        dateEnd.set(DateEndRange.UNTIL_EOY);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("2024-02-04 Security 1 EUR 0,33 Ex-Tag\r\n" //
+                        + "2024-02-05 Security 2 USD 13,33 Ex-Tag\r\n" //
+                        + "2024-03-05 Security 1 EUR 0,33 Zahltag\r\n" //
+                        + "Security 2 USD 13,33 Zahltag\r\n" //
+                        + "2024-04-05 Security 2 USD 0,55 Ex-Tag\r\n" //
+                        + "2024-04-09 Security 1 EUR 0,55 Ex-Tag\r\n" //
+                        + "Security 2 USD 0,55 Zahltag\r\n" //
+                        + "2024-04-12 Security 1 EUR 0,55 Zahltag\r\n" //
+                        + "2024-07-14 Security 1 EUR 0,22 Ex-Tag\r\n" //
+                        + "Security 2 USD 1,22 Ex-Tag\r\n" //
+                        + "2024-07-15 Security 2 USD 1,22 Zahltag\r\n" //
+                        + "2024-08-11 Security 1 EUR 0,22 Zahltag", getListAsString(list));
+        
+        dateType.set(DateType.EX_DIVIDEND_DATE);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("2024-02-04 Security 1 EUR 0,33 Ex-Tag\r\n" //
+                        + "2024-02-05 Security 2 USD 13,33 Ex-Tag\r\n" //
+                        + "2024-04-05 Security 2 USD 0,55 Ex-Tag\r\n" //
+                        + "2024-04-09 Security 1 EUR 0,55 Ex-Tag\r\n" //
+                        + "2024-07-14 Security 1 EUR 0,22 Ex-Tag\r\n" //
+                        + "Security 2 USD 1,22 Ex-Tag", getListAsString(list));
+
+        dateType.set(DateType.PAYMENT_DATE);
+        list = widget.getUpdateTask(now).get();
+        assertEquals("2024-03-05 Security 1 EUR 0,33 Zahltag\r\n" //
+                        + "Security 2 USD 13,33 Zahltag\r\n" //
+                        + "2024-04-09 Security 2 USD 0,55 Zahltag\r\n" //
+                        + "2024-04-12 Security 1 EUR 0,55 Zahltag\r\n" //
+                        + "2024-07-15 Security 2 USD 1,22 Zahltag\r\n" //
+                        + "2024-08-11 Security 1 EUR 0,22 Zahltag", getListAsString(list));
+    }
+
+    private String getListAsString(List<DividendItem> list)
+    {
+        return list.stream() //
+                        .map(entry -> {
+                            StringBuilder sb = new StringBuilder();
+                            if (entry.hasNewDate)
+                            {
+                                sb.append(entry.type.getDate(entry.div).toString() + " ");
+                            }
+                            sb.append(entry.getSecurity().getName() + " ");
+                            sb.append(entry.div.getAmount() + " ");
+                            sb.append(entry.type);
+                            return sb.toString();
+                        }) //
+                        .collect(Collectors.joining("\r\n"));
+    }
+}
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1320,6 +1320,17 @@ public class Messages extends NLS
     public static String OptionDateIsInTheFuture;
     public static String OptionDateIsInThePast;
     public static String YearlyPerformanceHeatmapToolTip;
+
+    public static String LabelEarningsDividendList;
+    public static String LabelEarningsDividendPeriodFrom;
+    public static String LabelEarningsDividendPeriodUntil;
+    public static String LabelEarningsDividendDateType;
+    public static String ColumnDividendsNextExDate;
+    public static String ColumnDividendsNextExDate_MenuLabel;
+    public static String ColumnDividendsNextPaymentDate;
+    public static String ColumnDividendsNextPaymentDate_MenuLabel;
+    public static String ColumnDividendsNextPaymentAmount;
+    public static String ColumnDividendsNextPaymentAmount_MenuLabel;
     static
     {
         // initialize resource bundle
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
@@ -49,6 +49,7 @@ import name.abuchen.portfolio.ui.util.viewers.CopyPasteSupport;
 import name.abuchen.portfolio.ui.util.viewers.ListEditingSupport;
 import name.abuchen.portfolio.ui.util.viewers.ShowHideColumnHelper;
 import name.abuchen.portfolio.ui.views.columns.AttributeColumn;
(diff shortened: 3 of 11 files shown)


Nirus2000:
Hello @kimmerin 
Thank you for this pull request. That is definitely a great idea.

Perhaps you could format the source code before we continue working on it. That would help us a lot.
Example: 
"nexpdd" change to "nextDividendsPaymentDate"
"nexdd" ...
""nexdpa" ...
... and so on.

We should also keep the translation correct and complete it with deepl.com and correct the capitalization.
Example: 
"ColumnDividendsNextExDate_MenuLabel" = "Next Dividend Ex-Date" --> "Next dividend ex-date"
... and so on.

We should also take care to keep the translations consistent.
Example: 
"ColumnDividendsNextExDate" = "Div Ex. Tag"
"ColumnDividendsNextExDate_MenuLabel" = "Next ex-dividend date"
"Div Ex. Tag" <--> "Next ex-dividend date"
... and so on.

Example: 
"LabelEarningsDividendList" = "Übersicht anstehender Dividenden" --> "Übersicht anstehender Dividendenzahlungen"
... and so on.

But it looks good... formatting the source. 👍🏻

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,170 @@
...
+            {
+                return null; //
+            }
+            return sec.getEvents().stream() //
+                            .filter(DividendEvent.class::isInstance) //
+                            .map(e -> ((DividendEvent) e).getDate()) //
+                            .sorted(Comparable::compareTo) //
+                            .filter(d -> d.isAfter(now)) //
+                            .findFirst().orElse(null);
+        };
```
Nirus2000, line 46: Maybe this is a more elegant solution?
```suggestion
LocalDate now = LocalDate.now();

Function<Object, LocalDate> dataProvider = element -> {
    Security security = Adaptor.adapt(Security.class, element);
    return Optional.ofNullable(security)
                   .map(s -> s.getEvents().stream()
                               .filter(DividendEvent.class::isInstance)
                               .map(e -> ((DividendEvent) e).getDate())
                               .filter(d -> d.isAfter(now))
                               .min(Comparable::compareTo)
                               .orElse(null))
                   .orElse(null);
};
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,170 @@
...
+                LocalDate date = dataProvider.apply(element);
+                if (date == null)
+                {
+                    return null; //
+                }
+                if (!date.equals(LocalDate.now()))
+                {
+                    return null; //
+                }
+                return Colors.theme().greenBackground();
```
Nirus2000, line 62:
```suggestion
if (date == null || !date.equals(LocalDate.now())) {
    return null;
}

return Colors.theme().greenBackground();
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,170 @@
...
+            {
+                return null; //
+            }
+            return sec.getEvents().stream() //
+                            .filter(DividendEvent.class::isInstance) //
+                            .map(e -> ((DividendEvent) e).getPaymentDate()) //
+                            .sorted(Comparable::compareTo) //
+                            .filter(d -> d.isAfter(now)) //
+                            .findFirst().orElse(null);
+        };
```
Nirus2000, line 90:
```suggestion
LocalDate now = LocalDate.now();

Function<Object, LocalDate> dataProvider = element -> {
    Security security = Adaptor.adapt(Security.class, element);
    return Optional.ofNullable(security)
                   .map(s -> s.getEvents().stream()
                               .filter(DividendEvent.class::isInstance)
                               .map(e -> ((DividendEvent) e).getPaymentDate())
                               .filter(d -> d.isAfter(now))
                               .min(Comparable::compareTo)
                               .orElse(null))
                   .orElse(null);
};
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,170 @@
...
+                LocalDate date = dataProvider.apply(element);
+                if (date == null)
+                {
+                    return null; //
+                }
+                if (!date.equals(LocalDate.now()))
+                {
+                    return null; //
+                }
+                return Colors.theme().redBackground();
```
Nirus2000, line 106:
```suggestion
if (date == null || !date.equals(LocalDate.now())) {
    return null;
}

return Colors.theme().redBackground();
```

Nirus2000 (changes requested):
Perhaps these points can also help you?

kimmerin:
> Perhaps you could format the source code before we continue working on it.

I've followed the setup of Ecslipe for contribution according to the provided documenation and my understanding is that there is an automated source formatter. At least my formatting gets changed all the time as soon as I save ;-)

> That would help us a lot. Example: "nexpdd" change to "nextDividendsPaymentDate" "nexdd" ... ""nexdpa" ... ... and so on.

These are unique identifiers used by the SWT table to distinguish the columns from each other and are used - at least I'm expecting that - to persist the state of the table-configuration. I assumed that they aren't required to be human readable after looking at other occurreces like `column = new Column("6", Messages.ColumnShareInPercent, SWT.RIGHT, 80)`. I can change these identifiers but camelcase would be a first for all of these, though.

> We should also keep the translation correct and complete it with deepl.com and correct the capitalization. Example: "ColumnDividendsNextExDate_MenuLabel" = "Next Dividend Ex-Date" --> "Next dividend ex-date" ... and so on.

The capitalization is correct in this case, because it's an entry for the menu where other entries are in uppercase as well, e.g.

```
ColumnCapitalGainsMovingAveragePercent_MenuLabel = Capital Gains % (moving average, current holdings)
```

> We should also take care to keep the translations consistent. Example: "ColumnDividendsNextExDate" = "Div Ex. Tag" "ColumnDividendsNextExDate_MenuLabel" = "Next ex-dividend date" "Div Ex. Tag" <--> "Next ex-dividend date" ... and so on.

To be honest I don't understand your example. Can you rephrase?

> Example: "LabelEarningsDividendList" = "Übersicht anstehender Dividenden" --> "Übersicht anstehender Dividendenzahlungen" ... and so on.

I cah change `Dividenden` to `Dividendenzahlungen` but both phrases are used in the properties-file already.

> We should also take care to keep the variable names consistent. Example: "Security sec" --> "Security security" "SecurityEvent se" --> ... ... and so on.

TBH: As long as they are consistently named within a class and are memorizable within the code block they are used in, I'd rather keep the names as they currently are.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
kimmerin: "Elegant" is a quite subjective term ;-) When writing code my goal is to produce left-documenting code and - this is biased of course - but I think that my version is explaining its logic more clearly to a reader: Take all DividendEvents, sort their ex-dates in an ascending order, throw away all dates that are in the past and take the first that remains.

Your version IMHO doesn't provide that and is a bit confusing with `orElse` that is needed to be used for the stream and the Optional. So TBH: I'd prefer to keep my version.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
kimmerin: We can change that, yes. I prefer these separate checks (easier to follow for people trying to make sense of it and you can see what part isn't covered in coverage-reports) but this particular case is simple enough to be put into one if-statement.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
kimmerin: Same answer as in https://github.com/portfolio-performance/portfolio/pull/3927#discussion_r1560601973

Nirus2000:
No problem... good job 👍🏻

kimmerin:
I've added texts for the other languages but this is definitely something to be looked over by people who actually speak these languages.

kimmerin:
@Nirus2000 checked in some small improvement today. The PR is still in status "Changes requested". What requests are still open?

buchen:
> The PR is still in status "Changes requested". What requests are still open?

I think it is on me :-) Will look at it later this week

Nirus2000 (approved):

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -215,6 +215,8 @@ public enum WidgetFactory
 
     HEATMAP_EARNINGS(Messages.LabelHeatmapEarnings, Messages.LabelEarnings, EarningsHeatmapWidget::new),
 
+    EARNINGS_DIVIDEND_LIST(Messages.LabelEarningsDividendList, Messages.LabelEarnings, DividendListWidget::new),
```
buchen, line 218: The name is very generic. And "earnings" and "dividends" is kind of duplicated - dividends are earnings, but not all earnings are dividends. Why not call it UPCOMING_DIVIDENDS_LIST. And the class UpcomingDividendsListWidget?

buchen (changes requested):
First, I think that is a great new feature which is making use of existing data. Thanks.

Then about the code review: I have added a couple of smaller comments in the code. Overall it looks good to me. Thanks.

The about the feature. My general proposal is to simplify the options.

I am not sure about the "period start" and "period end" selection options in the dashboard widget. Is this selection really relevant? Do we want the user to fiddle with these selection options? My proposal is:
* show all future dividend payments (assumption: this list is limited to at most one future payment per held instrument)
* show all dividend payments of the last 3 months for which we do not find a dividend transaction in the file

I think we should **not** show in the widget dividend payments for instruments that the user does not hold. For this, the user can use the newly added columns in the security list.

Then about the layout of the widget: I propose to make it one line per instrument (security) similar to the list of earnings. The line should include payment and ex-date and the dividend payment and the (expected) total payment account. That requires understanding the number of shares held at the ex-date.

And this layout removes the "type of date" selection. Having just one line also makes the "number of payments" easier to understand. Because if one selects "both", the number does not match to the number of lines.

Then about the sizing of the widget: It took me some time to understand that it has a fixed size and that (some) entries are just missing. I would propose to do it similarly to the EarningsListWidget which just uses as much space as needed. The fixed size makes sense for charts. Thinking about it, it also does not make sense for the "follow up" and "limit exceeded" widgets (didn't think about it when developing those widgets).

Beyond this code change (in a separate pull request!), we could add new feature:
* Right click on an upcoming payment and create the actual transaction
* Allow editing of the dividend events (right now one must use DivvyDiary, but I think it is a valid use case to allow adding such events)

What do you think?

name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidgetTest.java
```diff
@@ -0,0 +1,268 @@
...
+import name.abuchen.portfolio.model.SecurityEvent;
+import name.abuchen.portfolio.model.SecurityEvent.DividendEvent;
+import name.abuchen.portfolio.money.Money;
+import name.abuchen.portfolio.ui.Messages;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateEndRange;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateStartRange;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DateType;
+import name.abuchen.portfolio.ui.views.dashboard.DividendListWidget.DividendItem;
+
+public class DividendListWidgetTest
```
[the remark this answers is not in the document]
kimmerin: > My goal is always to have no warnings about not translated labels

I've set up Eclipe the way it's documented "here" and I get these no-NLS-warnings in non-test projects but - and I've just checked it right now to make sure - in test projects (here portfolio.ui.tests) these warnings are missing.

I've assumed that this is intentional because it's a test case where leniency can often be seen (e.g. I declare "throws Exception" most of the time while I hate to see this to happen in production code).

> If you annotate the class with @SuppressWarnings("nls")

I know that annotation ;-) but haven't seen any reason to use it here. I can add it, though. If I do I just have to ignore the then-showing warning that I'm using a non-required annotation ;-)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidget.java
```diff
@@ -0,0 +1,354 @@
...
+                            Policy.EXACTLY_ONE);
+        }
+    }
+
+    /**
+     * @deprecated This constructor is only used for testing, don't use it for
+     *             anything else
+     */
+    @Deprecated
+    DividendListWidget() // needed for testing
```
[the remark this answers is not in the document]
kimmerin: Trying to add it results into an access-restriction-error due to the fact that the package hasn't been added to the list of Import-Package in the manifest-file. I can add it now as part of this change if you want.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidget.java
```diff
@@ -0,0 +1,354 @@
...
+                    }
+                    DividendItem added = checkAndAdd(items, security, de, dateType, DateType.EX_DIVIDEND_DATE, fromDate,
+                                    untilDate, null);
+                    checkAndAdd(items, security, de, dateType, DateType.PAYMENT_DATE, fromDate, untilDate, added);
+                }
+            }
+
+            Collections.sort(items, new DividendItemComparator());
+            LocalDate prevDate = null;
+            DividendItem prevItem = null;
```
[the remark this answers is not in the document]
kimmerin: Found that in the list of current warnings myself ;-) This is a remnant of another attempt to group together entries of the same date.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidget.java
```diff
@@ -0,0 +1,354 @@
...
+        @Override
+        public int compare(DividendItem di1, DividendItem di2)
+        {
+            DateType type1 = di1.getFirstType();
+            DateType type2 = di2.getFirstType();
+            int ret = type1.getDate(di1.div).compareTo(type2.getDate(di2.div));
+            if (ret != 0)
+            {
+                return ret; //
+            }
```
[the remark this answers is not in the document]
kimmerin: I know but I've got burned so many times with if-statements without braces. Google "apple goto fail" and you'll see another spectacular example of things that got burnt because of this.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
kimmerin: When reading the code I had the impression that "EARNINGS" was use as a kind of "grouping mechanism" in order to see widgets that belong together being sorted together in the Outline. I can rename the enum-name if really needed but I'd keep the class name as it is.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
```diff
@@ -1320,6 +1320,17 @@ public class Messages extends NLS
...
+    public static String LabelEarningsDividendList;
+    public static String LabelEarningsDividendPeriodFrom;
+    public static String LabelEarningsDividendPeriodUntil;
+    public static String LabelEarningsDividendDateType;
+    public static String ColumnDividendsNextExDate;
+    public static String ColumnDividendsNextExDate_MenuLabel;
+    public static String ColumnDividendsNextPaymentDate;
+    public static String ColumnDividendsNextPaymentDate_MenuLabel;
+    public static String ColumnDividendsNextPaymentAmount;
+    public static String ColumnDividendsNextPaymentAmount_MenuLabel;
```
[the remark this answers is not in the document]
kimmerin: I've used ResourceBundle Editor as it supposedly was used by others as well (see e.g. `messages_fr.properties`)

kimmerin:
> The about the feature. My general proposal is to simplify the options.
> 
> I am not sure about the "period start" and "period end" selection options in the dashboard widget. Is this selection really relevant?

I think yes. When creating the widget I've imagined questions that users might want to get answered by this:

- When are the next payments to be expected?
- What payments have happened so far
- When do I have to buy additional stocks at the latest to still get the dividend

The options I've provided allow the configuration of individual widgets to answer these questions. There should be a useful default (at the moment all types of future dates till the end of the year), so the majority of users don't need to "fiddle" but for these more specific questions, they are necessary.

> I think we should **not** show in the widget dividend payments for instruments that the user does not hold.

That wasn't intended, so if it currently does I need to change this (after writing a test case for it of course ;-)

> Then about the layout of the widget: I propose to make it one line per instrument (security) similar to the list of earnings. The line should include payment and ex-date and the dividend payment and the (expected) total payment account. That requires understanding the number of shares held at the ex-date.

To be honest, I see usage for both types of widgets, so your proposed widget should be an additional one rather than "my" widget being changed to this new layout. I can do that as part of this PR but I think "out there" people are craving for this already, especially with the already started dividend season, so I think a new PR for the new widget should be in order as well.

> Then about the sizing of the widget: It took me some time to understand that it has a fixed size and that (some) entries are just missing. I would propose to do it similarly to the EarningsListWidget which just uses as much space as needed.

I've derived my widget from your abstract class so it's all your fault!!1! ;-) I can change it to behave like EarningsListWidget. Should I change "follow up" and "limit exceeded" as well while on it?

>     * Right click on an upcoming payment and create the actual transaction

That might look nice in the first place but you'd still end up using the PDF-import, because that event doesn't contain any information about taxes, fees, conversion rates, etc. that you'd need to provide manually in that case. And that's just the simple stuff, it gets more interesting with stock being held in multiple depots.

>     * Allow editing of the dividend events (right now one must use DivvyDiary, but I think it is a valid use case to allow adding such events)

If we add this, the DividendEvent should be extended to keep the record date as well (which is often different from the ex-date)

> What do you think?

I think you know now ;-)

buchen:
>> I am not sure about the "period start" and "period end" selection options in the dashboard widget. Is this selection really relevant?
> I think yes. When creating the widget I've imagined questions that users might want to get answered by this:
> 
> When are the next payments to be expected?
> What payments have happened so far
> When do I have to buy additional stocks at the latest to still get the dividend

The first and the third question can be answered with the widget having the heuristic that I proposed (show all in the future + last 3 months that are not found).

The second question is a different "thing". I believe the user is not interested in what dividend payments happened  in principle, but which payments actually happened. Thinking about it, I wonder if this really is a separate new widget or whether we should extend the earnings list widget with (optionally) future payments.

kimmerin:
> I believe the user is not interested in what dividend payments happened in principle, but which payments actually happened.

Well... as a user I can tell you that I'm interested into both ;-) I'd like to know when to expect sudden drops of prices (ex date) and when to expect money on my account (payment date). For the former the dividend amount for a single share is relevant because that's the expected drop in price. For the latter, both amounts (per share and all) might be of interest dependent on the context of the widget's usage.

I agree that for the information about the expected dividend amounts it makes sense to extend the existing widget that shows the earnings of a particular year to show an additional sum per month and entries for expected dividends, i.e. dividends that you owned - in lack of a record date - one day before the ex-dividend date. This should be done in another PR, together with a couple of UX-improvements (I'm not going into detail here, because that would be off topic).

TL;DR: I still see use for this widget, so I'm going to add the account filter like in the transaction-widget. Concerning the time range: What about removing the range for the future, i.e. all future dividend-events are shown and keeping the "from-range" but with the possible values "None", "1 Week", "1 Month", "1 Quarter", to allow people to see dates in the past that they otherwise missed because they don't fire up the app on a daily basis?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DividendListWidget.java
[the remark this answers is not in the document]
kimmerin: I've added the entry in the Manifest-file

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
kimmerin: I've renamed the enum. Better now?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,165 @@
...
+            Money m1 = getAmount(LocalDate.now(), (Security) o1);
+            Money m2 = getAmount(LocalDate.now(), (Security) o2);
+
+            if (m1 == null)
+                return m2 == null ? 0 : -1;
+            if (m2 == null)
+                return 1;
+
+            return m1.compareTo(m2);
+        }));
```
buchen, line 144: We can use functions of the column viewer sorter. We must also adapt the object as the row objects is passed here as well.

```java
        column.setSorter(ColumnViewerSorter.create(element -> {
            Security security = Adaptor.adapt(Security.class, element);
            return security != null ? getAmount(now, security) : null;
        }));
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
```diff
@@ -259,6 +260,7 @@ public String getText(Object element)
         portfolioColumns.addColumn(column);
 
         addAttributeColumns(portfolioColumns);
+        addDividendPaymentColumns(portfolioColumns);
```
buchen, line 263: This is on the level of a "portfolio".

I propose *not* to add this. First, the implementation always returns null at the moment. Second, I think the context is missing - which security inside the portfolio has the next dividend date?

buchen (commented):
I started to "divide and conquer" this pull request. I now cherry-picked the new dividend columns as they are fairly straightforward and separate from the widget. The code comments are apply only to the new columns. I have rebased and merged this functionality.

One more comment: it helps me reviewing/handling the pull requests if you *rebase* the pull request instead of *merging* changes from the master into the pull request. Rebasing means you review your changes against the master - that is easier than the other way around.

For my next block of uninterrupted time, I'll look at the widget.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,165 @@
...
+import name.abuchen.portfolio.ui.Messages;
+import name.abuchen.portfolio.ui.util.Colors;
+import name.abuchen.portfolio.ui.util.viewers.Column;
+import name.abuchen.portfolio.ui.util.viewers.ColumnViewerSorter;
+import name.abuchen.portfolio.ui.util.viewers.DateLabelProvider;
+
+public class DividendPaymentColumn
+{
+
+    static Column createNextDividendExDateColumn(Client client)
```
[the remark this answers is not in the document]
kimmerin: I've removed it. In an earlier iteration this method needed it.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
```diff
@@ -0,0 +1,165 @@
...
+                }
+
+                return Colors.theme().redBackground();
+            }
+        });
+        column.setSorter(ColumnViewerSorter.create(dataProvider::apply));
+        return column;
+    }
+
+    static Column createNextDividendPaymentDateColumn(Client client)
```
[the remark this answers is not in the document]
kimmerin: See previous comment

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
kimmerin: I agree. Trying to remember what has led me to do this in the first place I can only think of "the end of a long day".

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/columns/DividendPaymentColumn.java
kimmerin: I've changed it accordingly

buchen:
@kimmerin 

> Should I resolve these conflicts or is this going to be done by you anyways when deciding on the non-merged part of this PR?

I already have a local change that works on top of the master branch, so there is no need for you do push changes here. Unfortunately, I am quite busy these days. And if I only have a couple minutes here and an hour there, it is sometimes hard to tackle the big rocks. And then I review and merge smaller changes because those I can intellectually process in that time. After this weekend, I hope it will become better.

Sorry for the inconvenience and thanks for your patience.

kimmerin:
@buchen No problem. I just wanted to make sure that this PR isn't in limbo because everybody is waiting for the others to do something ;-)

stoeggich:
sorry for asking but what happens next? will the merg come?

kimmerin:
@stoeggich I'm curious myself but it's summer vacation time here in Germany, so things can slow down because of this.

kimmerin:
Hi @buchen, seven months into this ;-), is there a chance this PR is going to be integrated fully or is there something you need from me in order to do so, that I'm not aware off, leading to this stand-still?

kimmerin:
Hi @abuchen, after nearly a year, I assume this PR is never going to be merged. May I ask what's reason behind it? Because it seemed to me that there's demand out there to have an overview in the dashboard rather than just the columns.

stoeggich:
@kimmerin you wanted to write @buchen

kimmerin:
> @kimmerin you wanted to write @buchen

Sorry, yes. So here again ;-)

Hi @buchen, after nearly a year, I assume this PR is never going to be merged. May I ask what's reason behind it? Because it seemed to me that there's demand out there to have an overview in the dashboard rather than just the columns.

pfalcon:
@kimmerin: Since this patch, there were other developments for PP, e.g. a generic event list widget for dashboard: https://github.com/portfolio-performance/portfolio/pull/4431 . That widget can show **any** type of event, not just dividend events (a culprit is that somebody now needs to add more types of events). Likewise, there's already a type of column which can be used to track date of *any* event - that's custom attribute of "Date" time, which can be shown as "Date - days between" column, which can be sorted upon and allow users to pretty quickly see what're upcoming (or recently) passed events are.

Of course, it requires some legwork from users to set up and tie it all together, but then it allows to track any kind of event, be it dividend payments or ex-dates, earnings date, CEO birthday, or whatever. And it of course would be nice to have automagic support for dividends, but is it really easy to tie it all together and will it really work the way different people expect?

kimmerin:
> Since this patch, there were other developments for PP, e.g. a generic event list widget for dashboard: #4431 . That widget can show **any** type of event, not just dividend events (a culprit is that somebody now needs to add more types of events).

OK, I've tried that in the dashboard. Since the original issue is closed, here is some feedback: The widget itself is way to greedy in terms of real estate. The type of event is selected and fixed, i.e. it doesn't need to be repeated for each entry next to the date. If multiple entries share the same date, they should be grouped together.

Concerning the ex-dividend date: Currently they are missing completely as selectable event-type simply because the event-type "dividend" contains two dates and this generic widget only supports to be configured to show one of them. You aren't able to configure the widget to show dates of two types of events (e.g. ex-date and payment-date) in one widget, so you have to - again - use up precious real estate on your dashboard to see both dates.

So while this issue allows some solution to see upcoming dividend-dates I still see some use in the widget I've created because it's optimized to show these informations in a more optimized way concerning required space.

But it's not up to me and if your widget is considered sufficient for the people who voiced a demand for this kind of thing, @buchen should just close this issue for good so people don't wonder why it's in limbo for so long.

kimmerin:
> You can see if there's any similarity with your patch or no.

There is no real similarity in my eyes but YMMV.

> There's this patch, https://github.com/portfolio-performance/portfolio/pull/3927, which claims that it solves all things dividends, but doesn't show much, so one can only guess what exactly it does.

You're aware under what PR you've responded? ;-) If you want to see a sceenshot: https://forum.portfolio-performance.info/t/widget-gesucht-bekannte-kunftige-dividendenzahlungen/28020/4

> There's currently no "dividend" event type, currently there's only adhoc "divvydiary.com" proprietary event type. Which is useless for anyone who doesn't use divvydiary.com.

It's an event type that is currently only been populated with data by divvydiary. If tomorrow there were another source the created event type should be the same, so that widgets/columns/... don't need to be changed in order to work as they did before. Separating ex-date and payment date to two different event-types I don't see as practical because as a user I'd like to see both dates at the same time and you will create confusion with securities with regular dividends, where the payment of the last dividend happens after the ex-date of the next one (Prospect Capital as an example).

> It's absolutely not sufficient, doesn't claim to be such, and intended for long and progressive elaboration by different people. I also wish that @buchen gave specific instructions what to do to get patches merged as fast as possible, but so far I didn't see that. Whereas I had to figure out answers to such questions, both for my own patches and others', and just share my thoughts with a fellow contributor.

Let me rephrase then: If the new feature covers the features of my PR here, then - by all means - close this one because it's no longer requred and that way it's clearly documented for everybody.

derelvis:
I would really appreciate if we could get some improvements out of this thread/pr to show some more/other information and grouping.
Especially the missing payment date makes it useless for my own use case. 
Personally I would love to show it just like in the DivviDiary email "Deine Dividenden Zahltage für heute":

!!Hauptfenster 2026-02-06 09-25-21

Thank you so much everybody for your time and effort and letting us all participate <3

buchen:
The feature is now merged. Finally.

Some parts of this pull request were cherry-picked already some time back. I now salvaged the widget itself and refactored it to the latest code base, added translation labels, updated the use of colors, added an option to separate between all instruments, active instruments, and held instruments (which then can be filtered by securities account if needed).

PR #3941, opened 2024-04-19
escape ampersands in label texts that would otherwise be interpreted as special character

kimmerin:
Hi,

as shown in https://forum.portfolio-performance.info/t/widget-gesucht-bekannte-kunftige-dividendenzahlungen/28020/4 security names containing an ampersand show up in a wrong way if they are put into a label in a dashboard widget. This PR fixes that for exceeded limits and exceeded dates.

I've added two helper methods in AbstractSecurityListWidget that is also present in my other PR #3927.


buchen:
Thanks for fixing those. There is already a helper method used in the other placed: (not very helpfully named) ```TextUtil.tooltip```

kimmerin:
OK, I've changed the method in the abstract class to reuse code by calling the helper method

PR #3951, opened 2024-04-23
Prevent download of dividend data from Divvy Diary if security is inactive

kimmerin:
Hi,

while working on the dividend widget I've seen that there is no test for the parsing of the JSON data. So I've added one. While doing this I've realized that currently all securites are requested, even the ones that are configured ot be inactive. Prices aren't downloaded in that case, so dividend-informations shouldn't, either IMHO. As well I've changed the time where the result is cached after the parsing happened to prevent invalid data (e.g. some non-JSON result by a CDN with hickups) to end up in the cache.


buchen:
> Prices aren't downloaded in that case, so dividend-informations shouldn't

Actually historical prices are not downloaded if the configuration say "no download". That is independent from whether the instrument is "active" or "inactive". One can have an inactive instrument and yet have it updated the prices, i.e. it is not proposed in selection dialogs but updated.

I am aware that this sometimes confuses users. There are arguments for both ways - attaching the download to the "active" flag and attaching it to the "download config". (But the biggest argument is that I do not want to change the behavior for thousands of users now.)

Being the way it is, I wonder if the DivvyDiary download should be attached to the same logic - if the download of historical prices is configured, then also download dividend data. If not, then not.

That would avoid introducing a new logic: prices are downloaded based on the configuration, dividends based on the inactive flag.

Plus: I do not really understand what the issue is with loading dividend payments for all securities? Is it the time it takes? What is the problem that users have?

@kimmerin @Pnda87 @stoeggich What do you think?


(I leave the pull request open for the time being, but we would need a separate pull request)

---


Then about the code. I picked up the other change, i.e. not caching the response if it is an invalid JSON.  I have marked the methods with the ```@VisibileForTesting``` annotation. I removed the check for the getter/setter for the API key. *In general, we should be very, very cautious making methods visible just for testing purposes.* It typically exposes implementation details which then require us to change the test if the implementation changes - and then the test has reduced value. Let's avoid this in the future.

stoeggich:
I think the same logic should be used for updating.

It might make sense if you could mark stocks as dividend stocks and ignore the rest.

Either way, perhaps a test as an experimental function would be good.

buchen:
> I think the same logic should be used for updating.

Let me also give you some more background why it is implemented the way it is implemented. Originally, there was no currency. But then all indeces like Dax, Nasdaq, Dow Jones etc. would show up in the dialog for purchasing. one could also use the flag "inactive" to reduce the items in the dialog while at the same time still update the quotes (in order to use them as benchmarks).

Of course, now it is possible to setup this instrument has having no currency. But there can be (a lot) of users out there which still have this setup.

I like the idea about rolling this out via the instrumental flag!

kimmerin:
> Plus: I do not really understand what the issue is with loading dividend payments for all securities? Is it the time it takes? What is the problem that users have?

It's one request per security, so the more securities there are the higher the load on the server side. That might lead to the introduction of limitiations on that side if not enough PP-users subscribe to their monthly plan to finance the increased bill for the required bandwith. Loading data for all securities, even the ones you have forgotten to even exist in your data because yoiu've filtered them out in your views is like the freeloaders at buffets fillings up their plate and then fill up a second plate, just because they can.

kimmerin:
> In general, we should be very, very cautious making methods visible just for testing purposes.

It's the way Java works. To allow a test to be done you need some way access to the method/member and package visible is the one I've used as a compromise to limit the possible "users" to a minimum while having the ability to use them in a test to check for correct functionality. C++ knows "friend" classes that can access otherwise restricted methods but until Java gets something comparable (I'm not holding my breath for such a thing to happen) package visible methods are the best solution.

I suppose there is a reason why Guava introduced `@VisibileForTesting` and configure the use of an annotated method/member outside test-projects as an error would ensure that these methods/memers' access "leaks" into production code.

> It typically exposes implementation details which then require us to change the test if the implementation changes - and then the test has reduced value.

FTR: This assumtion is only valid for implementations that haven't unit-testability in mind and BTW: Unit tests are essentially supposed to fail if something changes in the CuT, integration tests on the other hand shouldn't.

> Let's avoid this in the future.

I will try in the future but TBH as I'm doing test driven development IRL it might occasionally happen again.

kimmerin:
> Being the way it is, I wonder if the DivvyDiary download should be attached to the same logic - if the download of historical prices is configured, then also download dividend data. If not, then not.
>
> That would avoid introducing a new logic: prices are downloaded based on the configuration, dividends based on the inactive flag.

Just in case my previous answer doesn't count as such: I thought that `isRetired` _is_ the configuration that controls if historical data is downloaded, so: I don't see a need for an additional configuration to control if dividend-data should be downloaded or not but it's using the configuration for the download of historic prices. If my implementation is using `isRetired`wrong in this context, can you point me to a class where I can see the correct way?

Pnda87:
> One can have an inactive instrument and yet have it updated the prices, i.e. it is not proposed in selection dialogs but updated.
> 
> I am aware that this sometimes confuses users. There are arguments for both ways - attaching the download to the "active" flag and attaching it to the "download config". (But the biggest argument is that I do not want to change the behavior for thousands of users now.)

In my opinion, this behavior is very unintuitive. As a user, I would expect that inactive securities are really inactive and thus no longer updated. Neither prices nor events from DivvyDiary.

What is the use case for a price download of a security that I do not want to actively use?

Is the statement that thousands of users use the workaround for benchmarks somehow verifiable? Personally, I don't agree that a large number of users would have a problem with a (reasonable) change.

buchen:
> In my opinion, this behavior is very unintuitive. As a user, I would expect that inactive securities are really inactive and thus no longer updated. Neither prices nor events from DivvyDiary.

Fair. I tend to agree.

My idea is:
* Use the isRetired flag to determine if they are in the drop down box and if instruments are updated (both prices and divvy diary)
* Check if users have inactive instruments that have an active download configuration and propose to activate them

I will close this pull request now (as the main change is merged already)

PR #3969, opened 2024-04-29
Securities Account chart with Invested Capital, Delta

mierin12:
Hello,

**Proposition**:
In the same way than Deposit Accounts have a pane with an Account Balance Chart, this PR is proposing to add a Securities Account Balance Chart (first commit).
Actually, the real goal is also to associate to those Portfolio Charts the Invested Capital and Delta metrics (+accumulated Taxes and Fees and more if relevant) that a lot of users are looking for (second commit).

It is limited in scope compared to https://github.com/portfolio-performance/portfolio/pull/3754 which should be the general correct solution for those common metrics, but I think here the implementation is avoiding the "multiselection" and UI questions of https://github.com/portfolio-performance/portfolio/pull/3754. 
The metrics are here associated to a Portfolio only (no association to Taxonomy, Custom Filter Portfolio+Cash account etc) but I think this is nevertheless already providing some useful data.

**Implementation** : similar to the `AccountBalancePane`,  `AccountBalanceChart` and to the `SecuritiesChart`.
Caveat: by using a similar structure than `SecuritiesChart`, the plotted data series are **not** customisable (color, area vs line, linestyle etc). 
I mostly used the color defined in `DataSeriesSet` but "less gray" colors could be used.

**Results** : 
Example from kommer file:
Capture d'écran 2024-04-30 002018
Capture d'écran 2024-04-30 002116
custom account
Capture d'écran 2024-04-30 011020

**Bug** : this PR is in Draft as there is at least one bug that I did not manage to fix. When the graph pane is selected from the Statement of Assets or Transactions panes of a Securities Account, the chart is somehow too big for the plotted data :
Capture d'écran 2024-04-30 002145
It updates correctly when changing the metric in the chart menu or when switching between Securities Account.
If someones has an idea on this bug or a feedback, it is much welcome. 
 
(Other ideas in case of next steps:  in the same way, Earnings/Interests could be added to the `AccountBalanceChart`, Delta could be a bicolor positive/negative area as is in the `SecuritiesChart`, "account events" may be added in those charts.)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioBalanceChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioBalanceChart.java
@@ -1,0 +1,90 @@
+package name.abuchen.portfolio.ui.views;
+
+import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+import java.util.Collections;
+import java.util.List;
+
+import org.eclipse.swt.widgets.Composite;
+import org.swtchart.ISeries;
+
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.Portfolio;
+import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.model.Transaction;
+import name.abuchen.portfolio.money.CurrencyConverter;
+import name.abuchen.portfolio.money.CurrencyConverterImpl;
+import name.abuchen.portfolio.money.ExchangeRateProviderFactory;
+import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.snapshot.PortfolioSnapshot;
+import name.abuchen.portfolio.ui.util.Colors;
+import name.abuchen.portfolio.ui.util.chart.TimelineChart;
+
+public class PortfolioBalanceChart extends TimelineChart // NOSONAR
+{
+    private Client client;
+
+    public PortfolioBalanceChart(Composite parent, Client client)
+    {
+        super(parent);
+        this.client = client;
+        getTitle().setVisible(false);
+    }
+
+    public void updateChart(Portfolio portfolio, ExchangeRateProviderFactory exchangeRateProviderFactory)
+    {
+        try
+        {
+            suspendUpdate(true);
+
+            for (ISeries s : getSeriesSet().getSeries())
+                getSeriesSet().deleteSeries(s.getId());
+
+            if (portfolio == null)
+                return;
+
+            List<PortfolioTransaction> tx = portfolio.getTransactions();
+
+            if (tx.isEmpty())
+                return;
+
+            Collections.sort(tx, Transaction.BY_DATE);
+
+            LocalDate now = LocalDate.now();
+            LocalDate start = tx.get(0).getDateTime().toLocalDate();
+            LocalDate end = tx.get(tx.size() - 1).getDateTime().toLocalDate();
+
+            CurrencyConverter converter = new CurrencyConverterImpl(exchangeRateProviderFactory,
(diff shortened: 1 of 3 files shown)


buchen:
> this PR is in Draft as there is at least one bug that I did not manage to fix. When the graph pane is selected from the Statement of Assets or Transactions panes of a Securities Account, the chart is somehow too big for the plotted data

That is indeed super strange. I was able to debug it to these lines below:

If the pane has buttons (and the chart pane has buttons as opposed to the account balance pane), then somehow the width of the chart element passed on to the adjust range method is way too small. And then adding a couple percentages margin makes the chart show up in the center. I fixed that by reversing the order - first showing the chart, then updating the toolbar. Not intuitive.

https://github.com/portfolio-performance/portfolio/blob/11be3facf1f82e19a29fe02ccbdf4638b5fd34a9/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/InformationPane.java#L187-L200


Apart from this, I'll like your proposal.

However, right now the calculation of the data series happens in the UI thread. For the account balance, this is not so dramatic because that it is fairly cheap. For the portfolio balance, this can become a problem because that can be an expensive calculation. Let me change that (possibly for both)

mierin12:
Ah great, well done for the fixing the bug !
If this is non UI calculation is also updated for AccountsBalance chart, then this means we could enhance it too with Interest metric for example !

One point to confirm maybe : the currency of the charts. I used `client.getBaseCurrency())` but I hesitated with `portfolio.getReferenceAccount().getCurrencyCode()`. I do not own myself multicurrency accounts or stocks, so I am rather short-sighted on this. The most important is to have all data serie/metric of a same graph at the same currency of course, I think this is ok on the PR, but which currency is the most relevant one if a portfolio is linked to a dollar cash ref ? If the portfolio has stocks in several currencies ? This I do not know.

buchen:
Hey @mierin12,

I have made a couple changes - not it is rebased and merged.
* it is not calculated in the background
* it is reusing the computed PerformanceIndex
* I extracted the toDouble methods into a helper class

Nice change.

buchen:
> the currency of the charts. I used ```client.getBaseCurrency()``` but I hesitated with ```portfolio.getReferenceAccount().getCurrencyCode()```.

I agree to using the base currency. I think all calculations should be in the "reporting currency" (= base currency). If the user wants to see it differently, it is easy to change via the menu or from the statement of assets view.

I think the currency of the reference account does matter here. Personally, I am thinking about getting rid of this reference account. It is used to collect dividend statements for the instruments in the brokerage account. But there could be better ways to do this (for example, right now it is possible to change the reference account, and then the link of the old payments is lost. Or what if payments are done in multiple currencies? Then only some are picked up). Therefore I try to avoid adding even  more dependency on the reference account.

For cash accounts it is different. There you want to see the balance in the currency of the account.

buchen:
> (-> no side effect on SecurityChart as SecurityPriceChartPane is also using InformationPanePage  and Buttons (but without the "bug"..) ?)

Intersting point. I do not really know. Maybe because initially no security is selected (but an account is selected). I do not want to spend the time debugging now - with the updated code, the security price chart pane continues to work as expected.

mierin12:
Hello @buchen , many thanks, I hope the feature will be liked. I will have a look on your update to better understand the 'calculated in the background' and PerformanceIndex advices to be able to do better next time.

One comment, are the transactions not sorted anymore ? I remember I add to specifically put the sort before calculating the start date from the tx list, because I had a case with a Account with a Transfert of security which somehow was taking the first place of the transaction list while not being the oldest transaction of the portfolio. So start date was at first this Transfert date and not the correct start date. In AccountBalanceChart the sort is after start date but the problem seems to arise only for Transfert of security so it was not impactful. I do not have my computer with me at the moment to test..

Issue #3972, opened 2024-05-01
Y axis does not automatically correctly reset after "save diagram"


mierin12:
**Describe the bug**
Small bug
When saving a chart, the Y-axis is not reset properly. For all type of charts.
In addition, in the Return/Volatility Chart, the X axis disappears and the X axis title is appearing twice.

**To Reproduce**
1. Go to Return/Volatility chart
2. Click on 'Export Data' (or right click on the chart)
3. Click on 'Save diagram' and save it

**Expected behavior**
After a save to png/jpeg, the chart should automatically get back to its prior save state.
(The bug can be quickly manually fixed by simply clicking on the chart, the apparition of the tooltip corrects the chart).

**Screenshots**
Before save:
![Capture d'écran 2024-05-01 120918](https://github.com/portfolio-performance/portfolio/assets/160436107/ae7142b2-ca92-447f-91fc-cd9593ca69f9)
After save:
![Capture d'écran 2024-05-01 121917](https://github.com/portfolio-performance/portfolio/assets/160436107/9350d86f-40ce-47c7-ae6a-c029a47c36f6)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: 0.68.4

PR #3973, opened 2024-05-01
clean some save chart behavior

mierin12:
First commit : add missing title to chart 

Even if charts titles are hidden, they appear when saving the charts.
For Deposit account balance chart and Taxonomy stacked charts, no title was provided. 
So, when exporting those charts, a default "Chart Title" was used as title and default filename, while for all other charts, a correct title is provided.

Second commit : 

Closes https://github.com/portfolio-performance/portfolio/issues/3972 
Fix the chart appearance just after saving it. fix looks ok on my Windows.

(should this have been two separate PR ?)


buchen:
> should this have been two separate PR

It makes it easier for me if each thing is a separate PR.
In this case, it is not a problem because the change is so tiny.
But if you can separate them, I would appreciate it.

PR #3977, opened 2024-05-02
PerformanceView: show a disabled version of a security's logo if there are no holdings to prevent user confusion

kimmerin:
When going to the performance view (Performance->Calculation) entries of securities that have a quantity of zero are currently presented with the generic logo of an inactive entry even if there is a configured logo. That confuses me but I'm not [alone](https://forum.portfolio-performance.info/t/logo-fehlt-in-performanceberechnung-bei-bestand-0/10321).

I've extended the ImageManager to provide disabled versions of an image and use that in PerformanceView to keep the actual logo showing up while it still allows to distinguish between "has holdings" and "has no holdings".

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
@@ -30,7 +30,12 @@ public final class LogoManager
 
     public Image getDefaultColumnImage(Object object, ClientSettings settings)
     {
-        Image logo = getLogoImage(object, settings);
+        return getDefaultColumnImage(object, settings, false);
+    }
+
+    public Image getDefaultColumnImage(Object object, ClientSettings settings, boolean disabled)
+    {
+        Image logo = getLogoImage(object, settings, disabled);
         return logo != null ? logo : getFallbackColumnImage(object);
     }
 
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
```diff
@@ -30,7 +30,12 @@ private LogoManager()
...
     public Image getDefaultColumnImage(Object object, ClientSettings settings)
     {
-        Image logo = getLogoImage(object, settings);
+        return getDefaultColumnImage(object, settings, false);
+    }
+
+    public Image getDefaultColumnImage(Object object, ClientSettings settings, boolean disabled)
+    {
+        Image logo = getLogoImage(object, settings, disabled);
         return logo != null ? logo : getFallbackColumnImage(object);
```
buchen, line 39: The issue is now: if the instrument has no logo (and hence the fallback image is used) then it is now the default image (either the active or the retired image). Basically, the "disabled" flag should also be applied to the fallback image (and probably the disabled fallback image should be cached somehow).

buchen (changes requested):
Hi @kimmerin,

great catch.

I like it for two reasons:
* First, it fixes a confusing thing: the logo is suddenly gone
* Second, it also makes the disabled image look different from a retired image. I think it was the wrong decision 8 years ago to reuse the retired image here.

For the change - see my comment that the disabled look is now not applied to instruments that have the default logo

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
kimmerin: Done a change and checked it in. Should now work as intended.

buchen:
merged. Thanks.

PR #3978, opened 2024-05-02
Fix Dashboard Widget often Hidden by Vertical Scrollbar

### LINKED ISSUE #3256: Dashboard Widget often Hidden by Vertical Scrollbar
**Describe the bug**

Vertical scrollbar can hide right side of dashboard widget.

**To Reproduce**

Drag window height up and down. Varies by precise width of window.

**Expected behavior**

Widget should not be displayed behind scrollbar

**Screenshots**

Display image in new full-screen tab.

PP-WrongMargin

Version: 0.62.0 (Apr 2023)
Platform: win32, x86_64

Mikerion:
Closes #3256

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DashboardLayout.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DashboardLayout.java
@@ -49,7 +49,7 @@ public class DashboardLayout extends Layout
 
         Rectangle availableBounds = composite.getBounds();
 
-        int widthPerWeight = (availableBounds.width - (children.length * SPACING)) / total;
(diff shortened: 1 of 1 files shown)


buchen:
Hi @Mikerion,

actually, looking at the layout, it makes the columns to small because it is subtracting on SPACING too much.

Now, we can throw in another 10 pixels, but I am not sure how this fixes the problem described in #3256.

There is code that - on Windows - adds additional space to make room for the vertical scroll bar. I think what we need to investigate is why that code does not work (or does not in all cases).

https://github.com/portfolio-performance/portfolio/blob/3ec2c690676d0d297e8dbab1062635e4a17721d2/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/DashboardView.java#L498-L504

PR #3983, opened 2024-05-03
Modify Directa SIM PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-from-directa-sim/27285/2


Nirus2000:
This test case has already been implemented. Closed without merge

Issue #2050, opened 2021-02-06
Request: Ability to search transactions within a single account


metal450:
If you go to Deposit Accounts->Transactions, you can see all transactions for just one account - but there is no search box.

If you go to All Transactions, there's a search box, but you cannot filter by account.

It would be nice to have either one of these, or the other, so that transactions could be searched, while limiting to just one account :)

PR #3985, opened 2024-05-03
Add Transaction Filters and Search box for Transactions Pane and Account Transactions Pane

Mikerion:
Closes https://github.com/portfolio-performance/portfolio/issues/2050


buchen:
Thanks @Mikerion. Nice addition. I have refactored the changes to extract the common code into a separate package.

mierin12:
hello @Mikerion  @buchen : I believe there is small bug : the filter icon disappears (but filter is still active) when switching tab, example on a Deposit Accounts : Transactions -> Account Balance Chart -> Transactions : logo filter has disappeared.
It then reappears if you leave Deposit Accounts then goes back. But it gives the feeling that transactions are missing since the filter is still active but not visible. Issue also present for Securities Accounts

PR #3987, opened 2024-05-03
Fix Trade Republic PDF-Importer to switch Save Back from Buy to delivery inbound

Nirus2000:
Fix Trade Republic PDF-Importer to switch Save Back from Buy to delivery inbound

https://github.com/portfolio-performance/portfolio/pull/3874#issuecomment-2093452630

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
@@ -581,8 +581,8 @@ public class TradeRepublicPDFExtractorTest
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
-        assertThat(countBuySell(results), is(1L));
-        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countBuySell(results), is(0L));
+        assertThat(countAccountTransactions(results), is(1L));
         assertThat(results.size(), is(2));
         new AssertImportActions().check(results, CurrencyUnit.EUR);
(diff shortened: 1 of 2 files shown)


christen90:
Hi Alex,
I think further adaptation is needed.

If we import Saveback as delivery inbound we should skip (or marks as NonImportable - already covered with other document) the saveback line 20-21.

Please find the PDF dump attached.
[Kontoauszug09.txt](https://github.com/portfolio-performance/portfolio/files/15219891/Kontoauszug09.txt)

I would do it myself, but I'm not at home this week and have no access to my eclipse pc. :-)

BR Christian

christen90:
Actually the saveback transaction needs to be imported anyway to get the details. I would prefer the inbound delivery and skipping the compensation payment.
Saveback document is generated on start of month, the correction at the end of month. So the users would be more up to date with using the inbound delivery and no buy transaction.
Does it make a difference at the end? Just one transaction in addition but no different result of the calculations. Right?

Nirus2000:
That's another fundamental discussion. The transactions exist on paper, so they are read in one-to-one, just like in all the other importers. 
It is then up to the user to decide whether to accept them or not. For this purpose, we have created the option of not importing individual transactions or converting them into deliveries (manually).

In the [comment](https://github.com/portfolio-performance/portfolio/pull/3874#issuecomment-2093452630) you wrote that there were no costs on a contra account and now they do exist. That confuses me a little.

I will withdraw the PR and integrate this PDF import in the account statement.

christen90:
> In the [comment](https://github.com/portfolio-performance/portfolio/pull/3874#issuecomment-2093452630) you wrote that there were no costs on a contra account and now they do exist. That confuses me a little.

Sorry I should have triggered the export before raising the question. The corresponding transactions take place at the same day. It didn't felt like costs.

Probably you are right and your approach is more robust.

Issue #3286, opened 2023-04-16
Generalize investment plans


funnym0nk3y:
**Is your feature request related to a problem? Please describe.**
I'm with a broker that offers a flatrate trading fee. IMHO this should be included in the fee calculation, but there is no possibility to create a fee transaction regularly. 

**Describe the solution you'd like**
Every now and the there are feature requests like #2563 or #2505 to add specific transactions to the investment plan functionality. To help with that I propose a generalization of the investment plan feature that allows for every possible manually created transaction to be scheduled. In my opinion there is no difference between the currently schedulable transactions and those which aren't, so why not allow them all to be scheduled? 

**Describe alternatives you've considered**
When a new transaction type is introduced there will be a request to make them schedulable at some point. Adding each transaction per request requires the devs to visit this feature over and over again.

PR #3995, opened 2024-05-05
Add Interest to Investment plan

Mikerion:
Related to #3286


name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
```diff
@@ -43,6 +43,9 @@ public enum Type
 
     private long amount;
     private long fees;
+    private long taxes;
+
+    private Type type;
```
buchen, line 48: I totally agree that we need to introduce a ```type``` property and stop deriving the plan type from the properties.

However, to complete this
* we need to update the version of [Client](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Client.java#L32) file because older version of PP will not handle the new properties (taxes and type)
* add a migration to the [ClientFactory](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java#L838-L841)
* persists the type also in the binary file format ([ProtobufWriter](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ProtobufWriter.java#L724)) - the XML format will include this automatically due to the reflection used by XStream

My proposal is to migrate the investment plans to always have a ```type```. In the current code, older plans do not have a type; only those that have been edited or newly created. This is very error prone when we change code in the future. I propose to have that "guessing logic" in the ClientFactory and from then on we can assume that type is a non-null property.

buchen (changes requested):

Mikerion:
Hello @buchen ,
I don't know why CI failed, but all things are commited, even `protos` `PInvestmentPlan` changes.

buchen:
@Mikerion I think you forgot to commit the [client.proto](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/client.proto) file. Then the build is re-generating the old code - and that of course fails due to missing methods.

Mikerion:
Ah, I'm not familiar with this so I wrote it manually. I'll rework it. Thanks.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/InvestmentPlanTest.java
```diff
@@ -47,6 +47,9 @@ public void setup()
     @Test
     public void testGenerationOfBuyTransaction() throws IOException
     {
+        investmentPlan.setType(InvestmentPlan.Type.BUY_OR_DELIVERY); // explicit
```
buchen, line 50: I removed the comment "explicit now". I don't think it helps the explain the current code (and the change can be seen in the history).

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
```diff
@@ -1572,6 +1575,34 @@ private static void fixNullSecurityProperties(Client client)
...
+            }
+            else
+            {
+                if (plan.getAccount() == null)
+                    throw new IllegalArgumentException("portfolio and account are not set for " + plan.getName()); //$NON-NLS-1$
+
+                if (plan.getSecurity() == null)
+                    type = (plan.getAmount() >= 0) ? InvestmentPlan.Type.DEPOSIT : InvestmentPlan.Type.REMOVAL;
+                else
+                    throw new IllegalArgumentException("security is set with a set account for " + plan.getName()); //$NON-NLS-1$
```
buchen, line 1599: I removed the exceptions here. The problem: if the exception is thrown, then the file is never opened and the user has no chance to "fix" the entries.

Previously, those checks were done when generating the transactions. Then the user could fix the plan, for example deleting the plan and re-creating one, or editing the plan. All error are anyway programming errors because they should never happen if the editing logic is correct.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
```diff
@@ -481,16 +488,22 @@ private TransactionPair<?> createAccountTx(CurrencyConverter converter, LocalDat
...
         else
         {
-            type = AccountTransaction.Type.REMOVAL;
+            transactionType = AccountTransaction.Type.REMOVAL;
             txAmount = -txAmount;
         }
+        if (type == Type.INTEREST)
+        {
+            transactionType = AccountTransaction.Type.INTEREST;
+            txAmount -= taxes;
```
buchen, line 504: The logic for all other transactions is:

the amount is the monetary amount of the transactions.
taxes are deducted from this amount implicitly.

If you remove the taxes here, then the taxes will be removed again.
And it is not consistent with the other transactions.

buchen:
Rebase. Merged. Done.

Thanks.

I added one more commit. It removes the "negative amount indicates a REMOVAL" logic. That is not needed anymore because we have the explicit type.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
Mikerion: Hello, interesting, this was there because otherwise transaction showed wrong value when I was testing it IIRC.

Mikerion: obrazek
@buchen Created transaction is adding taxes to the amount, not subtracting it.

buchen: Bildschirmfoto 2024-06-08 um 14 24 40

Bildschirmfoto 2024-06-08 um 14 24 59

Bildschirmfoto 2024-06-08 um 14 30 35


I believe, if you have a look at the cash account, you should see that 100 is added to the balance.

The view says:

* You got 120 of interest payments
* You had to pay 20 of taxes
* Your payout was 100

Is that also what you see?

That is the same behavior that is implemented for the investment plan of the purchase (only there the reverse logic applies because one pays an amount X plus the fees to purchase the instrument).

Now I agree that entering the values might be counterintuitive to the user. Do you think that adding gross and net values (and making sure both are in sync) is more helpful?

Mikerion: Yes, I see that.
image

I think problem is in consistency - if I create Interest transaction from Cash Account and from Plan I would need to insert different values. If I insert same values (as user assuming using gross everywhere and taxes) the result transactions are different:

Adding Interest through Account (100 as gross, payout 80):
image

Adding Interest through Plan (also 100 assuming it's gross, but it's actually net now, resulting in 120 gross as in first screenshot)
image


Result transactions with different net value payouts:
image

buchen: I think the best way is to have the very same input controls (gross value, taxes, net value) when creating the investment plan as when creating a single transaction. I can do this.

buchen:
Bildschirmfoto 2024-06-09 um 21 15 55

I am now using the same labels and show net/gross values.

PR #4005, opened 2024-05-08
Add portfolio filter in Trades view

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/1597 (except the Filter by date range (applied on the Close Date)..)
Issue: https://forum.portfolio-performance.info/t/weitere-filter-bei-trades-unter-dem-reiter-berichte-fur-einzelne-depots/21955
Issue: https://forum.portfolio-performance.info/t/trades-filtern/18171
Issue: https://forum.portfolio-performance.info/t/trades-nach-offen-vs-geschlossen-filtern/11208/5, fifth post

Hello this is a proposition to add to the Trades view a filter by account, in addition to the onlyOpen/onlyClose/onlyProfitable/onlyLoss already existing filtering options.

I have also added in the search box the possibility to search in the Portfolio column (second commit).

(I am thinking a little change to the `clientFilterMenu`, allowing to give the possibility to **only** fetch Portfolio account and custom filter account (instead of  fetching Portfolio account + (Portfolio +Deposit account) +custom filter account) may be good, .
I do not see the value of a (Portfolio + Deposit account) filter here for example, since, if I am not mistaken, it is the same filter as the Portfolio account only. Same for the `SecuritiesPerformanceView` (which is the reference for this commit's code). So the Portfolio +Deposit account takes a lot of place in the menu for nothing in those two views.
tradefilter
)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
@@ -14,6 +14,7 @@ import org.eclipse.e4.core.di.annotations.Optional;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.ActionContributionItem;
 import org.eclipse.jface.action.ControlContribution;
+import org.eclipse.jface.action.IMenuListener;
 import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.Separator;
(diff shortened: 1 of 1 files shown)


flywire:
This should be refactored to support current functionality with: three layer icon "Filter data by portfolio and reference account".

image

mierin12:
Yes indeed, there also has been also significant changes in the main code of TradesView since, so it will be easier for me to start from a new PR rather than managing all the rebase conflicts. I am closing this one and will try to work again on this.

PR #4006, opened 2024-05-08
Securities Account's Holdings chart

mierin12:
Hello,
This is a proposition to add a Pane in the Securities Account view showing the PieChart holdings of the individual Portfolios. 
So nothing new, same as if the Statement of Assets Holdings or associated widget was filtered on each of the Securities Account, one by one.
But the goal here is to quickly and simply get access to those charts, since you can very easily scroll through all your Securities accounts.
(Possible next step : add an option to add the referenced cash account to the holdings) 

**The Good:**

- Even if the feature is not validated, there would still be at least some little stuff fix that might be reusable (?) :

     - Circular Chart (used for SWT Holdings Pie Chart + Taxonomy SWT Holdings Pie Chart) are not displaying any inner nor outer label in case of 100% holding or very close to 100%. A tentative fix from the angle value itself is proposed.
     - SWT Holdings Pie Chart : added the total amount for the center of the Chart, similar to the one existing for Taxonomy SWT Pie Chart, also applicable for the general Statement of Assets Holdings chart.

- **SWT Pie Chart (light mode) :**
2024-05-08 23_59_32-Portfolio Performance
- **Browser Pie Chart (medium size)** 
 Capture d'écran 2024-05-08 231709

**The Bad:**
- (?) I tried to not duplicate code so I reused the existing class. Meaning we should be a bit more careful in the review.
- (?) snapshot is not instantiate at the creation of the chart but at the refresh. It stills seems to work. I do not know how to instantiate it with a non null portfolio in `createViewControl` of the new `PortfolioHoldingsPane`.  
- **Browser Pie Chart (small size) :** the ring quickly become small in this chart. and this pane may be small for some users.
Capture d'écran 2024-05-08 231559

**and The Ugly:**
- **SWT Pie Chart (dark mode):**

this is why the PR is in draft... I cannot figure out why the CSS is failing here, while it is not failing for the other Holdings chart of Statement of Assets. The label background are correct but the chart background specifically fails ! I thought the custom/dark.css "chart" properties would be automatically applied through `CircularChart`? Actually while investigating I put some crazy red/green color for .chart background-color in this css but they never appeared. I am at loss.


buchen:
@mierin12 writes:
> I cannot figure out why the CSS is failing here, while it is not failing for the other Holdings chart of Statement of Assets.

Me neither. It is not clear why it is not styling the component correctly. If wrap this with yet another composite, then it works as expected:

```java
        Composite container = new Composite(parent, SWT.NONE);
        container.setLayout(new FillLayout());
        chart.createControl(container);
        return container;
```

I think the styling is a little fickle as it was bolted on to SWT at a later point in time. Maybe it does not pick up the styling because it is now part of the PageBook with a stack layout?!? Anyway, I think adding the composite does not hurt.

buchen:
One more thing: I wonder if we just should use the SWT charts. In the dashboards, I also made the decision to only use the SWT widgets. They may not look so nice, but they are much better support than the HTML widget. I find it very painful to change the HTML charts because there is next to no debugging available with the embedded browser. And in this location, there was never the "older" HTML chart available.

What do you think?

mierin12:
Yes why not for swtchart only. People not using the widget and having by default the HTML chart may see this swtchart pie chart for the first time, but we can see if there is user feedback about it. I prefer medium and big size HTML over medium/big swtchart, but I greatly prefer small size swtchart over the small HTML, and at this pane location the chart may be small.

One little visual difference is that the slice boundaries of swtchart are always white, even in Dark mode, while HTML chart adapts them nicely. I have checked a bit the visual of dark slice on swtchart: not bad but it seems that the slice colors are getting out of the slice. With a slice thickness of 2 this behaviour is hidden, but it also gives a new "look" to the graph to get used to.

mierin12:
Is the behaviour of an empty chart for an empty portfolio ok ? I was thinking that maybe a messagepainter text, same as the Security chart, could be nice, but maybe in a next step if the currency behavior is good enough.

I think it is possible to generate an java error with : create a new empty portfolio, select the holdings tab, then delete this empty portfolio.. I also put it in the "next step" case ..

buchen:
> Is the behaviour of an empty chart for an empty portfolio ok ?

I think that is okay. It is unlikely anyway that the user will be on that tab if there the portfolio is empty. And if he is, he knows.

buchen:
> but I greatly prefer small size swtchart over the small HTML, and at this pane location the chart may be small.

That is also my gut feeling. And we could put two or three charts next to each other (in a future pull request): first the holdings, then by classification, etc.

mierin12:
> > but I greatly prefer small size swtchart over the small HTML, and at this pane location the chart may be small.
> 
> That is also my gut feeling. And we could put two or three charts next to each other (in a future pull request): first the holdings, then by classification, etc.

Yes very good idea !

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/holdings/HoldingsPieChartSWT.java
```diff
@@ -93,15 +114,17 @@ public Control createControl(Composite parent)
...
+            ((Composite) chart.getPlotArea()).addListener(SWT.MouseUp,
                         event -> chart.getNodeAt(event.x, event.y).ifPresent(node -> {
                             NodeData nodeData = id2nodeData.get(node.getId());
                             if (nodeData != null)
                                 financeView.setInformationPaneInput(nodeData.position.getInvestmentVehicle());
                         }));
 
-        updateChart();
-
+         updateChart();
```
mierin12, line 126: updateChart needs a non null Snapshot, but at this stage, Snapshot is not yet instantiated when the chart is created from the new information pane, so it is hidden in the finance view 'if' block (=this is the existing Statement of Assets/Holdings chart). I did not notice any issue, but to be noted.

buchen:
Merged.

I removed the labels to reuse the existing labels. And I removed the HTML pie chart (which will allow us more charts in the information pane in the future).

PR #4011, opened 2024-05-11
Add a widget to display the current Sharpe Ratio value

fkocik:
Hi guys,

First of all, thanks for this platform that is the best tool for investment management ;-)

To evaluate the global safety of my investments, I usually use the Sharpe Ratio (https://en.wikipedia.org/wiki/Sharpe_ratio).

The goal of this contrib is to display the value of this ratio on the performance dashboard as a simple widget.

Since this indicator is based on a personal risk free IRR that cannot be computed automatically (for example, if you are on TradeRepublic, it's 4% but some others brokers or country have different values). I wrote code to set it via global Preferences. However, perhaps it's not the good place to do it ?


fkocik:
@buchen Thanks, I tried to fix affected files according to Maven report. Hope that's OK now :-)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -173,6 +177,23 @@ public enum WidgetFactory
...
+                    (widget, data) -> IndicatorWidget.<Double>create(widget, data)
+                    .with(Values.PercentPlain)
+                    .withColoredValues(false)
+                    .with((ds,period) -> {
+                        IPreferenceStore store = PortfolioPlugin.getDefault().getPreferenceStore();
+                        PerformanceIndex index = data.calculate(ds,period);
+                        double r = index.getPerformanceIRR();
+                        double rf = store.getDouble(UIConstants.Preferences.SHARPE_RATIO_IRR)/10000;
+                        double volatility = index.getVolatility().getStandardDeviation();
+                        return (r-rf)/volatility;
```
Nirus2000, line 190:
```suggestion
            double volatility = index.getVolatility().getStandardDeviation();
            
            if (Double.isNaN(rf))
                return Double.NaN; // Handle invalid rf value
            
            double excessReturn = r - rf;
            return excessReturn / volatility;
```

buchen:
Hey @fkocik,

thanks for your contribution. It's a great addition.

One thing: the risk free return is currently **not** stored in the file itself but in the installation preferences. From my point, this has a couple disadvantages: first, if one synchronizes the file to another device (desktop or mobile), the risk free return is lost. Second, a typical troubleshooting is just delete the installation (which includes all preferences).

My proposal: let's configure this per file and save it inside the portfolio file.

Sn1kk3r5:
While reading this, I would like to place one thing. When setting the IRR in the global section of preferences, it must be clear to the user where to change, and how it works. As IRR depends on a reporting period which might defer from the period one is looking at, this could lead to a lot of confusion which then needs a lot of explaining.
Cheers

fkocik:
Hi @buchen,

I'm agree on this point but in my first approach I didn't find how to do that.

I just rework the widget to associate it with a Config object : is this the right way to do this ?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/SharpeRatioWidget.java
```diff
@@ -0,0 +1,37 @@
...
+            PerformanceIndex index = dashboardData.calculate(ds, period);
+            double r = index.getPerformanceIRR();
+            double rf = ((double)configuration.getRiskFreeIRR()) / 10000;
+            double volatility = index.getVolatility().getStandardDeviation();
+
+            if (Double.isNaN(rf))
+                return Double.NaN; // Handle invalid rf value
+
+            double excessReturn = r - rf;
+            return excessReturn / volatility;
```
Nirus2000, line 34: In the SharpeRatioWiget file they call the variable e.g. “riskFreeIRR” and here only "r" or “rf”... 
Let's name it immediately 👍🏻 

```suggestion
            double performanceIRR = index.getPerformanceIRR();
            double riskFreeIRR = ((double)configuration.getRiskFreeIRR()) / 10000;
            double volatility = index.getVolatility().getStandardDeviation();

            if (Double.isNaN(riskFreeIRR) || volatility == 0)
                return Double.NaN; // Handle invalid risk-free IRR or zero volatility

            double excessReturn = performanceIRR - riskFreeIRR;
            return excessReturn / volatility;
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/SharpeRatioWidget.java
```diff
@@ -0,0 +1,37 @@
...
+    {
+        super(widget, dashboardData, true, null);
+        SharpeRatioConfig configuration = new SharpeRatioConfig(this);
+        this.addConfig(configuration);
+
+        this.setFormatter(Values.PercentPlain);
+        this.setValueColored(false);
+        this.setTooltip((ds, period) -> {
+            return Messages.TooltipSharpeRatio;
+        });
```
Nirus2000, line 23:
```suggestion
        this.setTooltip((ds, period) -> Messages.TooltipSharpeRatio);
```

Nirus2000 (changes requested):

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/SharpeRatioConfig.java
```diff
@@ -0,0 +1,85 @@
...
+            try
+            {
+                this.riskFreeIRR = Integer.parseInt(riskFreeIRRStr);
+            }
+            catch (NumberFormatException ex)
+            {
+                this.riskFreeIRR = 0;
+                PortfolioPlugin.info(ex.getLocalizedMessage());
+            }
+        }
```
Nirus2000, line 36: If riskFreeIRRStr is null, the parseInt method will throw a NumberFormatException, then we jump from try to catch.

```suggestion
        {
        try
        {
            this.riskFreeIRR = Integer.parseInt(riskFreeIRRStr);
        }
        catch (NumberFormatException ex)
        {
            this.riskFreeIRR = 0;
            PortfolioPlugin.info(ex.getLocalizedMessage());
        }
```

Nirus2000, line 36: If riskFreeIRRStr is null, the parseInt method will throw a NumberFormatException, then we jump from try to catch.

```suggestion
        try
        {
            this.riskFreeIRR = Integer.parseInt(riskFreeIRRStr);
        }
        catch (NumberFormatException ex)
        {
            this.riskFreeIRR = 0;
            PortfolioPlugin.info(ex.getLocalizedMessage());
        }
```

Nirus2000 (changes requested):

fkocik (commented):
Thanks @Nirus2000 :-)

Nirus2000 (approved):

buchen:
@fkocik thanks for updating the pull request.

I also had a local change pending (forgot to post it here) and it is very similar.

The main difference is: this pull request now stores the risk free rate of return **per widget**. My local implementation is storing it **per file**. I tend to think that if I, as user, pick a risk free rate of return, it would apply to all share ratios. Therefore I tend to want to use my local implementation. @fkocik Any thoughts?


@Sn1kk3r5 writes:
> When setting the IRR in the global section of preferences, it must be clear to the user where to change, and how it works. As IRR depends on a reporting period which might defer from the period one is looking at, this could lead to a lot of confusion which then needs a lot of explaining.

Agree. For now, the configuration is triggered from the widget (it is not so easy to put it into the general settings sections. A dialog is easier because it has validation logic and an explicit "OK").

Because the sharpe ratio is using the IRR and the IRR is annualized, I think one risk free rate of return applies to all widgets. However, I agree that if one calculates the IRR on short periods of time, one can get numbers. My thinking was to put the calculation also into the tooltip - then users can get an understanding of *why* the number is so high (or low).

buchen:
I have merged your changes and added two changes on top: store the risk-free rate of return per file and enhanced the tooltip to show the calculation.

Sn1kk3r5:
@buchen 
> Because the sharpe ratio is using the IRR and the IRR is annualized, I think one risk free rate of return applies to all widgets. However, I agree that if one calculates the IRR on short periods of time, one can get numbers. My thinking was to put the calculation also into the tooltip - then users can get an understanding of _why_ the number is so high (or low).

Fully agree.

Another point why I raised this is IRR sometimes changes. So as an example: IRR for Jan. to March 3,25 / April to June 3,5 / July to Sep. 3,75 and Oct. to Dec. 4,00. Now assuming one would set an IRR between 3,25 and 4.00 in looking at the same reporting period from Jan. to Dec. the sharp ratio will not be correct. Just because sharp ratio isn't made for this.

But I agree, when this is clear in the tool tipp, users need to read this in order to understand.

Cheers

PR #4021, opened 2024-05-17
Update config for Launch Configuration DSL plug in

Nirus2000:
Updating the memory usage when loading.

See PP-Forum:
https://forum.portfolio-performance.info/t/pp-wird-immer-langsamer/8819/9

I don't know whether it makes sense to set the configuration file to these default values for new installations.
Existing installations, i.e. updating the configuration file, will certainly be too time-consuming.

### DIFF
--- a/portfolio-app/eclipse/launches.lc
+++ b/portfolio-app/eclipse/launches.lc
@@ -28,13 +28,15 @@ eclipse configuration PortfolioPerformance {
 	ignore org.junit;
 	ignore org.hamcrest.core;
 	ignore org.hamcrest.library;
-	
-    memory min=512M max=1G;
     
     execution-environment 'JavaSE-17';
     
     argument "-consoleLog";
     
+    vm-argument "-Xms2G";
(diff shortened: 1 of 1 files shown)


buchen:
This launch configuration is only used when starting the app from within Eclipse.

As such, we can change it, but I am not sure if it is really necessary. The start must be fast (because during development I do it all the time) and my test file are typically rather small. Therefore I would rather not change it here.

The "productive" configuration is in the ```name.abuchen.portfolio.product`` file.
We can change this, however, we first need to research the impact.

It is my understanding, if the Java Virtual Machine does not have the minimum space available, it does not even start. And then the user would have to start editing the ini files.
My understanding about the maximum value is that the virtual machine will allocate until the space available. If the compute does not have that much physical memory, it will start swapping - which will make performance very bad.

As said, that is my understanding. The latest virtual machines might react more reasonable to this. I'd love to learn about it.

buchen:
Closing - for the start of the development, I do not think we need more memory (use smaller test files instead)

PR #4025, opened 2024-05-21
Adding 4 new securities including transaction and dividends

Sn1kk3r5:
Deposit on USD cash account
Adding Amazon security in USD incl. purchase
	Adding 20:1 stock splitt
Adding Daimler Truck security
Adding Mercedes Benz Group incl. purchase
	Adding 3 dividend payments incl. tax
Buying NVIDA incl. purchase
	Adding 6 dividend payments incl. tax
Adding all new securities into existing taxonomy


buchen:
merged (with a minor change: I removed the latest security price as it is anyway just -1)

PR #4029, opened 2024-05-23
modify trade republic importer

christen90:
adding some 3 lines formats
fixed März
added Steuern Steueroptimierung transaction
corrected typo in Gebühren section
added setMaxSize to removal blocks

PR #4033, opened 2024-05-24
Improvement of the regular expression of the date in the PDF importers

Nirus2000:
Improvement of the regular expression in the date by replacing the regex pattern in the format MMMM (LLLL) and MMM (LLL). This is defined in the ExtractorUtils as a static variable.

Remove .replace("Mar", "Mar") (JDK7 vs. JDK8)

https://github.com/portfolio-performance/portfolio/pull/4029#discussion_r1612042728 and ff.

@buchen -> https://github.com/portfolio-performance/portfolio/issues/2683

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java
@@ -30,6 +30,41 @@ import name.abuchen.portfolio.money.Values;
 
 public class ExtractorUtils
 {
+    /**
+     * @formatter:off
+     * Regex pattern to match month names in various languages, including English, German, Spanish, Italian, and French.
+     * The pattern covers month names in both full and abbreviated forms.
+     *
+     * Supported month names:
+     * - English: January (Jan), February (Feb), March (Mar), April (Apr), May (May), June (Jun), July (Jul), August (Aug), September (Sep), October (Oct), November (Nov), December (Dec)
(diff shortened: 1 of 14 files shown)


buchen:
I am skeptical that this change will help us in the long run.

What problem exactly is this pull request trying to address?

My hunch is:
* by hard coding the translations, we will fail more often - date parsing is complex and we should leave it to Java to understand what a full month name is, what an abbreviated month, what a standalone month name and what not a standalone month name is
* error messages that are printing the regular expressions will be long and very hard to read
* we introduce this "mixed creation of regular expressions" discussed in 2683 - I still would argue that it does make maintenance of the regular expressions harder in the future
* while previously the parsing failed with an error message related to date parsing, now it fails with a very long pattern
* it makes it slower because the parsing of the regular expression is more complex (but, admittedly, it is only half a second when running the tests)

Where am I wrong?

Where I see the point is to move the "Mrz" and "Mär" handling to the ```ExtractorUtil.asDate``` method. Although I understand at the moment we know only that the Sutor bank that is abbreviating "März" as "Mrz". This change I will cherry-pick regardless of the other discussion.

buchen:
And: please, please, before make the change that takes a lot of work (I know looking up all month name in all languages takes a long of work and diligence), we can also make a draft change and then discuss :-)

Nirus2000:
Hello @buchen
I understand... the first goal has already been successful. 
The issue with the JDK 7 to 8... okay, your variant is smarter ;-)

The problem is that with this pattern, we get a fail faster than if we work `.*` or `[\w]{3,4}` or `[\wä]{3,4}` and then java checks if this is a date. I therefore do not believe that this is slower.
We pattern first and then we check if it's a date, right?

This only applies to the names of the months, there are no other date differences in my opinion.

Alex 🔢

PR #4044, opened 2024-05-31
Update to new Portfolio Report API

buchen:
Portfolio Report now

* includes cryptocurrencies
* allows to retrieve all available crypto currencies
* can provides quotes in different currencies

This pull request changes
* remove the now obsolete API to retrieve prices by market
* removes the obsolete properties from the client
* uses Portfolio Report when creating crypto currencies (either via search, the explicit dialog or during the PDF import)

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/baaderbank/BaaderBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/baaderbank/BaaderBankPDFExtractorTest.java
@@ -36,7 +36,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.collection.IsEmptyCollection.empty;
 import static org.junit.Assert.assertNull;
 
-import java.io.IOException;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Iterator;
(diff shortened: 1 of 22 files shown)


buchen:
@Nirus2000 please have a look at c1c763147e34c58ab5996d71e693883ec19fcb76 as I needed to change the tests slightly

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/baaderbank/BaaderBankPDFExtractorTest.java
```diff
@@ -67,27 +67,29 @@
...
                 {
                     return List.of( //
-                                    new Coin("bitcoin", "BTC", "Bitcoin"), //
-                                    new Coin("ethereum", "ETH", "Ethereum"));
+                                    new Coin("bitcoin", "BTC", "Bitcoin").asResultItem(), //
+                                    new Coin("ethereum", "ETH", "Ethereum").asResultItem());
                 }
-            };
+            });
         }
```
Nirus2000, line 93: Can we look into making this a global list of coins tested so far?
Maybe outsource the list?

Nirus2000 (approved):
I think looks good 👍🏻

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/baaderbank/BaaderBankPDFExtractorTest.java
buchen: good Idea. I have extracted into a separate helper class with a static method.

Unfortunately, it is a method of the AbstractPDFExtractor which means we still have to override the method in each test.

PR #4087, opened 2024-06-21
Corrected translation of TTWROR and TTWRORAnnualized widgets labels i…

netsho:
The TTWROR and TTWROR labels in fr resource file were in English. 
Translated both the labels in the French.


netsho:
While using the app, i came by this translation error. 
First time creating a pull request, i don't know if i should have created an issue before, I saw the error and thought about correcting it quickly.
I'm open to correcting my approach if it does not respect the workflow.

buchen:
Thanks @netsho for sharing the fix.

If you want to fix further translations issues, you can also [join the POEditor project](https://poeditor.com/join/project/4lYKLpEWOY).

For code, pull requests is the way to go.

PR #4088, opened 2024-06-22
add CSV export to Taxonomy stacked charts

mierin12:
Issue: https://forum.portfolio-performance.info/t/daten-exportieren-button-ohne-funktion-unter-klassifizierungen/27371

Hello, this is a proposition to add (or re add??) the CSV export of chart data to the Taxonomy stacked charts.
Currently the export button is doing nothing for those charts. 
I duplicated `TimelineChartCSVExporter` with some simplifications (I do not think we can expect `discontinuousSeries` here). dateSeries is a bit different too since its a categorized graph.


buchen:
> this is a proposition to add (or re add??) the CSV export

Thanks. I think the button never had a functionality. I believe I did not know how to hide buttons (here: the export button) if the current view has nothing attached to it. But having an implementation is better.

PR #4089, opened 2024-06-22
Fixed date format pattern for local Canada

Nirus2000:
Fixed date format pattern for local Canada
Add @SuppressWarnings("nls") and remove all $NON-NLS-1$ and $NON-NLS-2$ ...


buchen:
Rebased, don't merge ;-)

PR #4093, opened 2024-06-23
Performance improvement on TTWROR calculation: incrementally apply tr…

buchen:
…ansactions


Nirus2000:
This is a great change for the performance of Portfolio Performance... 🥇 💯 
WOW ! 👍🏻

henfri:
Thanks also from my Side. I noticed the performance increase and traced it to this one!

PR #4099, opened 2024-06-24
ClientFactory: Remove stale unused load() methods

pfalcon:
The load/save functionality is now implemented using a contained ClientPersister class. However, top-level load() methods which just assume plain-text XML format weren't removed and just sit there, unused.


pfalcon:
Ah, I see, they're used for tests only. Maybe should be described as such ;-).

PR #4102, opened 2024-06-24
Add search with Ctrl+F

### LINKED ISSUE #4051: Search with Ctrl+F
Currently if you want to search (on any of the pages), you have to manually take the mouse and click in the "Search" box at the top left - the standard "Ctrl+F" (or F3) keyboard shortcuts do not work.  It would be nice if you could just hit Ctrl+F, and it would put the cursor in the "Search" box, so you could search without needing to rely on the mouse.

netsho:
Closes #4051 

**Test cases verified manually :** 

1. Initial View Without Search Bar: Verify Nothing Happened => **Passed**.
2. Initial View With Search Bar: Verify The Focus on the Search Bar => **Passed**.
3. Switch From View Without Search Bar to View With Search Bar: Verify The Focus on the Search Bar => **Passed**.
4. Switch From View With Search Bar to View Without Search Bar: Verify Nothing Happened => **Passed**.
5. Switch Between Two Views With Search Bars: Verify The Focus on the Search Bar of The Second View => **Passed**.
6. Switch View With Active Search Bar: Verify The Focus on the Search Bar of Each View And No Errors =>**Passed**.
7. Switch Between Multiple Views Sequentially: Verify The Focus on the Search Bar of Each View => **Passed**.
8. Switch Views Rapidly: Verify The Focus on the Search Bar of Each View => **Passed**.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/AbstractFinanceView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/AbstractFinanceView.java
@@ -77,10 +77,10 @@ public abstract class AbstractFinanceView
     private List<Menu> contextMenus = new ArrayList<>();
 
     /**
-     * Boolean used to verify whether the listener for "CTRL + F" is added to
-     * the display.
+     * Listener used for detecting whether CTRL + F shortcut is pressed on the
+     * keyboard.
      */
-    private static boolean isCtrlFListenerAdded = false;
+    private static Listener keyListener;
 
     protected abstract String getDefaultTitle();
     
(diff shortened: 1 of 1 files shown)


buchen:
I like the proposal.

Two questions:
* why not have a key listener on the e4 model and then "find" the text?
* what about the search field in the lower information pane?

netsho:
* I did not know about the e4 model. I have read about it and i saw that it's better for flexibility to handle UI bindings. I'll have a look at it and try to implement this solution if it's in my capacity.
* The "CTRL + F" is not implemented for the lower information pane. I'll try to do it. In pages where there is only one search box between the main view and information pane, it is fine. But in the other case where there is 2 search box, i have to find a way to make a "priority" for the search box in the main view.

This pull request will go back into draft mode.

buchen:
Maybe what we can do:

* we tag a composite if it contains a search field (say the AbstractFinanceView or the InformationPane)
* if the user presses Ctrl-F, we take the focused control, go up the composite hierarchy until we find the tag, then search for the search box and activate it

This way one could also activate the search field in the information pane - if for example the lower table has the focus currently.

buchen:
Hi @netsho,

sorry, it took some time because I was busy.

I have now taken up your idea, but moved it to use the key handler of the e4 framework.
Additionally, it will also focus a search box in the lower information pane (if it is not hidden).
And if the search bar on the top is already focused, it will focus the lower one (if it exists).

I marked the both of use as co-authors. Thank you for your contribution (and patience).

Andreas.

PR #4103, opened 2024-06-24
Securities Account chart  : add colors to invested capital and delta

mierin12:
Hello, this is a proposition to add a bit of colors to the currently grey Invested Capital/Delta data series of the new Securities Account charts.

This PR is proposing to use the same colors as in the Securities chart: same yellow for invested capital than for "share held" and the same blue/red area for positive/negative Delta as for the pos/neg "market trend vs purchase price", through CSS to deal with light and dark theme.

**Notes :** 

- another totally different implementation would be to let the user chose the area/line/colors setup for this Securities Account chart, same as Statement of Assets chart. However it did not seem obvious to implement for me, so I am proposing what I managed to do, so with fixed color (copy from Securities chart)
- with a bi-color delta, this is adding "fake" data series from the graph that should not be exported in CSV, so there is a small change on the CSV exporter /!\
- with both invested capital and Delta in "area" mode, the colors can overlap. In the examples below, I am showing examples where Delta is kept as a line and not a bi-color area in case this is preferred. 

Some examples for different kind of portfolio and theme:
**Before:**
grey2

grey3

grey4

**After (delta is bi-color) :** current PR
Light_InvArea_Delta_Area_2

Light_InvArea_Delta_Area_3

Light_InvArea_Delta_Area_4

Dark_InvArea_Delta_Area_2

Dark_InvArea_Delta_Area_3

Dark_InvArea_Delta_Area_4

**After (delta is kept as a line)** PR to be simplified
Light_InvArea_Delta_Line_2
Light_InvArea_Delta_Line_3
Dark_InvArea_Delta_Line_2
Dark_InvArea_Delta_Line_3

### DIFF
--- a/name.abuchen.portfolio.ui/css/shared/dark.css
+++ b/name.abuchen.portfolio.ui/css/shared/dark.css
@@ -72,6 +72,13 @@ SecuritiesChart {
 	shares-held-color: #ffaf4d;
 }
 
+PortfolioBalanceChart {
+	invested-capital-color: #ffaf4d;
+	absolute-delta-color: #4ddaff;
+	delta-area-positive-color: #23b0d7;
+	delta-area-negative-color: #e25b5a;
+}
+
 * {
 	font-family: "#systemfont";
 }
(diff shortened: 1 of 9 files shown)


buchen:
tested and merged.

PR #4115, opened 2024-06-29
Improve autoscale of icons for high definition screens 4K

### LINKED ISSUE #411: Zu kleine Symbole bei großen Bildschirmauflösungen
Bei großen Bildschirmauflösungen (mein Laptop hat z.B. 3200x1800px unter Windows 10) ist das Programm nicht mehr bedienbar, weil die Icons viel zu klein angezeigt werden. Mit ca. 2mm Größe verschwinden die Symbole dann regelrecht im Bild und somit verschwinden wichtige Bedienelemente des Programms.

Es fehlen m.E. Icon-Sets (z.B. 48x48px) für große Bildschirmauflösungen. Diese könnte man entweder in den Einstellungen selbst auswählen oder das Programm erkennt die Auflösung selbst und verwendet dann entsprechend größere Icons zur Anzeige.

Nirus2000:
Closes #411
Improved automatic scaling of icons for 4K high-resolution screens or additional scaling for font sizes > 100%

Hello @buchen 
Maybe it would be a cool idea to create a "HowTo" for the settings of the PortfolioPerformance.ini and to integrate a possibility in the "Settings" where we can easily select options by checkboxes. e.g. the memory from 
- 512Mbyte to 2Gbyte ... (Xms, Xmx, Xsss --> [here](https://forum.portfolio-performance.info/t/pp-wird-immer-langsamer/8819/9))
- Scale resolution
in the *.ini

**Question:**
_More simply, display the PortfolioPerformance.ini in a text field so that it can be edited or integrate a custom.ini for special modifications if necessary._

---
These settings can be added subsequently in PortfolioPerformance.ini.
```
-Dswt.enable.autoScale=true
-Dswt.autoScale=quarter
-Dswt.autoScale.method=nearest
```
---
It is also possible to set the autoscaling to a specific value, e.g. 150
```
-Dswt.autoScale=150
```

### Before:
grafik
grafik

### After:
grafik
grafik

---

###Another example:

Your screen settings
grafik

with `-Dswt.autoScale=quarter`
grafik
grafik

with `-Dswt.autoScale=200`
grafik
grafik

### DIFF
--- a/portfolio-product/name.abuchen.portfolio.distro.product
+++ b/portfolio-product/name.abuchen.portfolio.distro.product
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="Portfolio Performance" uid="name.abuchen.portfolio.distro.product" id="name.abuchen.portfolio.bootstrap.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="0.69.1.qualifier" useFeatures="true" includeLaunchers="true">
+<product name="Portfolio Performance" uid="name.abuchen.portfolio.distro.product" id="name.abuchen.portfolio.bootstrap.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="0.69.1.qualifier" type="features" includeLaunchers="true" autoIncludeRequirements="true">
 
    <configIni use="default">
    </configIni>
@@ -13,6 +13,8 @@
       </vmArgsLin>
       <vmArgsMac>-XstartOnFirstThread -Dorg.eclipse.swt.internal.carbon.smallFonts
       </vmArgsMac>
+      <vmArgsWin>-Dswt.autoScale=quarter -Dswt.autoScale.method=nearest
+      </vmArgsWin>
    </launcherArgs>
 
    <splash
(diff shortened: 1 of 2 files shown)


buchen:
This change proposes to add the properties for **all** installations.
But the text suggests to "integrate a possibility in the "Settings" where we can easily select options by checkboxes".

Now I am not sure if you suggest to merge this for all - or if this is how it could look like for *some* users?


About: ```swt.enable.autoScale``` appears to be deprecated and [it appears to not work anymore](https://bugs.eclipse.org/bugs/show_bug.cgi?id=493462#c17).

Not sure if I can test ```-Dswt.autoScale=quarter``` because it appears to apply only to Windows and Linux. But if I understand it correctly, the default value is ```integer```:

> integer: scale factor depends on the current display resolution, but only uses integer multiples of 100%. The detected native zoom is generally rounded down (e.g. at 150%, will use 100%), unless close to the next integer multiple (currently at 175%, will use 200%).

So if it improves it, it sounds like setting it to "quarter" can be okay for all users:

> quarter: scale factor depends on the current display resolution, but only uses integer multiples of 25%. The detected native zoom is rounded to the closest permissible value. (This used to be the default in the last two pre-release milestones.)

buchen:
> with -Dswt.autoScale=200

This part I really do not understand. The icons include the default image and the 2x image (200). Therefore the image should be there. Why is it not loaded? (And to scale up the 100% image is an option but of course looks ugly)

Nirus2000:
Self note... --> look @ e8c8c40215ce612f794182bfa381f402bb82629b

portfolio-product/name.abuchen.portfolio.distro.product
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?pde version="3.5"?>
 
-<product name="Portfolio Performance" uid="name.abuchen.portfolio.distro.product" id="name.abuchen.portfolio.bootstrap.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="0.69.1.qualifier" useFeatures="true" includeLaunchers="true">
+<product name="Portfolio Performance" uid="name.abuchen.portfolio.distro.product" id="name.abuchen.portfolio.bootstrap.product" application="org.eclipse.e4.ui.workbench.swt.E4Application" version="0.69.1.qualifier" type="features" includeLaunchers="true" autoIncludeRequirements="true">
```
buchen, line 4: Looks like the product file format has changed.
I checked: the ```autoIncludeRequirements``` is the default Tycho behavior so I do not expect changes in the Maven build.

But it makes sense to update it.

buchen:
merged.

after reading the documentation, I have enabled this flag also for Linux

PR #4120, opened 2024-07-01
Separation of the numerical values in the security chart

Nirus2000:
* There is the problem that too many numerical values are separated when displaying buy, sell, deposit... overlap as soon as the chart has a certain format. This is how we separate the display of values from investment markers and dividend values.
* The same for high/low values.
* Sorting label in menu structure
* Rename SHOW_DATA_LABELS to SHOW_DATA_DIVESTMENT_INVESTMENT_LABELS

**Together:**
grafik

**Split:**
grafik


Nirus2000:
> **_Unfortunately, I can't think of a better translation for the investment figures.
"Zahlenwerte anzeigen" or "Show data labels" is not really correct!_**

@chirlu @OnkelDok 
Do you have a suitable new translation for the display of purchases/sales/deliveries/deliveries in general, in the settings?

buchen:
add to client migration

buchen:
merged

PR #4122, opened 2024-07-01
Modify Trade Republic PDF-Importer to support new transactions

christen90:
Hello @Nirus2000,
I added 2 tax refund formats and a saveback deposit booking.

Interesting is that Trade Republic labels tax payments as Tax Optimisation and removes 0,01 €. 😄Thats why I introduced the sign recognition switch also for tax refunds. After that Kontoauszug11 made some trouble - I fixed it manipulating the amounts after and start saldo.

There is still an issue with the recognition of deposit/removal. A credit card payment was recognized as refund because there was a sell booking before with bigger amount than payment.

See TODO note at name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java:1905

I think we might also need to search for other lines which will not imported (buy/sell).

I have no Idea how to accomplish this. Could you please help out?

BR Christian


Nirus2000:
Squash, merged and rebase

PR #4126, opened 2024-07-02
Improves the group width dynamically based on the widest element

Nirus2000:
Fixes the width calculation of the FormLayout for larger resolutions and fonts.

### Before:
grafik

### After:
grafik

---

Hello @buchen 
What I have not been able to customize so far is this problem here. Do you have any ideas?
[I think here](https://github.com/portfolio-performance/portfolio/blob/a01b34a49b738b5e510e8fb058409e7c384b4a6d/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/OpenPreferenceDialogHandler.java#L97:L113)

grafik
grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AttributeSettingsPane.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AttributeSettingsPane.java
@@ -215,11 +215,19 @@ public class AttributeSettingsPane implements InformationPanePage
             // before
             stylingEngine.style(gpColors);
 
-            int width = SWTHelper.widest(positiveExceeded.getFirst(), negativelyExceeded.getSecond());
+            // Compute the widest element width in the group using
+            int width = SWTHelper.widest(positiveExceeded.getFirst(), negativelyExceeded.getFirst(), button,
+                            resetButton);
 
-            FormDataFactory.startingWith(positiveExceeded.getFirst()).width(width + 20)
-                            .thenBelow(negativelyExceeded.getFirst()).width(width + 20).thenBelow(button)
-                            .left(new FormAttachment(0)).thenRight(resetButton);
+            // Set the width of the gpColors group to the widest element's width
+            // plus padding
+            gpColors.setSize(width, gpColors.computeSize(SWT.DEFAULT, SWT.DEFAULT).y);
+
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AttributeSettingsPane.java
```diff
@@ -215,11 +215,19 @@ private static void createColorView(Composite parent, IStylingEngine stylingEngi
             // before
             stylingEngine.style(gpColors);
 
-            int width = SWTHelper.widest(positiveExceeded.getFirst(), negativelyExceeded.getSecond());
+            // Compute the widest element width in the group using
+            int width = SWTHelper.widest(positiveExceeded.getFirst(), negativelyExceeded.getFirst(), button,
+                            resetButton);
```
buchen, line 220: "button" and "resetButton" are displayed next to each other.

I do not unterstand whether it makes to determine their size separately. Even if one of the buttons is the widest element, it does not help because both are displayed next to each other.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AttributeSettingsPane.java
```diff
@@ -215,11 +215,19 @@ private static void createColorView(Composite parent, IStylingEngine stylingEngi
...
 
-            FormDataFactory.startingWith(positiveExceeded.getFirst()).width(width + 20)
-                            .thenBelow(negativelyExceeded.getFirst()).width(width + 20).thenBelow(button)
-                            .left(new FormAttachment(0)).thenRight(resetButton);
+            // Set the width of the gpColors group to the widest element's width
+            // plus padding
+            gpColors.setSize(width, gpColors.computeSize(SWT.DEFAULT, SWT.DEFAULT).y);
```
buchen, line 224: I believe this is the relevant change to improve it.

buchen:
I merge because I cannot test this at the moment - on macOS the change makes no difference. It always looks the same.

buchen:
> What I have not been able to customize so far is this problem here. Do you have any ideas?

I have quickly played around with

```java
        themeEngine.applyStyles(dialog, true);
```

but it has no effect. No clear idea right now why the size remains fixed. I'll concentrate on the release for now (and this change is "in")

PR #4138, opened 2024-07-11
Add new account statements to Raiffeisenbank PDF Importer

ZfT2:
- added Mastercard account transactions
- extended interest charge

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
@@ -688,7 +688,7 @@ public class RaiffeisenBankgruppePDFExtractor extends AbstractPDFExtractor
                             if ("H".equals(v.get("sign")))
                                 t.setType(AccountTransaction.Type.DEPOSIT);
 
-                            setTransactionDate(t, v);
+                            dateTranactionHelper(t, v);
 
                             t.setCurrencyCode(v.get("currency"));
                             t.setAmount(asAmount(v.get("amount")));
@@ -744,7 +744,7 @@ public class RaiffeisenBankgruppePDFExtractor extends AbstractPDFExtractor
                             if ("H".equals(v.get("sign")))
                                 t.setType(AccountTransaction.Type.DEPOSIT);
 
-                            setTransactionDate(t, v);
+                            dateTranactionHelper(t, v);
 
                             t.setCurrencyCode(v.get("currency"));
                             t.setAmount(asAmount(v.get("amount")));
@@ -787,7 +787,7 @@ public class RaiffeisenBankgruppePDFExtractor extends AbstractPDFExtractor
                                                         .assign((t, v) -> {
                                                             t.setType(Type.INTEREST_CHARGE);
 
-                                                            setTransactionDate(t, v);
+                                                            dateTranactionHelper(t, v);
 
                                                             t.setCurrencyCode(v.get("currency"));
                                                             t.setAmount(asAmount(v.get("amount1")) + asAmount(v.get("amount2")));
@@ -805,7 +805,7 @@ public class RaiffeisenBankgruppePDFExtractor extends AbstractPDFExtractor
                                                             if ("S".equals(v.get("sign")))
                                                                 t.setType(Type.INTEREST_CHARGE);
 
-                                                            setTransactionDate(t, v);
+                                                            dateTranactionHelper(t, v);
 
                                                             t.setCurrencyCode(v.get("currency"));
                                                             t.setAmount(asAmount(v.get("amount")));
@@ -850,7 +850,7 @@ public class RaiffeisenBankgruppePDFExtractor extends AbstractPDFExtractor
                             if ("H".equals(v.get("sign")))
                                 t.setType(AccountTransaction.Type.FEES_REFUND);
 
-                            setTransactionDate(t, v);
+                            dateTranactionHelper(t, v);
 
                             t.setCurrencyCode(v.get("currency"));
                             t.setAmount(asAmount(v.get("amount1")) + asAmount(v.get("amount2")) + asAmount(v.get("amount3")));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -728,6 +718,37 @@ private void addAccountStatementTransactions()
...
+                        .documentContext("currency", "nr", "year") //
+                        .match("^(?i)[\\d]{2}\\.[\\d]{2}\\. (?<day>[\\d]{2})\\.(?<month>[\\d]{2})\\. (?<note>.*) PN:4444 (?<amount>[\\.,\\d]+) (?<sign>[S|H])$") //
+                        .match("^.*$") //
+                        .match("^ (?<note2>.*)$") //
+                        .assign((t, v) -> {
+                            // @formatter:off
+                            // Is sign --> "H" change from DEPOSIT to REMOVAL
+                            // @formatter:on
+                            if ("H".equals(v.get("sign")))
+                                t.setType(AccountTransaction.Type.DEPOSIT);
```
Nirus2000, line 742: Here we have a consistency problem between translation and action

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -1249,4 +1252,20 @@ protected BigDecimal asExchangeRate(String value)
...
+            t.setDateTime(asDate(v.get("day") + "." + v.get("month") + "." + year));
+        }
+        else
+        {
+            t.setDateTime(asDate(v.get("day") + "." + v.get("month") + "." + v.get("year")));
+        }
+
+        t.setCurrencyCode(v.get("currency"));
+        t.setAmount(asAmount(v.get("amount")));
+    }
```
Nirus2000, line 1270: The source code is also inconsistent here.
Either we apply the determination of the year to all transaction extractions or to none.

Furthermore, we should only create an "auxiliary class" that correctly reflects the name of the class and what it does. This means that if we use the auxiliary class, we should not use "date" and "amount" in the name and apply them. These are different treatments.

Remove the "currency and amount extraction" and leave them as they were and use/rename the function
"setDateAndAmountForAccountStatement" to "DateTransactionHelper"
If these are then applied to all transactions.

Translated with DeepL.com (free version)

Nirus2000 (changes requested):
Hello @ZfT2 
I have looked at your changes and commented on them. Please take a look at them.

Greetings
Alex

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
ZfT2: OK, but should I use the already existing class "ExtractorUtils" instead of creating a new one "DateTransactionHelper"?

Nirus2000: The "ExtractUtils" are global and are used across all importers.
So we would create the auxiliary class in this importer, because its here.

ZfT2: you mean as inner class? I think then we can still use one additional (renamed) method...

Nirus2000: No, I think we're talking at cross purposes.
The function:
- rename "setDateAndAmountForAccountStatement" to "DateTransactionHelper"
- Remove the amount and currency from the function
- Replace the source code with the function

ZfT2: > No, I think we're talking at cross purposes. The function:
> 
>     * Rename "setDateAndAmountForAccountStatement" to "DateTransactionHelper"

"DateTransactionHelper" sounds for me lik a static helper class. Or do you really mean a method named
"private void DateTransactionHelper(AccountTransaction t, ParsedData v)" ? Looks strange for me. 

>     * Remove the amount and currency from the function

no problem...

>     * Replace the source code with the function

is also clear to me.

Nirus2000: C'mom... Do we really want to make it that complicated?
- Rename "setDateAndAmountForAccountStatement" to "DateTransactionHelper"
---> `private void DateTransactionHelper(AccountTransaction t, ParsedData v)`

If you look at other importers, such as Comdirect, we handle it the same way there. The name reflects exactly what the function is supposed to do.

Well then... Let's go 👍🏻 
Alex

ZfT2: Sorry but I've never seen a method named starting with capital letter and ended with *Helper... Also not in the Comdirect Importer. This will give an SonarLint warning. 
But new proposal is committed now :)

Nirus2000:
"setTransactionDate" ... Seriously?

Are you setting a date? No, you determine the date of the transaction using the strings that are available.
It is set in the wrap class. It is better to name the classes what they do... 

So "helper class"... "dateTranactionHelper" ...

Nirus2000 (approved):

ZfT2:
> "setTransactionDate" ... Seriously?
> 
> Are you setting a date? No, you determine the date of the transaction using the strings that are available. It is set in the wrap class. It is better to name the classes what they do...
> 
> So "helper class"... "dateTranactionHelper" ...
> 
> Easy, or what you think?! :-D
> 
> Write "Helper"... in the searchbar in Eclipse an you will find other "helper"-classes... ;-)


Helper _classes_ of course but helper _methods_? I don't see any other helper _method_ in the whole source code, but OK.... maybe matter of taste :)

Issue #3780, opened 2024-02-08
French numpad emiting a dot as a decimal symbol.


DaveInDev:
On my french keyboard, when I use the numpad to type a decimal symbol, I get a dot "." and the software is unhappy because it seems that in the french settings, it's expecting a ",".
But then if I change the language to english, the numpad still emits a dot ".", as you expect in english, and there is no more problems...

**Expected behavior**
When language is set to french,the numpad decimal should be interprated as a ",".

**Desktop (please complete the following information):**
 - Windows 10 x64
 - Version 0.67.3

mierin12:
Hello @DaveInDev , I do not have a solution, however it looks similar to this : https://github.com/portfolio-performance/portfolio/pull/2826
Would the French + Belgian setting be similar to your need ? I am actually now using this setting myself as it is indeed rather practical.

I have quickly checked, in such Belgian setting, a dot is accepted as a decimal separator in _most_ (but not all) locations:

- [x] Deposit/ Removal / Interest / Interest Charge / Fees / Fees Refund / Taxes / Transfer : comma and dot are accepted as decimal separator 

- Buy/Sell : 
- [x] Share / Gross value / Fee /Taxe: comma and dot are accepted as decimal separator
- [ ] Quote : Only comma is accepted as decimal separator

- Dividende : 
- [x] Share / Gross value /Fee / Taxe : comma and dot are accepted as decimal separator
- [ ] dividend payment per share / dividente par part : Only comma is accepted as decimal separator
- Taxonomy:
- [ ]  Weight/Allocation : Only comma is accepted as decimal separator
- Security Historical Quote 
- [ ] direct modification : Only comma is accepted as decimal separator
- [x] Add.. : comma and dot are accepted as decimal separator

![2024-03-01 19_28_01-Portfolio Performance](https://github.com/portfolio-performance/portfolio/assets/160436107/c4783939-2913-4ecd-b138-d31bdd212546)
![2024-03-01 19_30_08-Portfolio Performance](https://github.com/portfolio-performance/portfolio/assets/160436107/cb082376-3f4c-4576-b0b6-6508b05dcefb)
![2024-03-01 19_34_35-Portfolio Performance](https://github.com/portfolio-performance/portfolio/assets/160436107/7b270c28-75d7-4ab8-bf4a-24a4a53c04a0)
[...]

PR #4143, opened 2024-07-12
allow '.' from numpad as French decimal separator

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/3780
Issue : https://forum.portfolio-performance.info/t/decimal-separator-on-french-keyboard/28939

Hello, this is a proposition to deal with the numpad decimal on French keyboard, which is a `.` and not the `,` official French decimal, therefore currently not accepted by PP as an input.

Notes :

- This is only dealing with the numpad `.`, and not dealing with the other `.` accessible under `SHIFT+;` but I do not see it as an issue.
- ~There are still some locations in PP where numpad `.` is not allowed, e.g. when manually editing a security Historical Price. Which I think is to be dealt into `NumberVerifyListener`. I am not sure yet what would be the clean way to do it there, but I think updating Historical Price manually happens a bit less often than creating/editing Transactions.~ -> done in second commit

edit: I am having second thought about the country condition `"FR".equals(Locale.getDefault().getCountry())`. This means it is only working for fr_FR, while if only the language is set but the country FR is not specified, it is not working. Would it be better to be more general and check for `Locale.getDefault().getLanguage()` instead ? On the other hand, there are so many possible countries possible under fr_XX, for which I have no idea if they should be considered.

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/utils/StringToCurrencyConverterTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/utils/StringToCurrencyConverterTest.java
@@ -145,6 +145,31 @@ public class StringToCurrencyConverterTest
         converter.convert("1.234,56");
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidFRfrAmount()
+    {
+        Locale.setDefault(new Locale("fr", "FR"));
+        StringToCurrencyConverter converter = new StringToCurrencyConverter(Values.Amount, false);
+        converter.convert("1.234,56");
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidFRfrAmount2()
+    {
+        Locale.setDefault(new Locale("fr", "FR"));
+        StringToCurrencyConverter converter = new StringToCurrencyConverter(Values.Amount, false);
+        converter.convert("123.45");
+    }
+
+    @Test
+    public void testValidFRfrAmount()
+    {
+        Locale.setDefault(new Locale("fr", "FR"));
+        StringToCurrencyConverter converter = new StringToCurrencyConverter(Values.Amount);
+        assertThat(converter.convert("12,34"), is(1234l));
+        assertThat(converter.convert("1 234,56"), is(123456l));
+    }
+
     @Test
     public void testValidArithmetic()
     {
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractTransactionDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractTransactionDialog.java
@@ -93,7 +93,7 @@ public abstract class AbstractTransactionDialog extends TitleAreaDialog
             // on French keyboard, the numpad decimal is '.'
             // which should be catched and replaced by a ',' as
             // decimal separator
-            if ("FR".equals(Locale.getDefault().getCountry()) || "BE".equals(Locale.getDefault().getCountry())) //$NON-NLS-1$ //$NON-NLS-2$
+            if ("FR".equals(Locale.getDefault().getCountry())) //$NON-NLS-1$
             {
                 char decimalSeparator = new DecimalFormatSymbols().getDecimalSeparator();
                 value.addKeyListener(new KeyAdapter()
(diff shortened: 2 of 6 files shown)


Nirus2000:
Hello @mierin12 
maybe helps to look [here](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/StringToCurrencyConverter.java)  with pull request #2826 

Regards
Alex 👍🏻

mierin12:
Hello Nirus, thanks for tips, I had seen this Pull request, but I thought it was incomplete : I had listed the locations where it was working and where it wasn't working in this comment : https://github.com/portfolio-performance/portfolio/issues/3780#issuecomment-1973750976

So here I am trying a different approach: to detect when numpad decimal is pressed, and use the official decimal separator instead. With the second commit, I think it now covers the Quote, Taxonomy's weight, and Attribute. I have added belgium country to this PR too.

Nirus2000:
Hello,
but we should add [testcases](name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/utils/StringToCurrencyConverterTest.java)... I think some changes in Java 22 are possible.

For example, we already know that there are changes in the date, but that's another topic.

I have the thought that if a user with FR has a US keyboard, then the problem continues... I think a different approach is needed, where we need to determine whether a dot or a comma is used as a decimal separator.

mierin12:
Hello,
I have added some test cases in the stringtocurrencyconvertertest, is this the kind of tests which were expected ? Maybe it makes it more robust, but to be honest I do not think that my new tests are so useful given that this PR is not changing the stringtocurrencyconverter. I assume meaningful tests would be 

- to test that the replacement of numpad decimal is done ONLY for number input fields, and not for other input (text). I do not know how to unit test that.
- in the number input fields case, that the replacement occurs correctly. I do not know either how to unit test a keyboard listener action, I guess first we would need to factorize the associated code (it is quite duplicated at the moment), and then test a single method.

> ### Question:
> 
> I have the thought that if a user with FR has a US keyboard, then the problem continues... I think a different approach is needed, where we need to determine whether a dot or a comma is used as a decimal separator or group separator.

I do not understand the concern. On US keyboard, isn't the numpad decimal use for decimal separator ? If so, this PR will permit the US user to keep using it as a decimal separator even in a PP french setting. Are people typing manually the thousands separator ? If yes is the numpad decimal key used to type thousands separator instead of decimal separator ? To get 1 234,56 I never need to manually type the space.

buchen:
Thanks @mierin12 for the change.


> I have added some test cases in the stringtocurrencyconvertertest, is this the kind of tests which were expected ? Maybe it makes it more robust, but to be honest I do not think that my new tests are so useful given that this PR is not changing the stringtocurrencyconverter.

I agree. The pull request is changing the character added to the text field - and then there is no change in parsing the number value.

If we add a test case, then for the listener itself. However, I right now we have no test cases that run SWT code (because it requires a Display) and I do think we should add them now.

BTW, personally I also cannot test this, because I do not have a keyboard with numpad.


> Would it be better to be more general and check for Locale.getDefault().getLanguage() instead ? On the other hand, there are so many possible countries possible under fr_XX, for which I have no idea if they should be considered.

I would go for the language.


I have two comments:
* first, there is a lot of code duplication. Let's refactor this into one helper class
* second, for the NumberVerifyListener - is there really a special handling necessary or can't we just rely on the KeyListener

I refactored the code little bit and pushed it as PR #4178
@mierin12 can you be so kind and test it on your French keyboard?

buchen:
merged with #4178

catrucho:
I am Spanish from Spain and I believe I have exactly the same problem as the Frech colleague.
I am trying to input decimals from my numeric keypad that should be comas but a dot is entered and I am forced to use numeric keypad to enter the first pad of a number, then go to the letter keyboard to press the comma and come back to numeric keypad to enter the decimal part.
I tested to change my preferences to french and the problem is solved.
I wonder if a more broad solution might be created in the preferences language tab to fix this problem to more people, or at least make the same solution to spanish keyboards.
Thanks a lot

mierin12:
hello @catrucho , I think it can be done quite easily. One question : 
"." is the thousands separator
Capture d’écran 2025-02-01 003322
You never use the numpad "." to type the thousand separator, right ? It just visually appears when you type more than 4 digits ? In French I never type the "space", for example.
To be sure that people do not use the numpad "." to write the spanish thousands separator that the fix would bother.

mierin12:
For example, currently "1.234" is allowed by PP as a Spain number and will be :  "1234". With the fix, "1.234" will be "1,23". So if people were using the first way, it is a change for them. (In French "1.234" was not allowed at all)

catrucho:
Yes absolutely, you never write the thousand separator, the system does it for you as you said. Nobody would write that. I think it is the same exactly as french.
If I want to write 1234 I would type that and the system ddisplay 1.234 as you said. You would only push the "." from the numpad if you want to write a decimal number.
Thanks for your help.

mierin12:
@catrucho : it should be now available in the new version 0.74.1.

catrucho:
I saw it already, working like a charm, I am extremely grateful for your help.

PR #4146, opened 2024-07-15
modifies dkb importer to support new transaction

christen90:
modifies dkb importer to support taxes for interest payed on tagesgeld account


Nirus2000:
Good job 👍🏻

PR #4147, opened 2024-07-16
Onvista PDF Extractor: reworked changes, added new transactions

ZfT2:
- reworked change transactions
- added account transactions
- added statements from forum thread: https://forum.portfolio-performance.info/t/pdf-import-von-onvista-bank/2076/145


Nirus2000:
Hello @ZfT2 

As in all PDF importers, CSV importers and in the manual dialog, it is not intended to post transactions with an amount of 0.00 as a transaction in Portfolio Performance.

It is also not possible to edit these transactions in the transaction dialog, nor does it make sense to post an amount of 0.00.

Furthermore, we do not support reversals. We have added these dialogs for this purpose, which you now want to remove. 

For a split & reversal etc. are treated as a failure.
- MsgErrorTransactionTypeNotSupported
- MsgErrorOrderCancellationUnsupported
- MsgErrorSplitTransactionsNotSupported

However, this is the wrong way and your pull request will not be merged into the master. 

We therefore set the pull request to draft until the adjustments have been reversed and the new transactions have been correctly completed with the tests.

ZfT2:
> As in all PDF importers, CSV importers and in the manual dialog, it is not intended to post transactions with an amount of 0.00 as a transaction in Portfolio Performance.

OK I will check this.
 
> Furthermore, we do not support reversals. We have added these dialogs for this purpose, which you now want to remove.
> 
> For a split & reversal etc. are treated as a failure.
> 
>     * MsgErrorTransactionTypeNotSupported
> 
>     * MsgErrorOrderCancellationUnsupported
> 
>     * MsgErrorSplitTransactionsNotSupported

I know the dicussion about reversals. But my point of view is, to let the user decide if they want to import all transactions (original and correction) or delete/modify the original (wrong) booking. This PR is based of hundreds of PDF documents from my long term depot and so I think it is really usefull to enrich the importer with this fuctionality to support bigger bulk imports. (and save a lot of manual work)

For splits, if they really took place in the depot, why should we not import them?

Nirus2000:
> I know the dicussion about reversals. But my point of view is, to let the user decide if they want to import all transactions (original and correction) or delete/modify the original (wrong) booking. This PR is based of hundreds of PDF documents from my long term depot and so I think it is really usefull to enrich the importer with this fuctionality to support bigger bulk imports. (and save a lot of manual work)

Yes, and we have decided that transactions will not be imported via the automatic PDF import. Therefore, the tests are marked as “Failture” and the user can book them manually.

Nirus2000:
A transaction with 0.00 is not possible and is therefore not allowed in the automatic import. It is also not possible to subsequently edit the transaction for an amount of 0.00.
grafik
grafik

ZfT2:
> A transaction with 0.00 is not possible and is therefore not allowed in the automatic import. It is also not possible to subsequently edit the transaction for an amount of 0.00. 

If I see right, there are only Inbound- and Outbound Deliveries with the amount of 0,00, this should be possible since longer time, exactly since [PP Version 0.27](https://forum.portfolio-performance.info/t/neues-nennenswertes/23/21). And not deposits and sales with 0.

PR #4149, opened 2024-07-17
add Savings (sum of performance-neutral transfers) to Statement of Assets chart

mierin12:
Issue : https://forum.portfolio-performance.info/t/aggregierte-einlagen-entnahmen-im-diagramm-vermoegensaufstellung/13147

Hello,

This is a proposition to add a new common data series in the Statement of Assets graph : the accumulation of performance-neutral transfers. This value is already available in the performance calculation ("Performance neutral transfers"), and in the widget "performance-neutral transfers", https://github.com/portfolio-performance/portfolio/pull/2381. This PR is proposing to see it over time in a graph.

This is similar to the "Invested capital for the current period" except that here the initial value of the portfolio is not taken into account. I believe users sometimes do not always understand why the "Invested capital for the current period" does not start at 0, so this would be this new common data. 

I named it "Saving" by default because this data is named this way in the Payment section, but maybe "Transferals (cumulative)" is good too ?, or something else ? This is the hardest part of the PR :)
Capture d’écran 2024-07-17 225931

In the similar previous addition of common data serie, https://github.com/portfolio-performance/portfolio/commit/ffe60347971e8a90ee692aa9339f02fe1a71a95a, I see that Client and Clientfactory were updated, but as I do not understand it, I did  not reproduce it.

(while checking it against the existing perf calculation, savings and transfer widget, I noticed that :

- Day 1 of reporting period is not taken into account for the calculation (ok, as explicited by the "excl" of the date menu, real start is end of Day 1), ok but
- Day 1 of reporting period IS taken into account in the statement of assets graph, so any plotted cumulative common data with a non-zero value in Day 1 will show a different final value that their performance/widget calculation for the same reporting period)


buchen:
> In the similar previous addition of common data serie, https://github.com/portfolio-performance/portfolio/commit/ffe60347971e8a90ee692aa9339f02fe1a71a95a, I see that Client and Clientfactory were updated, but as I do not understand it, I did not reproduce it.

In the other commit, I also changed the meaning of the existing data series. The "taxes" used to denote the "accumulated" series. When loading the file, this is fixed (so that "taxes_accumulated" denotes the accumulated taxes). Due to this change, I also had to update the client version. In this case, as we are only adding a new data series, that is not needed

buchen:
Rebased and merged. I took the liberty to rename the enum value (for consistency) and rename the label (because "LabelSavings" is used slightly different in the Payment view).

Sorry for taking that much time.

PR #4153, opened 2024-07-20
Fix: Problem with JSON provided dates which are given by strings containing unix timestamps

kingbtcvl:
I recently came across the following small issue:

When trying to get historic data with a JSON provider a problem can occur that gives an output like
`Text '1721433600' could not be parsed at index 0`.
The problem appears when the JSON path to the date yields a string that contains the date in form of a unix timestamp.

I guess this PR will fix this problem.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
@@ -344,7 +344,16 @@ public class GenericJSONQuoteFeed implements QuoteFeed
             return LocalDate.parse(object.toString(), formatter);
         }
         if (object instanceof String s)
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
```diff
@@ -344,7 +344,16 @@ else if (object instanceof String s)
             return LocalDate.parse(object.toString(), formatter);
         }
         if (object instanceof String s)
-            return YahooHelper.fromISODate(s);
+        {
+            try
+            {
+                return parseDateTimestamp(Long.parseLong(s));
```
buchen, line 350: In general I agree that we should try to parse the string also as a timestamp.
However, I propose to reverse this - because the "normal/default" behavior is that the string is actually an ISO date.

buchen:
merged

PR #4157, opened 2024-07-21
Improved label display for security chart diagrams

Nirus2000:
These changes improve the readability of the chart labels by ensuring that labels are not drawn on top of each other and are only in non-negative positions. This should contribute to a clearer presentation of the data.

In pull request #4120 we have the option to show and hide the individual labels and this prevents overlapping.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
```diff
@@ -1396,15 +1406,13 @@ private void addDividendMarkerLines(ChartInterval chartInterval)
...
-                    FontMetrics fontMetrics = event.gc.getFontMetrics();
+                    Color defaultForeground = Colors.theme().defaultForeground();
+                    event.gc.setForeground(defaultForeground);
 
                     // Three levels of the label
                     int[] labelExtendX = new int[3];
 
                     int symbolSize = border.getSymbolSize();
-                    int labelHeight = fontMetrics.getHeight();
-                    int halfLabelHeight = labelHeight / 2;
```
buchen, line 1406: Just out of curiosity, was there a specific reasons why the fontMetrics did not work?
I think the idea was that the label height never changes.
(But the current code is also okay because we anyway need to get the text extend)

buchen:
I merged the change. Thanks!

In general, I think we need to rethink our strategy with the labels.

Maybe instead of trying to place each series of labels separately, we should collect **all** labels first. And then have one place that is placing all labels. Apparently, there are some [label placing algorithms](https://en.wikipedia.org/wiki/Automatic_label_placement) we could think about. And then maybe it needs some priority (say purchase is more important to display than max).


That could avoid situations like this:
Bildschirmfoto 2024-07-28 um 08 05 55

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
Nirus2000: All labels in the Security Chart are created without fontMetrics, so it is more consistent if we keep this as long as it has no effect. 😎

PR #4158, opened 2024-07-21
add Drawdown chart widget

mierin12:
Issue : https://forum.portfolio-performance.info/t/drawdown-chart-als-widget/2523

Hello, this is a proposition to add a new widget related to Drawdown : this is a chart of the drawdown over time ("underwater" if this is the right word). I initially tried to add this in the performance chart, but I found that it was easier to implement as a widget. 
The widget is inspired from the existing `ChartWidget`.

I am thinking maybe to add an option to choose between "auto Y range" and a fixed "0 to -100%" range, which would help to compare different data ?

Capture d’écran 2024-07-21 130236


buchen:
What a great addition! 👍 

I have slightly edited the translated labels, but other than that I just merged it.

I hope to be able to release a new version later today.

PR #4178, opened 2024-08-02
Allow '.' from numpad as French decimal separator

buchen:
When entering a number of a French keyboard, the dot from the numpad is usually automatically converted into the French decimal separator ','.

We do not have to change the NumberVerifyListener. It is not triggered because the KeyListener is directly inserting the decimal. However, that is not a problem because we know it is a valid character for numbers.

Issue: #4143
Issue: #3780
Issue: https://forum.portfolio-performance.info/t/decimal-separator-on-french-keyboard/28939

[squashed commits; refactored code into utility class; rebased to master]


buchen:
Superseeds #4143

mierin12:
>  @mierin12 can you be so kind and test it on your French keyboard?

@buchen  Many thanks ! I have tested the branch feature_french_keyboard_support; it works like a charm !!
Transaction dialog : ok
Attribute (including Limit Price attribute) : ok
Taxonomy weight : ok
Cotation : ok

Issue #4135, opened 2024-07-06
Inconsistency with "source" attribute


pfalcon:
An idea to have a field which identifies a source/producer of some entry in the PP data model is very useful one. And the first impression is "how smart PP to already implement that!". However, closer look shows that it's more of an unloved one, due to issues like https://github.com/portfolio-performance/portfolio/commit/0617a9916d32b4d5d5cb36d6dd1d5b428b8345d0 .

Another sad case I'd like to share here. So, grepping thru models package, one can rejoice that SecurityEvent has a "source" attribute. But closer look shows that it's only for DividendEvent subclass! How could that be? Now, fixing that would be trivial, if not for [proprietary binary formats](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ProtobufWriter.java).

Just to prove that there're *inconsistency* the the "source" attributes, let's look what else has them. 
AttributeType has them, but why? Code like https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientSettings.java#L112 can only perplex. Like, source of "ter" attribute is ... "ter"? Really?

By common sense, what requires "source" is a **value**, not type. For example, you wrote a script which uses Site1's API to automatically query TER. Site1's API perhaps handles only subset of securities, and if you queried it once, you may want to re-query it again in the future. So, for values you actually set from Site1, you set source of "site1", and then during update operation, touch only values tagged like that (not touching values set by another script from Site2, and not blindly overwriting values manually entered by user (== empty source)).

Now next, Taxonomy also has "source", and one can again wonder why. By logical sense, even if some taxonomy is treated as "value", i.e. may be updated by automatic means, it can be done by the name of [root of] taxonomy. E.g., "Taxonomy from Site1". Everything under it would need to be synced with Site1.  But yeah, I get the idea that "everything under" requires a recursive search, so having a "source" on taxonomy entries may be an optimization to avoid that.

[...]

PR #4181, opened 2024-08-04
SecurityEvent: Move "source" field to the parent SecurityEvent class

pfalcon:
Previously, "source" field was available only in the DividendEvent subclass. But it makes sense to have it on the parent class, to streamline integration with external tools to manage events. Thus, "source" is a generic machine-readable field to record source and/or "ownership" of the event information. A typical workflow would be:

1. An external tool TOOL1 creates an event and sets its source attribute to "TOOL1" to designate this event is "owned" by the tool, which it's free to modify or even delete later.
2. This attribute eventually gets rendered as <source> element within <event> element in PortfolioPerformance XML.
3. PP preserves the value of the source attributes across load/save cycle of the portfolio.
4. On the next run, TOOL1 can identify events originally create by itself, and update them as needed. This is important, because event information provided by external source if often tentative and subject to change.

As mentioned above, the "source" field is intended to be machine-readable, and as such, it's currently now shown anywhere in the PP UI. It might be useful to address that at later time, if the need for that is found.

WARNING:
This change doesn't include steps which would be required to migrate Protobuf binary format. It's expected that developers familiar with this format would step in to help with the migration.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/SecurityEvent.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/SecurityEvent.java
@@ -37,7 +37,6 @@ public class SecurityEvent
     {
         private LocalDate paymentDate;
         private Money amount;
-        private String source;
 
         public DividendEvent()
         {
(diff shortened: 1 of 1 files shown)


pfalcon:
This is attempt to move forward to address some issues from https://github.com/portfolio-performance/portfolio/issues/4135 . As the commit message says, this doesn't touch Protobuf related code, and doesn't include migration steps which are almost certainly would be required for the Protobuf files.

buchen:
merged.

pfalcon:
Thanks for handling the ProtoBuf part!

Issue #4182, opened 2024-08-04
Security ticker symbol is not available in various places in the app


pfalcon:
I guess it's fair to say that quite many people are more familiar with the security symbols rather than name. To start with, some companies have symbols rather different from their names. But real issue is with funds (ETFs) - their names are oftentimes mouthful, regalia- and buzzword-ridden. So again, their tickers are more familiar to many people than their full names.

Yet, there're a number of places in PP UI where symbol is not shown, mostly in the Reports section.

One example already fixed https://github.com/portfolio-performance/portfolio/issues/3292, but I'd like to address this consistently throughout the app. Initial looking shows that there's no single place to fix it, so multiple patches will be required, to be done over time. So, I open this umbrella ticket to reference from individual patches.

The overall plan would be to address all of these places:

* Input elements. The idea is to use format "SYMBOL (Security Name)" instead of current "Security Name"
* Chart labels - likewise.
* For tabular data, introduce a separate column, to allow independent sorting on symbols vs name.

pfalcon:
Copying some discussion from https://github.com/portfolio-performance/portfolio/pull/4183 here:

From @buchen : 

---
the change is small - from a coding perspective - but has a big impact on the UI.

I agree there is value in improving the display of security names. But I do not think that the ticker symbol is the right way to go. At least not the way for everybody. In Germany for example, I don't think that the ticker is commonly used. Personally, I only configure it if I need it to retrieve historical prices from Yahoo Finance and similar sites. Adding the ISIN is then an option, but I do not think I want to force this on all users.

If you are very attached to the feature (and don't want to maintain it in your local branch), then I would be willing to merge a "configurable version" of this feature. Somehow along this user story.

> As a I user, I want to customize the display string for instruments used in dropdown boxes, and chart legends. I can go to the settings and configure the default display string of an instrument. I can choose to prefix it with ISIN, WKN, or Ticker (or maybe any other attribute configured for instruments).

What do you think?

---

pfalcon:
@buchen : 

> What do you think?


I'd like to start with addressing this:

> and don't want to maintain it in your local branch

I effectively a personal fork of PP with close to 100 commits on top of master. Of course, most of them aren't of the quality suitable for mainline. Now that I'm sure there's nothing around coming even close to PP in functionality (so it does make sense to invest more effort into it vs something else), I'd like to do motions to flush at least something to mainline (or find other reasons for not doing so besides just my laziness).

[...]

PR #4183, opened 2024-08-04
Security: Make toString() be "ticker (name)"

pfalcon:
For many people, ticker is more familiar than security name/title (this is especially true for funds). toString() is used e.g. in dropdown boxes for security selection.

Partiall addresses #4182

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java
@@ -880,7 +880,12 @@ public final class Security implements Attributable, InvestmentVehicle
     @Override
     public String toString()
     {
-        return getName();
+        if (getTickerSymbol() == null)
+            return getName();
+        else if (getTickerSymbol().equals(getName()))
+            return getTickerSymbol();
+        else
+            return getTickerSymbol() + " (" + getName() + ")";
     }
 
     public String toInfoString()


pfalcon:
Updated to handle null tickerSymbol.

@buchen : Seems like an easy patch to review? ;-)

buchen:
Hi @pfalcon,

the change is small - from a coding perspective - but has a big impact on the UI.

I agree there is value in improving the display of security names. But I do not think that the ticker symbol is the right way to go. At least not the way for everybody. In Germany for example, I don't think that the ticker is commonly used. Personally, I only configure it if I need it to retrieve historical prices from Yahoo Finance and similar sites. Adding the ISIN is then an option, but I do not think I want to force this on all users.

If you are very attached to the feature (and don't want to maintain it in your local branch), then I would be willing to merge a "configurable version" of this feature. Somehow along this user story.

> As a I user, I want to customize the display string for instruments used in dropdown boxes, and chart legends. I can go to the settings and configure the default display string of an instrument. I can choose to prefix it with ISIN, WKN, or Ticker (or maybe any other attribute configured for instruments).

What do you think?

(Not closing this PR as to have this discussion visible)

pfalcon:
@buchen :

> What do you think?

Thanks for the detailed reply. I started to write which turned out to be even more comprehensive reply, so I decided to to move it to parent ticket, https://github.com/portfolio-performance/portfolio/issues/4182 , because the idea is to indeed consistently change UX across PP, and this patch is just first step of it.

buchen:
Closing this one. Continuing the discussion in #4182

PR #4211, opened 2024-09-01
Security: Allow toString() be (in general) "ticker (name)"

pfalcon:
For many people, ticker is more familiar than security name/title (this
    is especially true for funds). toString() is used e.g. in dropdown boxes
    for security selection. So, if a corresponding preference is enabled by
    user, include ticker in the display string.
    
    If ticker symbol is not available, or matches the security name, toString()
    will return just "name", as before.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1151,6 +1151,8 @@ public class Messages extends NLS
     public static String PrefLabelDisplayPA;
     public static String PrefLabelEnableExperimentalFeatures;
     public static String PrefLabelNote;
+    public static String PrefLabelPreferSecuritySymbol;
+    public static String PrefLabelPreferSecuritySymbolTooltip;
     public static String PrefLabelProxyHost;
     public static String PrefLabelProxyPort;
     public static String PrefLabelQuoteDigits;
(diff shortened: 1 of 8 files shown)


pfalcon:
This is updated version of https://github.com/portfolio-performance/portfolio/pull/4183, based on (ongoing) discussion in https://github.com/portfolio-performance/portfolio/issues/4182 . A checkbox preference `Prefer 'symbol' to identify securities` is added, by default the behavior is the same as before, but if the pref is checked, `SYMBOL (name)` will be used.

pfalcon:
Example of security selection dropdown before the setting activation:
Screenshot at 2024-09-01 16-15-27

The setting activated:
Screenshot at 2024-09-01 16-15-48

Example of security selection dropdown after the activation:
Screenshot at 2024-09-01 16-16-19

pfalcon:
@buchen : What do you think about this solution?

buchen:
> What do you think about this solution?

I am just looking at it.

I'll push an alternative implementation that
a) stores the property in the file (in order to re-use it across installation and on the mobile app)
b) allows to pick other identifier as well (because ticker is a special case)

I am not done yet. For example the sorting is not correct yet (see also your screenshot above).
And the label and translations are missing.

But as I will break for lunch soon, I will push a draft pull request and would appreciate if you can test it.

pfalcon:
@buchen : 

> But as I will break for lunch soon, I will push a draft pull request and would appreciate if you can test it.

Thanks! Commented at https://github.com/portfolio-performance/portfolio/pull/4265#issuecomment-2392887069

Issue #1499, opened 2020-04-18
Allow manual updating of online quotes on several equities simultaneously


KopolJunam:
EDIT: OBSOLETE? I just saw that there is an online menu where one can do that what I want, so maybe adding this to the context menu is not really necessary, or is it?

As the users of the VIAC/CS quote feed know the Credit Suisse server sometimes doesn't answer and it takes several attempts or it has to be done at a non busy time (evening/weekend). This can be done easily by adding the quote update to the menu when there are multiple equities selected. At the moment a manual update is only possible on a single equity.
A keyboard shortcut could be welcome too for the poor VIAC users who sometimes have to try several times.

Btw, I tried several approaches to mitigate this problem but nothing really works so far. A comfortable manual updating feature might do the trick for most or at least for now.

@buchen I would implement this myself as it's probably a piece of cake if it's ok with you.

buchen:
There is the menu entry, yes, but even if you select multiple securities, it updates only one.

The reason is that my "central selection" can only have one security. This is not easy to change unfortunately.

What is probably much easier to enhance the context menu of the SecuritiesTable to also show the option "Update online quotes" if multiple securities are selected. Feel free to submit a change.

pfalcon:
> Allow manual updating of online quotes on several equities simultaneously

https://github.com/portfolio-performance/portfolio/pull/4212 implements just this.

mierin12:
Done on [release 0.71.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.71.0) with https://github.com/portfolio-performance/portfolio/pull/4212

PR #4212, opened 2024-09-01
Allow to update quotes for several selected securities 

pfalcon:
Some views in the app allow to select multiple securities, so allow to
perform operations on all of them. Without such support, behavior can
be confusing, for example, currently a user can select multiple securities
in "Securities" view, and choose "Update quotes (selected)" via main menu,
but only one will be actually updated.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UpdateQuotesHandler.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/UpdateQuotesHandler.java
@@ -34,8 +34,8 @@ public class UpdateQuotesHandler
             if ("security".equals(filter)) //$NON-NLS-1$
             {
                 selectionService.getSelection(client).ifPresent(s -> {
-                    new UpdateQuotesJob(client, s.getSecurity()).schedule();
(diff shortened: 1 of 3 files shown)


pfalcon:
@buchen : I believe in one of your comments here in the issues I saw your comment (was awhile, can't point to now) that supporting multiple selection is somehow challenging. Caught my attention because was one of the initial changes I did. This adds infra for multiple selection, and then specifically implements it for "Update quotes (selected)". Going thru the entire app and making sure it's consistently supported is a different matter of course. Probably why it sat in my fork for awhile. I guess it's better to land this initial support, then none at all.

pfalcon:
> that supporting multiple selection is somehow challenging.

Here it is: https://github.com/portfolio-performance/portfolio/issues/1499#issuecomment-616767156

buchen:
I agree. Not sure what made me write the comment 🤔

myrdd:
Thanks for the PR. I think we should update the `command.updateQuotesCurrentSecurity.name` text from singular to plural (”selected securities“):

https://github.com/portfolio-performance/portfolio/blob/bc41cc672cacafc2a1761d3ec1f39838aba4512e/name.abuchen.portfolio.bootstrap/OSGI-INF/l10n/bundle.properties#L75

This also needs to be changed in all translations.

PR #4214, opened 2024-09-03
adds ingoing, outgoing transfer and mark interest bookings in AccountStatements as NonImportable

christen90:
Hi together,
this PR adds incoming, outgoing transfer
It also marks german interest payments in account statement as nonimportable
reason: tax amount is not included in account statements. interest transactions can be imported via specific document.
In my case there were two bookings (Geldmarkt Dividende and Cash Zinsen) in Abrechnung-Zinsen-document.
Importing both (Kontoauszug and Abrechnung Zinsen) was causing that Booking gets imported twice. Once the seperate dividend and interest with the exact amount of tax and the other one from account statement including just the sum. Resulting in a wrong saldo at the end.

BR Christian


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
```diff
@@ -2085,9 +2110,19 @@ private void addAccountStatementTransaction_Format02()
                             t.setDateTime(asDate(v.get("date")));
                             t.setAmount(asAmount(v.get("amount")));
                             t.setCurrencyCode(asCurrencyCode(v.get("currency")));
+                            v.getTransactionContext().put(FAILURE, Messages.MsgErrorTransactionTypeNotSupported);
```
Nirus2000, line 2113: Hello @christen90 
Unfortunately I don't see a test for this change (FAILURE), why this line is added.
Can you add a test case that makes this change plausible?

Regards
Alex

Nirus2000 (changes requested):

Nirus2000:
Hello @christen90 
I understand the reason, but we should issue a different message here, which could possibly also be used for other PDF importers.

**I could imagine the following message:**
```
MsgErrorTransactionAlternativeDocumentRequired = "Die Buchung wird mit diesem Dokument nicht verarbeitet. Ihre Bank oder Ihr Broker stellt ein anderes Dokument dafür zur Verfügung."

MsgErrorTransactionAlternativeDocumentRequired = "This document does not process the booking. Your bank or broker will provide a different document for this purpose."
```

Please change the error message and add the missing translations via deepl.com.

Regards
Alex 😃

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
```diff
@@ -1351,16 +1351,22 @@ public void testKontoauszug05()
                         hasSource("Kontoauszug05.txt"), hasNote(null))));
 
         // assert transaction
-        assertThat(results, hasItem(interest(hasDate("2023-04-01"), hasAmount("EUR", 0.01), //
-                        hasSource("Kontoauszug05.txt"), hasNote(null))));
+        assertThat(results, hasItem(withFailureMessage( //
+                        Messages.MsgErrorTransactionTypeNotSupported,  //
```
Nirus2000, line 1355: Hello @christen90 
I understand the reason, but we should issue a different message here, which could possibly also be used for other PDF importers.

**I could imagine the following message:**
```
MsgErrorTransactionAlternativeDocumentRequired = "Die Buchung wird mit diesem Dokument nicht verarbeitet. Ihre Bank oder Ihr Broker stellt ein anderes Dokument dafür zur Verfügung."

MsgErrorTransactionAlternativeDocumentRequired = "This document does not process the booking. Your bank or broker will provide a different document for this purpose."
```

Please change the error message and add the missing translations via deepl.com.

Regards
Alex 😃

Nirus2000 (changes requested):

Nirus2000 (approved):

Issue #2570, opened 2021-12-05
Total TTWOR on "Statement of Assets"


wgebis:
**Is your feature request related to a problem? Please describe.**
Statement od Assets is the one of the greatest vie in PP. I don't know why in row where total values are presented there in no values of certain columns e.g. TTWOR, it should be pretty useful. 

<img width="277" alt="Screenshot 2021-12-05 at 23 18 56" src="https://user-images.githubusercontent.com/32294473/144766235-ea645e70-7a55-4f9a-a6a0-7f16a8af33b3.png">

mierin12:
Done since [release 0.71.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.71.0) with https://github.com/portfolio-performance/portfolio/commit/0f7eb387d416f556cef9a4429a82cde5f4373db7

PR #4217, opened 2024-09-07
Statement of Assets : add total perf for TTWROR, TTWROR p.a. and IRR

mierin12:
Closes  https://github.com/portfolio-performance/portfolio/issues/2570
Hello,

This is a proposition to add the TTWROR cumulated, TTWROR annualized and the IRR to the Statement of Assets totals and subtotals (main category of Taxonomy) lines.
I think there were debates about adding those totals for Abs Perf % or Capital Gains %, so those are not added (not defined). I do not know if those argument were also applicable to TTWROR and IRR, I am assuming not.

I have checked some cases with the results from the associated widget, looks ok.
Not tested are the "new values" : subtotal per category with a portfolio filter activated (I do not think such values previoulsy existed elsewhere as it requires a "double filter" taxonomy+portfolio).

I have kept two commits to highlight one issue to consider in the review : without the second commit it was not working properly because of java heap space issues. I do not get it anymore with the second commit, but maybe it is still not optimized enough ? Should a LazyPerfomanceIndex methodology be used, as it is done for Security ? 

Also, this is showing 0 % for the _Without Classification_ line, which seems wrong. Now that I think of it, _Without Classification_ is never available as a data series. Is performance index defined for it ?

**Before** 
Capture d’écran 2024-09-07 230019

**After**
Capture d’écran 2024-09-07 230055

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
@@ -70,6 +70,7 @@ import name.abuchen.portfolio.snapshot.AssetCategory;
 import name.abuchen.portfolio.snapshot.AssetPosition;
 import name.abuchen.portfolio.snapshot.ClientSnapshot;
 import name.abuchen.portfolio.snapshot.GroupByTaxonomy;
+import name.abuchen.portfolio.snapshot.PerformanceIndex;
 import name.abuchen.portfolio.snapshot.ReportingPeriod;
 import name.abuchen.portfolio.snapshot.SecurityPosition;
 import name.abuchen.portfolio.snapshot.filter.ClientFilter;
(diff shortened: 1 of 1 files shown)


mierin12:
> Also, this is showing 0 % for the Without Classification line, which seems wrong. Now that I think of it, Without Classification is never available as a data series. Is performance index defined for it ?

I did not manage to produce a result for the Without Classification category, so the third commit is making it returns nothing instead of a wrong 0% TTWROR or IRR.

mierin12:
Hmmm another comment. Maybe I was too naive.
- with this implementation, the performances for total and categories match the Performance Diagram and the different widgets. Good ! Or is it ? To be noted that depending on their reporting period, they can include securities no longer in the portfolio and therefore not listed in the Statement of Assets, at least not without using the time machine.
- if the definition is instead to consider the TTWROR/IRR of totals **restricted to the current holdings** (=listed in the statement of assets), then I guess this PR is not ok. 😞 

I like the first definition as it stays consistent with the values calculated everywhere else, and can give a quick and very nice overview of Securities Accounts' total performance, as the statement of assets is also used there (statement of assets to be read in this case by row rather than by column : "the last 3 years IRR of total portfolio is, well, the last 3 years IRR of the total portfolio and not the last 3 years IRR of the below lines only") 
On the other hand, for other columns such as P/L, Abs. Perf.,  I believe it is the second definition which is used, summing only the securities currently hold.

buchen:
> To be noted that depending on their reporting period, they can include securities no longer in the portfolio and therefore not listed in the Statement of Assets, at least not without using the time machine.

I have a similar issue with the "totals" row in the "Performance -> Securities" view:
* It is possible to add a filtered row (say "dividends of account X").
* Now I filter the view for account Y
* The sum is showing only showing the dividends for account X for securities which are also held in account Y
Is this what the user would expected?

Then the user can filter the rows - for example only show instruments where there are current holdings.
* The sum is showing only the sum of the now visible items


However, here I only have "sums" - simple adding of the values.

> I like the first definition as it stays consistent with the values calculated everywhere else

I tend agree.

It is also not trivial to calculate the performance using then only the instrument currently still held. We would have to write a special ClientFilter that is stripping all other transitions.

Maybe the right way is to add a tool tip that explains to the user that the calculated performance values is relative to the account and the time interval - and therefore can include the performance of instruments not held anymore.

buchen:
merged

PR #4219, opened 2024-09-09
improve label for creation of single transaction of investment plan

mierin12:
Hello,
This is a very small proposition to add a new label to be used when only 1 transaction from an investment plan is automatically created. Previously, only one label was available, therefore using plural even for 1 transaction.
 
**Before**
Capture d’écran 2024-09-09 230350

**After**
Capture d’écran 2024-09-09 230902


buchen:
Hi @mierin12,

the Java MessageFormat supports such formatting - singular and plural. One can use the choice pattern to adapt the text accordingly.

```
Created {1,choice,1#1 transaction|1<{1} transactions} for investment plan ''{0}''.
```

I'll change the text accordingly.

Issue #4224, opened 2024-09-13
[devel] OptionLabelProvider doesn't scale


pfalcon:
https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/OptionLabelProvider.java is flawed in its functionality: for example, it handles getText(), but not getToolTipText().

Instead of having a solution which will proxy-push "column option" as additisonal param to each method, IMHO it's better to have pull API, allowing labelProvider's methods to query it. Implementation-wise, needed data would be pushed into a LabelProvider instance.

After wrapping my head around differences between TableColumn and name.abuchen.portfolio.ui.util.viewers.Column (the latter should have been called ColumnTemplate or ColumnDescriptor IMHO), I came to a following solution:

* ColumnWithOptionLabelProvider class, subclassing ColumnLabelProvider and having getOption() method.
* Instances of this class are treated specially by ShowHideColumnHelper (should have been ColumnManager or something by now) TableViewerPolicy.create(), similar to already existing special handling of CellItemImageClickedListener: https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ShowHideColumnHelper.java#L297 . What's being done is injecting TableColumn into label provider instance.
* Any Column subclass with "option" support then absolutely must call with Column.setLabelProvider() with factory lambda (to ensure there're separate instances of LabelProvider attached to each concrete TableColumn, and each label provider injected a separate option object). I'd went as far as added runtime check for that (store bool regarding how setLabelProvider() and then check it when option param is parsed). For example, DistanceFromAllTimeHighColumn and DistanceFromMovingAverageColumn don't do that (which is ok now with OptionLabelProvider, but they don't support tooltips).

@buchen : Does that sound good to you or do you have better ideas? I'd try to implement this plan and convert DistanceFromAllTimeHighColumn to show a tooltip.

pfalcon:
ParameterizedColumnLabelProvider based on these ideas was merged in 
https://github.com/portfolio-performance/portfolio/pull/4225

PR #4225, opened 2024-09-13
DistanceFromAllTimeHighColumn: Add tooltip with info about ATH

pfalcon:
Show tooltip with absolute value of ATH and its date. To support tooltip,
    switch to ColumnWithOptionLabelProvider.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ParameterizedColumnLabelProvider.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ParameterizedColumnLabelProvider.java
@@ -1,0 +1,25 @@
+package name.abuchen.portfolio.ui.util.viewers;
+
+import org.eclipse.jface.viewers.ColumnLabelProvider;
+import org.eclipse.swt.widgets.TableColumn;
+
+/**
+ * ColumnLabelProvider subclass which allows to access "option value"
+ * associated with a column. This can be e.g. reporting period (and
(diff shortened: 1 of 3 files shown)


pfalcon:
Discussion about ColumnWithOptionLabelProvider: https://github.com/portfolio-performance/portfolio/issues/4224

pfalcon:
> Discussion about ColumnWithOptionLabelProvider: https://github.com/portfolio-performance/portfolio/issues/4224

Switched name to ParameterizableColumnLabelProvider per "discussion" in https://github.com/portfolio-performance/portfolio/issues/4226

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ParameterizedColumnLabelProvider.java
```diff
@@ -0,0 +1,25 @@
...
+ * there can be multiple instance of the same base column with different
+ * reporting periods set).
+ */
+public class ParameterizedColumnLabelProvider extends ColumnLabelProvider
+{
+    private TableColumn tableColumn;
+
+    public void setTableColumn(TableColumn tableColumn)
+    {
+        this.tableColumn = tableColumn;
```
buchen, line 18: **That** is a nice idea. It will simplify. I just wonder if we should throw an exception if the tableColumn is already set. Then developers get an obvious hint if they re-use label provider (instead of using the factory method).

Anyway, will merge when I am back at my development machine.

buchen:
merged.

PR #4236, opened 2024-09-17
Add "Ausschüttung Investmentfonds" to MerkurPrivatBankPDFExtractor an…

cubinet-code:
Add "Ausschüttung Investmentfonds" to MerkurPrivatBankPDFExtractor and associated test testWertpapierDividende02

New documents only seem to have a graphical company logo that the pdf text extractor is not able to read hence the BankIdentifier was changed to match the full address instead.

In addition new "Quellensteuer" was added and regexes extended to accommodate this variation of a dividend type.


cubinet-code:
Thank you Nirus!

PR #4260, opened 2024-09-30
Fix "absolute deviation" in the "delta percentage indicator"

myrdd:
`node.getRoot()` leaded to an "absolute deviation" calculated in relation to the total value of *all* (filtered) assets. The filtering worked, but the "root" TaxonomyNode includes the "Without Classification" class, which we don't want.

This commit fixes above issue, but goes further: multiple levels of asset classes can now be rebalanced correctly using the "delta percentage indicator", because only the current sub-class is taken into account, as opposed to the classification root.

Issue: https://forum.portfolio-performance.info/t/ohne-klassifizierungen-nicht-auswerten-filter-vergessen-ihre-einstellungen/655/9

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/DeltaPercentageIndicatorLabelProvider.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/DeltaPercentageIndicatorLabelProvider.java
@@ -36,7 +36,7 @@ public class DeltaPercentageIndicatorLabelProvider extends OwnerDrawLabelProvide
 
             Data data = new Data();
 
-            data.totalAmount = node.getRoot().getActual().getAmount();
+            data.totalAmount = node.getParent().getTarget().getAmount();
             data.targetAmount = node.getTarget().getAmount();
             data.actualAmount = node.getActual().getAmount();
 


buchen:
Hi @myrdd,

thanks for the contribution.

Let's keep in mind that the change only changes the color coding of the bar.

The color coding is based on the 5/25 rule ([for example see here](https://www.whitecoatinvestor.com/rebalancing-the-525-rule/)).

> The “5” portion of the rule means that if an asset allocation deviates by an absolute percentage of 5% of the portfolio then you rebalance it.  This refers to the big blocks in your portfolio. 

Now we can argue should the big blocks include the unclassified instruments or not.

For personally, that does not make a difference as I have everything classified. But I see the argument if you want to focus on a certain part of the portfolio (hence not classify other parts), the the analysis of big blocks can change. I will adapt the change to focus on the classified instruments only.

What I do not agree is that the 5% rule should apply to the parent category (as this pull request implies). The idea is to handle the big blocks in your portfolio. If you apply the 5% rule to the categories, many small things become big blocks and the indicator becomes useless.

myrdd:
> What I do not agree is that the 5% rule should apply to the parent category (as this pull request implies). The idea is to handle the big blocks in your portfolio. If you apply the 5% rule to the categories, many small things become big blocks and the indicator becomes useless.

Okay, I think you are right. :+1:  
Thanks a lot.

> Let's keep in mind that the change only changes the color coding of the bar.

FTR, it also changes the tooltip:

https://github.com/portfolio-performance/portfolio/blob/cad4b38461b7a9ec5954f92e0c127e1acf25eb16/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties#L2639

PR #4264, opened 2024-10-02
Add AccountBalancePane to StatementOfAssetsView

Mikerion:


buchen:
I like the idea of adding the account balance chart to the statement of assets.

I just wonder - instead of having two tabs - I think it makes sense to have one tab that either shows the security chart of the account balance chart.

One way to do it could be to extend ```SecurityPriceChartPane``` to have a ```StackLayout``` with one element being the price chart and the other being the account balance chart. Depending on the input (```Adaptor.adapt(XXX.class, input);``` we decide what to show.

What do you think?

Mikerion:
Yes, only one tab would be better. I wanted it too but I didn't know how to hide/switch tabs.

Mikerion:
@buchen

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/panes/SecurityPriceChartAndAccountBalancePane.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private boolean isSecurity = true;
+
+    @Inject
+    @Optional
+    public void onDiscreedModeChanged(@UIEventTopic(UIConstants.Event.Global.DISCREET_MODE) Object obj)
+    {
+        if (balancePane != null)
+            balancePane.onDiscreedModeChanged(obj);
+    }
```
buchen, line 47: This is not necessary. Because the AccountBalance registered itself for the event changes, this would result in duplicate calls.

buchen (commented):
Nice.

Thanks for tidying this up. With the one chart widget showing the right chart depending on the context it looks nice.

I remove the onDiscreedModeChanged method and renamed the classes slightly (I feld ChartPane is good enough ;-))

Merged.

PR #4265, opened 2024-10-03
Formatting: Optionally add instrument identifier as prefix to the name

buchen:
Issue: #4211


buchen:
@pfalcon

pfalcon:
From https://github.com/portfolio-performance/portfolio/pull/4211#issuecomment-2390965067 : 

> I am just looking at it.
>
> I'll push an alternative implementation that
> a) stores the property in the file (in order to re-use it across installation and on the mobile app)
> b) allows to pick other identifier as well (because ticker is a special case)
>
> I am not done yet. For example the sorting is not correct yet (see also your screenshot above).
> And the label and translations are missing.
>
> But as I will break for lunch soon, I will push a draft pull request and would appreciate if you can test it.

Thanks for looking into this! I tested this PR, it definitely would work for me, and trying other prefixes, I agree they may be useful for somebody too.

I note that this is much bigger change than mine. I plugged into Security.toString() because it allowed to cover (almost?) all security selection checkbox with a single change. With your approach, I see that each dialog needs to be updated separately, and so far only "Buy" and "Sell" seem to be covered, others like Dividend, Taxes, etc. aren't, hope you'll take care about them.

I surely noticed sorting issues with my patch either, but just trying to follow open-source software way of not trying to solve all issue at once and do incremental changes instead.

Btw, "Prefix name with..." choice in Preferences probably should be "Prefix security name with...".

So, thanks again for working on it and hope you can finish it soon!


> And the label and translations are missing.

That's something I want to ask about. I don't use Eclipse to hack on PP, doing all hacking in my favorite command-line editor and building with mvn, so I didn't do anything with translations so far. I understand that dealing with translations is an extra burden for the maintainer, which means longer processing time, and I'd like to streamline it as much as possible ;-). So, I looked at how to do it. My assumption was that there's some Eclipse plugin which, based on English properties file, uses some translation service to automatically add strings for all other languages. But https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#translations just says:

> Translate new labels into all existing languages using [DeepL](https://www.deepl.com/).

So, is the idea is to copy-paste English string into DeepL manually, then copy-paste it back? Oh, I'd say that puts quite a burden on contributors, while result would probably still be mediocre (I can appreciate that DeepL might translate short strings better than e.g. Google Translate, but still not ideal comparing to a native speaker). I'd think that just providing English strings in property files should be enough, and leaving translating them to dedicated/regular translation contributors (hopefully there're such, and they have infrastructure to track added/changed strings, like I hope the POEditor project provides).

Please let me know if you have any hints or suggestions with this matter.

buchen:
> I plugged into Security.toString() because it allowed to cover (almost?) all security selection checkbox with a single change.

I think that was a good idea. But the #toString only works if it is a global configuration which applies to all files. That's why I decided against it.

> I surely noticed sorting issues with my patch either, but just trying to follow open-source software way of not trying to solve all issue at once and do incremental changes instead.

Totally agree. The comment was more or less to myself: the todo list of what still needs to be done here.

> So, is the idea is to copy-paste English string into DeepL manually, then copy-paste it back?

To be honest, this is what I do all the time. With the resource bundle editor, there is one form. On DeepL, I select the language, copy the translation and paste it back.

Then I sync the translations with POEditor.com. The contributors there fix and polish the translations, and I sync it back.

I always thought about extending the resource bundle editor with a translation API call. The project itself is dormant. But I haven't found the time so far.


For the contributions, it is generally easier to start with only the English translation. Then I still have a chance to point out to other (existing) labels. If that changes, one has to redo the translation anyway. And then the style. The free-DeepL doesn't allow to use "informal" tone for the translations, etc. Just keeping the English is fine for contributions

pfalcon:
Great, thanks!

PR #4291, opened 2024-10-17
Improve Export of transaction history to DivvyDiary

mosswald82:
positions that have been completely sold are no longer filtered out when exported to DivvyDiary. 
So they are included in the calculations for historical portfolio values/dividends


mosswald82:
This is my first ever pullrequest on github. If I have done something wrong please forgive me and let me know what I need to change.
I have tried to make the changes in such a way that it does not affect other parts of the program.
The new parameter which deactivates the previous filter behavior is only used when exporting to divvydiary.

buchen:
@mosswald82 Can you test if it works as expected. With 45f8482fe0526df652bd64f9131e62c2df78f862 the changes are merged in master

mosswald82:
Thanks for the corrections.
The result looks good! I have the same number of “sold securities” at the end of my portfolio. As well as the same total amount of dividends in the dividend matrix over the entire period!

PR #4298, opened 2024-10-22
fix non hidden tooltips for some languages in securities chart

mierin12:
Hello, this is a proposition to fix an issue in the tooltips of securities chart, only happening for French, Dutch and Polish languages.
The string used to exclude the tooltip was generally the same of the tooltip label, but actually different for those three languages.
ex for French : `Messages.SecurityMenuBuy`: _Acheter_ (the verb) vs `PortfolioTransaction.Type.BUY.toString()`: _Achat_ (the noun).

**Before :**
2024-10-22 19_04_16-Portfolio Performance

**After :**
2024-10-22 19_01_27-Portfolio Performance


buchen:
good catch!

PR #4299, opened 2024-10-22
zero percentage value shown as white and without arrow

mierin12:
to be consistent with the format of Statement of Asset.

Hello, this is a proposition to apply the same color rule found in Statement of Assets : a 0% (performance, price delta) is not shown as green with green arrow, but white without arrow. Same also as for a 0.00 value when not a %.

**Before :**
Statement of assets: "white" behavior
Capture d’écran 2024-10-22 193459
Securities Lists: "green behavior"
Capture d’écran 2024-10-22 193555
Securities Performance View "green behavior"
Capture d’écran 2024-10-22 193701

**After :** 
Capture d’écran 2024-10-22 194115
Capture d’écran 2024-10-22 194230

(Also applicable for the Return column of Trades.)

0% IRR seems to be -0% so very very slightly negative ?


buchen:
I like it. Let's see the response from the users :-)

PR #4302, opened 2024-10-23
Fix yearly Payment barchart color with full change to eclipse.swtchart

mierin12:
Issue : https://github.com/portfolio-performance/portfolio/pull/1760
Closes : https://github.com/portfolio-performance/portfolio/issues/1949

Hello,

This is a draft proposition to fix the Payment/Year bar charts which are not colored in case of negative values (so always for Taxes/Fees, and sometimes for Closed Trades / Sum / Savings if one year is negative).

As per this comments : https://github.com/portfolio-performance/portfolio/pull/1760#issuecomment-698347352, new swtchart revisions have the `barOverlay` attribute which is exactly what is needed here and avoid the stack workaround. Once switched to `eclipse.swtchart`, the fix is trivial.

The switch to `eclipse.swtchart` could either :

- be limited to a few charts only, by duplicating some files : https://github.com/portfolio-performance/portfolio/compare/master...mierin12:portfolio:partial_change_to_swtchart14_to_fix_yearly_negative_payment_color
- be propagated to all of the charts. This PR. First commit is the switch to `eclipse.swtchart`, second commit the fix itself.

CSS looks ok on my Windows.

**Before :**
Capture d’écran 2024-10-22 221130

**After :** 
Capture d’écran 2024-10-22 221205

Capture d’écran 2024-10-22 221702

I put in draft as for the moment the switch to `eclipse.swtchart` only consists in resolving the Errors message. So mainly using `getPlotArea().getControl()` instead of `getPlotArea()` for Listeners. 
I followed a bit what was done in your branch https://github.com/portfolio-performance/portfolio/compare/master...feature_org.eclipse.swtchart. 
(It seems that removing entirely `getPlotArea().getControl()` and applying the Listener directly on the chart seems to work too. I do not understand if there is a difference between those two ways. or using `(Composite) getPlotArea()`

There are still Warnings. I do not know if they should be silenced or resolved (with <?> ?).
Capture d’écran 2024-10-23 214621

and I am not sure how to double check that all is indeed working as before. At first glance it looks ok, but this is changing a lot of different files so who knows.
~~One example : holding down and moving the tooltip across the chart does not feel as smooth as before ?~~ resolved

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
@@ -8,6 +8,7 @@ import org.eclipse.swt.events.PaintEvent;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
 
 import name.abuchen.portfolio.money.Values;
@@ -119,7 +120,7 @@ class MeasurementTool implements ChartTool
             var text = buffer.toString();
 
             Point txtExtend = e.gc.textExtent(text);
-            Rectangle plotArea = chart.getPlotArea().getClientArea();
+            Rectangle plotArea = ((Composite) chart.getPlotArea().getControl()).getClientArea();
 
             e.gc.setBackground(Colors.brighter(color));
             e.gc.setForeground(Colors.getTextColor(color));
--- a/name.abuchen.portfolio.ui/plugin.xml
+++ b/name.abuchen.portfolio.ui/plugin.xml
@@ -184,7 +184,7 @@
                class="name.abuchen.portfolio.ui.editor.Sidebar">
          </widget>
          <widget
-               class="org.swtchart.Chart">
+               class="org.eclipse.swtchart.Chart">
          </widget>
          <widget
                class="name.abuchen.portfolio.ui.util.Colors$Theme">
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ChartCSSHandler.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ChartCSSHandler.java
@@ -5,12 +5,12 @@ import org.eclipse.e4.ui.css.core.dom.properties.converters.ICSSValueConverter;
 import org.eclipse.e4.ui.css.core.engine.CSSEngine;
 import org.eclipse.e4.ui.css.swt.properties.AbstractCSSPropertySWTHandler;
 import org.eclipse.swt.graphics.Color;
-import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
-import org.swtchart.Chart;
-import org.swtchart.IAxis;
-import org.swtchart.ILegend;
-import org.swtchart.ITitle;
+import org.eclipse.swtchart.Chart;
+import org.eclipse.swtchart.IAxis;
+import org.eclipse.swtchart.ILegend;
+import org.eclipse.swtchart.IPlotArea;
+import org.eclipse.swtchart.ITitle;
 import org.w3c.dom.css.CSSValue;
 
 import name.abuchen.portfolio.ui.util.chart.ScatterChart;
@@ -70,7 +70,7 @@ public class ChartCSSHandler extends AbstractCSSPropertySWTHandler implements IC
 
             chart.setBackground(newColor);
 
-            Composite plotArea = chart.getPlotArea();
+            IPlotArea plotArea = chart.getPlotArea();
             if (plotArea != null)
                 plotArea.setBackground(newColor);
             ILegend legend = chart.getLegend();
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ChartElementAdapter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ChartElementAdapter.java
@@ -2,7 +2,7 @@ package name.abuchen.portfolio.ui.theme;
 
 import org.eclipse.e4.ui.css.core.engine.CSSEngine;
 import org.eclipse.e4.ui.css.swt.dom.CompositeElement;
-import org.swtchart.Chart;
+import org.eclipse.swtchart.Chart;
 
 @SuppressWarnings("restriction")
 public class ChartElementAdapter extends CompositeElement
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ElementProvider.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ElementProvider.java
@@ -10,7 +10,7 @@ import org.eclipse.e4.ui.internal.css.swt.definition.IFontDefinitionOverridable;
 import org.eclipse.e4.ui.internal.css.swt.definition.IThemesExtension;
 import org.eclipse.swt.widgets.Table;
 import org.eclipse.swt.widgets.Tree;
-import org.swtchart.Chart;
+import org.eclipse.swtchart.Chart;
 import org.w3c.dom.Element;
 
 import name.abuchen.portfolio.ui.editor.Sidebar;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/AbstractChartToolTip.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/AbstractChartToolTip.java
@@ -173,12 +173,12 @@ public abstract class AbstractChartToolTip implements Listener
 
     private Rectangle calculateBounds(Event event, Point size)
     {
-        Rectangle plotArea = ((Composite) getPlotArea().getControl()).getClientArea();
+        Rectangle plotArea = ((Composite) getPlotArea()).getClientArea();
 
         int x = event.x + (size.x / 2) > plotArea.width ? plotArea.width - size.x : event.x - (size.x / 2);
         x = Math.max(x, 0);
 
-        Point pt = getPlotArea().getControl().toDisplay(x, event.y);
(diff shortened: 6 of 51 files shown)


buchen:
Hi @mierin12,

give me some time to check this pull request.

> ~~One example : holding down and moving the tooltip across the chart does not feel as smooth as before ?~~ resolved

I have a (very, very old) branch that also switches over to the SWTChart library.
In the end, I did not merge the branch because the tooltip were not smooth.
(Back then my theory was that the new implementation is not working with arrays but a provider for points that includes unboxing of the numbers)
But I think time moved on. Let me check.

Andreas.

buchen:
@mierin12 I am not sure if it is visible in the screen recording below, but the tooltip has a visible lag.

I tried to some profiling but I am not able to pinpoint it. Compared to the previous implementation there seems to be a repaint event to the underlying chart when the tooltip is moved. If the chart contains a lot of data series, then these couple milliseconds create the lag (aka "not smooth moving").

In the "payments" view it does not matter, because there is not enough data.

https://github.com/user-attachments/assets/e02a0849-5e34-4990-aa61-eef216c33c20

I'll spend a little more time on digging around, but then I propose to upgrade to Eclipse SWTChart for the bar charts only (and fix the coloring - great work).

mierin12:
Hello @buchen , ah so maybe there are two sources of lag. There was one even worse before commit 5 :

- I did not use the correct `getPlotArea()` method in commit 1, tried to correct it in commit 3 but messed up and it added a lag.
- In the end, I think commit 5 fixed it and this can be sum up by :
Attaching the addlistener to `IplotArea.getControl()` is not the same (very laggy) as attaching it to `(Composite) IPlotArea` (better). At least for the listeners of AbstractChatToolTip

But if even the second case is still laggy, there has to be something else in addition.

mierin12:
To limit it to the Payment charts only, I can propose a PR from my other branch. 
https://github.com/portfolio-performance/portfolio/compare/master...mierin12:portfolio:partial_change_to_swtchart14_to_fix_yearly_negative_payment_color

Through PlainChart it is also applying the change to the Earnings and Activity Widget.

Then I duplicate TimelineTooltip to not impact the other chart. And a dedicated CSS handler is also created to correctly apply the CSS.

buchen:
@mierin12 - thanks for the branch. I can pick the changes from there.

buchen:
I am also wondering if we should consider a bigger "adoption" of SWTChart.

To my understanding, SWTCharts have a "wrapper" API that creates "extension" charts around the basic Chart.

https://github.com/eclipse/swtchart/tree/develop/org.eclipse.swtchart.extensions/src/org/eclipse/swtchart/extensions

Those charts support a couple more features, for example the buffering of the underlaying chart if for example the user does interactive selection. I just checked, even though the PlotArea has a setBuffered method, it is only used in the extension charts.

I never did that so far, because I don't want and don't need all the interactive things. But also basic capabilities like buffering are only implemented in the extension charts.

buchen:
@mierin12 I have created a pull request out of your branch in order to easier pick the change into master

mierin12:
Thanks for your time on this ! I will try to investigate this lag.

For the extension : it could be useful yes, maybe also for the pie chart extension ? It seems that it is adding a dynamic behavior. So this would put the SWT pie chart closer to the HTML pie chart (dynamic behavior when clicking on a slice, which does a "zoom").
https://github.com/eclipse/swtchart/wiki/PieCharts#extensions

buchen:
> For the extension : it could be useful yes, maybe also for the pie chart extension ?

Pie charts are already on org.eclipse.swtchart - so that should be easier.

buchen:
@mierin12 

One problem I faced that the getXDataSeries method is very inefficient now as it is (re-)constructing the array on the fly.
I therefore add a TimelineSeriesModel that provides access to the underlaying array (and changed to use epoch days in the double array).

```java
   public static class TimelineSeriesModel extends DoubleArraySeriesModel
    {
        private final double[] xdata;
        private final double[] ydata;

        public TimelineSeriesModel(double[] xSeries, double[] ySeries)
        {
            super(xSeries, ySeries);

            xdata = xSeries;
            ydata = ySeries;
        }

        public double[] getXData()
        {
            return xdata;
        }
    }
```

that allows the access the series for the binary search (find the closest neighbors to display the right tooltip)

```java
                var dataModel = (TimelineSeriesModel)series.getDataModel(); 
                var xData = dataModel.getXData();
                
                int line = Arrays.binarySearch(xData, ((LocalDate) focusedObject).toEpochDay());
```

Alas, that does not make a difference. It is still janky.

But I realized that also the current charts could be improved by switching from Data to LocaDate. But that is another change.

buchen:
@mierin12 And one more update: I enable a buffering like the BaseChart does, then there is no visible lag in the context menu anymore. I think that can be a good additional also for the measurement tool in the SecurityChart.

But - at least on macOS - the graph temporarily blurs (see the video). I think the reason is that the buffer (= image) is in a lower resolution on retina macOS than the original image. If we can fix this, then we can move to Eclipse SWTChart.

https://github.com/user-attachments/assets/7e58a0d9-ed54-4a61-a2a2-e5641b55f834




```java
    public void enableBuffering()
    {
        if (!getPlotArea().isBuffered())
        {
            var plotArea = getPlotArea();

            suspendUpdate(true);
            Image image = new Image(Display.getDefault(), plotArea.getImageData());
            ISeriesSet set = getSeriesSet();
            ISeries<?>[] series = set.getSeries();
            for (ISeries<?> serie : series)
            {
                serie.setVisibleBuffered(serie.isVisible());
                hideSeries(serie.getId());
            }
            /*
             * The image will be disposed when releasing the selection and
             * setting the background image to null.
             */
            plotArea.setBackgroundImage(image);
            plotArea.getControl().setData(IPlotArea.KEY_BUFFERED_BACKGROUND_IMAGE, image);
            plotArea.setBuffered(true);

            suspendUpdate(false);
            redraw();
        }
    }

    public void hideSeries(String selectedSeriesId)
    {

        ISeries<?> dataSeries = getSeriesSet().getSeries(selectedSeriesId);
        if (dataSeries != null)
        {
            dataSeries.setVisible(false);
            dataSeries.setVisibleInLegend(false);
        }
    }

    public void showSeries(String selectedSeriesId)
    {

        ISeries<?> dataSeries = getSeriesSet().getSeries(selectedSeriesId);
        if (dataSeries != null)
        {
            dataSeries.setVisible(true);
            dataSeries.setVisibleInLegend(true);
        }
    }

    public void disableBuffering()
    {
        if (getPlotArea().isBuffered())
        {
            System.err.println("Disabling buffering");
            var plotArea = getPlotArea();

            suspendUpdate(true);
            plotArea.setBackgroundImage(null);
            Object object = plotArea.getControl().getData(IPlotArea.KEY_BUFFERED_BACKGROUND_IMAGE);
            if (object instanceof Image image)
            {
                plotArea.getControl().setData(IPlotArea.KEY_BUFFERED_BACKGROUND_IMAGE, null);
                image.dispose();
            }
            plotArea.setBuffered(false);
            ISeriesSet set = getSeriesSet();
            ISeries<?>[] series = set.getSeries();
            for (ISeries<?> serie : series)
            {
                if (serie.isVisibleBuffered())
                {
                    showSeries(serie.getId());
                    serie.setVisibleBuffered(false);
                }
            }
            suspendUpdate(false);
            redraw();

        }
    }
```

buchen:
Hey @mierin12 - FYI, I pushed my experiments into the branch [feature_swtchart_buffering](https://github.com/portfolio-performance/portfolio/tree/feature_swtchart_buffering). The branch contains both changes - the move to the SeriesModel and the buffering. Alas, it does not make a difference. the change appears to be the extra paint event where I do not understand what triggers that.

buchen:
Hey @mierin12 - I found it. The SWTChart implementation is adding a mouse move listener that is forcing a redraw on any mouse move. That is causing the janky rendering. There is a workaround (although it requires to use internal classes). But it allows to move to eclipse SWTChart completely. Yiha!

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/MeasurementTool.java
```diff
@@ -119,7 +120,7 @@ public void paintControl(PaintEvent e)
             var text = buffer.toString();
 
             Point txtExtend = e.gc.textExtent(text);
-            Rectangle plotArea = chart.getPlotArea().getClientArea();
+            Rectangle plotArea = ((Composite) chart.getPlotArea().getControl()).getClientArea();
```
mierin12, line 123: I think I made a mistake here : getControl then casting to Composite is one action too much ?

((Composite) chart.getPlotArea()) should be sufficient ?

mierin12:
Great @buchen !! It is definitly more satisfying to do a full switch rather than separating the dépendancies through my ugly duplication of TimelineChartToolTip.

I put one comment where I think I made a mistake.

I do not have my laptop with me at the moment but is the following summary correct :
- The unduplication of mousemove listeners fixes the lag when moving tooltip (non regression)
- The buffer allows the measurement tool to be less laggy (improvement). On Retina, there can be a Blur.
- Timeseriesmodel replaces the deprecated getXserie method. More efficient.

Is that correct ?

mierin12:
Searching in their code, I think several times there is a check if getPlotArea is a Control or not. While I have automatically used getControl a bit everywhere. May for some cases it is non needed ?
```
BaseChart chart = getBaseChart();
		IPlotArea plot = chart.getPlotArea();
		if(plot instanceof Control control) {
			control.redraw();
		} else {
			chart.redraw();
		
}

buchen:
> The unduplication of mousemove listeners fixes the lag when moving tooltip (non regression)

Yes. I now have the same behavior before and after.

> The buffer allows the measurement tool to be less laggy (improvement). On Retina, there can be a Blur.

Well, on my Mac, the measurement tool behaves quite okay. But I know from reports that slower/older/Windows machines have problems. But to be honest, I would keep this change separate for the moment.

> Timeseriesmodel replaces the deprecated getXserie method. More efficient.

I think the default implementation of Eclipse SWTChart is inefficient because it got rid of the array of data. But we need that to efficiently find the closest neighbor. Maybe it is not the "performance killer" because we used that only when displaying the tooltip. 

> Searching in their code, I think several times there is a check if getPlotArea is a Control or not.

I don't really know. I can only assume that the there are cases where the PlotArea is replaced by something else. I am not aware that it is a case for our usage of SWTChart. Generally, I find their API quite confusing given the multiple types of charts and extensions.

buchen:
Created a new pull request #4312 that includes all changes including cleanup

PR #4303, opened 2024-10-24
[WIP] annualized quote change column

OnkelDok:
Hey, finally I found time to do a fresh installation of the development (because with installed one I get build errors). So here a first change proposal:

Sometimes for a security I want see the annualized quote change. Especially when looking at higher intervals (>1 year), then with an annualized value it's easier to get an idea of the yearly increase, which is normally a good indicator for financial products.

So here a first sample as draft. If @buchen generally accepts this new column then I would continue the work (translations etc.).

Suggestions for (better) naming of the column (label, header, description) are welcome.

grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesTable.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesTable.java
@@ -208,6 +208,7 @@ public final class SecuritiesTable implements ModificationListener
         addColumnDateOfLatestPrice();
         addColumnDateOfLatestHistoricalPrice();
         addQuoteDeltaColumn();
+        addQuoteDeltaColumnAnnualized();
         support.addColumn(new DistanceFromMovingAverageColumn(LocalDate::now));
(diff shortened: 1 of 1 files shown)


OnkelDok:
@buchen: I don't know if you noticed this draft PR. Please give a short feedback (thumbs up or down) if these kind of columns would find a way into the software or not. If yes, then I will continue and finish the work on this PR.
Thanks in advance!

stoeggich:
IZF is already there ?!

OnkelDok:
I've now added labels and translations.
According to the label of the annualized TTWOR column (``TTWROR p.a. 2 Jahre``) I used the same format (``𐤃 p.a. 2 Jahre``).
If the format like in my initial screenshot is better, then I can change this.

@buchen If you want see more reuse of code, then maybe we can create a class for the price change columns (normal an annualized) because a lot of code lines are the same. Only the final calculation and the labels and column ident is different.

OnkelDok:
@buchen: Alternatively, if a separate column is to much, then I thought about adding tooltip to the current existing "quote change column", where we then display the annualized value. But I have not found any way to add a tooltip provider for these kind of columns. It seems this "all securites" view has other column-types than other lists. Or am I miss something?

PR #4306, opened 2024-10-25
Add Color Definitions to ColorSchema Enum (HeatMap) (from #4283)

Nirus2000:
Add Color Definitions to ColorSchema Enum

1. **GREEN_YELLOW_RED**: Green for gains, Red for losses, Yellow for neutral values.
2. **GREEN_WHITE_RED**: Green for gains, White for neutral, Red for losses.
3. **GREEN_RED_GREY**: Green for gains, Red for losses, Grey for neutral.
4. **BLUE_ORANGE_GREY**: Blue for stable values, Orange for volatile values, Grey for neutral.
5. **BLACK_YELLOW_WHITE**: Black for extreme gains, Yellow for moderate returns, White for neutral.

Sorry...  Reopen this... i kill my fork :-(
https://github.com/portfolio-performance/portfolio/pull/4283


Nirus2000:
Here an example in light and dark theme

grafik

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/ColorSchema.java
```diff
@@ -24,46 +26,129 @@ private ColorSchema(String label)
...
+                        // 0.4 to 1.0 to 0 to 1.0
+                        float yellowFactor = (float) ((p - 0.4) / 0.6);
+                        color = Colors.interpolate(color, Colors.YELLOW.getRGB(), yellowFactor);
+                    }
+                }
+                else
+                {
+                    // Using -p for correct mapping of negatives
+                    color = Colors.interpolate(Colors.theme().defaultBackground().getRGB(), Colors.RED.getRGB(),
+                                    (float) -p);
```
buchen, line 67: By using ```-p``` all negative values will be white (= the color of the background) because the ```interpolate``` function clamps the value between 0 and 1 (and p is always between 0 and 1)

buchen:
Merged.

I have refactored the creation of the gradients into a separate class. And used multiple color points in order to increase the fidelity of the colors (for example, always interpolating between green and gray makes the lighter greens "dirty"). See 7ea526ee0290440ddd83360fb6553f06dc430456

@Nirus2000 
The only thing I am not yet happy with is the "background -> yellow -> black"  color schema. From my point of view, it does not work in dark mode. It becomes "black -> yellow -> black". Then it has a completely different meaning (color the outliers black) than in light mode.

We can:
* make it always "white -> yellow -> black". But then it might not look good in dark mode

You had an example somewhere (but I cannot find it anymore). How do they handle dark/light mode?

Nirus2000:
Hello @buchen 
**Light Theme:**
grafik


**Dark Theme:**
grafik

PR #4310, opened 2024-10-25
add purchase price including fees and taxes in Statement of Assets and Security performance views

### LINKED ISSUE #408: Einstandskurs wird nicht korrekt berechnet
in der Performance-Berechnung wird m.E. der Einstandskurs eines Wertpapiers nicht korrekt bestimmt, da Ordergebühren nicht in der Berechnung einfließen.

Der Einstandskurs wird in der Regel wie folgt definiert: "Der Kurs eines Wertpapiers, der zu seinem Kauf inklusive aller Nebenkosten aufgewendet wurde".

### LINKED ISSUE #792: Statement of assets' purchase price, but including commissions
Hi!
First of all, thanks for this project! I have installed at least 50 android apps, signed up on a dozen websites and looked at various other stuff and nothing was close to what I was looking for. I found this by chance in Archlinux's AUR and it's amazing!

I have a very very small amount of money, and a very expensive broker, so the commissions on my trades are very significant. For this reason, is it possible to have, in the statement of assets, an average purchase price that includes the broker fees? That figure looks deceptively good in my case, but if I add my transaction costs it becomes much more reasonable.

Thanks!

mierin12:
Closes #408
Closes #792
Issue : https://forum.portfolio-performance.info/t/entry-value-per-share-column-in-statement-of-assets-view/21680

Hello, this is a proposition to add the purchase price including the transactions fees and taxes in the Statement of Assets and Securities Performance view.
(is it correct to name it a "gross price" ?)

To not overload the menu (in particular in Statement of Assets), I have added a sub menu for purchase price, and to not have complicated names in the menu, ('Purchase price (moving average), Purchase price (moving average) (gross), Purchase price (FIFO), Purchase price (FIFO) (gross) ), I have added the possibility to add a LabelOnly message in the submenus. This could be reused.
I have added some MenuLabel for the FIFO purchase price to show "(FIFO)" in the menu. I think it looks cleaner.

**Results :**
- Statement of Assets :

- Securities Performance views :

This is the configuration that I prefer, but many other configurations are possible : 
<details>
  <summary>some tested config in the spoiler..</summary>
  
2024-10-26 01_10_34-Portfolio Performance

2024-10-24 22_37_34-Paramètres

</details>


mierin12:
Additionaly, this is moving down a little bit the 'Purchase Value' of Securities Performance view menu, to be similar with the order from Statement of Assets menu (below Market Value). Very optional and only a matter of visual personal preference.

Maybe isin/symbol/wkn could be harmonized between the two views too.

buchen:
Thanks for the pull request. I like the change.

Two small things I will change:
* in other places it says "before taxes and fees". So I will switch the "fees and taxes" label around
* rename the "labelOnly" to something like "sub-heading" of the column. That it is implemented as "LabelOnly" is one thing. But if we have a better dialog in the future, then heading can still work)

PR #4311, opened 2024-10-26
Fixes NPE in by expand and collapse menu

### LINKED ISSUE #4309: Null Argument Internal Error when right clicking table column header
The following dialog is presented when right clicking on the column header bar of the Calculation tab of the Performance Calculation pane in version 0.71.2 of the software.
image

The red X marks the spot
image

**To Reproduce**
1. Go to the 'Performance Calculation' pane
2. Select the Calculation Tab
3. Right click in the region shown by the red X
4. All going (well) wrong, you see the error.

 - OS: Windows 11
 - Version 0.71.2 (October 2024)
 - Platform win32, x86_64
 - Java: 17.0.5+8-LTS

Nirus2000:
Closes #4309 
Fix null pointer exceptions by adding null checks for `obj` and `targetNode`


buchen:
Thanks for the fix.

For the refactoring of the "Actions" I used the "SimpleAction" which already exists (as opposed to a local factory method addAction)

PR #4312, opened 2024-10-27
Switch to Eclipse SWTChart

buchen:
* Replace usages of org.swtchart by org.eclipse.swtchart
* Fixed bar chart colors in case negative values in payments view
* Fixed janky tooltip (due to unnecessary mouse move listener redrawing the chart)
* Use epoch day as value on the x-axis (instead of time millis) to have less conversions

Issue : https://github.com/portfolio-performance/portfolio/pull/1760
Closes : https://github.com/portfolio-performance/portfolio/issues/1949


buchen:
@mierin12 - can you be so kind and test this branch

mierin12:
Hello @buchen , tested, it looks flawless, congratulations !! The tooltip moves super fast.
I tested a cvs export, png export, the different Zoom shortcut, the measurement tool, the "Jump" function to go from Widget chart to the main charts for Statement of Assets/Perf/Earnings. I think all is good. 

There is one comment, but I think it does not come from this PR :
When moving the toolip fast, I can see a black background:
Capture d’écran 2024-10-27 195108
But I have this behavior from the current "master" too, so I am wondering if it arrived with the "short tooltip" update.
On the current published release, I have this black tooltip behavior on the SecuritiesChart when moving fast, but never on the other charts.

(The short tooltip is sometimes really small in Payment/Month(accumulated)
2024-10-27 19_53_18-Portfolio Performance )

mierin12:
FYI, they provided a way to bypass the mousemove listener :
https://github.com/eclipse/swtchart/issues/421

buchen:
@mierin12 thanks for the hint. Once SWTChart has released a new version, we'll switch to that version and update the code

buchen:
merged.

mierin12:
@buchen I have noticed something : on the Y axis, 0% does not seem to show up anymore. When you see it, you can't unsee it..
Securities chart second Y axis:
Capture d’écran 2024-11-18 004411
Performance chart:
Capture d’écran 2024-11-18 004426
But on Statetement of Assest chart, it is ok : 
Capture d’écran 2024-11-18 004442

I am assuming this is related to https://github.com/eclipse/swtchart/issues/375 ?
If I understood correctly, they have fixed it, but it will be available in the next release ? Thankfully it seems to be quite soon (~end of this month : https://projects.eclipse.org/projects/science.swtchart/releases/1.0.0)

mierin12:
Another point that I just noticed but this one is a good one I think : 
With the change to swtchart 0.14.0, the tick of logarithm scale looks much better in my view:
It seems to me that the label are now dynamically based on the size of the chart.

**Before**
Capture d’écran 2024-11-29 215926

**After**
Capture d’écran 2024-11-29 215851

Maybe this can help this PR about log scale on the perf axis chart where the label were an issue https://github.com/portfolio-performance/portfolio/pull/2246

PR #4327, opened 2024-11-02
set up more CalculatedQuote format in SecuritiesChart tooltips

mierin12:
for
- purchase cost when several holding periods exists
- calculated purchase/buy quote

Hello, this is a proposition to apply the CalculatedQuote format (number of decimal to be displayed as per the user setting) for those two data. 
For the purchase value (or rather purchase price), this format is already applied, but for the first holding period only. 

**Before :**
2024-11-02 18_41_34-Portfolio Performance

**After :**
 2024-11-02 18_39_36-Portfolio Performance

I was wondering if this should also be done for dividends ?
2024-11-02 18_49_46-Portfolio Performance


buchen:
merged.

PR #4339, opened 2024-11-08
Improve TradeRepublic PDF Importer

MonkeySon:
Hello,

this pull request improves the TradeRepublic PDF importer by adding support for an additional type of interest transaction pdf.
It is contains the following identifying line, which was not supported previously:

"IBAN BUCHUNGSDATUM GESAMT"

Due to how PDFs with "ABRECHNUNG ZINSEN" are currently handled, a change was required in both 
`addTaxesStatementTransaction()` and `addInterestStatementTransaction()`. When only added to one of these, the parse in the other function would fail.

BR,
MonkeySon


Nirus2000:
Done... 👍🏻 
Good job !

PR #4341, opened 2024-11-10
Improve sort of SecurityTaxonomyPage by weight

Nirus2000:
grafik


Morpheus1w3:
Sorry but this make updates more challenging because if you're updating a list of attributes, you have to read the whole list until you'll find the value who need to be updated. Alphabetical sort sequence is might preferable from my point of view.

Nirus2000:
Hello @buchen, @Morpheus1w3 and @auchri 
The background here is that the sense of the security editing is used correctly and is also displayed. It makes no sense to go by category when updating the weightings, as the total weighting is only 100%.

The point is to display a weighting correctly for a faster way to update the data, but not to sort the categories. 
These are already sorted alphabetically.

buchen:
One can have it now both ways - sort by taxonomy (the natural order created in the taxonomy, by name, or by weight)

Bildschirmfoto 2024-12-22 um 11 44 36

PR #4346, opened 2024-11-15
SecurityDetailsViewer: Another attempt to fix ScrolledComposite artifact

pfalcon:
Additional code to avoide situation when ScrolledComposite's scrollbar covers right edge of its internal container, obscuring information there (like last digits of quotes, etc.). Instead, we want to layout this container set so that inner content container took ScrolledComposite's width without scrollbar.

For reference, this is based on hint in this StackOverflow question: https://stackoverflow.com/questions/35123/prevent-swt-scrolledcomposite-from-eating-part-of-its-children


pfalcon:
@buchen, this a kind of follow-up to https://github.com/portfolio-performance/portfolio/pull/4329, now any artifacts with info pane hopefully should be gone.

PR #4347, opened 2024-11-15
Add weekly investment plan

### LINKED ISSUE #1396: Wöchentliche Buchungen für Sparplan
Echt tolles Programm, benutz es jetzt seit einem Jahr und alles geh super!

Eine Anregung: Würds praktisch finden wenn man für das Sparplan-Intervall auch "wöchentlich" als Option hätte. Überweisungen auf mein P2P-Konto sind zum Beispiel wöchentlich.

Beste Grüße!

### LINKED ISSUE #4330: Feature request: Weekly, Bi-weekly Interval (Investment Plan)
There is currently no option to schedule weekly or bi-weekly investment plans. The only option is the monthly interval.

Add the option to schedule weekly and bi-weekly interval using the investment plan. Or if possible custom interval.

**Describe alternatives you've considered**
None available

**Additional context**
N/A

mierin12:
Closes #1396
Closes #4330 
Issue https://forum.portfolio-performance.info/t/intervall-von-sparplaenen/5655

Hello, this is a proposition to add weekly (and bi-weekly) intervals to investment plan.
What I call "bi-weekly" is every two weeks = every two Mondays/Tuesdays etc. This is different than twice a month.

I think "twice a month" is also a wished interval but it seems more complicated to do, and a workaround with two monthly plans is already possible. But there was no workaround for weekly or bi weekly intervals. 

When an "each Monday" date is an holiday, the transaction is made on the next non-holiday date, and the following transactions are put back to "each Monday". 
If the start date itself is on a holiday Monday, the first start date is offset in the same way, and the next transactions are kept on Mondays.

A test is added in `InvestmentPlanTest`.

I have also changed the items shown in the column Interval when you want to switch the interval :
**Before :**
2024-11-15 22_46_07-Portfolio Performance

**After :**
2024-11-15 22_52_28-Portfolio Performance

Internally, I used `100` and `200` for the weekly and biweekly interval integers since I believe there will never be an "every 100 months" interval. It is very easy to add an "every three weeks" interval if relevant.


name.abuchen.portfolio/src/name/abuchen/portfolio/model/InvestmentPlan.java
```diff
@@ -49,6 +49,8 @@ public enum Type
 
     private List<Transaction> transactions = new ArrayList<>();
 
+    private LocalDate weeklyNominalDate;
```
buchen, line 52: If we are adding a new property (which is persisted in XML) then we also have to persist it in the binary (protobuf) format.

However, I think we work without this date. Why not use the "DayOfWeek" to determine if the transaction has been moved (similarly how we use the DayOfMonth to see if the monthly transactions moved)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/InvestmentPlanModel.java
```diff
@@ -45,6 +45,55 @@ public enum Properties
 
     private IStatus calculationStatus = ValidationStatus.ok();
 
+    public enum Intervals
+    {
+        WEEKLY(100), //
```
buchen, line 50: Great idea. I was always thinking about adding a new property (lots of work).
I propose to use (interval - 100) = week.
This way we leave the space >0 <100 for months, and >100 <200 for weeks.
And have enough "space" for future uses.

buchen (commented):
Great change.

I will make the two proposed changes and merge.

buchen:
merged.

mierin12:
Hello @buchen thanks ! But be careful : I tested with the changes but I think now there is an issue, I am not sure that the daysofweek method is exactly equivalent. I have an infinite loop when I try to generate the transactions if put a start date at Monday 18 march 2024 for example, to see the behaviour for Monday 1s April which is holiday. The transactions are never generated, it is stuck in a loop.

What is concerning me also is that the test should have catched that.

mierin12:
infinite loop :
Capture d'écran 2024-03-01 034901

mierin12:
I think that was the idea behind
```
        if (interval > 12) // weekly intervals are 100 or 200
            weeklyNominalDate = next;
```
We store what should have been the next date (=the next previous transaction) before the offset from calendar.
So when we want to generate on the next iteration we generate from the non offset date:
```
            if (weeklyNominalDate != null)
                previousDate = weeklyNominalDate; 
```
and we can advance in time. otherwise there is minus days, then plus days, then minus day again, it is stuck.

mierin12:
I think to keep day of week : 
```
                previousDate = transactionDate.minusDays(
                                transactionDate.getDayOfWeek().getValue() - start.getDayOfWeek().getValue());
```
should work ? without +7L to avoid going back in past too much ?

PR #4348, opened 2024-11-17
update Findependent PDF Extractor

fpschill:
Updated and extended PDF extractor for Findependent
- can process EN or DE PDF files
- new transaction types
- correct treatment of exchange rates


Nirus2000:
Hello @fpschill 
thanks for your pull request 👍🏻 
That looks good already.

So that we can validate your changes to the importer, we need so-called test cases. These are PDF debugs that are generated.
Please have a look at the video for generating the [Extract PDF documents for debugging](https://help.portfolio-performance.info/en/how-to/requesting-new-importer/) and of course the [Contributing Rules](https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#pdf-importer) for creating the test cases.

If you have any questions... let's help each other and learn from each other! 💯 

Regards
Alex

buchen:
@Nirus2000 Not sure if we will get a proper test case. Merge anyway because we have at least other test cases? I know that it increases the risk that it will break in the future.

Nirus2000:
We have no Tests about this... 😞

PR #4357, opened 2024-11-17
add comparison with "moving average" purchase price in securities chart

mierin12:
Hello, this is a proposition to add a "Market trend vs Purchase Price (moving average)" option in the Securities chart.
So far, both the FIFO and the Moving Average purchase prices could be plotted, but only the comparison with the FIFO one was possible.

**Results:**
Capture d’écran 2024-11-18 001605

2024-11-18 00_17_01-Paramètres

**Note :** the label I have added uses "Purchase Price" while the term "Purchase Value" is used in the securities chart option and legend. So not consistent. I can align to "Purchase Value" of course. But from the different columns available elsewhere, I thought that "Price" was the "per share" term, as opposed to "Value" which was the nb of shares*purchase price. 
And maybe the "Market trend vs Purchase Value" can be completed by "(FIFO)".


buchen:
> the label I have added uses "Purchase Price" while the term "Purchase Value" is used in the securities chart option and legend. So not consistent. I can align to "Purchase Value" of course. But from the different columns available elsewhere, I thought that "Price" was the "per share" term, as opposed to "Value" which was the nb of shares*purchase price.
And maybe the "Market trend vs Purchase Value" can be completed by "(FIFO)".


I agree. value = total valuation; price = price per share

I'll fix the label

PR #4366, opened 2024-11-24
Improve some PDF-Importer to standardized tax lost adjustment transaction

Nirus2000:
The tax lost adjustment transaction has been standardized for various PDF importers.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/quirinbankag/QuirinPrivatbankAGPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/quirinbankag/QuirinPrivatbankAGPDFExtractorTest.java
@@ -280,7 +280,7 @@ public class QuirinPrivatbankAGPDFExtractorTest
 
         // check taxes transaction
         assertThat(results, hasItem(taxes( //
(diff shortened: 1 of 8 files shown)


Nirus2000:
I have no issues... 
grafik

buchen:
The problem is the following.

* The tests run with the "en" locale.
* You write the shares to the context using String.format and hence the "en" locale (see line 177 in Direkt1822BankPDFExtractor), say "11.00000000"
* You then read the shares always with the German locale (that is hard coded in the AbstractPDFExtractor) which uses the comma as decimal separator

And, whoops, you have a mismatch because parsing does not see the dot as decimal separator.

It does not happen for you locally, because I assume you run with the German locale.

If you want to reproduce the problem locally, add this to your launch configuration
```
-Duser.language=en -Duser.country=
```

Bildschirmfoto 2024-11-29 um 10 37 07

I think the important thing is to use the same formatter for reading and writing (and hence the same locale setup),
for example,

```
context.put("shares", getNumberFormat().format(item.getShares() / Values.Share.divider()));
[...]
asShares(context.get("shares"))
```

or 

```
context.put("shares", Long.toString(item.getShares()));
[...]
t.setShares(Long.parseLong(context.get("shares")));
```

Nirus2000:
Hello @buchen 
Thanks for the hint.
I have changed it and moved it to the master.

Regards
Alex 👍🏻

PR #4371, opened 2024-11-29
Fix check locale

buchen:


pfalcon:
From https://github.com/portfolio-performance/portfolio/pull/4366#issuecomment-2507444512

> The tests run with the "en" locale.

@buchen : FYI, another issue with some tests is that they rely on particular set of timezones. Run in other timezones, they may fail because day wraparound happens.

buchen:
@pfalcon writes:

> FYI, another issue with some tests is that they rely on particular set of timezones. Run in other timezones, they may fail because day wraparound happens.

Which ones?

Because in this particular case it was indeed a programming problem - the code is writing the number in the current locale but parsing it with the German locale. That was most of the time working (it is a German bank PDF) but its good that the tests running in the English locale picked it up.

PR #4372, opened 2024-11-29
fix non visible 0% tick axis since the switch to swtchart 0.14

mierin12:
Hello,

Issue : https://github.com/portfolio-performance/portfolio/pull/4312#issuecomment-2481683966

Since switching to swtchart 0.14, the tick on chart axis for 0% does not appear anymore. As per the linked comment, it may be already fixed by swtchart but only available in the next release ? I am not sure.

I do not know why this works, but noticing than chart axis with the `ThousandAmountFormat` did not had this issue, I tried to use a similar approach. On my Windows it works. Maybe there is a simpler way to get it done ?

**Before :**
Capture d’écran 2024-11-29 214649   Capture d’écran 2024-11-29 214705   Capture d’écran 2024-11-29 214719   Capture d’écran 2024-11-29 214743

**After :**
Capture d’écran 2024-11-29 215029   Capture d’écran 2024-11-29 214404   Capture d’écran 2024-11-29 214307   Capture d’écran 2024-11-29 214439

Maybe the naming could be improved for maintainability point of view ? `PercentNumberFormat` -> `AxisTickPercentNumberFormat` ?

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/TimelineChart.java
@@ -1,6 +1,5 @@
 package name.abuchen.portfolio.ui.util.chart;
 
-import java.text.DecimalFormat;
 import java.time.LocalDate;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


buchen:
Very interesting fix. I really do not understand why this works, but it works.

I understand that SWTChart ist about to publish a 1.0.0 that would fix it, see eclipse/swtchart#410. But I plan to publish a new version this weekend. And that is a big change in an upstream library to include on such a short notice. I'll merge it and we can remove it once we adapt 1.0.0

I'll also take up your suggestion for renaming.

PR #4379, opened 2024-12-02
adding support for Ginmon VL Sparen

christen90:
Account Statements
Buys
Sells
Fees
Non Importables


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -1221,6 +1253,138 @@ private void addAccountStatementTransaction()
...
+                            if ("Wertpapierverkauf".equals(v.get("type")))
+                                t.setType(PortfolioTransaction.Type.SELL);
+                            t.setSecurity(getOrCreateSecurity(v));
+                            t.setDate(asDate(v.get("date") + v.get("year")));
+                            t.setCurrencyCode(v.get("currency"));
+                            t.setShares(asShares(v.get("shares")));
+                            t.setAmount(asAmount(v.get("amount")));
+                            v.getTransactionContext().put(FAILURE,
+                                            Messages.MsgErrorTransactionAlternativeDocumentRequired);
+                        })
```
Nirus2000, line 1380: As you know :wink:, we do not import securities bookings via account statements. In your example, you have already shown via Kauf14.txt that there are separate statements.

Nirus2000 (changes requested):
Hallo @christen90 
I would prefer that we process “Ginmon” as a (new) separate PDF importer, as sub-companies tend to change/convert their software more often. We would then have the problem of always looping “Ginmon” into a fixed importer.
It should be relatively easy to separate this.

Regards
Alex

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
christen90: Somehow they generate account statements with buys only. To have a better error message I tried it this way. Tagging it as non importance.

christen90:
> Hallo @christen90 I would prefer that we process “Ginmon” as a (new) separate PDF importer, as sub-companies tend to change/convert their software more often. We would then have the problem of always looping “Ginmon” into a fixed importer. It should be relatively easy to separate this.
> 
> Regards Alex

Currently the only Ginmon specific document is the invoice. And this will be marked as non importable. You mean that this one shall be moved to a new importer?

Best regards Christian

Nirus2000:
Hello @christen90 

> Currently the only Ginmon specific document is the invoice. And this will be marked as non importable. You mean that this one shall be moved to a new importer?

Yes, all documents created by “Ginmon Vermögensverwaltung GmbH”.

- FeeInvoice01.txt
- Kontoumsaetze08.txt
- Kontoumsaetze09.txt
- Kontoumsaetze10.txt
- Kontoumsaetze12.txt
- Kontoumsaetze12.txt

The way I see it, they all are. Except Verkauf14.txt is apparently a regular member of the DAB Group.

---

https://github.com/portfolio-performance/portfolio/pull/4379#discussion_r1866370582
> Somehow they generate account statements with buys only. To have a better error message I tried it this way. Tagging it as non importable.
Yes, we should keep it that way. Very good, I had overlooked “MsgErrorTransactionAlternativeDocumentRequired”. 👍🏻 

As a hint... if there are problems with the DAB, simply lock the document: PDFParser.java-->"_mustNotInclude_"
```
        public DocumentType(String mustInclude, String mustNotInclude,
                        Consumer<Transaction<DocumentContext>> contextBuilder)
        {
            this.mustInclude.add(Pattern.compile(mustInclude));
            this.mustNotInclude.add(Pattern.compile(mustNotInclude));
            this.contextBuilder = contextBuilder;
        }

        public DocumentType(String mustInclude, String mustNotInclude)
        {
            this.mustInclude.add(Pattern.compile(mustInclude));
            this.mustNotInclude.add(Pattern.compile(mustNotInclude));
        }
```


Regards
Alex

christen90:
I moved the invoice to a new importer.

Please see a picture of pdf document and confirm if you really would treat them as Ginmon documents instead of DAB. For me Ginmon is just mentioned in the addressfield but I understand Ginmon not as the publisher of PDF.

Kontoumsaetze09 pdf

Nirus2000 (approved):

Nirus2000:
Hello @christen90 
I have added the following changes:

For the Gimnon fee statement, we use the Gimnon statement document directly and not the account statement.
The Gimnon statement simply contains good information such as invoice number and period. If we have separate and detailed documents, then we will use these.

I have also tidied up the DAB a bit.

Regards
Alex

christen90:
Hello Alex,
thank you for the merge. One problem I'm seeing with Ginmon fee statement is, that no valuta date is given in the document. Ginmon just mentioning in the upcoming days...

I would still prefer the approach with fee-wise only considering the account statements.

BR Christian

Nirus2000:
Hello @christen90,

I understand the point.
I would suggest we start from 
“Invoice date: 31.10.2024”, which was set in the document.

christen90:
Hello @Nirus2000,
we can do this, but it is a workaround. No valuta date is provided in the invoice and in the test cases you can see that they don't match (invoice 31.10.2024 - last day of month <> account statement ~10 d later [Kontoumsaetze09.txt:18](https://github.com/portfolio-performance/portfolio/pull/4379/files#diff-aa4d62952660909f0cd211fb3f95b7ee0e68f9fd148cbd6bd89589fcc20590b7) ).

Which data you miss in the Kontoauszug to process the fees there?

Nirus2000:
> Hello @Nirus2000, we can do this, but it is a workaround. No valuta date is provided in the invoice and in the test cases you can see that they don't match (invoice 31.10.2024 - last day of month <> account statement ~10 d later [Kontoumsaetze09.txt:18](https://github.com/portfolio-performance/portfolio/pull/4379/files#diff-aa4d62952660909f0cd211fb3f95b7ee0e68f9fd148cbd6bd89589fcc20590b7) ).
> 
> Which data you miss in the Kontoauszug to process the fees there?

Hello @christen90 
I can't get the connection right now.
Can you explain this in more detail?

Regards
Alex

PR #4383, opened 2024-12-03
Fix PDF parser swallows error message about duplicate securities

buchen:


Nirus2000:
For me... looks good! 👍🏻

PR #4387, opened 2024-12-07
PortfolioListView: Add "Import CSV" context menu item

pfalcon:
Such is available in AccountListView, but randomly was not available for portfolio list. As for account list, CSV import dialog will have current portfolio and its reference account pre-selected.


pfalcon:
Busting random gaps in functionality.

Issue #4325, opened 2024-11-02
Incorrect account preselection for CSV import


verglor:
**Describe the bug**
For CSV imports accounts are pre-selected according to the last used account for any other CSV import so they are shared between imports from different banks/brokers. This may cause hard to recover mistakes of importing transactions to incorrect accounts.

Forum: https://forum.portfolio-performance.info/t/csv-import-improvements/29924

**To Reproduce**
1. Go to 'File > Import > CSV files' (or select CSV template for 'Broker A')
2. Select CSV file from 'Broker A'
3. Adjust CSV configuration and column mappings
4. Click on 'Next >'
5. Select desired cash and securities accounts (e.g. 'Broker A')
6. Click on 'Finish'
7. Go to 'File > Import > CSV files' (or select CSV template for 'Broker B')
8. Select CSV file from 'Broker B'
9. Adjust CSV configuration and column mappings
10. Click on 'Next >'
11. Cash and securities accounts are pre-selected to 'Broker A'

**Expected behavior**
The accounts should not be pre-selected for generic CSV imports or should be pre-selected according to CSV template used. 
Pre-selected accounts should not be shared between imports from different banks/brokers.

**Desktop**
 - OS: Windows
 - Version: 0.71.2

**Details**
There was [Remember the last selected account/portfolio by the type of PDF document](https://github.com/portfolio-performance/portfolio/commit/c527a219a6010de9610ccf6ecb0c94f209fb7759) implemented which does not fit well for generic CSV imports.
I tried to disable pre-selection according to `extractor` in [ReviewExtractedItemsPage.preselectDropDowns()](https://github.com/portfolio-performance/portfolio/blob/ccd521eb8a6f5fd9ea3e8ecf2ebba819414d0705/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java#L335) but instead of [CSVExtractor](https://github.com/portfolio-performance/portfolio/blob/ccd521eb8a6f5fd9ea3e8ecf2ebba819414d0705/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVExtractor.java) the private proxy is used.

Example of stored accounts:
```
import-target-account-Account\ Transactions=2bea6733-fe86-4e70-a0d2-507df4168945
import-target-account-Portfolio\ Transactions=2bea6733-fe86-4e70-a0d2-507df4168945
[...]

PR #4388, opened 2024-12-07
PortfolioListView: Add column with balance of reference account

pfalcon:
Thus, portfolio list view will now have complete information: securities account and its MTM balance, and reference account with its cash balance.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
@@ -2,6 +2,8 @@ package name.abuchen.portfolio.ui.views;
 
 import java.text.MessageFormat;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.util.List;
 
 import jakarta.annotation.PostConstruct;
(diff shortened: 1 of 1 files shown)


pfalcon:
https://github.com/portfolio-performance/portfolio/issues/4325#issuecomment-2525082005 describes a usecase. The patch has a non-externalized string.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
```diff
@@ -256,6 +258,23 @@ public String getText(Object element)
         }));
         portfolioColumns.addColumn(column);
 
+        column = new Column("ref_cash_bal", "Balance of reference account", SWT.RIGHT, 100); //$NON-NLS-1$
```
Nirus2000, line 261: Missing translations

Nirus2000 (changes requested):

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
```diff
@@ -256,6 +258,23 @@ public String getText(Object element)
...
+        column.setLabelProvider(new ColumnLabelProvider()
+        {
+            CurrencyConverter converter = new CurrencyConverterImpl(factory, getClient().getBaseCurrency());
```
buchen, line 264: The converter is not used. The balance will be shown in the account currency. Is this intended?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
buchen: I wonder if we should re-use the label from the account view: ColumnBalance

It works nicely in German: Depotvolumen vs. Kontostand

But maybe not so well in other languages.

buchen:
I added the translations and merged the change.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
pfalcon: > The converter is not used. The balance will be shown in the account currency. Is this intended?

Yeah, it's clearly was a copy-paste issue, where label provider was initially copied from security balance column, but the getText() then was copied from AccountListView, leaving converter unused.

Thanks for taking care of the patch!

PR #4393, opened 2024-12-12
ImportIBHandler: Use last saved dir to import IBKR XML files

pfalcon:
Use existing CSV_IMPORT_PATH saved path ID to not introduce another one. Previously, the open dialog started with some internal path to PP executables, so using the same dir as CSV is good enough improvement.


pfalcon:
@buchen: This is quick, "individual fork" level fix for the issue of not remembering a directory for IB imports and instead starting from some completely unrelated dir. As the description says, it reuses existing CSV_IMPORT_PATH for IB import path too. Potentially following further changes could be done:

1. Introduce a dedicated saved path id. IMHO, that's going to be too bread-crumb detailed. It's definitely good idea to save a last use import path, but it's pretty fair assumption that a user might want to import different types of files from the same dir (e.g. a Downloads dir).
2. But then, might make sense to rename CSV_IMPORT_PATH to just IMPORT_PATH.

I can't decide what's better, and then for my fork, I optimize just for the size of patch / minimize changes. Hope you either can make changes you see fit, or let me know how'd you like it to be elaborated. Thanks.

buchen:
Thanks. I agree it is a good idea to reuse the path. Not sure how often people mix and match

(to be honest, I always import PDF files by dragging them to the side bar. No opening dialog. No need to remember the path)

PR #4395, opened 2024-12-13
Infer the default currency based on the user's locale when creating a new portfolio file

liuhaoXD:


buchen:
Nice change!

Merged (the linter was complaining about the formatting. I fixed that).

PR #4235, opened 2024-09-16
Add derived data series in Statement of Assets Chart

mierin12:
Hello, this is a proposition to add derived data series to the statement of assets chart.

Closes #605
Closes #707
Closes #1529
Issue : https://forum.portfolio-performance.info/t/delta-diagramm-fur-ausgewahlte-aktien/6787

**Results :**
Capture d’écran 2024-09-15 202846

This is based on the PR https://github.com/portfolio-performance/portfolio/pull/3754 from @shivampaw. As it was without update since February, I am proposing some modifications taking into account the previous review.

**Main updates from #3754 :**

- Dedicated selection dialog for the derived data series.  

- In the new data series selection dialog, I am proposing a different tree order : 
  - _[derived data]/Accounts & Portfolio/Account_
 Capture d’écran 2024-09-15 202634
I find it more intuitive than 
  - _Accounts & Portfolio/Account/[derived data]_

- Securities are added as a possible subtype of the derived data. I guess this is well applicable/defined ?
- The derived data series are correctly plotted as bar or with an area when this is relevant. This was missing.

**Note on taxonomy labels :**
For the selection labels, two propositions (without/with commit https://github.com/portfolio-performance/portfolio/pull/4235/commits/8120e876c9a97435ee36bd9465f5967cd5681b03) :
- with fullPath that I prefer but which is different than the existing behavior,
Capture d’écran 2024-09-15 205300
- with the same behavior as the existing data series selection 
Capture d’écran 2024-09-15 204955

**Multiselection topic :**
There is no multiselection issue I believe. 
Multiselection allows here to plot several distinct derived data series.

**Legend label :**
Lots of `( )` maybe another format could be used ?
Delta (for the reporting period) (Portfolio A)
Delta (for the reporting period) [Portfolio A]
?

<details> 
  <summary><b>Other forum threads</b></summary><p>

Issue: https://forum.portfolio-performance.info/t/customize-filter-invested-capital/15474
Issue: https://forum.portfolio-performance.info/t/delta-fur-klassifizierung-anzeigen/15669
Issue: https://forum.portfolio-performance.info/t/graphs-of-investment-and-performance/23365
Issue: https://forum.portfolio-performance.info/t/filtermoglichkeit-allgemeiner-datenreihen/25460
Issue: https://forum.portfolio-performance.info/t/add-invested-capital-in-one-account-to-charts/10854
Issue: https://forum.portfolio-performance.info/t/investiertes-kapital-eines-depots-anzeigen-lassen/20766
[...]

PR #4396, opened 2024-12-17
add derived data series to statements of assets chart

mierin12:
Hello, same as https://github.com/portfolio-performance/portfolio/pull/4235 that I was not able to maintain by rebase, I manage to be up to date by squashing its commits.
So now this PR present a single clear commit rather than an history of WIP commits that I could not maintain.

Goal : derived data series are available in the associated widget but it does not allow to compare different data series and is not customizable (color). By adding the derived data series to the statement of assets chart, this is possible.
Exemple : comparison of Delta of different data series :
2024-12-17 19_45_24-Paramètres

Based on https://github.com/portfolio-performance/portfolio/pull/3754


buchen:
@mierin12 thanks for the hard work in refactoring the change (and the patience to see it through).

I like the approach (new type, separate dialog).

Please have a look at #4403. Based your code, I did some more refactored:
* First, I renamed the objects and types to be "derived" and "aspect" (similar to the widget on the dashboard)
* Second, the DerivedDataSeries is not an object that holds the reference to the underlying data series (base) and the aspect to be used. This way, we can re-used the (potentially) already cached data series.
* Third, I now create the data series based on the already created underlying data series.

If you have time, I would appreciate if you can have a look

Issue #2214, opened 2021-04-14
Save button disabled while selling a share


K1llsh0t:
**Describe the bug**
Unable to sell when credit note is 0.

**To Reproduce**
1. select any share and sell.
2. make sure credit note shows as 0 value 
3. See save button disabled

**Expected behavior**
It should allow saving when the credit note is 0.

**Screenshots**
<img width="657" alt="Screenshot 2021-04-14 at 6 38 44 PM" src="https://user-images.githubusercontent.com/28539587/114731015-6efebf00-9d52-11eb-8da2-70530b87861d.png">

**Desktop (please complete the following information):**
 - OS:  macOS

chirlu:
That’s intentional. There is a separate transaction type (not sure what it’s called in English) for selling a security without a corresponding account transaction.

K1llsh0t:
> That’s intentional. There is a separate transaction type (not sure what it’s called in English) for selling a security without a corresponding account transaction.

Thanks for the reply. I see a Buy, Sell and dividend options. Can you please help me with a screenshot where you were able to sell a security with 0 return?

EliasZepf:
> Can you please help me with a screenshot where you were able to sell a security with 0 return?

Rather late, but maybe it helps: You can use the "Delivery (Outbound)" transaction for this purpose.

![DeliveryOutbound](https://user-images.githubusercontent.com/77249873/118293679-2dd90680-b4da-11eb-96e2-ce8a0471a8d2.PNG)

mierin12:
In release [0.73.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.73.0,) a hint to use Outbound Delivery instead of a Sell in case of a zero sell amount has been added from https://github.com/portfolio-performance/portfolio/pull/4401

Issue #2103, opened 2021-02-25
Please allow "Sell" transactions totaling $0


metal450:
There seems to be a check in the software preventing any sale where the amount totals $0 (it shows the error "Input required for 'Total').  In the case of cryptocurrency, it would be much more convenient if this were allowed.  Why?  In PP, fees are always paid out in 'currency,' not in 'securities' - but in crypto, fees are paid in the crypto itself (i.e. ethereum gas).  To document this in PP, it takes 2 transactions each time: a Sell transaction to convert the crypto to USD, followed by a Fees transaction for the same amount.  However, if "Sell" transactions weren't prevented from having a total of $0, this could be documented as just one "Sell" where the amount equals the fees.  i.e. in the Sell dialog, the amount is $10, fees is $10, thus credit total is $0.  The result is the as desired: the crypto 'shares' were removed, you booked $10 in fees, and no change to your USD holdings - but it took half as many transactions.

As my transaction history grows increasingly cluttered with Sell, Fee, Sell, Fee, Sell, Fee, I think it'd be a lot more crypto-friendly to just remove this 'total can't be $0' check :)

Ragas13:
Duplicate of #2094

metal450:
Whoops, sorry - closing :)

chirlu:
This isn’t a duplicate of #2094, but it is a clear “won’t fix”. Use outbound delivery instead of sell.

metal450:
Then perhaps for UI clarity, it would be better to display that info on the dialog - i.e. "Sell with an amount of $0 should be input as Outbound Delivery," vs "Input required for sub-total."

Issue #1993, opened 2021-01-18
Fees based on Shares deduction


RomainD2:
Hi,

Thanks for this amazing program!

I would love to use it for all my financial products, however, there is a limitation because of some specific financial product which have fees based on securities shares deduction. For example :
![portfolio2](https://user-images.githubusercontent.com/4403383/104944018-e1ac0880-59b6-11eb-92df-97fc3e7a29f6.PNG)

You can see in "Nombre de Parts" which are deduced from my Shares. I found a way to do that, but with two operations. I need to sell this amount, and then, to add fees with the same amount, so two transaction.

Maybe the best way would to be able to add these fees with the same amount when selling this share, but it is not possible to save due to soft limitation :
![image](https://user-images.githubusercontent.com/4403383/104944516-9cd4a180-59b7-11eb-97b9-7fe7773db369.png)

Maybe there are a better way to implement that, adding for example the way to use Shares under the Fee option, for example (Quick photoshop ^^)
![image](https://user-images.githubusercontent.com/4403383/104944901-313f0400-59b8-11eb-94ec-cec3980fe1d3.png)

What do you think about this feature ?

Thanks

debegr92:
I think this feature will be also very helpful for some crypto investors. Most exchanges have fees in a specific crypto currency instead fiat. Maybe combine this with https://github.com/buchen/portfolio/issues/1960 to also allow share based fee for transactions between security accounts.

zr40:
This would be useful for my pension plan. They deduct their monthly fees directly as a percentage of shares.

zr40:
You can actually do this already. Go to the Securities account, and instead of "Sell" or "Fees", use "Delivery (Outbound)". Specify the number of shares that were deducted and make sure that the fees are equal to the value of the shares.

For example:

![Screenshot from 2021-01-30 16-44-54](https://user-images.githubusercontent.com/616299/106360739-88c76300-631a-11eb-9ddf-dffa7d4e0df6.png)

mierin12:
In release [0.73.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.73.0) a hint to use Outbound Delivery instead of a Sell in case of a zero sell amount has been added from #4401

PR #4401, opened 2024-12-18
add hint to use Delivery (Outbound) for a final null value in Sell

mierin12:
Hello, 

the idea comes from the comment https://github.com/portfolio-performance/portfolio/issues/2103#issuecomment-909314765

> Then perhaps for UI clarity, it would be better to display that info on the dialog - i.e. "Sell with an amount of $0 should be input as Outbound Delivery," vs "Input required for sub-total."

The question of "null value for Sell" does arrive from times to times here and on the forum :
https://github.com/portfolio-performance/portfolio/issues/2214
https://github.com/portfolio-performance/portfolio/issues/1993
https://github.com/portfolio-performance/portfolio/issues/2103

so the proposition is to add a hint to use an Outbound Delivery instead in this case.
Capture d’écran 2024-12-18 205945


buchen:
I slightly updated the translation. Thanks. That is a very helpful addition.

PR #4403, opened 2024-12-20
Add derived data series to statements of assets chart

buchen:
Uses DerivedDataSeries to hold a reference to the underlying data series (base) and the aspect to be shown (ClientDataSeries).

Issue: #3754
Issue: #4235
Issue: #4396


buchen:
@mierin12, as commented in the other PR:

> Based your code, I did some more refactored:
> * First, I renamed the objects and types to be "derived" and "aspect" (similar to the widget on the dashboard)
> * Second, the DerivedDataSeries is not an object that holds the reference to the underlying data series (base) and the aspect to be used. This way, we can re-used the (potentially) already cached data series.
> * Third, I now create the data series based on the already created underlying data series.
>
> If you have time, I would appreciate if you can have a look

mierin12:
Hello @buchen , thanks for checking it !!
I have one comment related to your third point. Building from the `availableSeries `works, except for the `isLine `and `isArea `value. Those parameters are set from the `baseDataSeries` but should actually depend here on the Aspect itself, not the baseDataSeries. Ex : Interest should always be Bar type : "Interest : Taxonomies A" should be a Bar even if by default Taxonomy A will be a Line.

I am not sure if there is a smart way to do it. For the Client ones they are set individually, so this can for sure be made case by case here also. Or to put parameters in DataSeries.ClientDataSeries ?

So far my only comment. I will continue to double check.

I think the naming "Aspect : data series" is definitely an improvement for the Legend and Tooltip over "Aspect (data series)". But I am wondering if it is too much to repeat the Aspect in the subfolders of selection dialog. 
Capture d'écran 2024-03-01 034901
In particular for Taxonomies that already have a long name.

vs
Capture d’écran 2024-12-20 210121
that I had in mind. 
But as it is purely a UI question, as you prefer.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dataseries/DataSeriesSet.java
```diff
@@ -68,6 +81,48 @@ public DataSeries lookup(String uuid)
...
+                                                ClientDataSeries.INTEREST_CHARGE_ACCUMULATED).contains(type))
+                    continue;
+
+                var label = type.getLabel() + ": " + baseDataSeries.getLabel(); //$NON-NLS-1$
+
+                var dataSeries = new DataSeries(DataSeries.Type.DERIVED_DATA_SERIES, baseDataSeries.getGroup(),
+                                new DerivedDataSeries(baseDataSeries, type), label, wheel.next());
+
+                dataSeries.setLineChart(baseDataSeries.isLineChart());
+                dataSeries.setShowArea(baseDataSeries.isShowArea());
```
mierin12, line 119: here : it will always be a Line without Area. we can check ClientDataSeries types before and assign a correct boolean for Line and Area ?

mierin12:
I do not know how to make code suggestion in the review but my proposition of update is an in this :
https://github.com/portfolio-performance/portfolio/commit/a7cda52ee0bd8dd1d6d6a264af6daf7a1bc8113a

mierin12:
No other comment than the one about the Line/Area settings, I think this works great !! Thanks a lot for looking into it, I think it is a super feature ! 

(for another feature in the future I am wondering if a similar approach could be used in the Performance Chart to use daily/weekly.. % for any series :) !!)

buchen:
Thanks @mierin12 for reviewing!

Both comments make sense. I missed the showArea/lineChart. The labels should be fixed as well.

> (for another feature in the future I am wondering if a similar approach could be used in the Performance Chart to use daily/weekly.. % for any series :) !!)

Yes!! Agree.

buchen:
The dialog should not be uncluttered (sorry for the German screenshot, but you get the idea).

Bildschirmfoto 2024-12-21 um 09 09 58

buchen:
Also, I have updated the isLineChart and showArea settings:
https://github.com/portfolio-performance/portfolio/pull/4403/files#diff-19190310c6609557cfdcae769065451dd00475f2e667a52f180b8f8c683dc445R133-R137

mierin12:
Perfect ! I have looked again, I do not see any issue, it looks good !

PR #4415, opened 2024-12-23
Cope with GBX/GBP variations in quotes from Yahoo

### LINKED ISSUE #4038: GBp (penny) quotes from Yahoo Finance interpreted as GBP (pound)
**Describe the bug**
While PP supports the GBX symbol for accounts denominated in British pence (https://github.com/portfolio-performance/portfolio/issues/406), brokers such as Interactive Brokers usually denominate accounts in British pounds and not pence. Importing GBX (also called GBp) or GBP transactions from the Interactive Brokers XML files to PP is therefore only possible to GBP accounts. Attempting to import to an GBX account does not import any transaction due to GBP not matching GBX. Importing to a GBP account works however correctly, with PP reading the correct GBP denominated purchase price from the XML file.

When PP queries current/historic quotes from Yahoo Finance, it receives a GBp response. See for example https://query1.finance.yahoo.com/v8/finance/chart/LSEG.L which begins with `{"chart":{"result":[{"meta":{"currency":"GBp","symbol":"LSEG.L","exchangeName":"LSE","fullExchangeName":"LSE","instrumentType":"EQUITY",` and contains quotes in pence (as indicated by the currency name "GBp". Unfortunately, PP interprets this GBp quote as GBP, leading to current and historic values being off by 100. 

For historic reasons, it seems that all quotes for London listed stocks are in GBp and there should be no situation in which Yahoo Finance actually returns a GBP quote. This issue is not only Yahoo Finance specific, though at least Yahoo Finance includes the "GBp" currency information in the API response. Other providers, like alphavantage.co do not provide the currency symbol in the response and simply return "GBp" instead of "GBP". https://www.alphavantage.co/query?function=GLOBAL_QUOTE&symbol=LSEG.L&apikey=demo

**To Reproduce**

Add a GBP stock purchase to a GBP denominated PP account. For example, LSEG.L. Provide the purchase price in GBP. Observe that PP obtains the GBp quotes from Yahoo and reports the value of the investment as 100x the true value.

**Expected behavior**
PP distinguishes between GBp and GBP. When Yahoo Finance provides a GBp quote for a GBP security, PP correctly converts from GBp to GBP and reports the accurate value.

**Workaround**
Obtain alphavantage.co API key. Setup manual JSON source for the security:

image

where the URL is https://www.alphavantage.co/query?function=GLOBAL_QUOTE&symbol=SYMBOL.L&apikey=APIKEY and disabling historical data.

 - OS: Linux
 - Version 0.68.4

bscattergood:
Scale GBX/GBP quotes to match the security currency when appropriate.

Closes #4038

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooFinanceQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/YahooFinanceQuoteFeed.java
@@ -62,6 +62,18 @@ public class YahooFinanceQuoteFeed implements QuoteFeed
                         .addParameter("corsDomain", "finance.yahoo.com").get();
     }
 
+    private static long scaleGB(long price, String quote_currency, String security_currency)
+    {
+        if (quote_currency != null)
+        {
+            if (quote_currency.equals("GBP") && security_currency.equals("GBX")) //$NON-NLS-1$ //$NON-NLS-2$
+                return price * 100;
+            if (quote_currency.equals("GBX") && security_currency.equals("GBP")) //$NON-NLS-1$ //$NON-NLS-2$
+                return price / 100;
+        }
+        return price;
+    }
+
     @Override
     public Optional<LatestSecurityPrice> getLatestQuote(Security security)
     {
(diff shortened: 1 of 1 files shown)


bscattergood:
I've some portfolios where securities were created in GBX, matching what Yahoo historically reported in quotes.  Their most recent rationalisations to GBP messed those up.

Empirically, this change (on top of commit 05f09c8f86efec646fa023d234d1f6b45e548673)  fixes those if I manually delete the incorrect historical quotes, and then "Update Quotes."

Rebasing to a more recent master I'm seeing build problems (but I see the same problems with master itself.)  Are those expected?

buchen:
Thanks for the change.

I think it is okay to merge. This is an incompatible change. Users, that previously got GBX might get GBP. But then this must also have been misconfigurations.

Thanks for the change. I'll change the variables names slightly ans the "underscore" is not a common naming.


> Rebasing to a more recent master I'm seeing build problems (but I see the same problems with master itself.) Are those expected?

Not sure which ones you mean. Maybe there have been temporarily build failures, but usually - and now - the build is successful.

PR #2623, opened 2022-01-07
Introduce new event list widget

buletti:
This PR introduces a new widget for the dashboard. The purpose of the new widget is to display a list of events of various types for a given time period. The events list is sorted in chronological order. Depending on the given reporting period the displayed events can be past events, current events or upcoming events.

The widget currently supports the following different types of events:
- Calendar events
  - Holidays
- Account events
  - Interest payment/booking
- Security events
  - Dividend declaration(*)
  - Dividend record date(*)
  - Ex-dividend date(*)
  - Dividend payday(*)
  - Dividend payment/booking
  - Shareholder meeting(*)
  - Earnings report(*)
  - Stock split
  - Note/custom event 

Changes:
- added the new event list widget
- extended the existing security event types of the model; new types marked by (*) above
- adjusted the UI to also create events for the new security event types
  - adjusted context menus
  - adjusted event creation wizard pages

Limitations:
- translation is quick and dirty; created with google translate
- for the event type markers I reused already existing icons. I think they are working ok, but using specifically designed icons would of course be nicer

Screenshots:

A demo dashboard with event list widgets of various reporting periods:
pp_event_list_widget

The adjusted context menu in the statement of assets listing:
pp_events_context_menu_soa

The adjusted context menu in the statement of assets pane:
pp_events_context_menu_soa_pane

An example creation dialog page for an earnings report event:
pp_earnings_report_event_dialog

The unchanged creation dialog page for a custom event with a free-form text input field:
pp_custom_event_dialog

Attachments:

- Demo portfolio [pp_test_depot_dax.zip](https://github.com/buchen/portfolio/files/7829662/pp_test_depot_dax.zip)

PR #4431, opened 2024-12-30
EventListWidget: Event list widget for dashboard

pfalcon:
Based on FollowUpWidget and offering similar functionality, just for securities' events.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -423,6 +423,7 @@ public class Messages extends NLS
     public static String EditWizardQuoteFeedTitle;
     public static String EODHistoricalDataMissingAPIKey;
     public static String ErrorKeyAlreadyUsed;
+    public static String EventListWidgetTitle;
     public static String EventWizardDescription;
     public static String EventWizardLabelMessage;
     public static String EventWizardTitle;
@@ -1059,6 +1060,7 @@ public class Messages extends NLS
     public static String MsgErrorTradeCollectionWithErrors;
     public static String MsgErrorUpdating;
     public static String MsgHintNoAttributesConfigured;
+    public static String MsgHintNoEvents;
     public static String MsgHintUseOutboundDeliveryForZeroTotal;
     public static String MsgIncorrectConvertedSubTotal;
     public static String MsgIncorrectSubTotal;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
@@ -838,6 +838,8 @@ EditWizardQuoteFeedTitle = Historical Quotes
 
 ErrorKeyAlreadyUsed = Key is already used.
 
+EventListWidgetTitle = Securities: Events
+
 EventWizardDescription = Add a custom event (date, free text) to a security.\nExamples: placed order/limit, chart signal, stock news, comments
 
 EventWizardLabelMessage = Event
@@ -2109,6 +2111,8 @@ MsgErrorTradeCollectionWithErrors = Error while collecting trades
 MsgErrorUpdating = Error on updating
 
 MsgHintNoAttributesConfigured = No attributes of type ''{0}'' configured.\n<a href=''settings''>Configure now</a>
+
+MsgHintNoEvents = No events.
 
 MsgHintUseOutboundDeliveryForZeroTotal = Note: Sales with a value of zero are not supported. Please create an outbound delivery transaction instead.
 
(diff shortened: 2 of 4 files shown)


buchen:
Hi @pfalcon, thanks for trimming down the big pull request. I started picking up this request, but wasn't able to finish it last weekend. I want do change a couple things I noticed: there are three types of events, it is not clear which is which as the current display does not tell. And I think for events it could make sense to use full reporting periods to decide which to show (I have 400+ "past" which makes it hard to use at the moment). Next weekend I hope to complete that work.

pfalcon:
@buchen: Thanks for looking into this!

> And I think for events it could make sense to use full reporting periods to decide which to show (I have 400+ "past" which makes it hard to use at the moment).

As mentioned, it's based on the FollowUpWidget widget implementation, and thus offers "show past vs future dates" and "sort dates in ascending/descending order" settings. That allows to configure it to show next upcoming events (whichever number fits into the chosen vertical size), or recently passed events. IMHO, that's good enough for starters (on par with FollowUpWidget, and it's pretty useful, though of course we can think how it can be extended further either).

What's clearly missing is ability to filter by event type, but first there're not enough event types to start with! And adding those is completely separate rabbit hole, which I wanted to avoid going into right away.

buchen:
@pfalcon do you have an opinion what to do with dividend events? They have two dates: the payment and the ex-dividend date. I am wondering if one wants to show both. Then we would have "synthetic" events which do not map to the underlying SecurityEvent class.

pfalcon:
@buchen : 

> do you have an opinion what to do with dividend events? They have two dates: the payment and the ex-dividend date

I think these should be 2 separate (and independent) event types, then individual people can decide whether they want to track only one of them or both.

If there's an idea to track "complete" dividend cycle (kinda linking ex-div and payment date in some sequence), then IMHO that should be treated as a separate task/layer from just having a well-rounded inventory of simple single-date events.

buchen:
> If there's an idea to track "complete" dividend cycle (kinda linking ex-div and payment date in some sequence), then IMHO that should be treated as a separate task/layer from just having a well-rounded inventory of simple single-date events.

Agree.

The current functionality is limited. One can download dividend events from DivvyDiary. That events contain both - payment date and ex-date. I hope to add news sources this year.

pfalcon:
@buchen: 

> I have now merged the change. A good start.

Thanks! But I'm not sure that tying it into "reporting period" was really a good idea. Because what's for example a "1 year" reporting period as typical configured in dashboard? It's a period *ending* today and starting 365 days ago. In full accordance with that, the widget after you changes shows:

* either events starting from 2024-01-12 (sorting direction: ascending)
* or, events up to today, in descending order (sorting direction: descending)

First of the above is rather specialized and not very useful info, 2nd is more or less useful (assuming a user doesn't look at dashboard every day and want to know what they missed over the last few days). But the most useful info is *upcoming* events (ex-dividends, earnings releases all affect the share price and thus investment decisions), and that's exactly what "reporting period" widget can't show right away. If you look at the previous "event widget patch", then it apparently also was tied into reporting period, you can tell that by the fact that there were workarounds applied against the issue as described above: [Fix current month and current year reporting period interval to include](https://github.com/portfolio-performance/portfolio/pull/2623/commits/a29cf0f4894123388cc299667d6a88de24a669cb)
dates in the future .

Apparently, the idea was to switch to "current month/year" reporting period specifically for the event list widget (away from "a period in the past" as normal for the entire dashboard). But that would still lead to problems on the period edges, like it wouldn't show events across a year transition like we just had.

But why all that? "Securities: Date reached" aka FollowUpWidget worked perfectly with it's "date in the future" vs "date in the past" setting, IMHO events shouldn't be tied into "reporting periods". (Reporting periods are to study performance across different timeframes, while events and not performance-related (at least per se)).

buchen:
@pfalcon

Fair comment. My thinking was this:

First, I felt that "date in the past" is of limited value. In my personal file, I have >1000 of events in the past. There I see nothing.

Second, I wanted to enable the use case outline in the other pull request #2623 (the events listed according to months, quarter, year, etc.) (I wasn't aware of the limitation to how the reporting period is translated into an interval.

Third, if there is more control needed, then introducing and teaching users of a new way to determine the interval felt also cumbersome.

I have now added to special reporting periods "in the past" and "in the future" which are only used here. It is the last 100 years or the next 100 years respectively. Then both options are available - past/future and a specific year or a specific reporting period.

pfalcon:
> First, I felt that "date in the past" is of limited value. In my personal file, I have >1000 of events in the past. There I see nothing.

> Second, I wanted to enable the use case outline in the other pull request https://github.com/portfolio-performance/portfolio/pull/2623 (the events listed according to months, quarter, year, etc.)

@buchen, I appreciate the response. I still don't see what value that usecase from #2623 brings. For example, as far as I understand, you're not interested in past events, then what, you're interest in (distant) future events, e.g. for 2025Q2? I don't see much value in that, simply because actual event dates for many securities aren't (exactly) known long beforehand.

Anyway, just thinking aloud about roundabout ways to get there, as FollowUpWidget logic already worked well, while the new intervals need further adjustments: https://github.com/portfolio-performance/portfolio/pull/4495 . (And just imagine what if you don't agree that today is not the past! ;-) )

And thanks for implementing the event type filter!

PR #4434, opened 2024-12-31
Added option to strip carriage returns from extracted PDF text

buchen:
Apparently, some Quirin Bank document contain single carriage returns. That carriage returns are not in the sample data, because the CreateTextFromPDFHandler strips all carriage returns from the raw text.

The difference has been introduced with commit
d88e9317bca8a370c271eec563ad5030da873e35

To limit the impact, this change applies it only to the Quirin extractor.


Nirus2000:
Let's make a separate branch with PDFBox 3.x or higher. 
We'll have to update it at some point and I'd like to do it now at the beginning of the year.
I'll certainly have to touch all the PDF importers anyway to edit the pre-packages etc. and at some point we'll have to update that too.
I would also run through all my test PDF documents to fix the first differences.

buchen:
I have updated the code to remove the carriage returns for all documents.
I checked: no example has a single carriage return in the text.


About_:
>Let's make a separate branch with PDFBox 3.x or higher.

I have created issue #4449 to discuss how best to upgrade to PDFBox 3.x incrementally.

Issue #4449, opened 2025-01-03
Upgrade Apache PDFBox to 3.x 


buchen:
PP uses [Apache PDFBox ](https://pdfbox.apache.org) to extract a text presentation of the PDF document for further processing.

By now, Apache PDFBox is released in **version 3.0.3**. We plan to use this issue to track an update of the library within PP.

Problem statement:

The process of importing works like this:

```   PDF ---(PDFBox)--> Plain Text ---(regex)--> Transactions```

The challenge is that we have almost no test cases that take the PDF as input. Why? Because users want to provide anonymous content, we generate the text in the desktop application, let the user anonymize and then share the plain text. That means a) we cannot test if the new version of PDFBox creates the same text output and b) potentially break many importers which would require new sample files to fix the code.

Options:

* Create a distribution of PP that contains both PDFBox versions: the old and the new (latest) version
* Attempt to import the PDF using the new PDFBox version
* If that fails, attempt to import the PDF the old PDFBox version
* When creating a debug text document, use the new PDFBox version (--> collect examples in new version)

Technical considerations:
* We must extract the PDFBox dependencies in separate bundles so that each bundle (old and new) and have a dependency to the old and new PDFBox

**Comment by buchen** [comment]
Collecting changes in the extracted text:

Delta in Baader Bank document of February 2023 --> only in the address, not a problem.

```diff
7,8c7
< 04.02.2023
< Muster Str. 42
---
> Muster Str. 42 04.02.2023
71c70,71
<  
\ No newline at end of file
---
>  
>
```

buchen:
Consorsbank from 2021 --> not a problem

```
1c1
< Consorsbank • 90318 Nürnberg
---
> Consorsbank  90318 Nürnberg
8,10c8
< Datum: 15.12.2021
< 12345 Musterstadt
< Seite: 1 von 1
---
> 12345 Musterstadt Datum: 15.12.2021Seite: 1 von 1
37c35
< Président du Conseil d‘Administration (Präsident des Verwaltungsrates): Jean Lemierre, Directeur Général (Generaldirektor): Jean-Laurent Bonnafé
---
> Président du Conseil dAdministration (Präsident des Verwaltungsrates): Jean Lemierre, Directeur Général (Generaldirektor): Jean-Laurent Bonnafé
```

buchen:
Degiro

```
--- 1.8.17
+++ 3.0.3
@@ -7,2 +7,2 @@
[...]

PR #4450, opened 2025-01-03
Feature upgrade apache pdfbox

buchen:
First draft for #4449

It switches to PDFBox 3.0.3 and falls back - in case of errors - to PDFBox 1.8.

What I am still thinking about: How to learn if and where there are differences between the PDFBox versions.

Right now, I am printing a log message. But users will most likely not notice it let alone inform us.

I wonder if we should create both versions of the extracted text, and, if there are differences, put both files into the debug text. At least we see documents that are different.


ZfT2:
Maybe just provide a separate "beta" Version of PP with (only) pdfbox 3, post it in the forums and let the community test it?
If after a time no (bigger) failures are reported, upgrade ist also in the main release...

Advantage: we will only deal with one pdfbox Version in future.

buchen:
> Advantage: we will (hopefully) only deal with one pdfbox Version in future.

Yes. That is the goal.

I have now made two more changes to help us learn about differences.

The extracted debug text indicates if the version is different. At least we will learn if there are differences.

Bildschirmfoto 2025-01-06 um 11 21 28

And I print the diff to the log file. In case we want to better understand what the differences are. The changes so far do not make a material difference (better extraction of special characters, additional line breaks in address, etc.)

Bildschirmfoto 2025-01-06 um 11 10 08

stoeggich:
Can you perhaps make it so that you can check many files at once? Then I would run everything through myself

buchen:
I added a menu option to create diffs for multiple files.
It is only visible if you activated the "experimental features" in the settings.

Bildschirmfoto 2025-01-10 um 10 07 59

What it does not support: anonymizing the data via mouse click. You might have to do that manually.

Please post meaningful deltas in this issue #4449

So far, I think the diffs are manageable. They do not look like material differences that have impact on the relevant regular expressions.

Nirus2000:
> I added a menu option to create diffs for multiple files. It is only visible if you activated the "experimental features" in the settings.
> Bildschirmfoto 2025-01-10 um 10 07 59
> 
> What it does not support: anonymizing the data via mouse click. You might have to do that manually.
> 
> Please post meaningful deltas in this issue #4449
> 
> So far, I think the diffs are manageable. They do not look like material differences that have impact on the relevant regular expressions.

Perhaps you should also add “Experimental”, as in the XML document menu item.
BTW. why is “XML documents” grayed out, although the “marked as experimental” is activated?

buchen:
> BTW. why is “XML documents” grayed out, although the “marked as experimental” is activated?

Because it is a "headline" - all supported XML documents should come afterwards. There is only 1 at the moment. Menus do not support a headline. They only support deactivated items.

the item is only visible when the experimental stuff is activated. I hope that is enough labelling. We'll remove it in a couple weeks anyway. (or maybe in years... who knows ;-))

Issue #4045, opened 2024-06-01
[Feature request] Show Currently Selected Filter Name in Rebalancing View of Classifications


mad-sysadmin:
**Is your feature request related to a problem? Please describe.**
The currently selected filter is only visible when clicking on the filter icon.

**Describe the solution you'd like**
Include the name of the currently selected filter in the header section at the top, maybe next to the name of the classification in brackets.

**Describe alternatives you've considered**
None.

**Additional context**
We have various filters defined due to the fact that we track many of our portfolios in one file.
Our classifications with the defined weights can thus be valid for multiple filters (=portfolios).
During rebalancing, it is currently a bit unclear which filter is applied.

PR #4451, opened 2025-01-04
update titles of Taxonomy/Payment/Calculation/Holdings views as per selected filter

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4045
Issue : https://forum.portfolio-performance.info/t/filter-name-im-diagramm-titel/6277
Issue : https://forum.portfolio-performance.info/t/anzeige-des-filters-bei-performance-berechnung/6715

Hello, this is a proposition to change the default title of the Taxonomy/Payment/Calculation/Holdings views in order to include the name of the selected filter in it. And in case "Whole Portfolio" is selected, the current default titles are used.

I did not put it in Securities Performance nor in Statement of Asset views as there can be many Tabs so the length for the title may be short.

The implementation is not exactly the same for all 4 views as sometimes we have access to ClientFilterMenu directly, sometimes ClientFilter. For HoldingsPieChartView, it seems that we do not need the call on line 109 in createBody, while we need it for PerformanceView. I kept it nonetheless.

What I have checked  : 
-title is correctly updated when there is a change of selected filter
-title is correct when the file opens on the view while a filter is already selected, or if we switch between the views

Capture d’écran 2025-01-04 020037
Capture d’écran 2025-01-04 020112
Capture d’écran 2025-01-04 020140
Capture d’écran 2025-01-04 020205
Capture d’écran 2025-01-04 022301

I have also taken the opportunity to add a //$NON-NLS-1$ in Taxonomy view, silencing the last remaining warning in Eclipse, but maybe a proper translated MessageLabel is best instead of "Sync".

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterDropDown.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ClientFilterDropDown.java
@@ -44,4 +44,9 @@ public final class ClientFilterDropDown extends DropDown implements IMenuListene
     {
         return menu.hasActiveFilter();
     }
+
+    public ClientFilterMenu getClientFilterMenu()
+    {
+        return menu;
+    }
 }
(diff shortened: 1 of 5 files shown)


Morpheus1w3:
@mierin12 @buchen 
Kindly revert this change as if several filters are selected, it's now impossible to change the filter as this overlap the designated button.

mierin12:
@Morpheus1w3 
I see, when users do not give a name to the filter, the default filter name consisting in all accounts name can be long indeed. 
Maybe there is a way to keep the feature but to fix the issue by truncating the title when it becomes too long. In views where several tabs can be created, it is the tabs that adapt themself when it becomes too long, maybe the opposite behaviour could be done. 
Capture d’écran 2025-02-10 203604

Issue #3999, opened 2024-05-06
more visibility for Accounts grouping Filter


mierin12:
**Is your feature request related to a problem? Please describe.**
Accounts filter is a very good and useful feature but maybe not very well known by users, and not accessible enough (accessible only inside filter dropdown menu).
https://github.com/portfolio-performance/portfolio/issues/3967
https://forum.portfolio-performance.info/t/performance-focused-only-on-some-security-and-deposit-accounts/28307

-> their creations are only accessible at the "calculation" stage: when the user want to see the performance of his data, while they should already be known as a feature to the user before that, and already constructed.

**Describe the solution you'd like**
Ideally account filter should be clearly accessible and managed at other locations too. 
Some ideas :
1 : add a little descriptive line on what are "New.." "Manage.." "Reset" are about. It should draw attention to them.
_"Group Deposit and Securities Accounts as new data serie"_ something like that
![Capture d'écran 2024-05-06 233230](https://github.com/portfolio-performance/portfolio/assets/160436107/fe4cf172-c8d3-4d41-a6bf-78e0df1dcabe)

2 : since those Accounts filters deal with Deposit & Securities Accounts, they could be accessible right when you are creating and managing such accounts: the "New.." & "Manage.." could be accessible just when you create a Deposit/Securities Account. Again with the little descriptive text.
![Capture d'écran 2024-05-06 233332](https://github.com/portfolio-performance/portfolio/assets/160436107/2fa421d0-7716-4efa-99d0-614f93274848) ![Capture d'écran 2024-05-06 235513](https://github.com/portfolio-performance/portfolio/assets/160436107/da04e749-c1e6-40ba-acd4-47fb93beef91)

3 : "New.." "Manage.." could also be present when you right click on a Account ? Well, maybe it would be a too long list and we loose visiblity.
![Capture d'écran 2024-05-06 233411](https://github.com/portfolio-performance/portfolio/assets/160436107/43dd3048-6d22-4a33-9744-fc605c2670a3)

4: alternative to 2 if the resulting list would be too long : add a new button next to the + to deal specifically with the Account Filter. 
Or + is used for New account+ Account filter and the list of operation is moved on a new button ?

Thought ? Other ideas ? I think those propositions are not very complicated to implement so I may be able to propose something if needed, but as this is UI related, everyone can have difference preferences so it is better to discuss about it before potential implementation.

So in short, not a incredible crazy new feature, but simply giving more visibility to those Filters would be a good added value in my view.
[...]

Issue #3967, opened 2024-04-29
Data series groups for Performance:Chart and Performance:Return/Volatility


sig-noi:
**Is your feature request related to a problem? Please describe.**
When a particular investing strategy is spread out over multiple accounts, it is desirable to view the performance of those accounts together. StockMarketEye allowed this with the Portfolio Group function.

**Describe the solution you'd like**
Preferably an Account group function, whereby Accounts can be grouped together for organizational and reporting purposes. Less ideally would be the ability to select combinations of Accounts as data series when modifying reports.

**Describe alternatives you've considered**
The current solution I have found is performing a Save As of the larger group of Accounts, removing the unrelated Accounts, and viewing only the related accounts together in the new more focused file. The issue with this approach is that it is cumbersome and time consuming. 

**Additional context**

sig-noi:
Yes, the custom filter worked for me! Thank you!

PR #4458, opened 2025-01-05
more visibility for account filter creation

mierin12:
with
- new labels
- creation and management of filter is possible from cash and securities account views

Hello, this is a proposition related to https://github.com/portfolio-performance/portfolio/issues/3999
My view : the filter feature is super useful, and even key, as it allows to model any accounts. Sometimes, 1 Security Account + 1 Ref cash account is not enough (you may want to map X Securities Account + Y Cash Account). And even when enough, I personally like to recreate them in filter to give a custom shorter name.

But, I think that this useful filter is a bit hidden : available only at the end of the filter icon, that is used to select a filter in the different views. But for new users, I believe you expect to find and see this account filter feature before having the need to select them. In the forum there are several threads asking how to combine accounts, while the feature already exists.
https://forum.portfolio-performance.info/t/performance-focused-only-on-some-security-and-deposit-accounts/28307
https://forum.portfolio-performance.info/t/dashboard-key-risk-indicators-does-not-allow-to-select-multiple-accounts/30160
https://forum.portfolio-performance.info/t/dynamic-summary-portfolio/30965
https://github.com/portfolio-performance/portfolio/issues/3967

This PR proposes to allows the creation and management of those filters in more location : 
- in Securities and Cash accounts view, in the "+" menu 
- in Securities and Cash accounts view, when right clicking on a account (not 100% sure about this as the menu becomes long and the filters creation does not depend on the account actually clicked)

I was thinking the Securities and Cash accounts view were good location to add those actions, as this is where you can create new accounts.

Also some new labels are proposed.
 

2025-01-05 13_37_38-Paramètres
2025-01-05 13_38_18-Paramètres
2025-01-05 13_38_39-Portfolio Performance


mierin12:
I was thinking to go even one step further, something like :
2025-01-05 13_50_06-Portfolio Performance
but this is maybe too much, and anyway not ready

buchen:
Hi @mierin12,

thanks for thinking about how to improve the "filter" feature.

I am hesitating the merge the pull request as it is right now. Why? First, it adds a "create filter" menu in places where there is no feature using the filter. For example, in the account list, there is nothing using the filter. Even more, the "filter" icon there is used to filter deactivated accounts. I am not sure if this increases confusion. Second, it introduces a new name "grouping". I think it is worth considering this, but then we should be more consistent.

I like your idea about the "grouped accounts" view. Essentially that is a "filter management UI" from the sidebar (instead of a dialog). I want to extend filters further - for example allow to include an account with a percentage (include a securities with 50% because it is shared with partners).

My take is:
* Let's not clutter the context menus in the cash and securities accounts view
* Let's convert the "manage filter" dialog into a "grouped account" sidebar

I like the idea of transitioning to "grouped accounts".

Maybe:
* Have a new item - that separates the "filter" from the "grouping"
* Usually menus are separated, but in "Performance -> Securities" or "mater data --> all transactions" we separate it out
* Rename it more consistently from "filter" to "group"

Is that something that resonates with you?

mierin12:
Hello @buchen 

thanks for the review !

> I am hesitating the merge the pull request as it is right now. Why? First, it adds a "create filter" menu in places where there is no feature using the filter. For example, in the account list, there is nothing using the filter. Even more, the "filter" icon there is used to filter deactivated accounts. I am not sure if this increases confusion. Second, it introduces a new name "grouping". I think it is worth considering this, but then we should be more consistent.

Yes I see and agree, it can be confusing because a filter is created, but not visible immediately. For a new user, he may not understand where to use it next.

> I like your idea about the "grouped accounts" view. Essentially that is a "filter management UI" from the sidebar (instead of a dialog). I want to extend filters further - for example allow to include an account with a percentage (include a securities with 50% because it is shared with partners).
> 
> My take is:
>    * Let's not clutter the context menus in the cash and securities accounts view
>    * Let's convert the "manage filter" dialog into a "grouped account" sidebar
> 
> I like the idea of transitioning to "grouped accounts".

"Filter management UI" exactly, and that we can complement nicely with dedicated columns or bottom Pane. In the end, Attributes could be attached to them too. 
For the percentage with an account, I do not see yet the full implication, it does not seem easy to do, so for the moment I will focus of imitating the "manage filter dialog" into the new view. 

> Maybe:
>    *  Have a new item - that separates the "filter" from the "grouping"
>    *  Usually menus are separated, but in "Performance -> Securities" or "mater data --> all transactions" we separate it out
>    *  Rename it more consistently from "filter" to "group"

Ah, why not ! One menu for the real filter (Security held <>0 ..), one new menu for "account selection" (default+custom account). The "book/account" icon could have been the perfect menu icon but the 'active' mode is already the color for Cash Account. It would seriously unclutter the All Transaction filter menu for sure. Actually, this separation of menu is independent of a new "grouped account" managing view I think. 

-> I will propose something as a Draft/POC this week. Not optimized and still with lots of code duplication but at least to see if the UI feels nice.

buchen:
> The "book/account" icon could have been the perfect menu icon but the 'active' mode is already the color for Cash Account.

About the icon we could use something like this: https://iconmonstr.com/layer-21-svg/
I generate them on that site as PNG in 16 and 32 sizes with 1 and 2 pixel border and the appropriate colors

buchen:
I am marking this as draft. This is not to be merged "as is", but I want to keep it open to be aware of the conversation. Once you have an alternative pull request, we can close this one.

buchen:
closing. Continue discussion in #4466

PR #4466, opened 2025-01-11
add Grouped Accounts view (more visibility for account filters)

mierin12:
Follow up from https://github.com/portfolio-performance/portfolio/pull/4458

This is what I reached. To be carefully tested!

Some notes:

- Drag & Drop in the tree (copy paste from EditClientFilterDialog)
- I have kept the EditFilter dialog (still a useful shortcut). What is important then is that both the new view and the Edit dialog are synchronized. I think this is the case.
- New and Edit can be called from the + icon for the new Grouped Account view
- Add element and delete element on the Tree items (~copy paste from EditClientFilterDialog)
- Expand state of the tree is remembered (copy paste from Taxonomy's AbstractNodeTreeViewer)
- Three bottom Panes, I tried to adapt the existing ones to avoid duplication. Maybe there are cleaner way to adapt them.
- I tried the 1px border 16 * 16 icon and the 2px border for 32 * 32 icon, but I think I may not use the correct one, it feels smaller
- If confirmed that we go this way, there are still "filter" vs "grouped account" inconsistencies to be standardize. I did not do the change in case other names option are found.
- proposition of a column with the cash balance for all of them. If cash account currency is not the global one, the specific currency is used. But for sorting this column, the converted value is used. Is this the correct behavior ?
- what should be the behavior if a cash/security icon is selected in the tree ? Should the bottom panes be updated ? I think no, but maybe it could be updated to their parent Grouped Account.
 - optionally in the second commit : removal of the menu option allowing to clear all the filter. Seems quite dangerous ! 

I put in draft because I feel some optimization of duplication could be made, but it is ready for review.

In addition to this, to be tested to see if the UI feels "right". Is it disturbing to have a Tree while the Cash and Security Accounts are Tables ? Is the goal of Grouped Account clear enough (not used for transactions as opposed to the Cash/Security Account).

Capture d’écran 2025-01-11 025545
Capture d’écran 2025-01-11 030431

Bug:
~~-I am just seeing there is a bug for the tooltip in the center of the Holdings chart in this new view.~~ ok
-Second bug I am just noticing, more complex I think : the Note field for Grouped Account nodes is not remembered when changing view. 
~~-when no grouped account exists, the bottom pane for Holding is not good looking~~ ok


mierin12:
>  I am just seeing there is a bug for the tooltip in the center of the Holdings chart in this new view.

Fix proposed. It also changes the main HoldingsSWT chart view : the center tooltip name is updated with the selected filter.
This is adding a non mandatory snapshotName property in ClientSnapshot, maybe more null check could be added, or a default name.

mierin12:
>  -when no grouped account exists, the bottom pane for Holding is not good looking ok

Proposition of fix : hide the holding chart when the pane is open while no account is selected https://github.com/portfolio-performance/portfolio/pull/4466/commits/1666c5f60d1bc28ba082c719e187576fa5e8974c

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/GroupedAccountListView.java
```diff
@@ -0,0 +1,535 @@
...
+        tree.setLinesVisible(true);
+
+        ColumnEditingSupport.prepare(groupedAccounts);
+        ColumnViewerToolTipSupport.enableFor(groupedAccounts, ToolTip.NO_RECREATE);
+        CopyPasteSupport.enableFor(groupedAccounts);
+
+        groupedAccountColumns = new ShowHideColumnHelper(GroupedAccountListView.class.getSimpleName() + "@top2", //$NON-NLS-1$
+                        getPreferenceStore(), groupedAccounts, layout);
+
+        Column column = new NameColumn("0", Messages.ClientEditorLabelClientMasterData, SWT.None, 100, getClient()); //$NON-NLS-1$
```
buchen, line 253: Historically, the columns did not have a name. It was just the order as configured in the menu. That was a problem, because I could only add columns add the end. Therefore, some years back I added a name. To be backwards compatible, some columns still have the index as name. But now we can give proper names to the column.

buchen:
@mierin12 great first change for grouped accounts. I like it.

I have created a separate branch [feature_grouped_accounts](https://github.com/portfolio-performance/portfolio/tree/feature_grouped_accounts) and also a pull request #4476.

For now, I have squashed your commits into one and added a couple of additional commits. I will squash them as well, but while reviewing, it might help to look at the changes individually.

I think it should be possible for you to create a pull request for this branch as well - in case you notice more things. Most likely, I will not find time to work on this the next couple days.

I am closing this PR to continue in the new one.

PR #4472, opened 2025-01-15
StatementOfAssetsViewer: Show number of securities in each category

pfalcon:
Show it after category title in parens, e.g. "Category (nn)".


pfalcon:
For a long time, I wondered why PP doesn't show such a basic info as number of securities in different views. Here's attempt to do something about that.

PR #4473, opened 2025-01-15
SecuritiesChart: Add "Show/hide markings" toolbar icon

pfalcon:
Chart markings are incredible useful, but they also may crowd a chart, especially at the longer timeframes. The idea is to allow user to quickly hide or show (default state) of markings, to either enjoy all the annotations that they provide, or vice versa, a pristine price chart.

Regarding what's being hidden are buy/sell operations, dividends, events, and additionally limit price lines. The idea regarding the latter is that the limit(s) may be considerably lower/higher than the level of prices on the chart shown, leading to skewed vertical scale. But other information, e.g. average purchase price, is shown in either case.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -704,6 +704,7 @@ public class Messages extends NLS
     public static String LabelHeatmapOrnament;
     public static String LabelHeatmapTaxes;
     public static String LabelHide;
+    public static String LabelHideMarkings;
     public static String LabelHistoricalReturnsAndVolatiltity;
     public static String LabelImportNotesFromSource;
     public static String LabelIncludeSecuritiesInPieChart;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
@@ -1400,6 +1400,8 @@ LabelHeatmapTaxes = Monthly taxes
 
 LabelHide = Hide
 
+LabelHideMarkings = Show/hide markings
+
 LabelHistoricalReturnsAndVolatiltity = Historical Returns and Volatility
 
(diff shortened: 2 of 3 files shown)


pfalcon:
@OnkelDok : I wonder if you have any comments here. I didn't see it fitting ChartToolsManager, because it seems to be for the tools which activate new action modes for chart, where this just simple checkbox-like setting.

---

From my PoV, ideally it would be an icon with stroken eye, tooltip "Hide markings", default off, so a user pressed it to hide markings. But there's no such existing icon, so I took one of the existing icons, which, with some imagination, can be seen as "pen drawing some annotations", so it's default state is pressed (show markings) and user would un-press it to hide them.

buchen:
merged.

I have added an icon. And added the translations. I like that the state is "transient".

Bildschirmfoto 2025-01-17 um 10 01 42

pfalcon:
@buchen : 

> I have added an icon.

Thanks, looks good!

> And added the translations. 

Thanks for continued taking care about this part, as discussed earlier, that's a part which I consider complicated for contributors (well, for myself). (If anything, I'd wish translations were handled as much as possible by translation teams, while your time was dedicated to processing patches to keep positive feedback loop for them).

> I like that the state is "transient".

Yeah, I think that's how it is intended to be: if user configured all the annotations, then apparently they want to see them by default. On occasions when they obscure chart, they can temporarily disabled them, but they auto-reset to enabled state even if user forgot to re-enable them. I see that the current behavior is that "disabled" state is kept while navigating within the same watchlist, but switching to another watchlist, or to completely different view, resets it. Again, all sounds good IMHO, and I'd let someone else argue that it should be different.

Otherwise, looking at it closely, I notice following:

* In the commit message, I wrote: "But other information, e.g. average purchase price, is shown in either case.". Well, the line showing current (averaged) purchase price isn't actually shown, and that's good, because it may be far away from the current price range on the chart, at that again may skew the vertical scale of the chart. What's shown is the chart shading whether it's below/above the average purchase price.
* I see that min/max annotation is not shown. Maybe it's worth showing after all (2 such markings per chart, shouldn't crowd much). Not a big deal, someone can look into it later.

PR #4487, opened 2025-01-24
AbstractSecurityListWidget: Show infopane on clicking a security

pfalcon:
Previously, there was a code to update infopane to show security on which user clicked in a dashboard list widget. But most people would probably never know about this functionality, unless they happen to have infopane enabled in the dashboard (which normally isn't). So, now explicitly show infopane on click, so this useful functionality didn't go unnoticed.


pfalcon:
@hug-sch: Btw, does manual mention that clicking security name in various dashboard widgets shows clicked security in the infopane? Does manual mention that Ctrl+L (show/hide infopane) is a useful keyboard accelerator throughout the various PP views?

PR #4478, opened 2025-01-17
ReviewExtractedItemsPage: Honor SecurityNameConfig

pfalcon:
I.e. render security name with one of prefixes selected by user.

PR #4497, opened 2025-01-27
Security list widgets: Use security name config

pfalcon:
I.e., if user configured a security name prefix like ticker symbol, ISIN, etc., show it like that.


pfalcon:
Another set of one-liners, following https://github.com/portfolio-performance/portfolio/pull/4478

PR #4499, opened 2025-01-27
Modify comdirect PDF importer to support foreign exchange transactions

gh-nw-pp:
First of all: Thanks for the great project and effort you've put into it!

I found that PP does not properly detect international bank transfers in comdirect PDF import (see test file)


buchen:
Bildschirmfoto 2025-02-02 um 20 18 19

BTW, that can make it pretty hard to understand the merge. I propose to do it the other way around: rebase the feature branch onto master.

Nirus2000:
I'll add some transactions... then i merge it.
Missing is taxes by 0,00 € (Transaction is not needed)

Nirus2000:
i'll add @gh-nw-pp to "contributers"... then i merge it...

PR #4500, opened 2025-01-29
SecurityListView: Properly show short holdings for "Shares held != 0" filter

pfalcon:
Don't forcibly clamp negative (short) holdings to zero. However, add comment that the calculation is still not entirely correct, as it may return 0 in case one account have positive number of shares, and another - same number, but negative amount of shares.


pfalcon:
@buchen : I would like to make some steps towards (better) support short positions. It's hard, because full short support is of course very distant, and it's hard to see where  even to start. Well, could start with small things, where there're ungrounded, arbitrary restrictions not compatible with short positions, like here.

PR #4505, opened 2025-02-01
adding further bookings to trade republic pdf importer

christen90:
adding inbound and outbound Echtzeitüberweisungen.


Nirus2000:
Good job 💯

Issue #4507, opened 2025-02-02
SecuritiesChart: FIFO vs MA price drawing is fully duplicated


pfalcon:
* createFIFOPurchaseLineSeries() vs createMovingAveragePurchaseLineSeries()
* addFIFOPurchasePrice() vs addMovingAveragePurchasePrice()

`-` are duplicated functions, differing only in dependent methods they call / values they use. Duplication means that a bugfix would be applied to only one of copy, that they diverge over time, etc. Instead, they should be refactored to be parameterized via arguments or use `switch` to capture differences between methods.

sumansuhag:
Hi!
Pfacon, I really appreciate you bringing this important issue of code duplication to our attention. You’re spot on; it’s a major contributor to our technical debt, which can really affect how maintainable and consistent our code is, especially when it comes to sensitive financial calculations.

We should aim to refactor this into what I like to call a 'Composable Financial Calculation Engine.' This would involve abstracting the core logic—perhaps using a strategy pattern or functional programming—so that we can easily inject the specific FIFO or Moving Average logic as needed. Not only will this tackle the current duplication, but it will also pave the way for smoother integration of any new accounting methods we might want to adopt in the future.

Let’s make this refactoring a priority to enhance our code health and ensure that any future bug fixes are applied consistently. Your thoughtful suggestion about using parameterization or switch statements is a great place to start!

PR #4509, opened 2025-02-03
SecuritiesChart: Factor out createPriceLineSeries()

pfalcon:
Instead of largely duplicated createFIFOPurchaseLineSeries() and createMovingAveragePurchaseLineSeries(), factor out reusable createPriceLineSeries() which accepts additional params to handle both cases.

As part of the refactor, largely duplicated-content messages like LabelChartDetailMarkerPurchaseMovingAverage vs
LabelChartDetailMarkerPurchaseMovingAverageHoldingPeriod were simplified: now there's one main message for series label, and reusable suffix message for holding period number.


pfalcon:
This addresses one case of duplicated code from https://github.com/portfolio-performance/portfolio/issues/4507

PR #4512, opened 2025-02-04
Modify WIR Bank PDF-Importer to support VIAC Invest transactions

gh-nw-pp:
VIAC launched a new product ([VIAC Invest](https://viac.ch/produkte/invest/)), which uses a PDF format that is very similar to its existing product / WIR Bank PDFs

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/Kauf12_ViacInvest.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/Kauf12_ViacInvest.txt
@@ -1,0 +1,29 @@
+PDFBox Version: 3.0.3
+Portfolio Performance Version: 0.74.1.qualifier
+System: win32 | x86_64 | 21.0.5+11-LTS | Eclipse Adoptium
+-----------------------------------------
+VIAC Invest AG
+Strasse 1
+1111 Stadt
+E-Mail info@viac.ch
+Telefon 0000 00 00 00
+www.viac.ch
+Vertrag 1.111.111.111
+Portfolio 1.111.111.111.11
+Herr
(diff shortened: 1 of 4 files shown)


Nirus2000:
Hello @gh-nw-pp 

thank you very much for this pull request.
To ensure a good overview and further support of different banks and brokers, it is important to keep the structure of the importers. ([Contributing rules](https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#pdf-importer))
Importers are created for each supported bank and/or broker. 
Therefore, a separate PDF importer must be created for _VIAG Invest AG_. If you need help, please let me know

Greetings
Alex

gh-nw-pp:
Hi @Nirus2000,

I understand your concern, but VIAC is part of WIR Bank (check the [VIAC product page in the WIR Bank website](https://www.wir.ch/de/produkte-loesungen/privatkunden/vorsorgen/3-saeule/viac), or check the [Impressum on the VIAC website](https://viac.ch/en/impressum/), which contains WIR Bank).

The "VIAC Säule 3a" ("Säule 3a" is a private pension plan in Switzerland) PDFs are already processed using the WIR Bank PDF importer, so I just added support for the "VIAC Invest" PDFs as well.
I don't have access to any PDFs issued by WIR Bank directly, but I assume, that they are (almost) identical. But for some reason, the wording on the PDFs for VIAC Invest is slightly different ("Kauf" vs "Zeichnung" or "Verkauf" vs "Rücknahme"), plus the bank identifer/header of the PDF differs:
grafik

Even though WIR Bank and VIAC Invest AG are two different entities, I would argue, that it would make sense to keep the same PDF importer for both of them. But if you insist, I can create a completely new PDF importer for VIAC Invest, which would pretty much be copy-paste of the WIR Bank importer.

Nirus2000:
Okay. you've convinced me :-)

Issue #3303, opened 2023-04-19
Rant: Please make Portfolio Performance friendly for geeks


pfalcon:
Portfolio Performance is incredibly powerful tool, it is also incredibly user friendly, as much as an open-source software can be end user friendly. But there're limits on how much software can be user-friendly, while being resilient to errors and similar. And as long as software depends on 3rd-party world (network services and APIs), it's in the constant state of breakage. https://github.com/buchen/portfolio/issues/3275, https://github.com/buchen/portfolio/issues/3302 are just couple of examples seen by a novice user. In both cases, discovering that something was wrong (beyond "it doesn't work!!11"), and what exactly was wrong, was pretty hard. Definitely harder than it could have been. So, can we please:

1. Have console logging enabled by default, and all the errors, etc. be logged there with enough context (like URLs accessed).
2. Actually more then that, it would be nice that errors lead to nice error dialogs in user's face with beautiful hundred-lines stacktraces. That way the errors won't go unnoticed. (Balancing that of course not easy, a user shouldn't get 100 dialogs in row, so that's not replacement for console log).

I.e. let those users who know how to deal with software errors, be able to deal with them without extraordinary effort on just getting access to them. And even users who don't know how to deal with them, at least would be able to report them with objective details.

Thanks for listening.

Nirus2000:
Hmm... if you don't quite like each other, then please stick to the technical, factual level.
This is getting childish... 
It is a 10 years old open source project and everyone can contribute. The time spent is more towards functionality and not making an error log more accessible or easier to read for beginners. 
I also see this point not as a BUG but as a FEATURE.

chirlu:
I don’t know @pfalcon from before. What I don’t like is this behaviour (in which he is not alone) of “I don’t immediately find some feature – let’s not ask nicely on the forums whether I might have overlooked it or whether there is some other way to achieve the desired result, let’s immediately file a bug report with some condescending remarks about how bad everything is”.

Nirus2000:
I'm right there with you that this is out of place and lacks netiquette.

pfalcon:
> “I don’t immediately find some feature
> let’s immediately file a bug report 

Trust me, I do my homework on looking for things, including in the source code, and post only then (in the worst case, just a reminder for myself that I have done some work - but of course, always with the hope that someone else may find it useful).

[...]

PR #4519, opened 2025-02-09
PortfolioLog: Log calling class for string messages

pfalcon:
Otherwise, it may be hard to identify where a problem occurs.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/PortfolioLog.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/PortfolioLog.java
@@ -92,7 +92,10 @@ public class PortfolioLog
      */
     public static void error(String message)
     {
-        log(new Status(IStatus.ERROR, PLUGIN_ID, message));
+        StackWalker walker = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
+        Class<?> callerClass = walker.getCallerClass();
+
+        log(new Status(IStatus.ERROR, callerClass.getName(), message));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


pfalcon:
This is change along the lines of https://github.com/portfolio-performance/portfolio/issues/3303 which I implemented quite a long ago, and I guess would be really helpful to other folks who may work on hacking PP.

Example of output:
```
!ENTRY name.abuchen.portfolio.snapshot.security.CostCalculation 2 0 2025-02-09 17:19:39.329
!MESSAGE Negative holdings during FIFO cost calculation. ...
```

Now it's clear where to look to understand a problem.

buchen:
I am worried about the performance impact. Throwable::getStackTrace was always a problem. StackWalker might be more efficient. But the typical logs I see have a many errors/warnings in the download of prices (no surprise, it is scraping after all). Do you have experience there?

The alternative would be to change the signature to ```error(Class<?> caller, String message)```

pfalcon:
@buchen : 

> But the typical logs I see have a many errors/warnings in the download of prices (no surprise, it is scraping after all).

Well, a price update operation would be network I/O bound, not logging bound, I guess. I also have a lot of logging errors (even with stacktraces) with Yahoo updates, and no, I wouldn't say there's any performance impact.

I have a test file with many short positions, so there're hundreds of logging messages from `CostCalculation` and other `*Calculation`, like given in a comment above, so that's purely CPU-bound processing, and still, with that logging, there's no noticeable performance impact.

> The alternative would be to change the signature to error(Class<?> caller, String message)

Do you mean to modify each call in the codebase to pass a reference to class where that call is located? Well, that would be too noisy a change, and overall making logging more complicated, whereas here the idea is how to get more from existing logging "for free" ;-).

buchen:
ok, if you haven't noticed anything, I take your word for it:-)

PR #4520, opened 2025-02-09
Trade: Add toString()

pfalcon:
Useful during developement to debug various algorithms and edge cases.


pfalcon:
Another changes along the lines of https://github.com/portfolio-performance/portfolio/issues/3303 . To debug this, and much more diagnostic output will be required. Actually, similar to https://github.com/portfolio-performance/portfolio/pull/3913, I'd propose to add toString() to as many internal classes as possible (this includes various "line items", etc.)

buchen:
I am not a fan of those debug statements. But I understand they are helpful to understand what is going on. And with the ```<``` and ```>``` one can also immediately see that this is not shown in the UI.

PR #4541, opened 2025-02-23
Add new BAWAG AG PDF-Importer

christen90:
adding bank identifier to enable barclays credit card bill starting February.

also see https://www.barclays.de/ueber-uns/newsroom/uebernahme-von-barclays-consumer-bank-europe-erfolgreich-vollzogen/


christen90:
Hello Alex,
ok. PR updated.

BR Christian

Issue #4503, opened 2025-02-01
Option between FIFO and MA selling stocks


MrTempiko:
It would be a nice feature, if we could choose between FIFO (First in, First Out) and MA (Moving Average) method when selling stocks. I am not from Germany, so my broker uses the MA method.

Because of this, my gross profit is different in PP and my broker.

Are there any plans for adding this feature in the future?

mierin12:
I agree it would be nice. The difference does not occur when selling (it is the same sell) but in the calculation afterwards.
In some area there are already the option to show MA realized/unrealized gains (columns Performance/Capital Gains MA), but I think it would be worth to have it in the Calculation tab/widget. I once tried to look into the code to see how to adapt this part to MA, but it was not super straight forward. I can try again but no promise of success.

MrTempiko:
Definitely would be nice, if you could try again, as I have no coding skills at all.

MrTempiko:
Definitely would be nice, if you could try again, as I have no coding skills.

MrTempiko:
> I agree it would be nice. The difference does not occur when selling (it is the same sell) but in the calculation afterwards. In some area there are already the option to show MA realized/unrealized gains (columns Performance/Capital Gains MA), but I think it would be worth to have it in the Calculation tab/widget. I once tried to look into the code to see how to adapt this part to MA, but it was not super straight forward. I can try again but no promise of success.

Any progress yet? Can I support you in any way?

mierin12:
> Any progress yet? Can I support you in any way?

Hello, yes I think I managed to get something !
Work in progress here :
https://github.com/portfolio-performance/portfolio/compare/master...mierin12:portfolio:calculation_FIFO_vs_moving_average

It applies to 

- The Calculation view (including Forex capital gains)
- The Securities Performance view with two new columns (including Forex capital gains)
- The Calculation widget

I still need to polish a bit the different labels for all languages and add more tests.
Then what remains are the "record", what you can see as tooltips in the Calculation view. But maybe for a first implementation it can be done without them.

MrTempiko:
Looks promising, can‘t wait to test ist.

One question: Is the moving average method also added in the Trades view?

mierin12:
Ah no, I missed it. Maybe not complicated, what I understand : the Trades themselves are the same, but it would mean duplicating the columns:

- Entry value

- Entry value per share

- Profit loss

- Gross profit loss

As per Moving average ? That's all ?

[...]

PR #4546, opened 2025-02-25
add Moving Average Capital Gains

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4503

Hello,

This is a proposition to add the Capital Gains as per Moving Average cost method in :

- The Performance Calculation view
- The Performance Calculation widget
- The Securities Performance view (more columns)

FIFO remains per default in the Performance Calculation view and widget.

There were already Calculation Gains as per Moving Average in the Securities Performance view, but only on the 'current holdings'.

**Performance Calculation view**
2025-02-25 00_07_01-Portfolio Performance

**Securities Performance view** 
2025-02-25 00_07_45-Paramètres

**Performance Calculation widget**
2025-02-25 00_08_17-Paramètres

Forex Gain : added, but I am not 100% confident on it. I wanted to add tests, but I did not managed to understand how to correctly represent the transactions in the test. Neither the transactions directly in the forex currency, neither the ones with an exchange rate.

Trail Record for the tooltip : I need to deep dive into it, to see first if the exact same structure can be reused or if it is custom made for FIFO. I propose to look into it in a second step.

Transfer : unsure if something has to be done here ? In the `CostCalculation.class`, the Transfer is only used for FIFO cost.
For the new class `CapitalGainsCalculationMovingAverage`, I used a bit of `CostCalculation `and `CapitalGainsCalculation `logics.

I took the opportunity to add Descriptions to those columns of Securities Performance view.


mierin12:
> Merging is blocked
> 
>     Merge is not an allowed merge method in this repository.
>     This branch must not contain merge commits. 
Did I make a mistake ? I rebased locally to the latest master, it was not a merge. I have an un rebased branch if this is preferred.

buchen:
Thanks @mierin12 as always. I know that also still have the "grouped accounts" change open. The week after next week I have vacation. That will give me the time to work on your pull requests. I will now create a new release but don't have the time to pick up this change.

buchen:
> Forex Gain : added, but I am not 100% confident on it. I wanted to add tests, but I did not managed to understand how to correctly represent the transactions in the test. Neither the transactions directly in the forex currency, neither the ones with an exchange rate.

I think we first have to define how to handle forex gains when using the moving average.

In the case of FIFO it is this:
* When a sale occurs, the code checks what if the user just held the foreign cash instead of purchasing the instrument with the cash. This amount is taken as forex capital gains.
* The forex capital gains are only influenced by changes in the exchange rate between the base currency and the foreign currency of the instrument.
* If the instrument is traded in the reporting currency, then the forex gains are always 0

**What does it mean for "capital gains on moving average"?**

I would argue this:

* Because we averaged the purchase price, we also need to average the exchange rate.
* When a sale occurs, the code checks what if the user held the foreign cash (based on the average exchange rate and the current exchange rate). This is the amount taken as forex capital gains.

Let's consider those two purchase transactions:
```
Date       Shares Price (USD) FX Rate (USD/EUR) Cost (EUR)  Total Shares  Avg Cost (EUR)
2024-01-01 100    50          0.90              4500        100           45.00
2024-02-01 50     60          0.85              2550        150           47.00
```

We have an **average cost** of: 47 EUR = (4500 EUR + 2550 EUR) / 150 shares

Now we add a sale:
```
Date       Shares Sold  Sale Price (USD)  FX Rate (USD/EUR)  Revenue (EUR)
2024-03-01 50           70                0.88               3080
```

**Averaged Purchase price in EUR**:
47 EUR * 50 shares = 2.350 EUR

**capital gains on the moving average**:
Revenue in EUR - Averaged purchase price in EUR = 3080 EUR - (47 EUR * 50 shares)  = 730 EUR

**average exchange rate**:
( (0.90 * 100 pieces) + (0.85 * 50 pieces) ) / 150 pieces =  0.883333333333

**forex gains**
( (Average purchase price EUR * average exchange rate) * current fx rate) - average purchase price EUR
( (2350 / 0.88333333) * 0.88) - 2350 = -8,8679156938

Because the average exchange rate went from 0.88333333 to 0.88 (at the point of sale), roughly 8.8 are lost due to exchange rate fluctuations. In other words: If the user had invested the 2350 EUR only in USD (assuming the average exchange rate because we are averaging here), then he would have lost 8.8 EUR due to exchange rate.


@mierin12 @MrTempiko What do you think of this definition?

If we agree, I believe we have to update CapitalGainsCalculationMovingAverage and maybe check whether it makes sense to calculate the averages on the total or on the position in order to minimize rounding errors.

mierin12:
Hello @buchen, I admit that forex are not clear for me. 
I had not thought about the average exchange rate. I do not know what is correct, but for comparison I am writing what I had in mind :

instead of
 ( **(Average purchase price EUR * average exchange rate)** * current fx rate) - average purchase price EUR
( (2350 / 0.88333333) * 0.88) - 2350 = -8,8679156938

what I had in mind was 

 ( (**Average purchase price USD**) * current fx rate) - average purchase price EUR
( (8000 USD / 150 * 50) * 0.88) - 2350 EUR = -3.3333 EUR
 ( 2666.67 USD * 0.88) - 2350 EUR = -3.3333 EUR
50 * [ (8000 USD / 150) * 0.88) - 47 EUR ] = -3.3333 EUR
which is equivalent to
50 * [ (100 * 50 USD) * (0.88-0,9) + (50 * 60 USD) * (0.88-0,85) ] / 150 
I am not sure if this make financial sense.

Average cost is saying "all my shares costed 47 EUR each, whatever their purchase date"
For forex, should we say "all my shares costed 8000 USD / 150 = 53.33 USD each, what if I held this 50*53.33 = 2666.67 USD cash" or "all my shares costed 47 EUR each at a identical (averaged) exchange rate, what if held that USD cash". 

(to be noted that I am not even getting this -3.33 realized forex gain from the PR, as my PR is not able to get the fx rate defined in the transactions but only the historical fx rate; and in this example they are different)

mierin12:
Using the wording found in the [manual ](https://help.portfolio-performance.info/en/reference/view/reports/performance/calculation) :
For the moment I do not consider the sell, but I am checking the unrealized forex gain on March 1 2024 (FX Rate at this date : 0.88)

|Type| Date       | Shares | Price (USD) | Cost (USD) | FX Rate (USD/EUR) | Cost (EUR) |
|--| ------------- | ------------- |---| --- |--- | --|
|Buy| 2024-01-01 | 100    | 50 | 5000 |0.90 |4500|
|Buy| 2024-02-01 | 50     | 60 | 3000 |0.85 |2550|

The 5000 USD purchased on Jan 1, 2024, were valued at 4500 EUR, whereas on March 1, 2024, they were valued at only 4400 EUR. Thus, quote price remaining unchanged, the investment (expressed in EUR) decreased already by 100 EUR due to the difference in the exchange rate. 

The 3000 USD purchased on Feb 1, 2024, were valued at 2550 EUR, whereas on March 1, 2024, they were valued at only 2640 EUR. Thus, quote price remaining unchanged, the investment (expressed in EUR) increased already by 90 EUR due to the difference in the exchange rate.

Total : -10 EUR of unrealized capital gain on March 1, 2024 due to the difference in the exchange rate. 
The 8000 USD were valued at 4500 + 2550 = 7050 EUR, whereas on March 1, 2024 were valued at only 7040 EUR . 
Over my whole 150 shares.

Now I am selling 50 shares on March 1, 2024. I can directly do -10 * 50/150 = -3.333 EUR.
or 
My 50/150 * 8000 USD = 2666.67 USD were valued at 50/150 * 4500 + 50/150 * 2550 = 2350 EUR, whereas on March 1, 2024 were valued at only 2346.667 EUR. Thus, quote price remaining unchanged, the investment (expressed in EUR) decreased already by -3.333 EUR due to the difference in the exchange rate.
And I have a new unrealized forex capital gain of -10*100/150 = -6.6667 EUR 

That is how I see it.

buchen:
Hi @mierin12,

I agree to your way to calculate the average exchange rate. In this example, it should be 0.88125.

(My example was assuming that I can weight the exchange rates according to the number of shares. That is wrong. It must be weighted according to the value of the transaction.)

Let me go through the rest of the code. Maybe I will also add the calculation we used here as a test case.

BTW, I really enjoyed this conversation. Thanks for taking the time to think through the examples. It really makes the implementation better. Thanks.

buchen:
> Trail Record for the tooltip : I need to deep dive into it, to see first if the exact same structure can be reused or if it is custom made for FIFO. I propose to look into it in a second step.

Yes. It makes sense as a second step. The same way we now took quite some time to "unpack" the calculation, it would help the user to understand what happens there.



BTW, this is now the calculation of forex:
https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationMovingAverage.java#L84-L113

And this is our example from this discussion put into the test case:
https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java#L169-L248


It is now merged :-) Thanks

In my personal portfolio, it does not make any material difference which method I use. I have only one position that changes. It is roughly 5K valuation and a difference of 6 EURs. Maybe it is because I am "buy and hold" and therefore do not have so many sale transactions (when rebalancing)....

MrTempiko:
Hey guys, sorry for my late response. Had a busy last week...
I personally do no Forex trading, so I didn't look into it deeply till now.

One question I have: Is the moving average also added in the Trades view for now?
For me, it makes the most sense in this view.

Nevertheless I think it's a good step, as the MA was always a feature many people missed, I think.

mierin12:
hello @buchen , I checked a bit and I am wondering if we made a mistake..
Because, if we keep this example but without the sale, this gives different results for forex unrealized gain in FIFO vs in moving average. Which is suspicious.
FIFO
2025-03-22 00_32_48-testforex2 - Excel

Moving average
Capture d’écran 2025-03-22 001515

It arises because different exchange rates are used : the 0.9 and 0.85 exchange rate of the transactions are different than the historical exchange rate for the same those days.

| Date| Transaction exchange rates| Historical exchange rates |
| ------------- | ------------- |--|
| 2024-01-01| 0.9   (=1/1.1111)|0.905  (=1/1.105)|
| 2024-02-01| 0.85 (=1/1.17647)|0.9247  (=1/1.0814)|

```
(4500/0.9+2550/0.85)/1.0827-(4500+2550)=338.93
(4500*1.105+2550*1.0814)/1.0827-(4500+2550)=89.62
```
with 1.0827 the rate for 22/03/2025

So in case the exchange rates put in the transactions are not the same as the historical ones, which one should be used ?

Further check based on the same example : two securities with the same USD cost on the same day, one directly from a USD cash account (security `test2`), one without (security `tst`, =transaction exchange rate defined).
For test2, historical exchanges rates are used both by the FIFO and moving average method, similar result
For tst, the results are different.

2025-03-22 01_01_54-Portfolio Performance
2025-03-22 00_55_51-testforex2 - Excel

Capture d’écran 2025-03-22 005654
Capture d’écran 2025-03-22 005851

Capture d’écran 2025-03-22 011221

PR #4555, opened 2025-02-27
import Time in function Import from CSV in case it is available

### LINKED ISSUE #4554: Time is ignored in CSV Importer
**Describe the bug**
After exporting a CSV with PP, and reimporting in another portfolio the time is ignored. 
**To Reproduce**
1. Export e.g. depot as CSV file. 
2. Close portfolio
3. Open another portfolio
4. Import the last CSV file

**Expected behavior**
Date and Time is taken from the PP CSV file
It is ignored.

**Screenshots**

**Solution**

Image

 - OS: [e.g. Windows]
 - Version [11]

pyBlockDev:
Closes #4554

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractorTest.java
@@ -112,6 +112,90 @@ public class CSVPortfolioTransactionExtractorTest
     }
 
     @Test
+    public void testTransferTransactionDateContainsTime()
+    {
+        Client client = new Client();
+        Security security = new Security();
+        security.setTickerSymbol("SAP.DE");
+        client.addSecurity(security);
+
+        CSVExtractor extractor = new CSVPortfolioTransactionExtractor(client);
+
+        List<Exception> errors = new ArrayList<>();
+        List<Item> results = extractor.extract(0,
+                        Arrays.<String[]>asList(new String[] { "2013-01-01T14:05", "", "DE0007164600", "SAP", "",
+                                        "SAP SE", "100", "EUR", "11", "10", "", "", "", "1,2", "TRANSFER_IN",
+                                        "Notiz" }),
+                        buildField2Column(extractor), errors);
+
+        assertThat(errors, empty());
+        assertThat(results.size(), is(1));
+        new AssertImportActions().check(results, CurrencyUnit.EUR);
+
+        PortfolioTransferEntry entry = (PortfolioTransferEntry) results.stream()
+                        .filter(PortfolioTransferItem.class::isInstance).findAny()
+                        .orElseThrow(IllegalArgumentException::new).getSubject();
+
+        PortfolioTransaction source = entry.getSourceTransaction();
+        assertThat(source.getType(), is(PortfolioTransaction.Type.TRANSFER_OUT));
+        assertThat(source.getMonetaryAmount(), is(Money.of(CurrencyUnit.EUR, 100_00)));
+        assertThat(source.getNote(), is("Notiz"));
+        assertThat(source.getDateTime(), is(LocalDateTime.parse("2013-01-01T14:05")));
+        assertThat(source.getShares(), is(Values.Share.factorize(1.2)));
+        assertThat(source.getSecurity(), is(security));
+        // security transfers do not support fees and taxes at the moment
+        assertThat(source.getUnitSum(Unit.Type.FEE), is(Money.of("EUR", 0)));
+        assertThat(source.getUnitSum(Unit.Type.TAX), is(Money.of("EUR", 0)));
+
+        PortfolioTransaction target = entry.getTargetTransaction();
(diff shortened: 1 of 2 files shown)


pfalcon:
Normal time format is `HH:mm:ss`. This change adhocly applies some random time format instead.

pyBlockDev:
Hi @pfalcon,

yes you are right. In my usecase if I export the  depot, then PP generates this:

image

This is now possible to read back with PP.
But if somebody has the T10:33:11  then we get an exception. If I use HH:mm:ss then my examples get an exception.
Maybe we could count the number of ":" and make a case decision.
What do you think?

Nirus2000:
Hello @pyBlockDev 
Thank you for the pull request.
If you take your time again, it would be good to have a look at the Contributing rules and of course check the test case for your change in your branch or create a suitable one.

Regards
Alex 😀

pyBlockDev:
Hi @Nirus2000,

thanks for your hint. I have added the testcases. Additional I have rechanged the code, use the existing structure and tried to simplify it.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractorTest.java
```diff
@@ -111,6 +111,90 @@ public void testTransferTransaction()
...
+        assertThat(source.getSecurity(), is(security));
+        // security transfers do not support fees and taxes at the moment
+        assertThat(source.getUnitSum(Unit.Type.FEE), is(Money.of("EUR", 0)));
+        assertThat(source.getUnitSum(Unit.Type.TAX), is(Money.of("EUR", 0)));
+
+        PortfolioTransaction target = entry.getTargetTransaction();
+        assertThat(target.getType(), is(PortfolioTransaction.Type.TRANSFER_IN));
+        assertThat(target.getUnitSum(Unit.Type.FEE), is(Money.of("EUR", 0)));
+        assertThat(target.getUnitSum(Unit.Type.TAX), is(Money.of("EUR", 0)));
+    }
```
Nirus2000, line 196:
```
Arrays.<String[]>asList(new String[] { "2013-01-01T14:05", "10:07", "DE0007164600", "SAP", "",
                                        "SAP SE", "100", "EUR", "11", "10", "", "", "", "1,2", "TRANSFER_IN",
                                        "Notiz" }),
```
Test result
```
assertThat(source.getDateTime(), is(LocalDateTime.parse("2013-01-01T10:07")));
```


you set the time to "T14:05" but in the result is "10:07"

Nirus2000 (changes requested):

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVPortfolioTransactionExtractorTest.java
pyBlockDev: Yes that is right, the reason is the Time column has a higher priority then the (Date+T) column in the PR implementaion. I thought about the possibility that somebody 
1. defines a Time in the date column with the "THH:MM" notation and 
2. **also** in the Time column.
 
In my opinion this is an error in the csv file, but I thought maybe it is better to handle this with a priority of the Time column and the current beahvior of the PP. Because at the moment in PP if the Time column is defined then this is used.

The PR has the following new behavior:

- if the Date column contains the "THH:mm" text and **no** Time column is defined, then the time is used from the Date
- if the Date column contains the "THH:mm" text and  Time column is defined, then the time is used from the Time.

pyBlockDev:
Thanks for everybody,

yes it is an adhoc implementation because it is only a bugfix. For me maybe took a little bit longer as I am unexperienced in PP :smile:.

In basics I wanted just give me found solution back to the community, I like the SW a lot. The code of PP is clean so I could change it, aslo without experience.

My basic wish would be that after a portfolio export in CSV, I can import this CSV file again in another portfolio. This was not working. Therefore I made the change.

In my CSV export from PP I see alwasys the following format:
image

I would have expected to import this again. The fix uses only one format and this is e.g. 2023-11-23T11:17. Because this one
is used by the export function.

I tried to implement this in a way that it has as less side effects as possible. 

I executed all the Testcase in JUNIT, all are passed.

buchen:
> Naturally, support for timed date formats would go into that array too. Patterns are tried in the order of appearance (and as we know, shorter patterns match, ignoring the rest of input string), so timed formats would need to go before the current format.

I agree. That is the right approach.

I'll push a change that is in line with that:
* change the default format use for date fields in tests (in order to not break the old tests)
* change the new test to set a specific date format string

> But if you make such a change, you'll find that existing tests in CSVExtractorTestUtil start to fail, because it hardcodes that 0th entry in the list should be used for tests, and now 0th entry as a different one.

I think that we change easily to keep the original one as default:
https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVExtractorTestUtil.java#L42

buchen:
merged. Thanks for your contribution!

PR #4576, opened 2025-03-05
sort for fifo also for transfers (#4446)

### LINKED ISSUE #4446: Incorrect FIFO order used when doing multiple transfers
I'm reporting [this bug from the forum](https://forum.portfolio-performance.info/t/berucksichtigung-des-initialen-anschaffungsdatums-bei-mehrmaligem-wertpapiertransfer/24435) here on GH for visibility (apologies if it already exists somewhere in German - I did my best to search but didn't find it).

Summary: transferring an asset between accounts more than once doesn't seem to reflect the proper FIFO order. Repro:

* Buy 1 unit on a crypto exchange (corresponding to one account in PP). Call it lot E1.
* Buy 1 unit in a wallet (corresponding to another account in PP). Call it lot W.
* Buy another on the same exchange. Call it lot E2.
* Transfer everything from the exchange to the wallet (2 units).
* Transfer 1 unit from the wallet back to the exchange. _It should pull E1 as it’s oldest._
* Sell 1 on the exchange. _It should sell E1_

However, it sold lot W.

8fdaa192ebb63f9e048f91bf47d84211643a9f3a
4bdce65eec3b1e02da96c600444a9f20e28cd42b
[test portfolio.xml.txt](https://github.com/user-attachments/files/18296070/test.portfolio.xml.txt)

There's more context & examples in the forum thread. This is quite impactful in a variety of scenarios - here's some that have come up on the forum:
* In crypto, it's extremely common to move assets around between wallets & exchanges. The US requires separate cost basis tracking per-wallet for tax purposes. This bug means that PP ultimately sells the wrong lot & shows the wrong cost basis, as illustrated above.
* A number of users have tried to use transfers to get around the limitation that PP doesn't yet support custom orderings (i.e. HIFO / LIFO / specific lots / etc) - i.e. move shares out of an account, sell the desired lot, then move them back. This also causes the bug.
* Moving equities around for custody transfer bonuses

ralisi:
Include sorting of Transactions when transferring from one portfolio to another in the same manner as is already done when selling. This fixes #4446

### DIFF
--- a/name.abuchen.portfolio.tests/src/issues/Issue4446FIFOMultipleTransfers.java
+++ b/name.abuchen.portfolio.tests/src/issues/Issue4446FIFOMultipleTransfers.java
@@ -1,0 +1,44 @@
+package issues;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.junit.Test;
+
+import name.abuchen.portfolio.junit.TestCurrencyConverter;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.ClientFactory;
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.money.CurrencyConverter;
+import name.abuchen.portfolio.money.CurrencyUnit;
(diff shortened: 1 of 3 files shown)


buchen:
thanks for the contribution. Fixed and merged.

PR #4577, opened 2025-03-06
Fix NullPointerException in TextUtil.compare

OnkelDok:
https://forum.portfolio-performance.info/t/fehlermeldung-internal-error/32110

I dont know why, but it seems at least one security name is ``null``. So I added a global fix for all users of ``TextUtil.compare`` and also created simple unittest.

---

One additional point I thought about: In the code there are plenty of this nullchecks for comparisons (e.g. https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java#L53-L58).
Would it be helpful/better to create a commonly used method where this nullcheckss are executed and use this everywhere?
Of course this is only possible by using ``Optional`` to also realize the path when both objects/values are not null.
Does that make sense?

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TextUtilTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TextUtilTest.java
@@ -191,4 +191,12 @@ public class TextUtilTest
         assertEquals("first-second", TextUtil.concatenate("first", "second", "-"));
     }
 
+    @Test
+    public void testCompareNullCheck()
+    {
+        assertEquals(-1, TextUtil.compare(null, "a"));
+        assertEquals(1, TextUtil.compare("a", null));
+        assertEquals(0, TextUtil.compare(null, null));
+    }
+
 }
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
@@ -325,6 +325,13 @@ public final class TextUtil
      */
     public static int compare(String left, String right)
     {
+        if (left == null && right == null)
+            return 0;
+        else if (left == null)
+            return -1;
+        else if (right == null)
+            return 1;
+
         return COLLATOR.compare(left, right);
     }
 


name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
```diff
@@ -325,6 +325,13 @@ public static String limit(String text, int maxLength)
      */
     public static int compare(String left, String right)
     {
+        if (left == null && right == null)
+            return 0;
+        else if (left == null)
+            return -1;
+        else if (right == null)
+            return 1;
```
Nirus2000, line 333:
```suggestion
        if (left == right) return 0; // Both are null or refer to the same object
        if (left == null) return -1;
        if (left == null) return 1;
```

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TextUtilTest.java
```diff
@@ -191,4 +191,12 @@ public void testConcatenate()
         assertEquals("first-second", TextUtil.concatenate("first", "second", "-"));
     }
 
+    @Test
+    public void testCompareNullCheck()
+    {
+        assertEquals(-1, TextUtil.compare(null, "a"));
+        assertEquals(1, TextUtil.compare("a", null));
+        assertEquals(0, TextUtil.compare(null, null));
```
Nirus2000, line 199:
```suggestion
        assertEquals(-1, TextUtil.compare(null, "a"));
        assertEquals(1, TextUtil.compare("a", null));
        assertEquals(0, TextUtil.compare(null, null));
        assertEquals(0, TextUtil.compare("abc", "abc"));
        assertTrue(TextUtil.compare("a", "b") < 0); // "a" should be less than "b"
        assertTrue(TextUtil.compare("z", "y") > 0); // "z" should be greater than "y"
        assertTrue(TextUtil.compare("abc", "ABC") != 0); // Depends on COLLATOR
        assertTrue(TextUtil.compare("café", "cafe") != 0); // Accents matter
```

Nirus2000 (changes requested):
Hello @OnkelDok 
Awesome thing from once again... :astonished:
I read your entry and think about it for a moment and then the aha! moment... :eyes:
Maybe the change fits better... take a look at it at your leisure.

Geil 👍🏻 

Regards
Alex

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/util/TextUtilTest.java
OnkelDok: Interesting idea. I don't though about it before. Mabe for strings this is fine. But I think I want only check for ``null`` and let the underlaying comparer make the decision for not-null objects.
It's up to @buchen what to use.

buchen:
merged.

PR #4582, opened 2025-03-08
FollowUpWidget: Group items by date

pfalcon:
Previously, the widget showed security name, and attribute name/date below it. Now, we start with showing attribute name, followed by date, and list all securities having the same attribute name and date. This saves vertical space, and clearly groups all securities having the same date attribute.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/lists/FollowUpWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/lists/FollowUpWidget.java
@@ -81,6 +81,9 @@ public class FollowUpWidget extends AbstractSecurityListWidget<FollowUpWidget.Fo
         }
     }
 
+    private LocalDate lastDate;
+    private AttributeType lastAttrType;
+
     public FollowUpWidget(Widget widget, DashboardData data)
     {
         super(widget, data);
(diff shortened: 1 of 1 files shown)


pfalcon:
Related discussion: https://github.com/portfolio-performance/portfolio/pull/3927#issuecomment-2706104520

## Screenshot before

followup-before


## Screenshot after

followup-after

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/lists/FollowUpWidget.java
```diff
@@ -133,13 +143,20 @@ protected Composite createItemControl(Composite parent, FollowUpItem item)
...
         name.addMouseListener(mouseUpAdapter);
-        date.addMouseListener(mouseUpAdapter);
 
-        FormDataFactory.startingWith(logo).thenRight(name).right(new FormAttachment(100)).thenBelow(date);
+        if (!item.date.equals(this.lastDate) || !item.type.equals(this.lastAttrType))
+        {
+            Label date = createLabel(composite, item.type.getName() + ": " + Values.Date.format(item.date)); //$NON-NLS-1$
+            FormDataFactory.startingWith(date).thenBelow(logo).thenRight(name).right(new FormAttachment(100));
+            this.lastDate = item.date;
+            this.lastAttrType = item.type;
```
buchen, line 154: In your example, if there is a refresh, wouldn't it show no headline at all?
There are two items that are collated into one. In the next rundown the "lastDate" and "lastAttrType" match also the first one.

I propose to not use a member variable. For reasons of simplicity, I added a "previous" item to the method.

buchen:
merged.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/lists/FollowUpWidget.java
pfalcon: > In your example, if there is a refresh, wouldn't it show no headline at all?

Yeah, apparently it required to override update() method to re-initialize those vars.

> I propose to not use a member variable.

Sure, if there's an easy way to avoid storing state, it's good idea. Thanks for the refactor!

PR #4600, opened 2025-03-27
Tooltip for quote change column to show annualized quote change

OnkelDok:
This is additionally or as alternative to #4303.

grafik

To achieve this the ``QuoteReportingPeriodLabelProvider`` now derives/extends ``ParameterizedColumnLabelProvider`` instead of ``OptionLabelProvider``. Reason for this change is that the tooltip cannot be set inside the ``update`` method of the ``OptionLabelProvider`` (``ViewerCell`` does not provide method to set the tooltip).
Additionally I did some refactoring to make ``ParameterizedColumnLabelProvider`` generic.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ParameterizedColumnLabelProvider.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ParameterizedColumnLabelProvider.java
@@ -8,7 +8,7 @@ import org.eclipse.swt.widgets.TableColumn;
  * with a column. This can be e.g. reporting period (and there can be multiple
  * instance of the same base column with different reporting periods set).
  */
-public class ParameterizedColumnLabelProvider extends ColumnLabelProvider
+public class ParameterizedColumnLabelProvider<O> extends ColumnLabelProvider
 {
     private TableColumn tableColumn;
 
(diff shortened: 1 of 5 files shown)


buchen:
@OnkelDok thanks for the change.

I noticed one thing:

Bildschirmfoto 2025-04-06 um 18 53 10

Notice that the change for 1 year in the table is 12.31% but the annual rate is 24.14%.
The reason is that this (test) file does not have prices for the full range available.
The column shows the value for the selected period, the annualized value is calculated on the actual difference between first and last price.

Either we calculate
* the annualized change on the selected period
* of we calculated as given, but provide some more details in the tool tip (say: this is the first and the last price point used which makes it X days and annualized this is Y percent).

@OnkelDok  what do you think?

OnkelDok:
@buchen Ah I see what you mean.
I dont have a favorit solution for this. Both would make sense in their own way. But also both can be a bit confusing.

Maybe option two will confuse less than the first one, because we will display all information how both values are calculated. And both values (quote changed and annualized one) are based on the same time range.

When we would calculate the p.a. value based on the configured/selected period then we would have to display all dates: the dates for the quotes for raw quote change and the dates for the annualized quote change. And they will differ.

Alternatively we could display both annualized values, and have to add information about the period they are based on. But This could overload the tooltip and maybe can confuse much more.

So I'm leaning a bit more to obtion b. Adding the details of the dates in the tooltip is a good thing in general to give more information about the displayed value.

Maybe we can also print the prices itself in the tooltip. So all values: prices, their dates and the annualized value.
grafik

buchen:
Personally, I also tend to lean to option b.

Adding some more information on the tool tip should also not be the problem. It is only calculated/retrieved when the tool tip is shown.

I like the proposal for the tooltip.
Just add the original value (for the selected period of the column) as well.

buchen:
Looks nice now.

PR #4605, opened 2025-03-29
Improve Usage of convertType in YahooSymbolSearch for Standardized Security Types

Nirus2000:
Refactored YahooSymbolSearch to use the convertType method from SecuritySearchProvider for standardizing security types, improving maintainability and consistency across search providers.


buchen:
Thanks. I think leeway and Twelvedata use EODHD data. Not sure if Yahoo  Finance uses the very same types.

I am also not sure what they are using. The OpenFIGI types are different: https://api.openfigi.com/v3/mapping/values/securityType (but of course with some overlap.)

PR #4611, opened 2025-03-30
Adding test cases for currency unit and currency symbol

Nirus2000:
When creating #4610 I noticed that the RegEx \p{Sc} can cause problems with the currency identification.
These test cases should solve this, but now there are new problems.


buchen:
Well, the symbols are not necessarily unique.

See here for the source: https://en.wikipedia.org/wiki/List_of_circulating_currencies

If you search for F.CFA, then this symbol is used for the West African CFA franc and for the Central African CFA franc.

BTW, the same is true for the $ sign. There are many currencies that use the $ sign to shorten it (United states dollar, singapore dollar, Candadian dollar).

I think we just cannot assume that the symbol is unique.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/money/CurrencyUnitTest.java
```diff
@@ -0,0 +1,199 @@
...
+    {
+        Set<String> currencyCodes = new HashSet<>();
+        for (CurrencyUnit currencyUnit : currencyUnits)
+        {
+            var currencyCode = currencyUnit.getCurrencyCode();
+
+            // Using a HashSet to ensure there are no duplicates.
+            assertTrue("Duplicate currency code found: " + currencyCode, currencyCodes.add(currencyCode));
+        }
+    }
```
buchen, line 49: This test does not make sense. The way the CurrencyUnit is implemented, there cannot be duplicate codes.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/money/CurrencyUnitTest.java
```diff
@@ -0,0 +1,199 @@
...
+            {
+                // Use HashSet to ensure there are no duplicates
+                var isAdded = currencySymbols.add(currencySymbol);
+
+                // Using a HashSet to ensure there are no duplicates.
+                assertTrue("Duplicate currency symbol found for currency code " + currencyCode + ": " + currencySymbol,
+                                isAdded);
+            }
+        }
+    }
```
buchen, line 77: I think this test is wrong, because we should allow multiple symbols.
The symbols could be used to display it.

I think for the importer, one has to understand from the context of the document which currency we are talking about. For example, if you just read a $ and it is from the United states, then it is the USD. But we cannot make that assumption for Canadian documents.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/money/CurrencyUnitTest.java
```diff
@@ -0,0 +1,199 @@
...
+     */
+    @Test
+    public void testCurrencyCodeLength()
+    {
+        for (CurrencyUnit currencyUnit : currencyUnits)
+        {
+            var currencyCode = currencyUnit.getCurrencyCode();
+            assertEquals("Currency code should have exactly 3 characters", 3, currencyCode.length());
+        }
+    }
```
buchen, line 110: Really? Did Copilot generate this test? 😄

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/money/CurrencyUnitTest.java
```diff
@@ -0,0 +1,199 @@
...
+        List<String> invalidCodes = List.of("US", "USD1", "1234", "EURO", "A1");
+        for (String invalidCode : invalidCodes)
+        {
+            var currencyUnit = CurrencyUnit.getInstance(invalidCode);
+            assertNull("Invalid currency code should not exist: " + invalidCode, currencyUnit);
+
+            assertFalse("Currency code should not be cached: " + invalidCode,
+                            CurrencyUnit.containsCurrencyCode(invalidCode));
+        }
+    }
```
buchen, line 149: What is the material difference to testInvalidCurrencyUnitsDoNotExistInCache?

name.abuchen.portfolio/src/name/abuchen/portfolio/money/currencies.properties
```diff
@@ -97,7 +94,7 @@ BRL.symbol = R$
 
 BSD = Bahamian dollar
 
-BSD.symbol = B$
+BSD.symbol = Bh$
```
buchen, line 97: According to Wikipedia, it is just $

Bildschirmfoto 2025-04-06 um 17 50 18

buchen:
So in short, I think we should not test for the uniqueness of symbols.

Is that a problem with the merged change?

PR #4616, opened 2025-04-02
Modify FlatEx PDF-Importer and implement getTickerSymbolForCrypto

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/277 
https://forum.portfolio-performance.info/t/pdf-import-von-flatex/1752/278 

------------------

Enhancement: Implement getTickerSymbolForCrypto Method Description

This enhancement introduces the getTickerSymbolForCrypto method, which maps cryptocurrency names to their respective ticker symbols. The method accepts a cryptocurrency name as input. It normalizes the input (trims whitespace and converts it to uppercase). A predefined mapping (cryptoTickerMap) returns the corresponding ticker symbol. If the name is missing or not found, an empty string ("") is returned.

Hello @buchen 
if you have time, can you please check the new method in the ExtractorUtils?

Thx 👍🏻 
Alex


buchen:
> if you have time, can you please check the new method in the ExtractorUtils?

In think the change looks okay.

There are no definitive ticker symbol for crypto. So using Coingecko makes sense. One could argue that the method rather belongs to Flatex because it is the names that Flatex uses. But I assume there will be future changes.

PR #4623, opened 2025-04-06
Modify ArkeaDirectBank PDF-Importer to support new transaction

mierin12:
Hello Nirus,

As per https://forum.portfolio-performance.info/t/pdf-import-from-arkea-direct-bank-fortuneo-banque-france/28131/22 and https://forum.portfolio-performance.info/t/pdf-import-from-arkea-direct-bank-fortuneo-banque-france/28131/25 I have added the Sale case for the importer. This is the only "Sale" example so far, it is the last transaction of Achat09.txt (it may be renamed as AchatVente01.txt ? Achat=Buy, Vente =Sell).
So this is also correcting the GrossAmount for this transaction, I understand that for a Sale, gross amount > net amount. And this new gross amount matches the value of Achat09.txt line 55.


Nirus2000:
Thx @mierin12 
I'm a bit busier than usual at the moment.

PR #4634, opened 2025-04-08
Adding Rebalancing charts

itshukychen:
Adding the following charts:
- New Stacked chart (bars over different categories)
- Rebalancing Stacked chart to view Actual vs Target over different taxonomy classes
- Rebalancing donut chart to see target allocation
- The ability to include/exclude securities in Taxonomy donut chart


buchen:
Thanks for the contribution. Give a couple days to have a look

buchen:
While going through the code, I have updated the code accordingly and merged it.

For easier readability, I have split the three features into three separate commits.

PR #4640, opened 2025-04-09
fix moving average calculation widget

mierin12:
Hello, I had some null exception when using the widget calculation with moving average method.


buchen:
My bad. Good spotting.

PR #4643, opened 2025-04-11
Modfiy BSDEX PDF-Importer to support new transaction

Nirus2000:
https://forum.portfolio-performance.info/t/pdf-import-von-bsdex-borse-stuttgart-digital-exchange/20155/17


Selevra:
Hi @Nirus2000,
just tried the importer in the released PP tool and have the same issued I already saw during development, that the wrong crypto asseets are matched to "BTC", "ETH" and Co.:
image
I think the logic needs to be improved to resolve ambigous results.

Also I wonder why this isn't an issue for other importers like Trade Republic etc.?!

PR #4652, opened 2025-04-13
Implement timezone field for JSON quote parser

jan-moeller:
The date format already provides a way to parse the time zone from the string via appropriate patterns such as V or Z (see `DateTimeFormatter` for available patterns), if it is part of the date representation chosen by the server.
However, if the server doesn't provide an explicit timezone, yet the timestamp is not in UTC, then there was no way to interpret it appropriately.

This issue has previously been discussed on the forums, for example [here](https://forum.portfolio-performance.info/t/how-can-i-adjust-date-extracted-from-json-as-unixtime/20054).

This PR allows the user to supply a fixed timezone if they know the server uses it to represent dates.

I have marked it as draft for the following reasons:
1. It's missing unit tests for the new functionality.
2. It's missing translations for the 2 new strings.
3. I'm looking for feedback whether this is an appropriate way to solve this issue. In particular, I'm not sure what timezone PP expects its dates to be in. I suspect UTC, in which case this change should make sense. However, there are other plausible candidates, and the use of `LocalDate` throughout the codebase makes me want to clarify that.

I'm going to work on 1. and 2. after 3. has been clarified.


buchen:
@jan-moeller Thanks for the contribution. It looks good to me. If you could add a unit test to safeguard the feature, that would be helpful.

> In particular, I'm not sure what timezone PP expects its dates to be in.

That is difficult question to answer.

In the file format, the dates of historical prices are just stored as ISO dates, say 2025-04-18. Since PP does not do intraday calculations, it assumes those are the end of day prices. And it depends on the exchange which date this is - say XETRA at 18:00 CE(S)T and New York at 22:00 CE(S)T. And because the timezone is stripped, this leads to certain blur. When XETRA closes, there is already a new day in Australia. But I still think it kinda works to assign the EOB close to that day.

It becomes more difficult with instruments which are continuously traded, say bitcoin. As "end of day" price, do you take UTC or do you take the local timezone of the user? And what if the user travels? Storing more than one price point per day would have a significant impact on the file size. Here it is UTC at the moment.

jan-moeller:
Thank you, @buchen for the feedback! After reading your comment I realized that (as is often the case with time zones) it seems the problem isn't quite as straight forward, as I initially thought. There are two reference points:

1. The timezone of the exchange
2.  a. In case of calendar dates: the timezone of the date in the JSON
    b. In case of UNIX-style timestamps: the epoch timezone

Here, 1) is implicit in PP, so its up to the user to make sure the parsed date is in that timezone. 2a) could either be given as part of the JSON data, or provided by the PP user. 2b) ought to always be 1970-01-01 00:00:00 UTC, but [in practice](https://forum.portfolio-performance.info/t/aktuelle-kurse-uber-ing-api-abrufen/14494/17) non-UTC epochs are sometimes used.

Since the exchange timezone is not tracked, there is no way to automatically compute the correct time based on the timestamp timezone alone. I therefore opted to make the assumption that exchange timezone and timestamp timezone are, by default, the same. While I don't have any data on this, plausibly this might be the most common case.

Under this assumption, "2011-12-03" and "2011-12-03+01:00" are the same date (the second one is just explicit that the TZ offset of the exchange is +01:00). This PR therefore ignores any timezone information of the timestamp except for ensuring its validity.

Instead, all calendar dates are now parsed into a `LocalDateTime` first, then offset by any user-supplied timezone/offset information, then converted to `LocalDate`. If the calendar date doesn't come with time information, then the **end** of that day is assumed, in line with how PP interprets the dates.

For example, "2011-12-03" is assumed to be 2011-12-03T23:59:59 when the timezone transformation is applied to it, while "2011-12-03T10:15:30" is used as-is.

Unix-like timestamps work analogous, except that they always carry time information.

If no timezone/offset is provided, behavior should be identical to how it was before.

Since I'm having a hard time to explain this in words, here is a table showing input date, format and timezone configuration, and what the result is:


| date (JSON)                              | format (GUI)           | timezone (GUI) | result                                                                        | comment                                 |
| ---------------------------------------- | ---------------------- | -------------- | ----------------------------------------------------------------------------- | --------------------------------------- |
| 2011-12-03                               |                        |                | 2011-12-03                                                                    |                                         |
| 2011-12-03+01:00                         |                        |                | 2011-12-03                                                                    |                                         |
| 2011-12-03T10:15:30                      |                        |                | 2011-12-03                                                                    |                                         |
| 2011-12-03T10:15:30+01:00                |                        |                | 2011-12-03                                                                    |                                         |
| 2011-12-03T10:15:30+01:00[Europe/Berlin] |                        |                | 2011-12-03                                                                    |                                         |
| 2011-12-03 10:15:30                      | yyyy-MM-dd HH:mm:ss    |                | 2011-12-03                                                                    |                                         |
| 2011-12-03 10:15:30 -13:00               | yyyy-MM-dd HH:mm:ss VV |                | 2011-12-03                                                                    |                                         |
| 2011-12-03                               |                        | +01:00         | 2011-12-04                                                                    | end of day + 01:00 is next day          |
| 2011-12-03-01:00                         |                        | +01:00         | 2011-12-04                                                                    | timezone information of date is ignored |
| 2011-12-03T10:15:30                      |                        | -10:16         | 2011-12-02                                                                    | -10:15 is still 2011-12-03              |
| 2011-12-03T22:45:00                      |                        | Europe/Berlin  | 2011-12-03 (standard time +01:00)<br>2011-12-04 (daylight saving time +02:00) |                                         |
| 2011-12-03 10:15:30                      | yyyy-MM-dd HH:mm:ss    | -10:16         | 2011-12-02                                                                    | -10:15 is still 2011-12-03              |
| 1322907330                               |                        |                | 2011-12-03                                                                    | Refers to 2011-12-03T10:15:30Z          |
| 1322907330                               |                        | -10:16         | 2011-12-02                                                                    | -10:15 is still 2011-12-03              |
| 1322952300                               |                        | Europe/Berlin  | 2011-12-03 (standard time +01:00)<br>2011-12-04 (daylight saving time +02:00) | Refers to 2011-12-03T22:45:00Z          |

I have also added unit tests to cover these cases.

If the assumption that 1) and 2) are in the same time zone doesn't hold, then it doesn't make sense to provide an actual timezone like "Europe/Berlin", because the offset would have to be computed from both the timestamp timezone and the exchange timezone, both of which might or might not have daylight saving time. However, users can still somewhat workaround that issue by providing an appropriate fixed offset. This would not always be technically correct, but since PP doesn't track sub-day precision timestamps, it should be possible to provide some offset that at least places the result in the correct day the vast majority of times.

I yet have to figure out how to add translations for the added strings, but other than that I hope this should work as-is now.

jan-moeller:
@buchen I just rebased this onto latest master. From my POV this is complete, I'm not actively working on it. If you have any questions or concerns related to the timezone handling let me know. I've tried to make the timezone behavior consistent independent of how the timestamp is given (i.e. number of seconds since epoch or string), but if you disagree with some of the choices I made, feel free to let me know.

buchen:
Thanks @jan-moeller for rebasing - and the nudge ;-)

This is one of the pull requests where one needs quiet time to concentrate. It is merged now. I also understand it does not have an impact on existing configuration (ZoneOffset.UTC is zero). I have updated the translated texts: now there is also Vietnamese. And I now tend to use informal vs formal language (Du vs Sie)

If you have time, can you also make a pull request to the documentation.
There is a page for the JSON importer: https://help.portfolio-performance.info/en/how-to/downloading-historical-prices/json/

jan-moeller:
Thank you @buchen, appreciate it!

Issue #3585, opened 2023-10-07
Improvements to (English) terminology used in the app


pfalcon:
Something I wanted to bring up for a while (but was shy too, because English is not my native language, nor I'm too familiar with financial terminology specifically). I know that Portfolio Performance started in German, and it's great to see it's being translated to many languages. I'll specifically touch English translation. So, I suspect that some terminology used in English translation might have heritage in German original. At the very least, some of the terms (in the context used in PP) I barely heard in finance-related contexts, while other terms are usually used in that context. A few examples:

* "Absolute performance %" - it seems that in English, "Total return %" is used almost universally
* "Deposits and Removals" - sorry, that's "Deposits and Withdrawals"

So, overall, I'd like to open this ticket to both bring up the issue, and collect feedback from more users, especially native ones (while keeping in mind and tracking distinction UK- vs US-originated terminology, etc.)

pfalcon:
Other related issue is consistency of terms/description. For example, it's important to know whether figures (amounts, percentages, etc.) are based on "gross" or "net" data (i.e. whether fees/taxes/etc. are accounted for or not). PP is actually pretty good handling that, as many report-like pages have a "Use gross" setting to let user explicitly switch. However, for many parameters, it's still unclear for end user whether gross or net calculation would be used. An example is "Absolute performance %" above.

So, I would like to propose that for any data value (like column/field name, etc.), it was explicitly stated whether it's net or gross. That can be done e.g. in a tooltip.

Morpheus1w3:
@pfalcon 
Join the translation team

[POEditor.com](https://poeditor.com/join/project?hash=4lYKLpEWOY)
Join translations
Login with your account
Start your translation in your language
Have fun....

buchen:
Thanks @pfalcon for bringing this up. The example "Withdrawals" drove your point home. I immediately realized that this is the right English term.

I also like the proposal to bring up significant changes of terms in this issue first. For example, the change to withdrawal will tough hundreds of terms in POEditor. But there is no way to revert the change, no meaningful history of the translation, etc. (Of course, if you find a typo, an improved translation, etc., feel free to use POEditor directly).

### Total return

> "Absolute performance %" - it seems that in English, "Total return %" is used almost universally

Sounds reasonable.

The PP definition is:

> Absolute performance = market value + sell/outbound deliveries + dividends - taxes - fees - initial valuation - buy/inbound deliveries

[...]

PR #4657, opened 2025-04-19
Double-click a cell in Statement of Assets to copy its value

### LINKED ISSUE #4085: Ability to copy value from a single cell
Related to https://github.com/portfolio-performance/portfolio/issues/2372, but posting as its own issue as it's more specific.

The ability to copy the value from a single cell would be very useful - specifically, from the Statement of Assets screen.

Since on that screen none of the cells are editable via double-clicking (except for Notes), that could be used for an easy, UI-free way to copy a single cell. i.e. double-click a row's Shares, and the share count gets copied to clipboard. Double click a row's Purchase Value, and that value gets copied. (that's how KeePass XC works - you can copy i.e. a username or password or the like by clicking on that value).

(my current workaround is to copy the entire row, then paste it to a textfile, then in the textfile, copy the value I actually want, then paste it where I want. Simply double-clicking the value itself to have it copy would be a big timesaver)

metal450:
Implement the ability to copy any single value from the Statement of Assets screen, by double-clicking its cell.

Closes #4085

Peek 2025-04-19 07-03

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
@@ -31,20 +31,26 @@ import org.eclipse.jface.resource.FontDescriptor;
 import org.eclipse.jface.resource.JFaceResources;
 import org.eclipse.jface.resource.LocalResourceManager;
 import org.eclipse.jface.viewers.ArrayContentProvider;
+import org.eclipse.jface.viewers.CellLabelProvider;
 import org.eclipse.jface.viewers.ColumnLabelProvider;
 import org.eclipse.jface.viewers.ColumnViewerToolTipSupport;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.viewers.TableViewer;
+import org.eclipse.jface.viewers.ViewerCell;
 import org.eclipse.jface.window.ToolTip;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.dnd.Clipboard;
 import org.eclipse.swt.dnd.DND;
+import org.eclipse.swt.dnd.TextTransfer;
 import org.eclipse.swt.dnd.Transfer;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Menu;
 import org.eclipse.swt.widgets.Shell;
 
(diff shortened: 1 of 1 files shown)


metal450:
Added screenshot

buchen:
I feel this is confusing. Double-click activates in place editing of the cell.

What happens if the cell is editable? Is it then also copied to the clipboard?

Copy to clipboard on double-click is not a UI pattern that I user would expect.



What I could imagine is, that we give users the options to switch from "row selection" (only the full row is selected) to "cell selection". And if he presses Ctrl-C, then in the first case the full row is copied to the clipboard (as today) or only the cell is copied to the clipboard.

metal450:
> I feel this is confusing. Double-click activates in place editing of the cell. What happens if the cell is editable? Is it then also copied to the clipboard?

Ah - yeah, I can see the argument for making this only apply for non-editable cells. I think only Name & Note are editable? The intention was really for copying values anyway (as if you double-click a cell that becomes editable, you can just Ctrl+C to copy). Happy to make this only apply to non-editable if you prefer.

> Copy to clipboard on double-click is not a UI pattern that I user would expect.

I've definitely seen it. You can do it in DBeaver, for example (tho I believe it's configurable in settings).

> What I could imagine is, that we give users the options to switch from "row selection" (only the full row is selected) to "cell selection". And if he presses Ctrl-C, then in the first case the full row is copied to the clipboard (as today) or only the cell is copied to the clipboard.

I suppose that's a valid alternate implementation - my thought here was just that this 1) doesn't require any additional options or UI, 2) doesn't affect existing navigation/behavior, 3) is transparent/non-invasive. i.e. since double-clicking a (non-editable) cell currently has no effect/action, it's a very easy way to get the desired behavior.

(As a side note, of all the PRs I recently submitted, this is the one I've been by far most desperate for. Every month as I reconcile my finances, I need to manually retype to the calculator the number of shares of each holding - which in the case of crypto can be many decimals long, which is very easy to mistake. So even just being able to copy "Shares" would be great, tho I figured if you can copy the other numeric values too, couldn't hurt! :) )

buchen:
> I need to manually retype to the calculator the number of shares of each holding

BTW, what you can already do today: Copy the full line to the clipboard (Ctrl-C) and paste it to a text editor; then select the shares. The line contains all digits of the shares regardless how many digits are displayed in the table.

> I think only Name & Note are editable?

More. Symbol, all configured attributes. Essentially all values should be editable that a) are not calculated and b) do not require complex validation.

> my thought here was just that this 1) doesn't require any additional options or UI, 2) doesn't affect existing navigation/behavior, 3) is transparent/non-invasive

I agree with all those intentions. And, yes, switching to cell navigation would have bigger implications.

I am not convinced yet that "overloading" the double-click is such a good idea.

What about adding this to the context menu? The context menu could detect on which row and cell it is clicked and then offer to copy the row or cell (say to entries "Copy row" and "Copy cell").

metal450:
> Copy the full line to the clipboard (Ctrl-C) and paste it to a text editor; then select the shares

Thanks - yea, I did know about that workaround (mentioned it in the "Closes" issue above, but probably should've mentioned it here too). Sometimes I do that, sometimes retype, depending on how many decimals there actually are.

> More. Symbol, all configured attributes.

Ah, gotcha. So this should probably only apply to Shares & Money columns.

> What about adding this to the context menu?

Better than nothing, but IMO double-clicking is just much quicker/more seamless vs needing to right-click, locate the entry in the menu, click again. Personally I don't think there's anything wrong with contextual double-clicking - i.e. another common example that does this is [KeePassXC](https://keepassxc.org); double-clicking a username or password field auto-copies it (which I use all the time); double-clicking a title goes to the editor, and double-clicking the URL it opens it in a browser.

Putty Telnet copies text on double-click, too.

My order of preference would be:
1) Double-click to copy (fastest, most seamless)
2) Cell-based keyboard navigation & Ctrl+C to copy (tho as mentioned, more complex to implement)
3) Having to pick from context menu

buchen:
> Double-click to copy (fastest, most seamless)

I understand your use case. But I do not agree to the user experience. As a user, I find it confusing that sometimes the double-click starts the editing and sometimes it copies the value to the clipboard. And one cannot see this just by looking at the column. One would have to just try. And then double-clicking would fill up the clipboard history - which I would find annoying.

My order of preference would be:
* Having to pick from context menu
* Cell-based keyboard navigation & Ctrl+C to copy (tho as mentioned, more complex to implement)

About the menu: I understand it requires at minimum two clicks (right click to open the menu, click to select the option), but is the "copy cell" operations really something that one would do very often? I am not sure where you do the "reconciliation" but even that is only once a month.

metal450:
> one cannot see this just by looking at the column. One would have to just try.

If u think it's really that confusing, could stick it in the tooltip, so hovering says i.e. "1.3642 (double-click to copy)"?

I mean if ur strongly against it then ur strongly against it, I'll just have to defer to u. But I do think having the behavior of double-clicking be contextual is pretty common and reasonable. Just like in KeePass: In each column, double-clicking performs the most sensible action for that column, from a workflow & efficiency perspective: if ur double clicking the username or password u probably want to copy it, if ur double clicking the URL u probably want to visit it, if ur double clicking the entry name u probably want to edit it. In this case, it's as simple as "double clicking a text field edits, double clicking a numeric field copies." The alternative here is double click does...nothing. Respectfully, I really don't feel like "_because we've done one action over here, it has to have no action anywhere else._" Tho again, it's your software so ultimately only you can decide ;)

Just out of curiosity, I asked Claude & it gave a long list of examples (its summary at the bottom of the response in particular reflects my opinion).

image
image

buchen:
It took some time, but the feature is now available.

I implemented it slightly different: if the users presses the 'Alt' key ('Option' on macOS) then the content of the cell is copied to the clipboard. If the user is not pressing the Alt key, then the in place editor is activated (if the cell as an editor attached).

One thing I am wondering about: how to let the user know that such a feature exists? In the application. Subtle.

JGKle:
> if the users presses the 'Alt' key ('Option' on macOS) then the content of the cell is copied to the clipboard

Sounds good :)

Hmm...only thing I can think of is the tooltip, & if there's an existing tooltip (i.e. number of shares), append it as a final line. So the tooltip would be the amount, newline, then "Alt+Double-Click to Copy"

pfalcon:
@buchen 

> One thing I am wondering about: how to let the user know that such a feature exists? In the application. Subtle.

IMHO, PP should really get inline help, with question mark icons sprinkled around the UI. Of course, this should be done after easier things, like making terminology non-confusing and consistent: https://github.com/portfolio-performance/portfolio/issues/3585

PR #4658, opened 2025-04-19
Monthly performance-neutral transfers widget

### LINKED ISSUE #2587: Add widget "Performance neutral transfer" on a monthly basis and per average
Hello!

Currently I really like to use the three widgets "Monthly % relative", "Monthly returns absolut" and "Monthly investments".

Very much I would like to use here also an overview of the "monthly performanceneutral transfers", gladly also together with an average.

Best regards
gulfwarden

JGKle:
Implement a new Dashboard widget to show "Monthly Performance-Neutral Transfers" (with similar appearance to the "Monthly Earnings" widget).

Closes #2587

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
@@ -1,0 +1,68 @@
+package name.abuchen.portfolio.ui.views.dashboard;
+
+import java.time.LocalDate;
+import java.util.stream.LongStream;
+
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.Dashboard.Widget;
+import name.abuchen.portfolio.snapshot.PerformanceIndex;
+import name.abuchen.portfolio.ui.views.dashboard.heatmap.AbstractMonhtlyHeatmapWidget;
+import name.abuchen.portfolio.ui.views.dashboard.heatmap.HeatmapModel;
+import name.abuchen.portfolio.ui.views.dataseries.DataSeries;
+import name.abuchen.portfolio.util.Interval;
+
+public class MonthlyPNTransfersWidget extends AbstractMonhtlyHeatmapWidget
+{
+    public MonthlyPNTransfersWidget(Widget widget, DashboardData data)
+    {
+        super(widget, data);
+        addConfig(new DataSeriesConfig(this, false));
+    }
+
+    @Override
+    protected void linkActivated()
+    {
+        // Nothing to do when clicking the title for now
+    }
+
+    @Override
+    protected void processTransactions(int startYear, Interval interval, HeatmapModel<Long> model,
+                    Client filteredClient)
+    {
+        Client client = getDashboardData().getClient();
+        DataSeries series = get(DataSeriesConfig.class).getDataSeries();
+
+        for (int year = interval.getStart().getYear(); year <= interval.getEnd().getYear(); year++)
+        {
+            int row = year - startYear;
+
+            for (int month = 1; month <= 12; month++)
+            {
+                int col = month - 1;
+
+                // Skip if this month is outside our interval
+                LocalDate startOfMonth = LocalDate.of(year, month, 1);
+                LocalDate endOfMonth = startOfMonth.plusMonths(1).minusDays(1);
+                if (endOfMonth.isBefore(interval.getStart()) || startOfMonth.isAfter(interval.getEnd()))
+                    continue;
+
+                // Calculate performance index of the month (including the day
+                // before for proper calculation)
+                Interval monthInterval = Interval.of(startOfMonth.minusDays(1), endOfMonth);
+                PerformanceIndex monthlyIndex = getDashboardData().calculate(series, monthInterval);
+
+                // Get transferals and sum them, skipping the first element (day
+                // before start)
+                long[] transferals = monthlyIndex.getTransferals();
+                long monthlyTransfers = transferals.length > 1 ? LongStream.of(transferals).skip(1).sum() : 0L;
+
+                // Set the value in the model
+                Long oldValue = model.getRow(row).getData(col);
+                if (oldValue != null)
+                {
+                    model.getRow(row).setData(col, monthlyTransfers);
+                }
+            }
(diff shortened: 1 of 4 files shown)


metal450:
Added screenshots

metal450:
Probably could be implemented to do that too, but I'll leave that as a different/additional feature that someone could choose to implement if needed

JGKle:
I wish we could get these merged so I could start updating again at some point :pray:

JGKle:
Does anyone other than buchen have commit access that could merge this? It's only 75 lines, fully self-contained, & has been sitting for more than 3 months

JGKle:
@buchen  🙏 🙏 🙏

JGKle:
Maybe @Nirus2000 has commit access? I really wish we could get this very short, self-contained & thus low-risk PR in - it's been finished & sitting for 19 weeks. Not even 100 lines.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
```diff
@@ -0,0 +1,68 @@
...
+                // before start)
+                long[] transferals = monthlyIndex.getTransferals();
+                long monthlyTransfers = transferals.length > 1 ? LongStream.of(transferals).skip(1).sum() : 0L;
+
+                // Set the value in the model
+                Long oldValue = model.getRow(row).getData(col);
+                if (oldValue != null)
+                {
+                    model.getRow(row).setData(col, monthlyTransfers);
+                }
```
Nirus2000: I think the logic was backwards—you only set the value if a value already exists. Now it always sets the calculated monthly transfer value. But I could be wrong here.

Maybe only
```
model.getRow(row).setData(col, monthlyTransfers);
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
```diff
@@ -0,0 +1,68 @@
...
+
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.Dashboard.Widget;
+import name.abuchen.portfolio.snapshot.PerformanceIndex;
+import name.abuchen.portfolio.ui.views.dashboard.heatmap.AbstractMonhtlyHeatmapWidget;
+import name.abuchen.portfolio.ui.views.dashboard.heatmap.HeatmapModel;
+import name.abuchen.portfolio.ui.views.dataseries.DataSeries;
+import name.abuchen.portfolio.util.Interval;
+
+public class MonthlyPNTransfersWidget extends AbstractMonhtlyHeatmapWidget
```
Nirus2000, line 14: Fixed typo: @buchen 
AbstractMonhtlyHeatmapWidget → AbstractMonthlyHeatmapWidget (missing 't' in Monthly)

Nirus2000 (changes requested):
Hello @JGKle 
Thanks for the contribution! Good work on the monthly transfers widget implementation. 

Check your source and clearly formatting...

Regards
Alex

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
JGKle: Ah thanks - will apply & re-test this weekend! :bow:

JGKle: You're right - updated, re-tested, and also fixed the conflict.

Not quite sure what you meant tho by _Check your source and clearly formatting..._ ?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/MonthlyPNTransfersWidget.java
buchen: thanks, yes, needs fixing, but not in this pr :-)

buchen:
Sorry @JGKle for the delay.

About the change. The feature makes sense.

My comments:
* it offers the user to filter by client, but that is not used at all
* instead it allows selection of a data series
* it then calculates the performance, and uses the transfers collected on the way - we could just use the transactions right away
* it shows a "link" to users that then has no action

My first question: **can we use the client filter as the selection criteria?**
The client filter is used in 'Monthly fees', 'Monthly taxes', 'Monthly earnings', and 'Monthly investments'. From a user point of view, it would be consistent with the other heat maps.

What is the difference compare to client filter?
* With a data series, the user could pick on instrument. The monthly performance-neutral transfers are then all purchases/sales. It is not possible to build a client filter just for one instrument unless there is an account with only that instrument.
* With a data series, the user could pick a category. The monthly performance-neutral transfers are then all transfers (if a cash account is part of the category) or all purchases/sales (unless if the cash account of that transaction is not part). Partial assignments would be resolved, say the cash account is assigned only to 50% but the instrument for 100%, then 50% of the purchase are a performance neutral transferal "into the category".


One argument for the data series is obviously the indicator widget for "performance neutral transfers" uses the data series. I admit, there is an inconsistency here between the selection criteria of the heat maps and the indicators (the single number in your screenshot above). I also understand now where why you implemented it the way you implemented it. I am just not sure what the meaning of a performance-neutral transfer for a particular security is.

JGKle:
Hi @buchen 

The answer to most of these is honestly just that I'm not that familiar with the codebase & did the best I could to figure out how to get the desired functionality, based on other examples I could find :) For my own purposes, I solely use it on the "Entire Portfolio" dataseries - no filters/dataseries selections are needed. Not sure if those would be useful for anyone else, but I'm all for just not having those options.

Will take another look after work or over the weekend & see if I can figure it out.

Thanks!

JGKle:
@buchen Ok, I pushed 3 commits that I believe should address the 3 issues:
* Uses Filter rather than dataseries
* Links to the transactions page
* Calculates directly from the transactions rather than from performance

How's that?

buchen:
looks good. will merge. will add some chores like the translations, but do not have the time today. sorry.

buchen:
merged

PR #4663, opened 2025-04-19
Improve to support lookup by currency display name (e.g. "Euro")

Nirus2000:
Adds support for resolving currencies by their localized display name using CurrencyUnit.getInstanceByDisplayName(String). Also updates asCurrencyCode(String) to include display name lookup as a fallback.

Update PDF-Importers...

---

Hello @buchen 
sooo... please check out `AbstractPDFExtractor.java` and `CurrencyUnit.java` ...

Thx
Alex

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ABNAMROGroupPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ABNAMROGroupPDFExtractor.java
@@ -6,6 +6,7 @@ import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Transaction;
 import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.Transaction.Unit;
+import name.abuchen.portfolio.money.CurrencyUnit;
 import name.abuchen.portfolio.money.Money;
 
 @SuppressWarnings("nls")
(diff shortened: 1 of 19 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ABNAMROGroupPDFExtractor.java
```diff
@@ -36,15 +37,15 @@ private void addAccountStatementTransaction()
...
+                                                                        .match("^Tagesgeldkonto \\(alle Betr.* in (?<currency>[A-Z]{3})\\).*$") //
                                                                         .assign((ctx, v) -> ctx.put("currency", asCurrencyCode(v.get("currency")))),
                                                         // @formatter:off
                                                         // Tagesgeldkonto 100,00 24.10.2011 100,00
                                                         // @formatter:on
                                                         section -> section //
                                                                         .attributes("currency") //
                                                                         .match("^Tagesgeldkonto (?<currency>.*)$") //
-                                                                        .assign((ctx, v) -> ctx.put("currency", asCurrencyCode("EUR"))))
+                                                                        .assign((ctx, v) -> ctx.put("currency", asCurrencyCode(CurrencyUnit.EUR))))
```
buchen, line 48: Instead of ```asCurrencyCode(CurrencyUnit.EUR))``` we can directly say ```CurrencyUnit.EUR```.
If we have a valid currency code (it is a static variable), then why try to resolve it it itself?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScorePriorityIncPDFExtractor.java
```diff
@@ -190,12 +190,12 @@ private void addAccountStatementTransaction()
...
                         .assign((t, v) -> {
                             v.put("date", v.get("day") + " " + v.get("month") + " " + v.get("year"));
-                            v.put("currency", CurrencyUnit.USD);
+                            v.put("currency", asCurrencyCode(CurrencyUnit.USD));
 
                             t.setDateTime(asDate(v.get("date")));
                             t.setShares(asShares(v.get("shares")));
                             t.setAmount(0L);
-                            t.setCurrencyCode(asCurrencyCode(CurrencyUnit.USD));
+                            t.setCurrencyCode(v.get("currency"));
```
buchen, line 198: And here we should use ```asCurrencyCode``` because we cannot necessarily know what is in the property currency (yes, it is set a couple lines earlier... but how knows where this is manipulated

buchen:
Merged (split into two commits: one for the change with the "display name lookup" and one for "fixing the constants")

PR #4666, opened 2025-04-20
Fix transaction search for numerics

### LINKED ISSUE #2345: Search in 'all transactions' for amounts with commas
**Describe the bug**
In 'All Transactions,' it's often useful to search for a specific amount (i.e. in dollars). But if I have an amount with a dollar value over $1,000, its behavior is inconsistent with regards to "commas"

* If I search for "1000", it will find Buy and Sell transactions worth $1000.
* If I search for "1,000" (with the comma), it will find Deposit and Transfer transactions worth $1000.

It should not require manually typing the comma in the search box, and it certainly should not be inconsistent, returning only one particular subset of transactions in one case, and a different subset of transactions in the other.

**To Reproduce**
1. Add some buy, sell, deposit, and transfer transactions worth $1000.00
2. Go to "All transactions"
3. Type "1000" in the search box. See results.
4. Type "1,000" in the search box. See reults.

**Expected behavior**
It should not require manually typing the comma in the search box - search for "1000" should return all transactions of this value, regardless of type.

 - OS: Linux
 - Version 0.53.3

metal450:
The Transactions search often returns goofy/incorrect results - for example:
* Searching for "100000000" would return every transaction for 1 share
* Searching for "5000000" would return transactions for $500, $2500, etc (basically any amount ending in "500").
* Searching for "2000" would NOT return transactions for $2000 (you have to explicitly type a comma, i.e. "2,000").

This PR should fix it.

Note that it retains the ability to search with commas (i.e. 2,000) to avoid changing prior behavior.

Closes #2345

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
@@ -1,6 +1,7 @@
 package name.abuchen.portfolio.util;
 
 import java.text.Collator;
+import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.util.Locale;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/util/TextUtil.java
```diff
@@ -362,4 +363,82 @@ public static String escapeHtml(String input)
...
+
+        // Split both strings into whole and decimal parts
+        boolean searchHasDecimal = cleanedSearchText.contains(String.valueOf(decimalSeparator));
+        boolean valueHasDecimal = formattedValue.contains(String.valueOf(decimalSeparator));
+
+        String[] searchParts = searchHasDecimal
+                        ? cleanedSearchText.split(Pattern.quote(String.valueOf(decimalSeparator)), 2)
+                        : new String[] { cleanedSearchText };
+        String[] valueParts = valueHasDecimal ? formattedValue.split(Pattern.quote(String.valueOf(decimalSeparator)), 2)
+                        : new String[] { formattedValue };
```
buchen, line 406: You will noticed in the merged code, that I replaced the regular expression with an indexOf. Then we do not have to have the "contains" and the regular expressions (which is compiled on every search).

buchen:
merged

PR #4668, opened 2025-04-20
Current Drawdown widget

### LINKED ISSUE #1640: Request: Peak Delta (And/Or Relative Peak Delta) Widget
At last I've got PP fully setup to replace my old "manual" portfolio-tracking Excel spreadsheet.  There's just one metric it doesn't seem to have, which I'd love to request in the form of a widget:

A "Delta Relative To Peak Delta."

Basically, in my spreadsheet, I show both the current delta, as well as the difference between that delta & the maximum delta we've ever seen.  This tells me when my accounts are at an all-time high, and if not, how far away I am from returning to that all-time high. (When we're at an all-time high, the value will be zero - since it's a peak.  When the account drops, it shows how far we are below that peak).

I hope this makes sense :)

metal450:
Implement a widget showing the portfolio's current drawdown from its high watermark.

Closes #1640


buchen:
Thanks for the contribution.

I noticed a difference between the "drawdown chart widget" and the this "current drawdown".  The former uses the "accumulated performance" data series to calculate the drawdown. In my personal case, the drawdown there is higher because I actually added fresh capital at the beginning of April. That leads to a higher drawdown in the chart compared to the drawdown on the "totals" data series (used in current drawdown). Just looking at the totals data series will include the performance-neutral transfers and therefore underestimate the risk.

For example, let's say one has 1K at the beginning of the year. It goes down to 600 Euros by mid of April. And then today, you add another 1K to the portfolio. The total is now 1.6K and the current drawdown is zero.

If we call this widget "distance from all time high" then I think it would be okay (similar to the already existing widget for the "Distance from ATH" for instruments).

What do you think? Or am I barking up the wrong tree?

Andreas.

metal450:
Ohh good catch, your right! This was intended to be drawdown from high watermark (per the tooltip), so performance-neutral transfers should not affect it, but it looks like they do currently. i.e. if the account climbs to 1M & this value is 0%, then the account retraces to 900k, this value will be -10%. Then if you contribute 100k, the value should still be -10%...it should be unchanged. But it looks like it's affected by performance-neutral transfers right now, which it shouldn't be....

metal450:
@buchen If I'm supposed to do something else to get all these PRs merged, I'd appreciate if you could tell me. I've been rebasing & rebuilding my own binaries locally like 6x in the last month -  not sure if there was still something wrong with all these, but seems like I just need someone with commit access to click one button. 🙏

buchen:
merged.

@metal450 - was super busy and on a business trip. Now working on the backlog...

PR #4672, opened 2025-04-21
Improve PortfolioPerformanceSearchProvider

Nirus2000:
Convert the security type using the SecuritySearchProvider instance

Remove OpenFIGI.java


buchen:
I wasn't quite sure that categorization the others are using (it is not OpenFIGI - see "crypto" which is not defined in the SecuritySearchProvider). But I agree that at the current point in time a separation is not really meaningful. Will extend the SecuritySearchProvider and merge.

name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/PortfolioPerformanceSearchProvider.java
```diff
@@ -84,7 +87,7 @@ public static Optional<ResultItem> from(JSONObject json)
                 answer.markets.sort(new ByExchangeComparator());
             }
 
-            return Optional.of(answer.markets.size() == 1 ? answer.markets.get(0) : answer);
+            return Optional.of(answer.markets.size() > 0 ? answer.markets.get(0) : answer);
```
buchen, line 90: The idea here was: if the instrument is traded on one market only, then simplify the UI (= do not show the instrument, then then a list of one market --> instead just show the one market directly). However, if there are multiple markets, then of course we need to return the full list.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/PPIDPreferencePage.java
```diff
@@ -48,11 +47,9 @@ protected Control createContents(Composite parent)
         this.oauthClient.addStatusListener(updateListener);
         parent.addDisposeListener(event -> oauthClient.removeStatusListener(updateListener));
 
-        Composite area = new Composite(parent, SWT.NONE);
+        var area = new Composite(parent, SWT.NONE);
         GridLayoutFactory.swtDefaults().numColumns(2).spacing(5, 10).applyTo(area);
 
-        // If we set the description on the preference page itself, then the
-        // layout of the page is broken.
```
buchen, line 55: I spend two hours debugging this... I need some reminder in the code why the 'description' is a custom field here instead of just using #setDescription. Otherwise I fear in a couple years I will attempt to refactor this again. Let me update the comment.  😄

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/PPIDPreferencePage.java
```diff
@@ -28,17 +28,16 @@
...
 public class PPIDPreferencePage extends PreferencePage
 {
+    private static final String EMPTY_USER_TEXT = "-"; //$NON-NLS-1$
     private final OAuthClient oauthClient = OAuthClient.INSTANCE;
-
     private Label user;
     private Button action;
-
-    private Runnable updateListener = () -> Display.getDefault().asyncExec(this::triggerUpdate);
+    private final Display display = Display.getDefault();
```
buchen, line 35: It's okay to use this as a member variable, but generally I prefer the direct use of the static method unless the display is actually context dependent (e.g. for the window management). Less instance variables.

buchen:
Split into two separate commits and merged.

PR #4676, opened 2025-04-24
Make RawResponsesDialog resizable

elho:
This PR makes the RawResponsesDialog that is used e.g. to display server responses for quote feeds resizable, so that the user can make it larger to more conveniently inspect  HTML and JSON responses that happen to be particularly wide or long.


buchen:
Thanks for your contribution!

PR #4680, opened 2025-04-26
Add toString() method to LatestSecurityPrice that prints all security attributes

elho:
This PR adds a toString() method to LatestSecurityPrice, that asso prints the high, low and volume attributes of the security.
Previously, only date and price where printed by the toString() method of the super class, leading to confusing and not helpful output from the test suute, when (e.g. in the tests of a quote feed) an assertion comparing two prices failed due to them only differing in the high, low and volume attributes.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/model/LatestSecurityPrice.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/model/LatestSecurityPrice.java
@@ -3,6 +3,8 @@ package name.abuchen.portfolio.model;
 import java.time.LocalDate;
 import java.util.Objects;
 
+import name.abuchen.portfolio.money.Values;
+
 public class LatestSecurityPrice extends SecurityPrice
 {
     private long high;
@@ -79,5 +81,12 @@ public class LatestSecurityPrice extends SecurityPrice
             return false;
         LatestSecurityPrice other = (LatestSecurityPrice) obj;
         return super.equals(other) && high == other.high && low == other.low && volume == other.volume;
+    }
+
+    @Override
+    @SuppressWarnings("nls")
+    public String toString()
+    {
+        return String.format("%s %,10.2f %,10.2f %,10d", super.toString(), high / Values.Quote.divider(), low / Values.Quote.divider(), volume);
     }
 }


Nirus2000:
I don't think that's a good idea.
ToString as a name says something completely different than what the function actually does. And where do you call this function?

elho:
> ToString as a name says something completely different than what the function actually does.

Uhm? Not sure what else you would expect it to do, the function does return a string representation (not a serialisation though, maybe that is what you are thinking of? :thinking:) of the object, after all, but (the developers of the) Java language came up with the name `toString`, and according to any documentation on it, AIUI it is supposed to output a string representation that describes the object when it is printed. :shrug:

Also, the super class of `LatestSecurityPrice`, `SecurityPrice` does implement a very similar `toString()` already, that this just adapts to work in the same vain.

> And where do you call this function?

I do not call it directly. But e.g. `YahooFinanceQuoteFeedTest.java` has asserts like 'assertThat(prices.get(0), equalTo(price));`,  which when the two `LatestSecurityPrice` objects difffer and thus the assertion faiss, the assert prints objects causing the the `toString()` method to be called to get the "foo" and "bar" parts of the "expected: foo" "but was: bar" output.
Without this PR the  `toString()` method of  `SecurityPrice` is used, which just yields "date: price", with the PR it is  "date: price high low volume".

> And what, if some values not available?

Good that you ask, but it will not fail, because class attributes will always be initiasized by Java, so for `high`, `low` and `volume` being longs, they will be `0L`, so the output being 0.0.
When they are set to LatestSecurityPrice.NOT_AVAILABLE`, which is `-1L`, the output IIRC is `-0.0`  (something small like `-0.0...01` being displayed with 2 fractional digits) - it could indeed be caught and output as "NOT_AVAILABLE" instead, nothing the existing code does, but maybe nicer. :thinking:

Nirus2000:
```
LatestSecurityPrice price = new LatestSecurityPrice(LocalDate.of(2025, 4, 27), 1234567890L, LatestSecurityPrice.NOT_AVAILABLE, LatestSecurityPrice.NOT_AVAILABLE, 500000L);
System.out.println(price.toString());

LatestSecurityPrice price = new LatestSecurityPrice(LocalDate.of(2025, 4, 27), 1234567890L, 1235000000L, 1234000000L, LatestSecurityPrice.NOT_AVAILABLE);
System.out.println(price.toString());
```

name.abuchen.portfolio/src/name/abuchen/portfolio/model/LatestSecurityPrice.java
```diff
@@ -80,4 +82,11 @@ public boolean equals(Object obj)
         LatestSecurityPrice other = (LatestSecurityPrice) obj;
         return super.equals(other) && high == other.high && low == other.low && volume == other.volume;
     }
+
+    @Override
+    @SuppressWarnings("nls")
+    public String toString()
+    {
+        return String.format("%s %,10.2f %,10.2f %,10d", super.toString(), high / Values.Quote.divider(), low / Values.Quote.divider(), volume);
+    }
```
Nirus2000, line 91:
```suggestion

@Override
@SuppressWarnings("nls")
public String toString()
{
    String highStr = (high == NOT_AVAILABLE) ? Messages.LabelNotAvailable : String.format("%,10.2f", high / Values.Quote.divider());
    String lowStr = (low == NOT_AVAILABLE) ? Messages.LabelNotAvailable : String.format("%,10.2f", low / Values.Quote.divider());
    String volumeStr = (volume == NOT_AVAILABLE) ? Messages.LabelNotAvailable : String.format("%,10d", volume);

    return String.format("%s %10s %10s %10s", super.toString(), highStr, lowStr, volumeStr);
}
```
You would still have to add the translations to the label, but you could copy them from the [name.abuchen.portfolio.ui.Messages;](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java#L782)
The translations from the resource bundle as well.
It's just an idea...

Nirus2000 (changes requested):

buchen:
> leading to confusing and not helpful output from the test suite

I believe #toString is not used in the UI and @elho suggests that for easier debugging. We should not add a translation label in this case.

buchen:
merged

PR #4683, opened 2025-04-28
Now feed provider is initialized during securities import from a csv file

mikaelpeltier:
Feed provider is automatically assigned to YAHOO when ticker symbol is defined on the imported securities

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
@@ -51,9 +51,6 @@ import name.abuchen.portfolio.online.impl.YahooFinanceQuoteFeed;
             String note = getText(Messages.CSVColumn_Note, rawValues, field2column);
             s.setNote(note);
 
-            if (s.getTickerSymbol() != null)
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
```diff
@@ -64,8 +61,12 @@ void extract(List<Item> items, String[] rawValues, Map<String, Column> field2col
...
                             0);
 
-        // nothing to do to add the security: if necessary, the security item
-        // has been created in the callback of the #extractSecurity method
+        // The security item is created in the callback of the #getSecurity
+        // method, into this callback, symbols are not yet initialized, thus
+        // moves feed initialization from the callback to here by taking care of
+        // not overriding an existing value
+        if (security.getFeed() == null && security.getTickerSymbol() != null)
+            security.setFeed(YahooFinanceQuoteFeed.ID);
```
buchen, line 69: By moving the change outside of the "getSecurity" this now applies to all existing securities as well. That can override the configuration that the user has chosen.

I have changed it to set the ticker also in the factory function above. It will be set again, but that should not hurt

buchen:
merged.

PR #4694, opened 2025-05-07
Imrprove YahooFinance quote feed to fetch up to 30 years of historical data

elho:
While the `validRanges` metadata field of the server response has 10y as the higest numeric entry in its list of possible values, other values like 20y or 30y do work as well, as also used in an example of the JSON quote feed in the documentation.


buchen:
let's do it while it lasts ;-)

PR #4695, opened 2025-05-07
Extend YahooFinance quote feeds to also provide daily low & high and volume

elho:
The values for daily low and high values as well as trade volume are typically provided within the server response (nowadays, I guess), so use them, if they are present. The code does handle them missing and the old tests with server responses without these extra attributes are left in place to test this.

Also included is a small cleanup of the tests, they defined `security` objects that were never used and in some case used a completely different ticker symbol than the test data. I assume that being a result of some too generous copy & paste from other tests, so dropped them.


Morpheus1w3:
@elho 
Let me ask one question, for what purpose? Nothing in PP is calculated with it? This will just increase the file size and decrease the performance,  as everything is loaded into the memory at runtime.

elho:
> @elho Let me ask one question, for what purpose? Nothing in PP is calculated with it? This will just increase the file size and decrease the performance, as everything is loaded into the memory at runtime.

The data source tab of edit security dialog (where you select the data source) has entries for low, high and volume in the table that remain empty/zero otherwise, and the graph tab at the bottom of the securities list on the right hand side shows the (latest) High, Low and Volume (showing as n/a otherwise).
In the portfolio file the low, high, and Volume values apparently are stored only for the latest price. (And the XML tags for them, which typically take up more space than the actual values, are always present anyway.) So you will be hard pressed to even see a difference looking at the file size when using human readable numbers.

sumansuhag:
The aim of a detailed response is to shift the focus from just code tweaks to bigger ideas about system design and preparing our financial data systems for the future. 

Let's tie in what elho said about 'Data Contract Evolution &amp; Observability for Financial APIs.' We should think about ways to manage outside data changes in these unpredictable market data situations.

Elho, you made some great points on dealing with data changes and doing thorough test cleanups. It's really important to handle missing info in server responses well, especially for financial systems. This highlights the need for a solid plan for evolving data contracts with outside APIs, so we can stay on top of changes in data formats. How do you stay ahead of these shifts in external data, particularly in fast-changing market situations, rather than just waiting for errors to pop up?

PR #4699, opened 2025-05-09
Add more moving average columns in trades

mierin12:
Hello,

This is to continue on https://github.com/portfolio-performance/portfolio/issues/4503 and https://github.com/portfolio-performance/portfolio/commit/ba6651ce64b567648228e187f8d3d2c92e22bd54
This is a proposition to add in the Trades view, commit by commit : 

- **Gross Profit/Loss for FIFO for Open trades** : was left empty until now, I am not sure why, in the manual it is stated : 

> This equals the value of the previous column plus taxes and fees. Because for closed trades, the taxes and fees of the sell transaction are included in the Gross Profit/Loss column, **this field cannot be calculated for the open trades**.

However I believe it makes also sense for open trades, as you can have fees (and in rare cases even taxes) on buy transactions.

- **Gross Profit/Loss for Moving average for Open trades** : for this one it is not complicated, but there I reached a naming definition issue on "net/gross cost" (whatever fifo or moving avg), either I am missing something, either the naming does not seem consistent across the different part of the code

In PortfolioTransaction the definition is
  ```
  /**
     * Returns the gross value, i.e. the value before taxes and fees are
     * applied. In the case of a buy transaction, that are the gross costs, i.e.
     * before adding additional taxes and fees. In the case of sell
     * transactions, that are the gross proceeds before the deduction of taxes
     * and fees.
     */
```
In CostCalculation the definition is
   ```
 /**
     * net investment, i.e. without fees and taxes
     */
```
I understand the latter. My understanding :
Value=G+C+F
G=net Gains, C=net Cost, F=acquisition Fees&Taxes
Value-C=G+F=gross gain (shows price movement only)
Value-G=C+F=gross cost
gross gain+net cost=net gain+gross cost

But in Trades, the naming is gross profit = gross exit value-gross entry value. which is a bit confusing to me
Anyway, I think the calculation themselves are correct, but it is a bit confusing, as at some points the two naming clash in what I have added :
 ```
   /**
     * moving average cost of shares held without fees or taxes
     */
    private final LazyValue<Money> movingAverageGrossCost = new LazyValue<>(
                    () -> costCalculation.get().netMovingAverageCost());
```

- **Entry value per share for moving average**

- **Return for moving average**

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
@@ -256,6 +256,11 @@ public class Trade implements Adaptable
         return (exitValue.getAmount() / (double) entryValue.getAmount()) - 1;
     }
 
+    public double getReturnMovingAverage()
+    {
+        return (exitValue.getAmount() / (double) entryValueMovingAverage.get().getAmount()) - 1;
+    }
+
     /**
      * @brief Checks if the trade is closed
      * @return True if the trade has been closed, false otherwise
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
@@ -98,6 +98,7 @@ public class Trade implements Adaptable
                             .setScale(0, RoundingMode.HALF_UP).longValue();
 
             this.exitValue = converter.at(now).apply(Money.of(security.getCurrencyCode(), marketValue));
+            this.exitGrossValue = exitValue;
 
             this.holdingPeriod = Math.round(transactions.stream() //
                             .filter(t -> t.getTransaction().getType().isPurchase())
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java
@@ -45,6 +45,7 @@ public class Trade implements Adaptable
     private double irr;
 
     private LazyValue<Money> entryValueMovingAverage;
+    private LazyValue<Money> entryGrossValueMovingAverage;
 
     public Trade(Security security, Portfolio portfolio, long shares)
     {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by mierin12** [comment]
In the last comment, I changed some my introduced variables names but I also propose to change one existing variable name in Trade (a very local change) : 
`entryGrossValue `-> `entryValueNet`
because, even if this value comes from the method `getGrossValue`, in the cases of purchases transaction, `getGrossValue `returns the value before/without the taxes & fees, so what we can call a net purchase value (if I understand correctly). So I am proposing to "uncouple" the name of this variable and the method to have the variable name closer to what it really represent instead. 
I had to make myself some schematic to be sure, this is what I understand :
<details>
  <summary>schematics</summary>
  
Capture d’écran 2025-06-20 151806
and considering a total sale:
Capture d’écran 2025-06-20 152036
  
</details>

or maybe my definition / understand of net/gross in this context is incorrect.

File: name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/trades/Trade.java [hunk 1]
```diff
@@ -262,7 +247,14 @@ public Money getGrossProfitLoss()
...
         if (exitGrossValue == null)
             return null;
-        return exitGrossValue.subtract(entryGrossValue);
+        return exitGrossValue.subtract(entryValueNet);
+    }
+
+    public Money getGrossProfitLossMovingAverage()
+    {
+        if (exitGrossValue == null)
+            return null;
```
buchen, line 256: I think we do not need this check anymore since exitGrossValue is now always non-null

buchen:
I agree that the use of net and gross is confusing. It sometimes depends on the direction on which to look at the transaction.

I wonder if we use "withoutTaxesAndFees" in the appropriate places. It is longer, but it then it is directly understandable. And I have started to use the label "withoutTaxesAndFees" in other places as well.

Any thought?

(I was starting to review and cherry pick your changes, so no need to change the PR)

buchen:
And maybe instead of the boolean in the methods, we use an enum:

```java
enum TaxesAndFees { INCLUDED, NOT_INCLUDED }
```

then the code reads as

```java
this.entryGrossValueMovingAverage = new LazyValue<>(() -> getMovingAverageCost(client, converter, TaxesAndFees.NOT_INCLUDED));
```

buchen:
Or instead of "withoutTaxesAndFees" we use "beforeTaxesAndFees" (1 letter less)

buchen:
I have not merged your changes into master.

From a UI perspective, I also wonder if "without taxes and fees" isn't a global selection. So instead of every column have a variant "with or without taxes", "with FIFO or moving average", "with reporting currency or instrument currency" whether those are orthogonal options in the view itself. I understand that is a big change so I am hesitant to plunge right into it. It would also have impact on other views because we would want to have the behavior consistent. What just as a thought - what do you think?

mierin12:
Thanks !
I think at some point it could be a global selection. Even more so maybe for Fifo vs Moving average choice. To have a triple choice : show me Fifo columns only, moving average only, or keep showing me both (current way). Same for with/without Taxes and Fees. 
But before going there, I think we need to be sure that both options do exist for each relevant columns/values. I think we are not far at all for Fifo vs Moving average. For with/without Taxes and Fees I am not sure. Would it be confusing if a user use 'with' global option but the taxonomy calculation which are 'without' by default right ?

mierin12:
Hello, 
I think this PR can be closed as all have been introduced in master.
Maybe the discussion of "global selection" can be continued elsewhere, my two cents is :
* If there is a global preference parameter introduced, we have to be sure that the impact is exhaustive, otherwise it can become very confusing (eg : "I chose "moving average" gobal parameter but this column Profit/Loss only exist in FIFO mode")
* **The option to be able to show several columns should still exist** (global selection should not be an exclusive choice only) : so instead of have 1 column parametrized by the global parameter, maybe all the columns should continue to exist as they do now, but some are hidden from menu selection if the associated global selection is not made ? I guess it would be a less drastic code change https://forum.portfolio-performance.info/t/trades-ermittlung-mittels-gld/30688
* First easy step could be to be more exhaustive on the description of the different column,s for the user to be sure of what is shown, is it gross or net ? etc I think it is not that far from being complete actually.

PR #4702, opened 2025-05-10
WIP: Multiple securities chart

OnkelDok:
First working version:
grafik

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -1,19 +1,12 @@
 package name.abuchen.portfolio.ui.views;
 
-import java.text.DecimalFormat;
-import java.text.MessageFormat;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.Period;
-import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.List;
-import java.util.Optional;
-import java.util.function.Supplier;
 
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.IAction;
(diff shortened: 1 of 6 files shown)


Sn1kk3r5:
Oh wow!

Is my assumption correct, you mark the securities you like to see together in the chart and go for the chart tab?

pfalcon:
This sounds cute, but how does it look with multiple markings enabled for the chart (like multiple purchase value types, etc.)?

OnkelDok:
> but how does it look with multiple markings enabled for the chart (like multiple purchase value types, etc.)?

In the multi-security mode all markers are disabled.

OnkelDok:
I think this basic function is ready to be reviewed @buchen.

What was introduced:
- multi-security selection support for [SecurityPriceChartPane](https://github.com/portfolio-performance/portfolio/pull/4702/files#diff-9801459873408ff32900ace1b521f34831f0e804402291143043863d3285bf49)
- [SecuritiesChart](https://github.com/portfolio-performance/portfolio/pull/4702/files#diff-8ceb17565e5cf6cb04295ac1cd15ebfc31fb8d4bf7a2fae46b201d7add60596e) now can display up to 10 security charts as benchmark in case more than one security is selected.
- In Benchmark-mode all chart options are disabled except ``ChartDetails.SCALING_LINEAR``.
- If more then 10 securities were selected a hint is displayed in the chart.
- at the moment only the "all securities" view supports this benchmark function

I reverted the changes described in my first comment regarding outsourcing chart painting in separate file.

A short screen capture:

https://github.com/user-attachments/assets/c1a0957d-0dfb-4c17-9df0-83d9a2ab28c1

OnkelDok:
> See for example the screenshot below: I have selected three instruments where one has data for a very long period, while the two others are younger. In such a case, does it really help if the later securities start at 0% a couple years later?

I think I have also seen this behavior in tools on the internet.

> Similarly, if the user has selected the "H" (= holding period) then only the holding period is shown even if there are prices available for the rest of the chart. Take the screenshot below: The lila instrument is bought 2013, the green instrument early 2017. However, there are prices for the green instrument available for 2016 and earlier - they are just not shown.

Oh, I did not tested this scenario. I think the holding should only be used to determine the maximum time range and prices should also be displayed also for intervals where the security was not held. But maybe the community thinks different.

BBLady00 (commented):
preview

mierin12:
Hello @OnkelDok , if you select one security and then select one of the Transaction/Quote/Trades tab etc, does it work ? They are now empty for me : https://forum.portfolio-performance.info/t/problem-with-new-update/33418/4
Could it be related to this PR ?

PR #4739, opened 2025-05-23
[WIP] support red-up-green-down theme

liuhaoXD:
* support red-up-green-down theme
* add (net profit/loss column)
* csv importer add support time format(yyyy.MM.dd)


liuhaoXD:
In most parts of Asia, Red is commonly used to represent scenarios such as buying and rising in stock trading, while green is used for scenarios like selling and falling. 
Based on the existing CSS, I added a new theme. Since I only have a MacBook at present, I can't test how the new theme perform on other platforms. Thus the new theme is only effective on the Mac platform.

In addition, for reasons unknown to me, in the previous dark mode, the contrast between the red and green colors and the background color was too low in my usage scenarios, making reading quite difficult. 
Therefore, in the new theme, I adjusted the color codes of red and green in the dark mode, while leaving the previous dark mode unchanged.

image

PR #4767, opened 2025-06-06
PDF Parser for Boursobank - Fix Foreign ccy  parsing

FlorentG74:
This change fixes the PDF parser regular expressions so that Boursobank transactions executed in a foreign currency load properly.


Nirus2000:
Hello @FlorentG74 
Thank you for this pull request... super.
I have added or corrected a few small changes.

Regards
Alex

PR #4775, opened 2025-06-09
fix single quote issue in new label

mierin12:
Hello,

One of the new label from 0.77.0 somehow escaped the double '' automatic correction from releng.
This is also the case for _TooltipCurrentDrawdown_, however for this one I changed a little bit the translation, the new translation does not contain a ', so it will work at the next sync of Poeditor.
I haven't seen other case for French, but it is strange that the automatic fix of last year did not catch them.

**Before :**
Capture d’écran 2025-06-09 194129

**After :**
Capture d’écran 2025-06-09 194037


Nirus2000:
Are there any more?
I thought we had a test case about this... 

Good catch...

Issue #4777, opened 2025-06-09
Securities - missing information after installing version 0.77.0


Helloitsme:
When I select a (any) security in the "all securities" menu

the graph is shown correctly but all the other tabs like "historical quotes, transactions, trades, etc are empty!

In the menu accounts -> all transactions   the transactions are still there, also in the  "statement of assets" part, but as mentioned they are not visible anymore when selecting a security via the "all securities" menu.

windows 11, software version 0.77.0

mierin12:
Fixed in [release 0.77.1](https://github.com/portfolio-performance/portfolio/releases/tag/0.77.1) by https://github.com/portfolio-performance/portfolio/commit/a273d565cde82004197df61686dfea1b9d315c60

PR #4778, opened 2025-06-09
fix InformationPane input for 1 selected Security

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4777
Issue https://forum.portfolio-performance.info/t/9-6-2025-bestimmte-features-sind-nach-update-weg/33425/

Hello this is a proposition to fix https://forum.portfolio-performance.info/t/problem-with-new-update/33418/2

I tried to do what you did in https://github.com/portfolio-performance/portfolio/commit/ea860fd53840168ccd1cfc6cc69c484795776308 in _StatementofAssetView_ because there is no such bug in the Statement of Asset (only in _SecurityListView_ and in _SecuritiesPerformanceView_)


mierin12:
already fixed in [0.77.1](https://github.com/portfolio-performance/portfolio/releases/tag/0.77.1)

PR #4787, opened 2025-06-13
Added 'S T P In' and 'SIP Insure' as new purchase transactions

buchen:
At least two transaction types I could identify:

https://forum.portfolio-performance.info/t/import-pdf-kfintech-india/32633/6


Nirus2000:
For me looks good :-)

Issue #4782, opened 2025-06-11
New IBKR importer wrokng conversation


seg-on:
Hi!
I tried to import the whole history from IBKR through Flex Report, but now it needs a EUR account, USD account and EUR account for transfers.
All currency conversions from EUR to USD were imported as transfers from a USD account to a EUR account.
This makes double expenditures in the USD account.
For example, I bought a stock in USD for 456 USD, and a cash transfer of 456 USD was made from USD account to the EUR account. Therefore, the USD account are in debt by -912 (-456-456).

buchen:
Thanks. Sounds like it is the wrong direction...

Nirus2000:
@itshukychen

Do you have time to have a look?

itshukychen:
Sure, @seg-on Can you share the file and the exact version?

buchen:
@seg-on, as @itshukychen said, we need an example.

I have to honestly admit I do not understand the IB flex format well enough.

Looking at [testIBFlexStatementFile10.xml](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/testIBFlexStatementFile10.xml) as an example, it contains to trades (transactions in PP speak): a purchase and a cash transfer.

In the [test case](https://github.com/portfolio-performance/portfolio/blob/6420bd98ac46eb9a4e3f699bf8c88d88307193b2/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java#L1641), we check that there is
* a purchase in EUR with a gross amount in USD
* a transfer from USD to EUR

That would mean: USD was converted to EUR to purchase the instrument in EUR (but traded in USD).

But I am not sure if that is the right way to interpret it.
Looking at the XML, the trade is fully in USD (even though it belongs to an account in EUR).

Is it maybe a purchase in USD and the transfer is a SELL of EUR to create that USD?
Then it would be exactly the other way around.

```
<Trade acctAlias="Cash" assetCategory="STK" buySell="BUY"
       cost="140.83125725" currency="USD" ...
<Trade acctAlias="Cash" assetCategory="CASH" buySell="SELL"
       cost="0" currency="USD"
       symbol="EUR.USD"
       fxRateToBase="0.94302"
       tradePrice="1.0596"
       tradeMoney="-140.9268"
       quantity="-133"
       ...
```

But:
[...]

PR #4792, opened 2025-06-19
IBFlexStatementExtractor: Cash xacts: Properly deal with diff. currency pairs

pfalcon:
There can be different currency pairs, e.g. EUR.USD vs USD.CAD, and it can be exchanged in one or another direction.


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -428,6 +430,20 @@ else if (element.hasAttribute("dateTime"))
             String targetCurrency = parts[0];
             String sourceAmount = element.getAttribute("tradeMoney");
             String targetAmount = element.getAttribute("quantity");
+            BigDecimal exachangeRate = asExchangeRate(element.getAttribute("tradePrice"));
+            // If "quantity" is negative, then we're dealing with pair like
+            // USD.CAD, exchanging USD to CAD, or maybe EUR.USD, exchanging
+            // EUR to USD.
+            if (targetAmount.startsWith("-"))
```
buchen, line 437: Can you contribute an example where the targetAmount does not start with a "-"?

Because all examples we have is buySell="SELL". 
And then it sounds to me that is the more reasonable attribute to use to determine the source and target pair.
If it is "SELL", and the symbol is "EUR.USD" then it is EUR is source and USD is target.
If it is "BUY" (but I do not have an example, and the symbol is "EUR.USD", then USD is the source and EUR the target.

pfalcon: > buySell=
> And then it sounds to me that is the more reasonable attribute to use to determine the source and target pair.

I can confirm that this will work too. Preparing example may take some time though.

buchen:
merged and adapted as discussed in #4782

PR #4804, opened 2025-06-23
Update Flatex documents to format generated by PDFBox 3

buchen:
This is a "companion" pull request to #4803

This is how the document would be generated if the user would have had PDFBox 3.
It is retrofitting the document to how it - most likely would have been created.

This sample files will not be generated anymore - also not for old documents - because PDFBox 1.8 is gone.

Here you can see how PDFBox 1.8 generated the document diffed against how PDFBox 3.0 generates the same document.

```diff
--- 1.8.17
+++ 3.0.3
@@ -11,2 +11,3 @@
-F Auftragsdatum      16.06.2025rau
-Vorname N hna e Handelstag         16.06.2025ac me
+Auftragsdatum      16.06.2025
+Frau Handelstag         16.06.2025
+Vorname Nachname
```

What do you say? Retrofit? Or keep in mind, that these documents will not be created this way anymore?

Because this is a fairly small change, I think both ways are possible.

(The pull request fails because I did it on the master which does not have the adapted code)


buchen:
Not wanted by @Nirus2000

PR #4827, opened 2025-07-04
SecurityListView: Allow to filter by options and non-options

pfalcon:
In security list view, add "Securities - options" and "Securities - non-options" filter choices beneath "Only securities". Option are time-limited and new-always-coming type of security, so over time, a large number of symbols can accumulate in portfolio, which clutter the view. It's expected that people may want to have "Securities - non-options" selected most of the time, and switch to option view as needed.


pfalcon:
As discussed in https://github.com/portfolio-performance/portfolio/pull/4825#issuecomment-3035766535

pfalcon:
@buchen : Can you please review this? Code-wise, it's pretty simple. But it adds initial support for options, where there was none before. But that's again the whole idea - to start adding it in simple and gentle manner.

PR #4832, opened 2025-07-05
Initial step of short transactions support: Short trades 

pfalcon:
Based on experimenting with adding support for short transactions in PP, this looks like the place to start - adding it on the level of individual trades. This is not complete support even on the level of trades, but the idea is to start with very simple and clean patchset, easy to review (and thus hopefully fast to merge). A lot more work will need to be done, but after experimenting with it last couple of years, I really think that initial footprint can land in the mainline, to hopefully gather interested parties to make further steps.

So again, this patch allows to match short trades, like initially selling 10 shares with later buying 5 shares. A unit test is included (a large part of the patchset is actually convenience test methods to make writing tests a bit less cumbersome).

### DIFF
--- a/name.abuchen.portfolio.junit/src/name/abuchen/portfolio/junit/PortfolioBuilder.java
+++ b/name.abuchen.portfolio.junit/src/name/abuchen/portfolio/junit/PortfolioBuilder.java
@@ -13,6 +13,7 @@ import name.abuchen.portfolio.model.Transaction;
 import name.abuchen.portfolio.model.Transaction.Unit;
 import name.abuchen.portfolio.money.CurrencyUnit;
 import name.abuchen.portfolio.money.Money;
+import name.abuchen.portfolio.money.Values;
 
 public class PortfolioBuilder
 {
@@ -29,6 +30,21 @@ public class PortfolioBuilder
     {
         this();
         this.portfolio.setReferenceAccount(referenceAccount);
+    }
+
+    public static long amountOf(double amount)
+    {
+        return Values.Amount.factorize(amount);
+    }
+
+    public static long quoteOf(double amount)
+    {
+        return Values.Quote.factorize(amount);
+    }
+
+    public static long sharesOf(double shares)
+    {
+        return Values.Share.factorize(shares);
     }
 
     public PortfolioBuilder inbound_delivery(Security security, String date, long shares, long amount) // NOSONAR
--- a/name.abuchen.portfolio.junit/src/name/abuchen/portfolio/junit/PortfolioBuilder.java
+++ b/name.abuchen.portfolio.junit/src/name/abuchen/portfolio/junit/PortfolioBuilder.java
@@ -100,6 +100,17 @@ public class PortfolioBuilder
         return portfolio;
     }
 
+    public PortfolioBuilder buyPrice(Security security, String date, double shares, double price)
+    {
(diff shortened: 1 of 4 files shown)


buchen:
Small and easy is a good idea.

I think I understand the idea in the trade collector. Instead of opening a trade with a purchase and closing it with a sell, check the first (not matched) transaction and determine the sell from there.

Three things come to my mind:

BY_DATE_AND_TYPE sorts the transactions on the same day first to purchase and then sales. Which means the short sale intra day would not work (unless given a time). I think that is okay. Do you see more restrictions?

One cannot mix normal trades with short trades. Is that okay? (one could have a separate transaction type SHORT_SALE but I find it hard to see the impact).

Is it fair to argue that previously successful trade matching will not change? The logic "flips" only if the first trade is a sale, but that was leading to an exception in the old code.

pfalcon:
@buchen : 

> I think I understand the idea in the trade collector. Instead of opening a trade with a purchase and closing it with a sell, check the first (not matched) transaction and determine the sell from there.

Right, just keep accumulating transactions of the same purchase-vs-liquidation type, and when an opposite type is spotted, create a trade out of this "a few open and one closing transaction" list, with createNewTradeFromSell() working as-is for the case of a short trade (the only change to it is renaming to emphasize that it's no longer tied with just "sell" closing transaction). That doesn't cover some corner cases, but should be good enough to start.

> BY_DATE_AND_TYPE sorts the transactions on the same day first to purchase and then sales. Which means the short sale intra day would not work (unless given a time). I think that is okay. Do you see more restrictions?

Yes, that's an issue. The best situation is simply when each transaction has as-unique-as-possible, e.g. second-resolution timestamp. But I guess, people importing from PDF usually don't have that, so we would need to think how to deal with that case. I didn't want to bring up this matter right away, because a solution would require some discussion/consideration, and then certainly your help with implementing.

> One cannot mix normal trades with short trades. Is that okay?

I think yes, because standard brokerage rules are that one can't have both long and short positions of the same instrument in one account. So, if one had a long position, "opening short" would mean first selling out existing position, and then opening actual short for the rest, if any. Ditto in the opposite direction.

> (one could have a separate transaction type SHORT_SALE but I find it hard to see the impact).

That was approach taken by https://github.com/portfolio-performance/portfolio/pull/1736 , and there was criticism of it, with which I agreed, so I tried to implement short trades using existing transaction types. I cannot really say that's the only and best approach, but IMHO, a good one to start with, and introduce additional transaction types only if some fundamental issue is seen, or it just becomes clear that with additional transaction types, the code becomes more clear/maintainable (so far, I didn't get that impression).

> Is it fair to argue that previously successful trade matching will not change? The logic "flips" only if the first trade is a sale, but that was leading to an exception in the old code.

Yes, as you see, changes are rather minor, contained, and easy to understand. Existing tests pass, new tests which should offer enough coverage are added. There always can be some "what if" reservations, but only improving test coverage can help with that. That's why changes to hopefully improve/ease writing testcases are also added as part of this patchset. (Many existing PP testcases are portfolio XML based, and that's really powerful way to test complex/peculiar cases, but somehow I (as someone not enough familiar with the codebase) have impression that basic unit test coverage worth being improved, and I added a test just along these lines).

pfalcon:
@buchen : What would be the next step here?

pfalcon:
@buchen : Ping

buchen:
merged

pfalcon:
Thanks!

Issue #4828, opened 2025-07-04
[RFC] New type of Cost / Purchase Price calculation - "break even"


pfalcon:
PP historically had "FIFO" method of calculating "Purchase Price", and recently, another way of calculating "Purchase Price" was added - "moving average" (https://github.com/portfolio-performance/portfolio/pull/4546). This addition led to some questions/challenges bother UI and API wise:

* UI wise, as there're actually multiple values/parameters available per method, this leads to large duplication of various choices/fields/columns, often cross-multiplied by other choices (like with/without taxes). This issue was already raised e.g. in https://github.com/portfolio-performance/portfolio/pull/4699#issuecomment-2993390827 . So, assuming that the freedom for user to choose any setting/combination is desired (which makes sense for flexible application like PP), there's still a question how to structure UI. Currently it's effectively enumerations of all possible combinations. Would it make sense to target for more scalable (but more effortful to implement) UI "constructor", like user choosing that they want column of FIFO method, then without taxes.
* On the API side, there's a similar situation - there're large duplication of lazy "getter" methods in https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/LazySecurityPerformanceRecord.java . This problem is actually exacerbated by that "automagic" lazy calculation support, where caching happen on a method level, leading to "dot product" method enumerations like movingAverageCost(), movingAverageCostWithoutTaxesAndFees(). A question is whether it would make sense to rather have a method like getCost(CostCalcType type, boolean withFeesAndTaxes) (which then would have to handle caching explicitly).  

I guess to answer those questions and have a direction for future, it would require to answer a question "Could there be other useful cost calculation methods?". I would like to present answer to that question as "yes", with the example of "break even" cost/price.

The intuition behind beginnings of break-even price should be simple:

1. Suppose you bought some number of shares for 100 units. After some time, you sold half of that number for 200. That means you recouped the original investment. You cannot lose money on this investment. Your break-even price is 0.
2. Suppose you bought a share for 200. Some time later you bought a share for 40 and sold it for 50. Your break even price is now 190.

[...]

PR #4833, opened 2025-07-05
SecuritiesChart: Factor out addPurchasePrice()

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4507

Hello,
This is a proposition to factor out some methods in SecuritiesChart to avoid duplication.
I followed your approach of an Enum file in model, same TaxesAndFees as https://github.com/portfolio-performance/portfolio/commit/e1b515bfc01559bbabf9af778d234c7d9d2b2a35
To be noted that I am also adding a CostMethod enum in PR https://github.com/portfolio-performance/portfolio/pull/4830, but at a different place, if both PR are merged as it is I wonder if this could cause issue

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -51,6 +51,7 @@ import com.google.common.primitives.Doubles;
 import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.AttributeType;
 import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.CostMethod;
 import name.abuchen.portfolio.model.LimitPrice;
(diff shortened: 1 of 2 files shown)


buchen:
> To be noted that I am also adding a CostMethod enum in PR https://github.com/portfolio-performance/portfolio/pull/4830, but at a different place, if both PR are merged as it is I wonder if this could cause issue

I don't think it will cause an immediate problem. But maybe it makes sense to merge them? Have one CostMethod. It would mean - unfortunately - moving labels.

And maybe we can then even move the use of CostMethod into the lazy security performance record... maybe another day.

Thank you very much for cleaning up the SecurityChart class. The class really has grown over time. I had started some time back to have separate package (securitychart with only SharesHeldChartSeries for now).

pfalcon:
@mierin12 : Thanks for this. This covers enough points from https://github.com/portfolio-performance/portfolio/issues/4828 that I felt motivated to make a follow-up adjustment - don't assume that there can/will be only 2 choices and !FIFO == MOVING_AVERAGE: https://github.com/portfolio-performance/portfolio/pull/4834

PR #4838, opened 2025-07-08
fix ClientClassificationFilter when client is already a filtered client

mierin12:
Hello,
This is a proposition of fix of the following topic :
From https://github.com/portfolio-performance/portfolio/pull/4217 and https://github.com/portfolio-performance/portfolio/commit/0f7eb387d416f556cef9a4429a82cde5f4373db7, the possibility to see the TWR and IRR at classification levels was added in the statement of asset.
But there is a strange behavior when the classification node contains Account while a grouped account filtering the whole client portfolio is used : example with kommer, add the TWR column over 10y, add a filtered grouped account containing all of the cash and securities account (so equal to the full portfolio). Similar results are expected with and without the portfolio filter, but this is not the case.

**Before :** 
2025-07-08 20_04_41-Paramètres
2025-07-08 20_05_54-Portfolio Performance
while the grouped account is actually composed of all accounts, no different than Total Portfolio

**After :** 
2025-07-08 20_09_42-Paramètres

Regarding the fix itself : I think this is because a filtered client's accounts are already ReadOnlyAccount and no longer the original accounts, while the `CalculationState state `(`categorizedAccounts `and `vehicle2weight`) are only made of the original accounts. So when they are compared for example at `if (state.isCategorized(account))`, it return false in this filter case, it sees a mismatch.
Not sure if unwrap is necessary, getSource was working too in this kommer example, or if the methods should take care of the fix vs the input account of the methods should be "sanizited unwrap" beforehand instead.
Maybe a test case would be good here ?

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/filter/ClientClassificationFilter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/filter/ClientClassificationFilter.java
@@ -66,12 +66,14 @@ public class ClientClassificationFilter implements ClientFilter
 
         public boolean isCategorized(Account account)
         {
-            return categorizedAccounts.contains(account);
+            return categorizedAccounts
+                            .contains((account instanceof ReadOnlyAccount readOnly) ? readOnly.unwrap() : account);
         }
 
         public BigDecimal getWeight(InvestmentVehicle vehicle)
         {
-            BigDecimal w = vehicle2weight.get(vehicle);
+            BigDecimal w = vehicle2weight
+                            .get((vehicle instanceof ReadOnlyAccount readOnly) ? readOnly.unwrap() : vehicle);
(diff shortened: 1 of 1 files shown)


buchen:
Good catch.

> Not sure if unwrap is necessary, getSource was working too in this kommer example, or if the methods should take care of the fix vs the input account of the methods should be "sanizited unwrap" beforehand instead.

Probably getSource works for now, but if we wrap more (first to strip taxes, then to group by classification), then unwrap works better.

I think the fix is at the right place. The CalculationState "knows" that it extracts the mapping from the taxonomy per object identity, it can also know that accounts should be unwrapped.

buchen:
> Maybe a test case would be good here ?

Good idea. I asked Claude Code to write one: 2c9dfcd59f5cdaa3ccbbb2c95711760c04c6e44b 🙃  I had to do some editing, but not so bad either.

PR #4842, opened 2025-07-11
Add new BBVA (es) PDF importer

Antonio171003:
I added the support for importing PDF files from BBVA (es).

Please review, thanks!


Nirus2000:
Hello @Antonio171003 
thx for this pull request...

I have a question.

**The purchase entry states the following:**
```
Exchange rate: 1.1448 USD/EUR
The gross value in USD is: 2,100.00 USD
The gross value in EUR is: 1,834.38 EUR
```

BOLSA (comisión fija) EUR 20.00000000 20.00

The final amount is **1,854.38 EUR** according to the PDF debug

```
IMPORTE TOTAL GASTOS EUR 28,13 <-- ?!?!??
```

But what is this?
```
SUBTOTAL EN ORIGEN) EUR 8,13 <--- ?!?!??
```

1854,38 EUR - 28,13 EUR is not equal to 1834,38 EUR

(1854.38 EUR - 20,00 EUR  - 8.13 EUR) / 0.87351502 = 2090.67 USD
2090.67 USD is not equal to 2100,00 USD

Somehow the 8,13 EUR are not correct or the amounts have been manipulated.
I think the EUR 8.30 is included in the EUR 20.00 fee, right?

The same in the sale...

Nirus2000:
@Antonio171003 
I have also implemented some errors and improvements in the importer, as well as renaming the files etc.
So.. pls check the importer if the test case now correct and the importer works...

@buchen 
I know... could you please check the ExtractorUtils.java to see if we can implement this?
The problem was that the time format was separated with a dot instead of a colon.
I edited the test cases and inserted the format.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java
```diff
@@ -214,7 +221,19 @@ private static DateTimeFormatter createFormatter(String pattern, Locale locale)
...
+                    };
+
+    // Time formatters with case-insensitive support for various locales
+    private static final DateTimeFormatter[] TIME_FORMATTER = { //
+                    createFormatter("HH:mm", Locale.GERMANY), //
+                    createFormatter("HH:mm:ss", Locale.GERMANY), //
+                    createFormatter("HH.mm", Locale.UK), //
+                    createFormatter("HH.mm.ss", Locale.UK), //
+                    createFormatter("HH.mm", AdditionalLocales.SPAIN), //
+                    createFormatter("HH.mm.ss", AdditionalLocales.SPAIN) //
```
buchen, line 235: Does it really make a difference between UK and SPAIN?

I could see that there might be a difference between "am/pm" and 24hour time format

buchen:
> I know... could you please check the ExtractorUtils.java to see if we can implement this?

Ich stehe gerade auf dem Schlauche ("I am standing on the pipe") - what exactly you want me to have a look at. There are many changes - but most seem cosmetic and unrelated.

Antonio171003:
Hello @Nirus2000

Thank you for your review.

About your question

This is the total amount of all fees and taxes:
> IMPORTE TOTAL GASTOS EUR 28,13 <-- fees + taxes

The breakdown is:

Fixed BBVA commission
> TARIFA BBVA TRADER
> BOLSA (comisión fija) EUR         20,00000000               20,00

Currency management fee
> GESTION DIVISA(SOBRE
> SUBTOTAL EN ORIGEN) EUR                8,13

So I don’t think the €8.13 is included in the €20.00 since they are two different fees, and their sum matches with total expense amount:
> IMPORTE TOTAL GASTOS EUR 28,13 <-- 20,00 + 8,13

However, the numbers actually don’t add up, because we have the following:
> The gross value in EUR is: 1,834.38 EUR
> Fees + Taxes in EUR is: 28,13 EUR

The final amount should be **1862,51** but the document says **1854,38**

But what happens is that BBVA uses a spread exchange rate, so that amount isn’t added on to the final figure because it’s already included in the gross value in EUR, as the exchange rate was unfavorable.

image

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ExtractorUtils.java
Nirus2000: Thats true 👍🏻

Nirus2000:
Add with pr #4850

PR #4851, opened 2025-07-16
Update KBCGroupNVPDFExtractor.java

MikiDi:
Hi,
I noticed pdf extraction wouldn't work for my example pdf. On closer inspection one of the latest updates (https://github.com/portfolio-performance/portfolio/commit/9f48cca1bcefcdb67329dfa0f5f3a04b35805b04) comes closes, but is a bit too specific for the ISIN extraction. I believe my proposal works more generally.
The "waardecode" value isn't necessarily always an ISIN. (may depend on market or someth if it is)
Works with https://github.com/Nirus2000/portfolio/blob/904a65387d25dafca41d768aa2cee42b6524d13c/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/kbcgroupnv/Dividende02.txt


Nirus2000:
Hello @MikiDi 
Thx for this pull request, but we need a PDF debug and a test case for this changes.

So please read the contributing rules... 😀 ... an add the changes.

Regards
Alex

MikiDi:
Thanks for considering! Will add a PDF debug next time.

PR #4863, opened 2025-07-27
fix CSV export of Tax, Fees and Shares of Cash Account

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/38780
Issue https://forum.portfolio-performance.info/t/export-von-kontoumsatzen-bei-zweitem-konto-z-b-tagesgeldkonto-werden-beim-typ-zinsen-keine-steuern-in-die-csv-datei-ubertragen/27855
Issue https://forum.portfolio-performance.info/t/export-anschliessend-import-von-csvs-mit-zinsen/33973

Hello, 

This is a proposition to fix CSV Exporter for Cash Account transactions (File/Export/CSV/Cash Account Transaction) of the following issues: 

- Taxes column was only exported for Dividend, while it can exist for Buy and Sell and Interest
- Fees column was never exported, while it can exist for Buy, Sell, Dividend
- Shares column was only exported for Dividend while it can exist for Buy and Sell

so it was impractical to export/ re import transactions.
**Example :** 
Capture d’écran 2025-07-27 201628

**Before :** 

2025-07-27 20_18_05-No_Name - Excel
Incomplete re-import : 
Capture d’écran 2025-07-27 202110

**After :**

2025-07-27 20_25_54-Test - Excel
2025-07-27 20_24_38-CSV Import

One difference also : previously 0 value Taxes of dividends were exported as "0", now as an empty cell. (= behaviour of the current Table CSV exporter). 

One note also : in this CSV exporter, transactions are NOT sorted by their operation date but by their creation date in PP I think, not sure if this is a feature or not.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVExporter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVExporter.java
@@ -18,6 +18,7 @@ import org.apache.commons.csv.DuplicateHeaderMode;
 import name.abuchen.portfolio.Messages;
 import name.abuchen.portfolio.model.Account;
 import name.abuchen.portfolio.model.AccountTransaction;
+import name.abuchen.portfolio.model.BuySellEntry;
 import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.PortfolioTransaction;
 import name.abuchen.portfolio.model.Security;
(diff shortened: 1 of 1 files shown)


buchen:
> One note also : in this CSV exporter, transactions are NOT sorted by their operation date but by their creation date in PP I think, not sure if this is a feature or not.

I do not think that is a feature... fixed 531688a89b507bf3142818f504b32b49d8351a58

Issue #4889, opened 2025-08-16
Incorrect GBX conversion calculation in sell transactions


Rich28:
**Describe the bug**
This issue occurs when PP is configured with a GBP deposit account and specifically relates to securities which have been acquired priced in GBX. 

In such a situation the GBX conversion calculation PP automatically performs in respect of populating the 'quote' field of a sell transaction provides an incorrect result - the GBX quoted price is unnecessarily multiplied by 100 in the calculation, which is believed to be the cause of this.       

**To Reproduce**
The behavior is best illustrated from the attached screenshots (taken from a modified version of Kommer.xml and comprising fictitious details).

**Expected behavior**
In these situations there is absolutely no need to multiply by 100.

**Screenshots**

In this screenshot note the quote values stated in the two buy transactions, which on both accounts are massively lower than the figure PP automatically generates as being the quote value for the sell transaction, as caused by multiplying this by 100 - as per the figure highlighted in blue.
 
<img width="1128" height="586" alt="Image" src="https://github.com/user-attachments/assets/830936dc-a261-4a6b-8c2d-61c3f1f403f2" />

**Desktop (please complete the following information):**
 - OS: [Windows 11]
 - Version [e.g. 0.78.1]

mierin12:
Fixed in [release 0.79.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.79.0) from https://github.com/portfolio-performance/portfolio/pull/4896

Issue #4322, opened 2024-11-01
Error in auto-generated quote price for Sell/Out-bound transactions


Rich28:
**Describe the bug**

If the currency for the cash account is GBP and a sell or outbound delivery transaction is created for a security priced in GBX, the quote price PP automatically applies to this is incorrect – the GBX quote price data is treated as if it was somehow GBP and there’s a requirement for a further currency conversion, which is wrong.

Likewise, if a sell or outbound delivery transaction is created for a security priced in USD (with the currency for the cash account still in GBP), the quote price PP automatically applies to this is also incorrect for the same reason.  

It is unknown whether this issue only occurs if the cash account is in GBP

**To Reproduce**

1. Ensure cash account is in GBP
2. Right click on an existing transaction for a security priced in GBX or USD and select either ‘Buy’ or ‘Delivery (Outbound)'
3. In the dialog box that appears note the automatically generated quote price, which mistakenly has been subject to two currency conversions…

**Expected behavior**
Rather than generating an incorrectly inflated quote price, for both types of transactions (‘Buy’ and ‘Delivery Outbound') should use the unaltered as obtained from the data source

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 0.71.2

mierin12:
Fixed in [release 0.79.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.79.0) from https://github.com/portfolio-performance/portfolio/pull/4896

Issue #2543, opened 2021-11-21
Wrong quote suggested when selling GBX securities from a GBP account


jfitzell:
Reported in version 0.55.0 on Mac OS X

**To reproduce:**
1. Create 2 securities accounts in GBP (only 1 is necessary but 2 helps make the bug clearer below)
2. Add a security priced in GBX (I use 0P0000TKZP.L below)
3. Create a "Buy" transaction for any number of units in one account
4. Create a "Sell" transaction

**Expected behaviour:**
Suggested quote should always be filled with the latest historical price for the security

**Actual behaviour:**
✅ When no securities account selected: quote is correct
![image](https://user-images.githubusercontent.com/2814934/142766136-24602284-f72d-416a-8b95-2065aa525c1a.png)

✅ When a securities account containing no units of the security is selected: quote is correct
![image](https://user-images.githubusercontent.com/2814934/142766159-0c549d69-8b1e-4c1e-bb4f-d08c5de807ff.png)

⚠️ When a securities  account containing units of the security is selected: quote is multiplied by 100
![image](https://user-images.githubusercontent.com/2814934/142766179-c84d01b7-d668-49fc-973e-21dd2237efa0.png)

mierin12:
Fixed in [release 0.79.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.79.0) from https://github.com/portfolio-performance/portfolio/pull/4896

PR #4893, opened 2025-08-20
fix automatic quote and total in foreign currency in Sell and Outbound delivery transaction dialog

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/2543
Closes https://github.com/portfolio-performance/portfolio/issues/4322
Closes https://github.com/portfolio-performance/portfolio/issues/4889

Hello,
This is a proposition to fix the above issues. They all mention GBX/GBP but the issue applied to any currency exchange.
The problem is easy to reproduce from the detailed steps of #2543. And was also existing with a time change or a date change or security change. 

In `updateSharesAndQuote`, at `setTotal(position.calculateValue().getAmount())`,  `position.calculateValue().getAmount` is in the security currency, but `setTotal` was using it directly to set the total value of the transaction, which is in the portfolio currency.

`setTotal` logic is slightly changed to first setup the `this.grossValue` (which is in the security currency), then all the rest are computed. I think it actually makes a bit more sense, before the logic was to estimate first the total final sell value from the market value and then deduced all the other value and quote from it  (market value->total->converted gross value->gross value-> quote), but with taxe & fees, the final total value will be different than market value, while `this.grossValue` is the one equivalent to market value.

Additionaly, I think the `updateExchangeRate` should be called before `updateSharesAndQuote` in `setSecurity`.

Edit : maybe the setTotal method name should be modified to better reflect the new Logic ? Also, I quickly check, I do not think this method is called elsewhere,  but if it is it could be an issue.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AbstractSecurityTransactionModel.java
@@ -276,8 +276,8 @@ public abstract class AbstractSecurityTransactionModel extends AbstractModel
         firePropertyChange(Properties.inverseExchangeRateCurrencies.name(), oldInverseExchangeRateCurrencies,
                         getInverseExchangeRateCurrencies());
 
-        updateSharesAndQuote();
         updateExchangeRate();
+        updateSharesAndQuote();
     }
 
     protected void updateSharesAndQuote()
@@ -302,6 +302,8 @@ public abstract class AbstractSecurityTransactionModel extends AbstractModel
                 if (position != null)
                 {
                     setShares(position.getShares());
+                    // position.calculateValue().getAmount() is in the security
+                    // currency
                     setTotal(position.calculateValue().getAmount());
                     hasPosition = true;
                 }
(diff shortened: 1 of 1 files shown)


buchen:
> maybe the setTotal method name should be modified to better reflect the new Logic ? Also, I quickly check, I do not think this method is called elsewhere, but if it is it could be an issue.

The #setTotal method is used if the user enters a value into the dialog.

Unfortunately, the use is indirect. But the framework that is binding an object property to a dialog input field, is doing this on Java Bean properties. The bean property "total" is then converted into "setTotal" and "getTotal" calls. But of course the code search does not see that anymore.

https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/SecurityTransactionDialog.java#L193-L194

If the user enters the total, now actually the gross value is changed.

> before the logic was to estimate first the total final sell value from the market value and then deduced all the other value and quote from it (market value->total->converted gross value->gross value-> quote), but with taxe & fees

The idea was:
* if the user enters "from the top" - shares ->  prices -> gross - then the total is updated
* if the user enters data "from the bottom" -> taxes and then total, the gross values are updated

Maybe that logic does not work (or is not necessary). A better logic would be to keep track of the values that the user explicitly entered, and then calculate the rest. Sometimes only the total is available. Sometimes the gross values are available.

buchen:
@mierin12 I have adapted your change with pull request #4896 in order to not change the setTotal method

What do you think?

buchen:
Continue in #4896

PR #4894, opened 2025-08-20
add tax field in Purchase investment plan

mierin12:
Hello,

This is a proposition to add the Tax field in the Purchase/Delivery Inbound investment plan. This field is available in the normal Purchase/Delivery Inbound, but was not present in the investment plan.


buchen:
We continue here: #4897

PR #4896, opened 2025-08-22
Fix automatic quote and total in forex (security transaction dialog)

buchen:
Issue: #4893
Issue: #2543


mierin12 (approved):
Hello,
Understood about the setTotal that should not be changed, yes I think your implementation works, thanks !

PR #4920, opened 2025-08-29
Add support for a new fees file type in BBVA PDF Extractor

Antonio171003:
I added support for a new fees type for BBVA bank.

Please review, thanks!


Nirus2000:
Hello @Antonio171003 

We do not extract taxes for normal fee transactions, so I have removed them and 
merged them into a new pull request. (#4923)

PR #4947, opened 2025-09-06
"Y" and "N" hotkeys for the "Save" dialog

JGKle:
Make the "Save" dialog respond to the standard "Y" and "N" hotkeys

### DIFF
--- a/name.abuchen.portfolio.bootstrap/src/name/abuchen/portfolio/bootstrap/messages.properties
+++ b/name.abuchen.portfolio.bootstrap/src/name/abuchen/portfolio/bootstrap/messages.properties
@@ -1,17 +1,17 @@
 
-LabelCancel = Cancel
+LabelCancel = &Cancel
 
 LabelCloneWindow = Clone
 
 LabelCloseWindow = Close
 
-LabelNo = No
+LabelNo = &No
 
 LabelSaveAll = All
 
 LabelSaveNone = None
 
-LabelYes = Yes
+LabelYes = &Yes
 
 SaveHandlerMsgSelectFileToSave = Select the files to save:
 


buchen:
I never knew that Y and N are standard hotkeys. Never tried it (typically use *Enter* as the default option or *Esc* to decline). I just tried it on macOS on another dialog, but that does not support it.

What is the typical behavior? Adapt it to the translation, e.g. "J" for "Ja" (Yes) in German or have a fixed binding to Y?

JGKle:
> typically use Enter as the default option or Esc to decline

The action for Escape is to cancel - there's no hotkey for "No" (i.e. to exit but don't save).

> I just tried it on macOS on another dialog, but that does not support it.

Interesting - maybe it's more widespread on Windows? I do occasionally encounter something that doesn't support it, but it's definitely the exception by a pretty large margin.

> What is the typical behavior? Adapt it to the translation, e.g. "J" for "Ja" (Yes) in German or have a fixed binding to Y?

That I'm not sure, but my guess is that the accelerator would match the first letter of the word, similar to activating entries in a menu.

Tbh tho "Y" isn't really that important here (as you mentioned, you can just hit enter). The "No" key is really what I was trying to solve.

buchen:
> The "No" key is really what I was trying to solve.

You are right. Esc cancels.

> That I'm not sure, but my guess is that the accelerator would match the first letter of the word, similar to activating entries in a menu.

Out out curiosity, does it work if you change the label to "&No"? At least on macOS that is stripped, but on Windows it might work (but also might require Alt-N)

I then have to check how this works for languages like Chinese. But, yes, also those little convenience things make sense

JGKle:
> Out out curiosity, does it work if you change the label to "&No"? At least on macOS that is stripped, but on Windows it might work (but also might require Alt-N)

I actually tried that, but it didn't work on Linux (which is the platform where I was building). I use it on both Windows & Linux.

wahlbrink:
The usual solution is the '&' in the labels. For the Eclipse IDE it works for me on Windows (also without ALT) and Linux (only with ALT).
(Didn't tested it with PP)

JGKle:
> The usual solution is the '&' in the labels. For the Eclipse IDE it works for me on Windows (also without ALT) and Linux (only with ALT). (Didn't tested it with PP)

Ok I'm dumb - I guess it does work just by altering the labels, I just didn't realize u have to use Alt on Linux (I was building on Linux but am more accustomed to Windows, so I just tested "N" only. Will push a fix :)

JGKle:
Done - tested & works on both. Easy peasy.

buchen:
Now it looks like a very nice and small change.

Merged.

Issue #4586, opened 2025-03-16
Problem importing CSV file - byuing with currency exchange


psyfreaky:
**Describe the bug**
It is not possible to import CSV file when you try to buy USD stock with EUR with exchange rate (or any stocks when you mix currencies).

**To Reproduce**
1. Open portfolio settings (attached). All deposit and securities accounts are set.
2. Import CSV (attached). All columns are prepared.
3. 1 import line is ok (buying USD with USD), second line is with error.
4. It shows "Exchange rate of gross value is missing"

**Expected behavior**
Import both lines.

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 0.74.1

testing import file
[test-import.csv](https://github.com/user-attachments/files/19273811/test-import.csv)
testing settings file (had to rename the extension)
[Portfolio-test-import_xml.txt](https://github.com/user-attachments/files/19273815/Portfolio-test-import_xml.txt)

psyfreaky:
It seems that the trieck is to choose "Type of data" as "Portfolio Transactions" if anyone searches.

mookkari:
The importer "Portfolio Transactions" is limited and only supports buy and sell types. 
I have types like Deposit, Fees, Dividend, etc in same CSV file, so I should use "Type of data" as "Account Transactions". This is working fine for all other lines except buying with currency exchange.

I still think this is a bug that must be fixed to "Account Transactions" importer.

Issue #3000, opened 2022-10-05
CSV Import of Buy/Sell Securities fails - "Exchange rate of gross value is missing"


matt-winfield:
**Bug Description**
When trying to import security transactions from CSV where type is buy or sell, and the transaction currency is different from the security currency (in my case, GBP and GBX), then those rows fail to be imported. The error indicates "Exchange rate of gross value is missing (transaction currency GBP and security currency GBX)"

**To Reproduce**
1. File -> Import -> CSV Files
2. Select CSV file containing transaction with type "Buy"
3. Data type is "Account Transactions"
4. Map the "Date", "Type", "Shares", "Value", "Exchange Rate", "Gross Amount" and "Currency Gross Amount" fields to the correct columns
5. Click next
6. The row is crossed out, right click to see the error

Adding the transaction manually through the UI works fine for this account

Example CSV that doesn't work:
```
Date,Type,Shares,Symbol,Description,Value,Contract Ref,ISIN,Sedol,Exchange Rate,Gross Value,Currency
17-Dec-2018,Buy,12.34,XYZ,EXAMPLE STOCK GBP,500.0,ZYX,GB00123,ABC,100,50000.0,GBX
```

**Expected behavior**
The transactions should be created successfully. It seems there was an update made that allowed this to work for dividend payments: https://forum.portfolio-performance.info/t/dividend-csv-update/15977
But it doesn't seem to work when the type is buy/sell.

This issue is discussed on the forum here: https://forum.portfolio-performance.info/t/exchange-rate-of-gross-value-is-missing-when-importing-csv-transactions-with-currency-conversions/20633/5?u=mattw

**Screenshots**
Screenshot of the error (Ignore “transaction does already exist”, that shouldn’t affect this):
![image](https://user-images.githubusercontent.com/15931560/194054084-107f55f1-4bb8-4490-a5a2-9c050ae6426a.png)

And the field mapping - I’ve tried mapping currency, gross amount, currency gross amount:
![image](https://user-images.githubusercontent.com/15931560/194054110-37702aca-efd1-401d-a9aa-0ab7c2989d11.png)

**Possible solution**
I had a look through the code and identified what may be causing the issue. I don't have local development set up to test the changes, but hopefully this points someone who's worked with this project more in the right direction:

It seems the bug might be caused by missing transaction.addUnit for buy/sell types.
Here is the code that imports Dividends and most other transaction types:
[...]

PR #4948, opened 2025-09-06
fix CSV importer of Buy/Sell account transaction with exchange rate

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/3000
Closes https://github.com/portfolio-performance/portfolio/issues/4586
Issue https://forum.portfolio-performance.info/t/import-csv-when-currency-conversions-include/34306
Issue https://forum.portfolio-performance.info/t/csv-import-wahrungproblem/37349

Hello,

This is a proposition to fix the following topic :

In CSV importer, when using the Account Transaction (which is default mode of the CSV importer), it is not possible to create buy/sell transactions when a exchange rate is required. While it is possible in Portfolio Transaction mode. So far, the workaround has been to recommend to use this Portfolio Transaction mode instead.
But, it is a workaround only, and since Account Transaction allows to create Buy/Sell and its CSV importer already accepts an exchange rate, I assume for foreign dividends or transfer, I think they should also allow the creation of those foreign buy/sell too.

Some notes :
- In the first commit, one little modification to have Portfolio Transaction as the default importer when the wizard is coming from PortfolioListView right click "import csv", so when we have a default portfolio account selected (but unsure if this is really necessary now).

- I also proposed that the CSV export of Account Transactions include the exchange/gross amount/currency, to be able to have compatibility with between Export-Import csv files.
- Two tests added, those are mostly the same as in CSVPortfolioTransactionExtractorTest.

To be noted that now CSVPortfolioTransactionImporter/Exporter and CSVAccountTransactionImporter/Exporter are quite similar.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import org.junit.Test;
 
 import name.abuchen.portfolio.Messages;
+import name.abuchen.portfolio.datatransfer.Extractor.BuySellEntryItem;
 import name.abuchen.portfolio.datatransfer.Extractor.Item;
 import name.abuchen.portfolio.datatransfer.Extractor.SecurityItem;
 import name.abuchen.portfolio.datatransfer.Extractor.TransactionItem;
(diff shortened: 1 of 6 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
```diff
@@ -621,4 +622,93 @@ public void testInterestWithTaxesTransaction()
...
+                        "2015-09-15", "XX:XX", // Date + Time
+                        "LU0419741177", "", "", // ISIN + TickerSymbol + WKN
+                        "56", "EUR", // Amount + Currency
+                        "BUY", // Type
+                        "", "-0,701124", // Security name + Shares
+                        "Notiz", // Note
+                        "", "0,14", // Taxes + Fee
+                        "", "", "", // account + account2nd + portfolio
+                        "", "USD", // Gross + Gross currency
+                        "1,1194" }), // Exchange rate
```
buchen, line 694: Here the exchange rate is used as:
base currency (EUR) * exchange rate = value (USD)
56 EUR * 1.1194  = 62.53 USD

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
```diff
@@ -621,4 +622,93 @@ public void testInterestWithTaxesTransaction()
...
+                        "2013-01-02", "", // Date + Time
+                        "DE0007164600", "SAP", "", // ISIN + TickerSymbol + WKN
+                        "-100", "EUR", // Amount + Currency
+                        "SELL", // Type
+                        "SAP SE", "1,9", // Security name + Shares
+                        "Notiz", // Note
+                        "12", "", // Taxes + Fee
+                        "", "", "", // account + account2nd + portfolio
+                        "110", "USD", // Gross + Gross currency
+                        "0,9091" }), // Exchange rate
```
buchen, line 648: Here the exchange rate is used as:
forex value (USD) * exchange rate = value (EUR)
110 USD * 0.9091 = 100 EUR

buchen (commented):
@mierin12 I think we are using the exchange rate differently. I think the issue is not introduced with this pull request, but exists in the portfolio-transaction importer as well.

What I mean if the gross amount is forex is given, it is interpreted as (forex) x (rate)=(transaction currency). If only the exchange rate is given, it is interpreted as (forex)/(rate)=(tx currency), e.g. calculated as (tx currency) x (rate)=(forex).

Or am I missing something?

I wonder how to "fix" it. Maybe it was never reported, because CSV data just contains either rows with gross value or only rows with exchange rates. Then maybe it is "okay" to live with it (and not an incompatible change)? But: the export of course we produce consistent exchange rates.

Of course, it would be better if the user could then provide the "notation" (Mengennotierung vs. Preisnotierung). But that is a bigger change.

buchen:
Just as a side node (and for me to keep the patch somewhere), we can formulate that tests as:

```java
        // check that gross value is fixed taken taxes into account
        assertThat(results, hasItem(sale( //
                        hasDate("2013-01-02"), hasShares(1.9), //
                        hasSecurity(hasTicker("SAP.DE")), //
                        hasAmount("EUR", 100.00), //
                        hasGrossValue("EUR", 112.00), //
                        hasForexGrossValue("USD", 123.20), // 112.00 * 0.9091
                        hasTaxes("EUR", 12.00), hasFees("EUR", 0.00))));
```

and 

```java
        // check that gross value is created with given exchange rate
        assertThat(results, hasItem(purchase( //
                        hasDate("2015-09-15"), hasShares(0.701124), //
                        hasSecurity(hasIsin("LU0419741177")), //
                        hasAmount("EUR", 56.00), //
                        hasGrossValue("EUR", 55.86), //
                        hasForexGrossValue("USD", 62.53), // 1.1194 * 55.86
                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.14))));
```

mierin12:
Yes indeed. I also think it was already present in the Portfolio transaction, I only wanted here to extension the functionnality to AccountTransaction. I am also not sure if something should be done or not. Exchange rate topics are never simple :)
I did not deep dice on this, some quick thought 
- Could a change break a user created template ? I think typical csv importer template can be created in PP.
- it seems to me the case where the gross value is present might be the clear one ? I think this is used also for Dividend. 
- 'gross value is present' is the case that will be exported, so the compatibility export-import should be working for this case.
- How is it done for the PDF or xml importer ? Is there such case where gross value is not present but only the exchange ? In which case, a consistent behaviour between the different importers would be the best way ? Maybe some ExtractorUtils are even shared between them ?
- At least, even if nothing is changed, I would say it has to be documented, or made clear at the import time. Something like a tooltip, or an example ? To be clear which exchange rate is expected in which case.

buchen:
I merged it now as is. I think it is not breaking it. But mid-term we could improve the CSV import by allowing to configure such things.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/ImportCSVHandler.java
```diff
@@ -87,8 +87,10 @@ public static void runImport(PortfolioPart part, Shell shell, IEclipseContext co
         if (account != null)
             wizard.setTarget(account);
         if (portfolio != null)
+        {
             wizard.setTarget(portfolio);
-
+            wizard.setExtractor("portfolio-transaction"); //$NON-NLS-1$
```
pfalcon, line 92: @mierin12 : I've been investigating why after recently update, "Portfolio Transaction" CSV format is being forced on me every time, which is otherwise something I never use (so, I must switch back to normal "Account Transaction" all the time). 

Here's the story: it's overall problem that doing simple things like performing the same import again and again in PP requires too many clicks. Many small precise clicks to be exact, like setting cash and securities accounts all the time. The way to avoid that is to invoke "Import CSV..." context menu in "Securities Account" list. That has pairs of (securities_acc, default_cash_acc), so in the case of a single cash acc (which is true for 90% of setups I guess), we achieve the aim of no need to perform additional clicks to select those accounts.

Hopefully you see the problem now - the fact that user pre-selected securities acc. as described above has absolutely no bearing of what *format* a user wants to import (because even for "Account Transactions", you still need both). And if users where happy with "Account Transaction" before, then this change is a regression requiring them to perform unneeded clicks again.

mierin12: Apologies, I think this is easy to revert. 
I thought an "import csv" from a securities account right click meant that only portfolio transaction were meant to be imported as they cannot be used for deposit, removal or Dividend.

pfalcon: Thanks!

PR #4959, opened 2025-09-08
Add Nordax Bank PDF-Importer cash transfer transaction

Nirus2000:


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NordaxBankABPDFExtractor.java
```diff
@@ -60,5 +61,29 @@ private void addAccountStatementTransaction()
...
+        // 24.08.2024 Übertrag des verlängerten 10.000,00 0,00
+        // @formatter:on
+        var transferBlock = new Block("^[\\d]{2}\\.[\\d]{2}\\.[\\d]{4} .bertrag des verl.ngerten [\\.,\\d]+ [\\.,\\d]+$");
+        type.addBlock(transferBlock);
+        transferBlock.set(new Transaction<AccountTransferEntry>()
+
+                        .subject(() -> {
+                            var accountTransferEntry = new AccountTransferEntry();
+                            return accountTransferEntry;
+                        })
```
Nirus2000, line 76: I'm not sure if we need to set any type of transaction (transfer_in or transfer_out)... its a transfer...

buchen: Good question. The "direction" of the transfer is determined by the source (where the money is removed) and the target account (where the money is added).

The ```AccountTransferItem``` has a property isOutbound. You currently set it to ```true```, but you could also reverse the direction if you want.

Issue #4965, opened 2025-09-09
Portfolio Time Machine for more menus


CanePlayz:
**Is your feature request related to a problem? Please describe.**

I would like to be able to see certain menus such as my dashboard or the Statement of Assets diagram at a certain point in the past. In the Statement of Assets, this is already possible by using the "Portfolio Time Machine".

**Describe the solution you'd like**

Extend the Time Machine feature to other menus.

**Describe alternatives you've considered**

You can technically use filters that range from the beginning of your investment history to the day you desire to see, but having the Time Machine available in a menu makes viewing these things way easier and more accessible.

mierin12:
I think it makes sense for the Statement Of Asset/Holdings' pie chart diagram.
For the other views, I do not see yet for which chart or values this could be relevant, do you have an example ?

CanePlayz:
These are the main ones, perhaps also the dashboard and the performance chart

mierin12:
I will propose a PR for the Holdings Pie Chart view, I think I manage to get it working.

For the performance chart, I still do not see, the performance chart is to draw series over a duration period, I do not see what a time machine with a single date would change ?
For dashboard, I guess it depends on each widget ? Which widget would be impacted by a time machine setting ?

CanePlayz:
Thanks.

Yeah, you're right about the performance chart.

An example dashboard of mine that would be impacted is a collection of many performance widgets, each one for one type of asset. Going back in time would allow me to see how each asset has performed up to a special point in time.

mierin12:
With [release 0.80.0](https://github.com/portfolio-performance/portfolio/releases/tag/0.80.0), the time machine is available on the Statement of Assets/Holdings view.

<img width="1740" height="290" alt="Image" src="https://github.com/user-attachments/assets/e8e93398-6b2f-49a5-b154-7dda15706f9b" />

Regarding the "many performance widget", this is still unclear to me. For them, the reporting period can be set and can be set in the past, so really I do not see which widget exactly could be impacted by a time machine when they already have the reporting period parameter.

CanePlayz:
First of all, thank you 👍

Regarding the second topic: This is how the dashboard looks:

<img width="189" height="456" alt="Image" src="https://github.com/user-attachments/assets/863e1885-9188-4e34-9fae-2d29ff9a89ac" />

I would like to be able to easily see how the different performances compared against each other on a certain point in time.

[...]

PR #4975, opened 2025-09-10
add Time Machine in Statement of Assets/Holdings view

mierin12:
Partially Closes https://github.com/portfolio-performance/portfolio/issues/4965
Issue : https://forum.portfolio-performance.info/t/portfolio-time-machine-auch-fuer-bestaende-gewuenscht/10261

Hello,

This is a proposition to add the time machine in the holdings view.
The methods already present in the `StatementofAssetsView` are moved in a separate class to be reusable. Some of those methods are very slighty modified (`getActiveShell() `that I could not access anymore -> `Display.getDefault().getActiveShell() `)
When a date other than today is selected, I added it to the title of the Holdings view.

(By the way about the title of the different views, with the addition of selected filter name in the titles, users that do not name their filter/grouped account and therefore have the very long name ("account 1 + account 2 + ...") can have the issue that the title is longer than the window, making the different icons no longer reachable. I did not manage to implement a truncate method in those case.)

**Results :** 
2025-09-10 20_21_39-Portfolio Performance

Capture d’écran 2025-09-09 235938

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TimeMachineDropDown.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TimeMachineDropDown.java
@@ -1,0 +1,108 @@
+package name.abuchen.portfolio.ui.util;
+
+import java.time.LocalDate;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Consumer;
+
+import org.eclipse.jface.action.IMenuListener;
+import org.eclipse.jface.action.IMenuManager;
+import org.eclipse.jface.action.Separator;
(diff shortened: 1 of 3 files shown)


buchen:
merged.

> have the very long name ("account 1 + account 2 + ...") can have the issue that the title is longer than the window, making the different icons no longer reachable

The logic right now is: first display the title, then grab as much as possible for the views (but show a chevron if there more views than available), and finally show the tool bar at the right side.

But I guess we would need a logic like this:
* always show the buttons of the toolbar at the right side
* at least show one view item for the view (plus the chevron), but if the title is shorter, show more
* show as much as possible of the title
I think that would need a custom layout

Issue #2719, opened 2022-03-02
[BUG] Pie and Donut Chart under Asset Allocation aren't rendered under Debian 11


RequestPrivacy:
**Describe the bug**
The Pie and Donut chart in the Asset Allocation Section in Taxonomies aren't rendered if revisited after once viewed. 

**To Reproduce**
1. Open Portfolio Performance 
2. Click on Asset Allocation under section "Taxonomies" (if not existent follow steps to create an portfolio with an corresponding asset allocation)
3. Go to pie chart (it renders)
4. Go to donut chart (it doesn't render)
5. Go back to pie chart (it does render again or remains in cache)
6. Go to different section (e.g. Deposit Accounts)
7. Go back to Asset Allocation
8. Click on pie chart if not already selected
9. Pie chart doesn't render + donut chart doesn't render

**Behavior**
Donut chart in asset allocation isn't rendered at all and pie chart is rendered once, but if you leave for another section and come back it doesn't render again.

**Desktop (please complete the following information):**
 - Linux, Debian 11 bullseye
 - Version: 0.56.5 (Feb 2022)
 - Java: 16.0.1+9, AdoptOpenJDK

fabtagon:
Same here, 
* Linux, Ubuntu 20.04
* PP Version: master (86e947d596dcc5d719973f1015725e4daf7fad00)
* Java: openjdk 11.0.14.1 2022-02-08
 
When the chart is rendered for the first time, the following GTK error gets printed to stdout:
```
(Portfolio Performance:8358): Gtk-WARNING **: 00:59:21.859: Negative content width -4 (allocation 14, extents 9x9) while allocating gadget (node entry, owner GtkSearchEntry)
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug

*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug


(Portfolio Performance:8358): Gtk-CRITICAL **: 00:59:22.193: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar

(Portfolio Performance:8358): Gtk-CRITICAL **: 00:59:22.233: gtk_box_gadget_distribute: assertion 'size >= 0' failed in GtkScrollbar

```

[...]

PR #5004, opened 2025-09-22
Linux Pie Charts

mierin12:
Hello, this is just a draft POC from @hendrikm78 proposition here : https://github.com/portfolio-performance/portfolio/issues/2719#issuecomment-3005852373

> I figured out the following:
> 
> - `pp.getInitialHeight `& `pp.getInitialWidth `return -4, seems the dimensions can not be determined when the script is called
> 
> To avoid this problem I did two workarounds:
> 
>  1. Limit the output of these two functions to >= 50
>  2. add a EventListener "load" to the flare and pie chart
> 
> For details see attachment. (I am not familar with git, so I would do it this way)
> The Flare-Chart works fine, but the Pie-Chart is a little bit tricky. Sometimes you have to switch to another Taxonomy and go back.
> 

I do not have Linux myself, but as per the comments in the linked discussion, **it seems this is fixing the "sunburst" pie chart**, but not fixing the "donut" pie chart.
So this seems promising, I think this is worth a bit of visibility, in case Linux contributors want to further investigate.
On my side, both "sunburst" and "donut" pie chart keep functioning as intended on Windows 11.

If this goes somewhere, the contributor is hendrikm78, not me.

### DIFF
--- a/name.abuchen.portfolio.ui/META-INF/html/flare.html
+++ b/name.abuchen.portfolio.ui/META-INF/html/flare.html
@@ -1,11 +1,14 @@
 <!DOCTYPE html>
+<html>
 <meta charset="utf-8">
 <link rel="stylesheet" type="text/css" href="${/META-INF/css/THEME.css}">
 <body>
 <script src="${/META-INF/js/d3.min.js}"></script>
 <script src="${/META-INF/js/pp.js}"></script>
 <script>
+window.addEventListener('load', createFlare);
 
+function createFlare(){	
 pp.onResizeEnd(function() {
   width = pp.getInnerWidth();
   height = pp.getInnerHeight();
@@ -239,4 +242,7 @@ function brightness(rgb) {
 }
 
 flare.setContent(loadData());
+}
 </script>
+</body>
+</html>
--- a/name.abuchen.portfolio.ui/META-INF/html/pie.html
+++ b/name.abuchen.portfolio.ui/META-INF/html/pie.html
@@ -6,9 +6,11 @@
 <script src="${/META-INF/js/d3.min.js}"></script>
 <script src="${/META-INF/js/pp.js}"></script>
 <script>
+window.addEventListener('load', createPie)
 var data = JSON.parse(loadData());
 var total = d3.sum(data, function(d) {return d.value;} );
 
+function createPie() {
 pp.onResizeEnd(function() {
     thePieChart.resize();
 });
@@ -245,4 +247,7 @@ thePieChart.update(data);
 function brightness(rgb) {
   return rgb.r * .299 + rgb.g * .587 + rgb.b * .114;
 }
+}
 </script>
+</body>
+</html>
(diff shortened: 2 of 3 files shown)


buchen:
I tested in on Linux.

Interesting: I get to see the pie charts in the taxonomy, but only if that is the initial view to be shown. If the taxonomy view opens with something else, then I see a small (I assume 50 by 50 pixel) thing and then an empty screen.

buchen:
BTW, that fits to the behavior that @hendrikm78 reported:

> The Flare-Chart works fine, but the Pie-Chart is a little bit tricky. Sometimes you have to switch to another Taxonomy and go back.

buchen:
I tested it on macOS and Windows - there the scripts continue to work as expected. I think we can merge this, but I think the default on Linux should still be the SWT charts

mierin12:
> I tested in on Linux.
> 
> Interesting: I get to see the pie charts in the taxonomy, but only if that is the initial view to be shown. If the taxonomy view opens with something else, then I see a small (I assume 50 by 50 pixel) thing and then an empty screen.

Does this initial vs non initial behaviour happens for both charts, or only for the donut chart, and flare is working correctly unconditionally all the time ?

buchen:
> Does this initial vs non initial behaviour happens for both charts, or only for the donut chart, and flare is working correctly unconditionally all the time ?

Only for the pie charts (holdings and donut). The flare chart seems to work all the time.

buchen:
This is how it looks now (I merged it with fbe8eab4b0842cd49846de7fade35280b0397727).

You if one opens the donut the first time around, there is a small square of labels (I assume the 50 by 50). And then the chart is gone. If the chart is the default view to be opened for this taxonomy, then it is shown:

https://github.com/user-attachments/assets/33396b69-8d19-4718-bffe-e7fbc72c75a2

mierin12:
It is great then if it is an improvement for Linux for the Flare. It is a bit puzzling that the behaviour is different between those two charts, while the implementation seems quite similar.

I wonder if there is a small side effect after all on Windows. I think there a visible one frame flicker when changing the accounts or the option "with/without Unclassified", "with/without securities level" ? On the Holdings, the change of selected accounts seems smoother.

buchen:
> I wonder if there is a small side effect after all on Windows. I think there a visible one frame flicker when changing the accounts or the option "with/without Unclassified", "with/without securities level" ?

To be honest, I only briefly tested on Windows with this change.

You mean there is now additional flicker on Windows when applying a "filtering options" such as "with/without unclassified"? I am not sure if that is new. If the file changes (new prices, new filter), I need to reload the HTML page. And for one frame, it renders an empty page. One could get around this - maybe - by triggering a method to update the pie chart. I haven't tried. I find it super hard to develop this because I could not find a debug console for the embedded browser.

mierin12:
I am not sure if this is new either to be honest, I can double check on an older version.

I took a video where we see the two behaviors : 

- Taxonomy : reload can be seen as there is a empty view in between
- Statement of assets holdings : generally* without the "in between empty view" reload

*at the end of the video we can see it is here too. But it is not really reproductible and I think quite rare there. 

https://github.com/user-attachments/assets/5a46a2a3-3da8-43c2-8e28-c5847f1ebc47

I agree this seems complicated while it is only a very small topic.

mierin12:
closing at it was included with https://github.com/portfolio-performance/portfolio/commit/fbe8eab4b0842cd49846de7fade35280b0397727

PR #5007, opened 2025-09-23
Update SecuritiesChart.java

OnkelDok:
Chart colors remain unchanged for benchmark, as proposed here: https://forum.portfolio-performance.info/t/farben-von-chart-grafiken-unverandert-lassen/37593. But only when user is marking securities in same order as displayed in the list (see my comment in forum).

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecuritiesChart.java
@@ -917,7 +917,7 @@ public class SecuritiesChart
         chart.setRedraw(false);
         chart.suspendUpdate(true);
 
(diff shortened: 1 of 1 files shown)


buchen:
I think it is a valid feature request, but not so easy to implement.

Maybe:
* we remember - for the session - which color was assigned to the instrument the first time it was selected for "multi line charts". If it is selected again, then we assign the same color. (I am not sure if we have state attached to the SecuritiesChart. If we assign them one after the other, the colors still maybe distributed well enough.
* we take the hash of the UUID of the security to get to a the id of the color (and maybe the color wheel has the size of the total number of instruments). This way we do not need to remember anything. Not sure how well the colors will be distributed though.

I also believe we do not know the order in which the instruments were marked.

buchen:
@OnkelDok I have played around and now remember the colors assigned to the instrument. Additionally, I try to assign them on the color wheel better. See 6640f4c970f719386ab83558de754f5a71c662b6

Issue #4999, opened 2025-09-20
Shortcut for Edit in Statement of Assets


dylan09:
In **All Securities** view the **Edit** command has a Shortcut (CMD + E) on macOS.

It would be fine to have this shortcut in all views where the properties of a security could be edited.

mierin12:
Done with https://github.com/portfolio-performance/portfolio/commit/8d1db7fd4850ecd491be95e7562a57bfd34bb546, will be available in the next release.

dylan09:
Thrilled. Thank you.

Issue #4091, opened 2024-06-22
Strg-E shortcut for RMB "Edit" menu entry missing


DanielFeld2020:
I'm missing the Strg-E shortcut for the RMB "Edit" menu entry in "Reports/Statement of Assets" AND "Reports/Securities". We already have this shortcut working in "Securities/All Securities".  It's quite frustrating to hit Strg-E and sometimes it works and sometimes not...

Could some developer please add the Shortcut?

Thanks!

mierin12:
Done with https://github.com/portfolio-performance/portfolio/commit/8d1db7fd4850ecd491be95e7562a57bfd34bb546, will be available in the next release.

PR #5011, opened 2025-09-25
add Security Edit shortcut in views

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4091
Closes https://github.com/portfolio-performance/portfolio/issues/4999

Hello, 
This is a proposition to add the Ctrl+E shortcut to the action "Edit Security". This is already available in the Security List View, but only there. So the proposition is to add it to the other views where a Security Context Menu exists. The exception is for transactions, where the Ctrl+E shortcut is already related to the "Edit transaction" action.

One detail, on Tree (Performance/Calculation and Taxonomy), Windows gives me a "beep" sound when closing the Edit window. But it is actually already (before the PR change) giving me the beep if press a Ctrl+E or else with a Tree row selected.

**Before**
2025-09-25 21_06_41-Portfolio Performance

**After**
2025-09-25 21_10_33-Portfolio Performance

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PerformanceView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PerformanceView.java
@@ -28,6 +28,8 @@ import org.eclipse.jface.window.ToolTip;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CTabFolder;
 import org.eclipse.swt.custom.CTabItem;
+import org.eclipse.swt.events.KeyAdapter;
+import org.eclipse.swt.events.KeyEvent;
 import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.Image;
@@ -417,6 +419,7 @@ public class PerformanceView extends AbstractHistoricView
         item.setImage(Images.VIEW_TABLE.image());
 
         hookContextMenu(calculation.getTree(), this::fillContextMenu);
+        hookKeyListener();
     }
 
     private void fillContextMenu(IMenuManager manager) // NOSONAR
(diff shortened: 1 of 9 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/SecurityContextMenu.java
```diff
@@ -27,12 +29,19 @@
 public class SecurityContextMenu
 {
     private AbstractFinanceView owner;
+    private boolean fullContextMenu = false;
```
buchen, line 32: I renamed this to "showAccelerator" because this is what it does.

To be honest, I got confused by the fullContextMenu in the TransactionContextMenu. I am not sure how where it used, but I did not have time to research and possibly remove the flag. So I wanted to avoid another one here.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PerformanceView.java
```diff
@@ -432,6 +435,20 @@ private void fillContextMenu(IMenuManager manager) // NOSONAR
...
+        calculation.getControl().addKeyListener(new KeyAdapter()
+        {
+            @Override
+            public void keyPressed(KeyEvent e)
+            {
+                Object selection = ((IStructuredSelection) calculation.getSelection()).getFirstElement();
+                if (selection instanceof ClientPerformanceSnapshot.Position position)
+                    new SecurityContextMenu(PerformanceView.this).handleEditKey(e, position.getSecurity());
+            }
+        });
```
buchen, line 449: Switch to using the keyPressedAdaptor to save a couple lines...

```
calculation.getControl().addKeyListener(KeyListener.keyPressedAdapter(e -> {
        }));
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/AbstractNodeTreeViewer.java
```diff
@@ -909,6 +912,24 @@ protected void fillContextMenu(IMenuManager manager) // NOSONAR
...
+        {
+            @Override
+            public void keyPressed(KeyEvent e)
+            {
+                IStructuredSelection selection = nodeViewer.getStructuredSelection();
+                if (selection.isEmpty() || selection.size() > 1)
+                    return;
+
+                TaxonomyNode node = (TaxonomyNode) selection.getFirstElement();
+                if (!node.isClassification())
```
buchen, line 927: It is better to check if ```node.getBackingSecurity()``` is not null. An assignment can also be a cash account

buchen (approved):
merged.

Issue #2371, opened 2021-08-06
Filter on Security Performance Does Not Filter Transactions


flywire:
**Describe the bug**
Performance Calculation report with filter does not filter the transactions.

**To Reproduce**
1. Display Performance Calculation report with filter on Entire Portfolio
2. Select a security held in multiple accounts then display transactions
3. Set filter to only display one of the accounts
4. See error - transactions not filtered by account

**Expected behavior**
Filter should apply to transactions too.

**Screenshots**
![image](https://user-images.githubusercontent.com/11288701/128502415-01dfe182-187f-49d2-9ca7-ec96ccaca489.png)
Filter Entire Portfolio

![image](https://user-images.githubusercontent.com/11288701/128502481-bcd13979-abfb-4ca8-b9fb-10bd109475dc.png)
No change when filtered for Broker Securities/Broker Funds/Bank

Version: 0.54.2 (Jul. 2021)
Platform: win32, x86_64
Java: 11.0.4+11-LTS, Azul Systems, Inc.

chirlu:
The bottom area just gives general information about the security.

flywire:
No, the filter carries through to the bottom section of other reports but not this one.

flywire:
No, the filter carries through to the bottom section of other reports but not this one.

eg Filter works on Income & Expenses, Fees Subreport, Transactions Subreport.

Following reports support filter:
* Statement of Assets: Trades
* Performance Calculation: Trades
* Securities: Trades
* Income & Expense: Seems to filter every top-line subreport

Interestingly, Trades report has no Account filter.

flywire:
I don't know when, but fixed by Version: 0.81.0 (December 2025).

PR #5025, opened 2025-09-28
allow information pane for Transaction to be filtered by accounts

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/2371

Hello, this is a proposition to filter the transactions in the bottom Transactions pane of a security based on the accounts selected. This is the current behavior for the Chart pane and the Trades panes, so I guess we can apply it also for Transaction pane ? And this is also the current behavior of the "uneditable transactions pane" of Performance Securities view.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/panes/TransactionsPane.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/panes/TransactionsPane.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.List;
 
 import jakarta.inject.Inject;
+import jakarta.inject.Named;
 
 import org.eclipse.jface.action.Separator;
 import org.eclipse.jface.action.ToolBarManager;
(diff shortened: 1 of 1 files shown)


buchen:
Hi @mierin12,

thanks for the proposal.

I realize PP is already inconsistent. Take the "PaymentsView". If filtered, the transaction pane uses account#getTransactions to get the filtered transactions. But it uses Security#getTransactions(client) to get unfiltered transactions for the client. So already today there is a mismatch.

The "problem" I have: The user can now edit the "derived" transaction. Say we have a transfer from account A to account B. If the filter only contains account A, then the transfer is converted into a withdrawal (because account B is not part of the set). The withdrawal can now be edited.

The second thing is: some transactions are just copied over to the filtered client, some transactions are edited (the delivery to withdrawal, but also when creating a client before taxes and fees). So at the moment it is really hard to say what are "original" transactions that can be edited and what not.

That is also the reason why - for now - I kept the "uneditable transactions pane" of Performance Securities.

If I see this right, this change would impact following views (the others have no client filter):
* PaymentsView
* TaxonomyView
* HoldingsPieChartView
* StatementOfAssetsView
* PerformanceView

Two questions:

* Mid-term my thinking is to have some transient filed "original transaction" in the Transaction class. This could hold a reference to the original transaction. If given, then the pane would either not allow editing or open the original transaction. And we could add a little icon to indicate to the user that this is a derived transaction. What do you think? Other ideas?

* I am still torn if merging this change make it better (more consistent) or worse (allow editing the virtual transactions and having no indicator). Any thoughts on the trade off?

buchen:
I close this one. It is included in #5076, but the latter one also fixes the editing dialogs and menus.

Issue #5048, opened 2025-10-04
Removing cache for MEXC


earthjasonlin:
It sucks when fetching latest price and changing the historical price feed from others to this (like coingecko to mexc).
There seems to be no need to add cache anyway (Binance doesn't have is wither)

Morpheus1w3:
@earthjasonlin 

What is your understanding of caching in this circumstances?

earthjasonlin:
> @earthjasonlin 
> 
> What is your understanding of caching in this circumstances? 

@Morpheus1w3 Maybe i misunderstood something but I'm specifically talking about this commit https://github.com/portfolio-performance/portfolio/commit/725dc71d0fbad6d44659a46d85383fb4157b2dce
I don't know why this cache makes it act wierdly but i can't find other clues

earthjasonlin:
It just can't fetch latest price, volumes etc

buchen:
@earthjasonlin 

Can you explain what the issue is with the caching?

The "problem" I was trying to fix is this: if the user has the price feed configured for historical **and** latest prices, then the almost identical call is done twice right after each other. Because my understanding is that there is no different endpoint for historical and latest prices. With a cache of 1 minute, I was trying to avoid this extra call and avoiding the rate limits that the servers usually apply.

PR #5029, opened 2025-10-01
Add MEXC as online quote feed

earthjasonlin:
## Why add MEXC?
Some CEX coins like BGB, OKB is not listed on Binance, Bitfinex, etc. MEXC has them all, and it has quite a large coverage of coins

PR #5054, opened 2025-10-05
Rework MEXC to fix wierd performances

earthjasonlin:
Issue: #5048
PR: #5029
Commit: https://github.com/portfolio-performance/portfolio/commit/fdcf3f7e250b2ce41a34dc086804c4296a19696a https://github.com/portfolio-performance/portfolio/commit/725dc71d0fbad6d44659a46d85383fb4157b2dce

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MEXCQuoteFeed.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/MEXCQuoteFeed.java
@@ -4,7 +4,6 @@ import java.io.IOException;
 import java.net.URISyntaxException;
 import java.text.MessageFormat;
 import java.text.ParseException;
-import java.time.Duration;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


earthjasonlin:
Also, imo there's no need to add cache, as the code is directly edited from BinanceQuoteFeed.java and should work perfectly.
Sorry for my mistakes before

buchen:
@earthjasonlin See comment in #5048

buchen:
Hi @earthjasonlin,

first, I picked up the rework to the MEXC feed: 497706620139df4401844af2da57df310e2dbf3d **But** without the removal of the caching. As argued in #5048, I want to avoid duplicate calls right after each other. The caching for 1 minute avoids the "latest download" does not execute the same query that was done before. Keep in mind that the dialog to edit the feed provider (HistoricalQuoteProviderPage) also does some caching (the idea being that if the user flips back between providers, we are not requesting the remote data again). Maybe that resulted in the weird behavior you have seen?

If you disagree, feel free to open another pull request. I would also need a better understanding what the "weird" is exactly.

Second, I merged the new USDs-M feed. Thank you. I did some refactoring to pull the common pieces here: 594e7ff79cef77f8c74d15ec049a335dfd818c77 If you have the time, I would appreciate if you could test.

The Finance feed is also missing the caching... I will add that when I find the time.

PR #5063, opened 2025-10-08
Migrate ECB quote feed from SDW to EDP

wahlbrink:
The SDW API has been moved to EDP (ECB Data Portal).

Some translations still need to be updated.


buchen:
merged. Thanks.

I changed the translation for the other items also to "ECB Data Portal". It is kind of a brand name. I could not find translations on the web site. Local translators could adapt - if needed.

buchen:
I wonder if it is possible to reduce the data loaded. It seems always to be the full data series (say starting 2000) and takes quite some time. Do you happen to know?

wahlbrink:
> I wonder if it is possible to reduce the data loaded. 

The parameters are documented here https://data.ecb.europa.eu/help/api/data and https://github.com/sdmx-twg/sdmx-rest/blob/master/doc/data.md . Probably `updatedAfter` is suitable.
E.g. `https://data-api.ecb.europa.eu/service/data/ECB%2CEST%2C1.0/B.EU000A2QQF08.CI?updatedAfter=2025-01-01T00%3A00%3A00%2B00%3A00` works for me (loads the €STR data for 2025 only).

PR #5079, opened 2025-10-13
Implementation of support for Tel Aviv Stock Exchange (TASE) support 

arnonm:


Nirus2000:
What? Are you sure? Merge and restart this pull request.

Screenshot_20251013_215802_Firefox

arnonm:
Will do. The 46 files are ~ correct (There are a 2-3 false changes in there), but more or less correct.

PR #5080, opened 2025-10-13
Display a new unit risk line in the returns/volatility chart view

Lotendan:
The line can be enabled/disabled via the menu, and offers a visual way to assess the amount of risk taken for an investment, compared to a unit risk.

Here is the result (note: pictures are displayed in French, despite looking it up on the Internet I didn't find a quick way to override the default language setting from the VM).
pp_menu

pp_line

Hope this will be useful!


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/ReturnsVolatilityChartView.java
```diff
@@ -301,6 +320,45 @@ private void setChartSeries()
...
+                Color gray = Display.getDefault().getSystemColor(SWT.COLOR_GRAY);
+                e.gc.setForeground(gray);
+                e.gc.setLineStyle(SWT.LINE_DOT);
+
+                IAxis xAxis = chart.getAxisSet().getXAxis(0);
+                IAxis yAxis = chart.getAxisSet().getYAxis(0);
+                int x0 = xAxis.getPixelCoordinate(0);
+                int y0 = yAxis.getPixelCoordinate(0);
+                
+                int x1 = xAxis.getPixelCoordinate(9999);
```
Lotendan, line 345: Using 9999 here, I tried to use `Double.MAX_VALUE` but this didn't work for some reason. Any hint on what to use instead, I can adjust.

buchen:
> Here is the result (note: pictures are displayed in French, despite looking it up on the Internet I didn't find a quick way to override the default language setting from the VM).

In Eclipse, you can run the application with `-nl en_US` to run it with a different locale.

Bildschirmfoto 2025-10-19 um 08 38 56

buchen:
About the feature: I like it.

> Using 9999 here [...] Any hint on what to use instead, I can adjust.

I also have no better idea. Maybe it is possible to get the "slope" of the curve from both axis and then calculate the pixels. But it looks good enough for. We'll add a comment to remember our future self why this magic number is used...

Two questions:

First: Shouldn't the line extend into the negative area as well?

Second: I would have thought that the unit risk line represents the "risk free asset". Here it is a constant definition. I am just asking myself, if we should allow the configuration of the "risk free return" to determine the slope of the line. And for the share ratio widget, the user can already configure a risk free return that is stored with the file (not per widget, but globally).

Bildschirmfoto 2025-10-19 um 09 09 35

Lotendan:
> We'll add a comment to remember our future self why this magic number is used...

Ok, I can do that.

> First: Shouldn't the line extend into the negative area as well?

Since the volatility/variance/semivariance aren't supposed to be negative, we shouldn't add a line on the left part of the graph (x < 0). A future enhancement (in a separate PR) would consist in actually hiding this part entirely to the user, because in practice it's never used.

> Second: I would have thought that the unit risk line represents the "risk free asset".

I've been thinking about that yes, and I wasn't sure either. The "risk free return" normally is more a notion of the Sharpe ratio. And since here we can choose different metrics for X (variance/semivariance) and Y (TRI/TTWROR) then it is not obvious to consider that this graph is always linked with the Sharpe ratio.
That being said, the graph clearly expresses a notion of risk (and this is what my "unit risk line" aims to visualize) so that wouldn't hurt me anyway.
Let me know your opinion and I can adapt the code.

buchen:
took too long, but finally merged. Will be part of the next release

PR #5081, opened 2025-10-15
New PR - Added support for Tel Aviv Stock Exchange Stocks Bonds and Funds

arnonm:
Replaces https://github.com/portfolio-performance/portfolio/pull/5079 with a single Commit


Nirus2000:
Hello @arnonm 
Thank you for the pull request. 
Unfortunately, I can't wrap my head around committing 350k of source code for a price provider. Especially since a lot of the source code is uncommented, and comments are incorrect or wrong.
Some to-dos, etc. 
I don't think anyone will take the time for this pull request. We have also implemented other course providers, such as Leeway, Yahoo, etc., and they are more resource-efficient.
Also... the tests fail...

arnonm:
So addressing the comments. Removed the live API tests and the sample files, which reduced the commit to 11,000 lines. 5000 of which are in 2 mock response files. I can cut these as well.
Will do a review of the comments and ensure to clean up the comments.

buchen:
Hi @arnonm,

thanks for the contribution.

First, I am trying to understand which **API endpoints** this contribution is using and which **terms and conditions** apply.

I found these general [terms of service](https://www.tase.co.il/en/content/about/terms_of_service) linked on the footer of the web page. It makes no mention of an API but also does not prohibit the use of the data for personal purposes. I see there is also a "TASE Data Hub" with API documentation and an API key for authentication. There are other data providers where one can configure the API key in the settings. I wasn't able to figure out any details now, but it sounds like a reasonable option?

Second, on a first glance I noticed a couple things with the pull request. One issue is that the complete listing is downloaded upon the start of PP. Why? The constructor of the Feed is loading the data but feeds are constructed at the start of PP. That means tens of thousands of downloads per day. My proposal is that we unpack this contribution: first have the download of historical prices if the user has configured TAVE as a source. In a second step, we extend the search in a reasonable way (either search TAVE only if user has provided an API key, or some other way). With the search, I first have to better understand what kind of traffic would go to the side.

arnonm:
Thanks Andreas,

These terms of service are the ones.
It looks like they are in the process of moving to a new version of the API
with an API key. I was holding off on this until I figured out the API key
strategy. I'll look at it as a version 2.
Regarding the comments:
1. The current API uses two endpoints depending on the security type.
However, the type is not returned by the endpoint API. So the solution I
found was to download and cache the index, then query the index to get the
security type and use the correct endpoint
2. However, you are correct and I missed the timing of the constructor. I
will add some conditions so that the caching is done if the source is
configured, or on the first manual search
I'll travelling for a few days, so will get to it when I am back.

Arnon

arnonm:
Added a check so that getAllIndices is not kicked off on Construction, only when the first search is called, and added a test to ensure it happens

arnonm:
@buchen  - raising this to the top again..

arnonm:
@buchen  - anything blocking this?

Nirus2000:
We are closing this pull request because there are unresolved legal and technical uncertainties.
In particular, the API usage terms are unclear and the integration has not been sufficiently tested.

arnonm:
That is a shame. I have been using it for months on a daily basis without
any technical issues. There are clear integration tests in the test package
- so that comment is not clear to me
Regarding the legal issue, can you clarify? The legal pages are
referenced and are open for public use

Arnon

Nirus2000:
Conclusion

We are closing this pull request because it introduces a clear legal risk related to data rights and redistribution. The Tel Aviv Stock Exchange does not grant permission for public redistribution of its market data via third-party software without a proper license. Access to TASE market data, including through their Data Hub API, is controlled and requires explicit authorization or a license to distribute or process that data, and is not generally permitted for public dissemination in an open-source project without meeting those licensing requirements. 

In other words, there is no free permission given in the TASE terms of service to redistribute their data publicly through software, and doing so without appropriate rights creates a legal liability for the project. That uncertainty around data usage rights and redistribution is a clear legal problem, so we cannot merge this change.

arnonm:
Got it.

Issue #4978, opened 2025-09-12
IB Flex: can't import dividend with currency different than security


lmb:
**Describe the bug**

- I have an account with interactive brokers, in GBP as base currency.
- That account contains a security (VWRL) which I bought in EUR on a German exchange a long time ago. This was transferred to IB. The security is denominated in EUR both in the IB interface as well as in PP.
- VWRL is denominated in USD. Dividends paid by the fund are therefore in USD.
- IB does not automatically convert the USD dividend into either EUR or GBP.
- Importing such a USD dividend for VWRL gives an error message "Foreign currency of gross value USD does not match currency of security EUR: 30 Dec 2024 Dividend GBP XXX.YY Vanguard FTSE All-World".
- Manually changing the transaction Cash Account to a USD deposit account changes the error message to "Transaction currency GBP does not match account currency USD: 30 Dec 2024 Dividend GBP XXX.YY Vanguard FTSE All-World"

**To Reproduce**
1. File -> Import -> IB Active Flex Query
2. Import .xml containing a transaction as describe above

**Expected behavior**

In the best case the importer would automatically assign the transaction with a deposit account in USD (probably the one selectable at the top of the dialog?). The UI would show that transaction in USD, while performance calculations would use the base currency of the security.

The second best case would be to be able to manually override the Cash Account as described.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 0.79.1

lmb:
I'd love to help get this resolved somehow. In order of my preference:

1. I'd be happy to pay a regular contributor of PP to work on this.
2. I can provide .xml for a volunteer to take a stab at this.
3. I can fix this myself provided I get some guidance from a maintainer on how to approach this best.

cc @Nirus2000 

Thanks for making PP such a great tool!

Nirus2000:
Hello @lmb
We need a complete(!) XML file with the transactions in order to reproduce and fix the problem.

Regards
Alex

lmb:
Here is a redacted version of the export: [PP.xml](https://github.com/user-attachments/files/22335881/PP.xml) I've changed all transactions to be for 10 shares and removed PII. Importing this file into PP still gives me the error as described above.

arnonm:
Imb, I was curious about your bug. I tried to reproduce it on the latest PP and did not get the error  in the GUI
I created three accounts, GBP, USD and EUR. I imported the xml and it was imported to the GBP account 
This line:
[...]

PR #5094, opened 2025-10-20
Remove most currency conversions from IBFlex importer

lmb:
IBFlex: Refactor extractDate to work on all transaction types

    Use a single method to extract a LocalDateTime from a transaction element.
    This improves the result of some imports by adding LocalTime when there
    previously wasn't one.

IBFlex: Improve handling of currency conversions

    The existing exchange rate calculations are spread all over the place and
    are not entirely consistent. For example, fxRateToBase can only safely be
    used if the accountCurrency is known.

    Improve this by using fxRateToBase correctly and by parsing ConversionRate
    elements contained in the IB XML. This allows doing cross currency
    conversions even if the accountCurrency is not known.

IBFlex: Remove most currency conversions

    The IBFlex importer currently has a set of historically grown heuristics 
    around currency conversions:

    1. DIVIDEND and TAX transactions are converted to account base currency
      in buildAccountTransaction() 2. BUY and SELL transactions are converted to
    account base currency
      in buildPortfolioTransaction()

    As far as I can tell, Interactive Brokers doesn't actually work the way that
    these heuristics assume.

    - There are no automatic currency conversions, the platform creates
     forex balances on the fly.
    - Performing a trade in a foreign currency requires a balance in
     that currency.
    - Dividends are distributed in the "base" currency of a security.
    - fxRateToBase attributes give the exchange rate to the base currency
     of the account.

    It is possible to get into a situation where the internal currency of a
    security doesn't match the listed currency. For example by buying a security
    denominated in USD on a German exchange and then transferring this to
    Interactive Brokers. The current heuristics make it impossible to import
    transactions in this case.

    Fix this by removing most currency conversions outright. Users will have to
    maintain additional deposit accounts in the correct currency and then
    manually track forex conversions.

    We still create GROSS_VALUE units with forex amounts for securities where
    the internal currency doesn't match the listed currency.

Closes https://github.com/portfolio-performance/portfolio/issues/4978

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
@@ -329,54 +329,6 @@ public class IBFlexStatementExtractor implements Extractor
                     throw new IllegalArgumentException("unsupported type '" + type + "'");
             }
 
-            // @formatter:off
-            // If the account currency differs from transaction currency
-            // convert currency, if there is a matching security with the account currency
-            // @formatter:on
-            if ((AccountTransaction.Type.DIVIDENDS.equals(accountTransaction.getType())
-                            || AccountTransaction.Type.TAXES.equals(accountTransaction.getType()))
-                            && (this.accountCurrency != null && !this.accountCurrency.equals(amount.getCurrencyCode())))
-            {
-                // matching isin or wkn & base currency
-                boolean foundIsinBase = false;
-
-                // matching isin & transaction currency
-                boolean foundIsinTransaction = false;
-
-                // matching conid (wkn) & transaction currency
-                boolean foundWknTransaction = false;
-
-                for (Security s : allSecurities)
-                {
-                    // Find security with same isin or conid (wkn) and currency
-                    if (element.getAttribute("isin").length() > 0 && element.getAttribute("isin").equals(s.getIsin()))
-                    {
-                        if (amount.getCurrencyCode().equals(s.getCurrencyCode()))
-                            foundIsinTransaction = true;
-                        else if (this.accountCurrency.equals(s.getCurrencyCode()))
-                            foundIsinBase = true;
-                    }
-                    else if (element.getAttribute("conid").length() > 0 && element.getAttribute("conid").equals(s.getWkn()))
-                    {
-                        if (amount.getCurrencyCode().equals(s.getCurrencyCode()))
-                            foundWknTransaction = true;
-                        else if (this.accountCurrency.equals(s.getCurrencyCode()))
-                            foundIsinBase = true;
-                    }
-                }
-
-                // If the transaction currency is not found but the base
-                // currency is found, and there is an exchange rate, convert the
-                // amount.
-                if ((!foundIsinTransaction || !foundWknTransaction) && foundIsinBase && element.getAttribute("fxRateToBase").length() > 0)
-                {
-                    BigDecimal fxRateToBase = asExchangeRate(element.getAttribute("fxRateToBase"));
-
-                    amount = Money.of(accountCurrency, BigDecimal.valueOf(amount.getAmount()).multiply(fxRateToBase)
-                                    .setScale(0, RoundingMode.HALF_UP).longValue());
-                }
-            }
-
             // Set the amount in the transaction
             setAmount(element, accountTransaction, amount);
 
@@ -572,14 +524,14 @@ public class IBFlexStatementExtractor implements Extractor
                 Money amount = Money.of(asCurrencyCode(element.getAttribute("currency")), asAmount(element.getAttribute("netCash")));
 
                 setAmount(element, portfolioTransaction.getPortfolioTransaction(), amount);
-                setAmount(element, portfolioTransaction.getAccountTransaction(), amount, false);
+                setAmount(element, portfolioTransaction.getAccountTransaction(), amount);
             }
             else
             {
                 Money amount = Money.of(asCurrencyCode(element.getAttribute("currency")), asAmount(element.getAttribute("cost")));
 
                 setAmount(element, portfolioTransaction.getPortfolioTransaction(), amount);
-                setAmount(element, portfolioTransaction.getAccountTransaction(), amount, false);
+                setAmount(element, portfolioTransaction.getAccountTransaction(), amount);
             }
 
             // Set share quantity
@@ -589,12 +541,12 @@ public class IBFlexStatementExtractor implements Extractor
 
             // Set fees
             Money fees = Money.of(asCurrencyCode(element.getAttribute("ibCommissionCurrency")), asAmount(element.getAttribute("ibCommission")));
-            Unit feeUnit = createUnit(element, Unit.Type.FEE, fees);
+            Unit feeUnit = new Unit(Unit.Type.FEE, fees);
             portfolioTransaction.getPortfolioTransaction().addUnit(feeUnit);
 
             // Set taxes
             Money taxes = Money.of(asCurrencyCode(element.getAttribute("currency")), asAmount(element.getAttribute("taxes")));
-            Unit taxUnit = createUnit(element, Unit.Type.TAX, taxes);
+            Unit taxUnit = new Unit(Unit.Type.TAX, taxes);
             portfolioTransaction.getPortfolioTransaction().addUnit(taxUnit);
 
             portfolioTransaction.setSecurity(this.getOrCreateSecurity(element, true));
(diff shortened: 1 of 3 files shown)


pfalcon:
> Dividends are distributed in the listed currency of a security.

This is not correct assumption, taken literally. A security hosted on a particular exchange (and "listed" in currency of that exchange) can still pay dividends in another currency.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -382,7 +356,7 @@ else if (this.accountCurrency.equals(s.getCurrencyCode()))
                 return;
 
             cashTransaction.setDate(extractDate(element));
-
```
pfalcon, line 385: Unrelated change.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -403,8 +377,8 @@ else if (this.accountCurrency.equals(s.getCurrencyCode()))
             }
 
             // Ensure all required attributes are present
-            if (!element.hasAttribute("symbol") ||
```
pfalcon, line 406: Unrelated change.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -789,36 +697,34 @@ private void setAmount(Element element, Transaction transaction, Money amount, b
          * <p/>
          * If "tradeTime" is not present, then check if "tradeDate" and
          * "dateTime" are the same date, then set "dateTime" as the trading day.
-         * 
+         *
          * @return the extracted date or null if not available
          */
         private LocalDateTime extractDate(Element element)
         {
-            if (!element.hasAttribute("tradeDate"))
```
pfalcon, line 797: Another example of unrelated changes, and less benign than the whitespace ones: the commit description talks about currency conversion, but actually changes way transaction timestamp is parsed behind the scene. Again, please assume that leads to regressions, and someone who reads the description wouldn't even know this patch introduced it.

(Otherwise I certainly agree that getting transaction timestamp in IBFlexStatementExtractor leaves much to be desired, but these changes flip-flop one tangled way for another - that's clear, because the correct way to get transaction timestamp in modern IBKR XML is from "dateTime" attribute alone. And getting proper, second-resolution timestamp is imperative to properly group transactions into trades.)

lmb:
Thanks for taking a first pass! I squashed all of my changes into a single commit because I wasn’t sure what the policy re multiple commits in a single PR was. Happy to split things out again.



> > Dividends are distributed in the listed currency of a security.
> 
> This is not correct assumption, taken literally. A security hosted on a particular exchange (and "listed" in currency of that exchange) can still pay dividends in another currency.

I think we are trying to say the same thing, but I lack the correct vocabulary. I mean the currency that a fund is denominated in I guess? I also don’t think this has an effect on the code itself.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
lmb: Should I reverse the order then? (Broken out into a separate commit) start with date time and fall back to trade date etc.

lmb: I tried to improve this by centralising it all in a single function.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -710,20 +731,18 @@ private Unit createUnit(Element element, Unit.Type unitType, Money amount)
             }
             else
             {
-                // Calculate the FX rate to the base currency
-                BigDecimal fxRateToBase = element.getAttribute("fxRateToBase").isEmpty() ? BigDecimal.ONE
-                                : asExchangeRate(element.getAttribute("fxRateToBase"));
+                BigDecimal exchangeRate = getExchangeRate(element, amount.getCurrencyCode(), accountCurrency);
```
buchen: I believe getExchangeRate can return null.

As I understand it, getExchangeRate is also checking for fxRateToBase *and* it now attempts to calculate the cross rate using the account currency, to probably it is not a change. But maybe a check for null could make sense

buchen:
Hi @lmb,

thanks for the contribution.

About cbf4f78ee3649c733e6ef1543786fff6db0cbf13: The changes to the date parsing make sense.
About 20f5503f044ecdd8a0d67577fab322d215f7e51b:  Improving the currency conversion looks good to me. We could add two checks for null.

As I plan to a make a new release, I might cherry pick those two commits to be included in the release.
Let me comment on 4d4b293114d16c2646d07fd98d91f29f1f44de15 later - need more time.

buchen:
About removing currency conversions:

I believe there is value in cleaning up the code. In the past, it was only possible to import to one deposit account at a time (and a deposit account has exactly one currency). That is now much more flexible - one can pick a deposit account per transaction currency that is imported. So I guess some of the conversions can be removed.

However, PP makes the assumption that a dividend transactions contains a currency conversion between the transaction currency and the instrument currency. That can be different. Once can configure an instrument with EUR as currency (traded on a European exchange) but still receive dividends in a foreign currency on the deposit account. Then PP needs the exchange rate to properly import the entry.

I believe that was the reason for the code in the #buildAccountTransaction method to check for the instrument.

buchen:
@lmb 

Ok, thinking about it some more I see that the currency conversion to the instrument currency is included.

The change imports transactions always in the currency as denoted in the IBFlex XML file. It is an incompatible change. Users trading in multiple currencies have to create deposit accounts in the additional currencies.

In general, the change looks good to me. The change make sense - it is closer to what a user would see in IBFlex. 

You write:
> and then manually track forex conversions.

Can the forex conversions also be part of the IBFlex XML file? Then they would also be imported.

lmb:
> Can the forex conversions also be part of the IBFlex XML file? Then they would also be imported.

It does look like it:

```
<FxTransaction accountId="A" acctAlias="A" model="" assetCategory="CASH" reportDate="20241219" functionalCurrency="GBP" fxCurrency="EUR" activityDescription="CASH: XYZ.22 EUR.USD" dateTime="20241219;073909" quantity="XYZ.22" proceeds="XYZ.460865" cost="-XYZ.460865" realizedPL="0" code="O" levelOfDetail="TRANSACTION"/>
<FxTransaction accountId="A" acctAlias="A" model="" assetCategory="CASH" reportDate="20241219" functionalCurrency="GBP" fxCurrency="EUR" activityDescription="CASH: -1.8 EUR.USD" dateTime="20241219;073909" quantity="-1.8" proceeds="1.497573" cost="-1.497578" realizedPL="-0.000005" code="C" levelOfDetail="TRANSACTION"/>
```

As usual it's hard to know what the codes mean.

> As I plan to a make a new release, I might cherry pick those two commits to be included in the release.

Please let me know how you want to proceed here. Should I make changes to this branch, or wait for you to cherry-pick and then rebase?

auchri:
@lmb 

> As usual it's hard to know what the codes mean.

Do you mean `code="O"` and `code="C"`?

According to [this repository](https://github.com/csingley/ibflex/blob/df672d2b9048c524119149940c642ecee3593fa5/ibflex/enums.py#L50C7-L50C11) it's opening and closing trade

buchen:
@lmb I have merged the pull request.

I applied two minor changes:

First, I applied the standard formatting where reasonable. Yes, I know the formatting is messy (I missed to apply a standard formatter). Also this file has mixed formatting as well. I want to avoid diffs only because of formatting, but at least the new code complies.

Second, I changed the test case to use the custom matchers. That makes it much more readable.

I merged this change _after_ yesterday's release. That gives us a chance to also support the import of the currency conversions ```FxTransaction```. **Do you want to give it a try?**

I close this PR. Let's open a new one for the FxTransaction.

lmb:
> That gives us a chance to also support the import of the currency conversions `FxTransaction`. **Do you want to give it a try?**

Do you have some example in the code base which does something analogous? Not very familiar with what open and closed means.

Happy to take a stab at this if you can give me some pointers.

lmb:
> I merged this change _after_ yesterday's release. That gives us a chance to also support the import of the currency conversions `FxTransaction`. **Do you want to give it a try?**

As far as I can tell these transactions are already imported, just via a different tag in the XML. The flexquery duplicates some information after all. So nothing to do here I think.

PR #5103, opened 2025-10-26
Fix .getLatestGroupingCriterion for HTMLTableQuoteFeed and GenericJSONQuoteFeed

wahlbrink:
Shouldn't they always use the URL as grouping criterion?


name.abuchen.portfolio/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeed.java
```diff
@@ -91,7 +91,7 @@ public String getGroupingCriterion(Security security)
     public String getLatestGroupingCriterion(Security security)
     {
         String latestFeed = security.getLatestFeed();
-        return getCriterionFrom(latestFeed != null && !latestFeed.isEmpty() ? latestFeed : security.getFeedURL());
+        return getCriterionFrom(latestFeed != null && !latestFeed.isEmpty() ? security.getLatestFeedURL() : security.getFeedURL());
```
buchen, line 94: good catch!

Yes, the grouping criterion should be the host name. The idea is to limit parallel requests to the same host (otherwise the host is limiting).

PR #5106, opened 2025-10-27
Feature: add taxonomy support to Trades view

georgemac-labs:
Hi maintainers,

Firstly, thanks for Portfolio Performance, which I find to be an impressive and useful tool.

In particular, the taxonomy feature is essential for me, but I need to see the assignments and aggregated metrics for completed trades. I could only find a way to do this for securities currently held.

**Thus I am proposing a PR to group and summarise entries in the Trades view in the same way.**

I am keen to have this functionality in the master branch, have endeavoured to adhere to the contribution guidelines and am happy to take onboard any input regarding how it should be implemented.

I look forward to your feedback.

Screenshot 2025-10-28 at 00 34 53

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/trades/TradeCategoryTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/trades/TradeCategoryTest.java
@@ -1,0 +1,373 @@
+package name.abuchen.portfolio.snapshot.trades;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.closeTo;
+
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import org.junit.Test;
+
+import name.abuchen.portfolio.junit.AccountBuilder;
+import name.abuchen.portfolio.junit.PortfolioBuilder;
+import name.abuchen.portfolio.junit.SecurityBuilder;
+import name.abuchen.portfolio.junit.TaxonomyBuilder;
+import name.abuchen.portfolio.junit.TestCurrencyConverter;
+import name.abuchen.portfolio.math.IRR;
+import name.abuchen.portfolio.model.Account;
+import name.abuchen.portfolio.model.Classification;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.model.Taxonomy;
+import name.abuchen.portfolio.model.TransactionPair;
+import name.abuchen.portfolio.money.CurrencyUnit;
+import name.abuchen.portfolio.money.Money;
+import name.abuchen.portfolio.money.Values;
+
+@SuppressWarnings("nls")
+public class TradeCategoryTest
+{
+    @Test
+    public void testAggregation() throws Exception
+    {
+        Client client = new Client();
+
+        Security security = new SecurityBuilder() //
+                        .addPrice("2020-01-01", Values.Quote.factorize(100)) //
+                        .addPrice("2020-02-01", Values.Quote.factorize(110)) //
+                        .addTo(client);
+
+        Account account = new AccountBuilder() //
+                        .deposit_("2020-01-01", Values.Amount.factorize(20000)) //
+                        .addTo(client);
+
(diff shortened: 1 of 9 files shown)


buchen:
Thanks for the contribution.

I believe allowing to group the trades by a taxonomy is a valid new feature.

Briefly looking at it:

@georgemac-labs  Is it fair to say the changes in the Trade class are separate change from the actual grouping by taxonomy? Fixes what is being tested with testShortMovingAverageCostDoesNotDivideByZero?

About the grouping change - let me make up my mind
* I wonder if using the TradeElement always could simplify the TradeTableViewer - less instanceof
* Adding a column "weight" to display the weight used could make sense. If I have split an instrument across categories, then it is nowhere to see in this view (or did I overlook something?)
* If the file does not contain a taxonomy, then trades are not grouped and we have the "original" view. I am thinking to allow this also as an option even if the user has taxonomies created. And possibly, one could show the totals nevertheless.

georgemac-labs:
Hi @buchen, thanks for taking the time to review the changes – I know it's not a small one.

> Is it fair to say the changes in the Trade class are separate change from the actual grouping by taxonomy? Fixes what is being tested with testShortMovingAverageCostDoesNotDivideByZero?

Yes – those are just calculation improvements. They are relevant to the taxonomy functionality (e.g. IRR values for categories), but not truly part of it. Would you prefer them in a separate commit or PR?

> I wonder if using the TradeElement always could simplify the TradeTableViewer - less instanceof

Yes, no question. Non-trivial change, but would be cleaner. I will attempt to do that without introducing any bugs!

> Adding a column "weight" to display the weight used could make sense. If I have split an instrument across categories, then it is nowhere to see in this view (or did I overlook something?)

I based the UI on Accounts → Securities Accounts, as that seemed the closest analogue. I didn't include Weight, as I don't see it as a column in that view, and personally I don’t use weights. But I’m happy to add it.

> If the file does not contain a taxonomy, then trades are not grouped and we have the "original" view. I am thinking to allow this also as an option even if the user has taxonomies created. And possibly, one could show the totals nevertheless.

Yes, I considered that, too, but I opted for consistency with what I saw in Accounts → Securities Accounts.

However, I now notice that Reports → Statement of Assets does indeed offer taxonomy selection in the cog dropdown, although I don’t see any way to disable grouping. Is that difference between those two views intentional or just historic?

Anyway, if I'm going to amend that, my idea would be to offer both manual selection of taxonomy – as in Statement of Assets – and the ability to disable taxonomy grouping altogether. What do you think?

buchen:
> Yes – those are just calculation improvements. They are relevant to the taxonomy functionality (e.g. IRR values for categories), but not truly part of it. Would you prefer them in a separate commit or PR?

Yes, please.

> I based the UI on Accounts → Securities Accounts, as that seemed the closest analogue. I didn't include Weight, as I don't see it as a column in that view, and personally I don’t use weights. But I’m happy to add it.

Thinking about it: let's leave the column "weight" for the moment. It is possible to add this with a later pull request.

> However, I now notice that Reports → Statement of Assets does indeed offer taxonomy selection in the cog dropdown, although I don’t see any way to disable grouping. Is that difference between those two views intentional or just historic?

I want to add the "no taxonomy" option also to Statement of Assets. I have not found the time yet. Adding this option to the "statement of assets" is another commit. But I do not want to take away the "not grouped by taxonomy" option here only to add it later again.

> Anyway, if I'm going to amend that, my idea would be to offer both manual selection of taxonomy – as in Statement of Assets – and the ability to disable taxonomy grouping altogether. What do you think?

Sounds good to me. The "cog drowdown" contains another option "no grouping" or "none".

georgemac-labs:
> Thinking about it: let's leave the column "weight" for the moment. It is possible to add this with a later pull request.

OK!

> Sounds good to me. The "cog drowdown" contains another option "no grouping" or "none".]

To ensure I'm not overlooking anything: dropdown contains or _should_ contain? Currently, I cannot find a none option.

image

georgemac-labs:
@buchen hope I have now made all the changes correctly:

- Split this PR into the two commits – (1) Trade calculations and (2) adding taxonomies to Trades view
- PR for taxonomy selection in Securities Accounts view https://github.com/portfolio-performance/portfolio/pull/5114
- PR for disable taxonomies option in SoA https://github.com/portfolio-performance/portfolio/pull/5115

buchen:
merged.

WIth 37ffdddf683ac2cda3ea374c44ce4d9f0073c06d I added the option to select no taxonomy also to the TradeDetailsView

georgemac-labs:
@buchen thanks, overlooked that!

PR #5115, opened 2025-11-01
Feature: add option for flat view (disable taxonomy grouping) to Statement of Assets

georgemac-labs:
@buchen here is a PR to add a "(None)" option to the taxonomy selector in SoA view, as mentioned here https://github.com/portfolio-performance/portfolio/pull/5106#issuecomment-3475940289


buchen:
merged

PR #5120, opened 2025-11-05
Template for CSV imports of Targobank bank account transactions.

hporten:
This CSV template will ease imports of the CSV exports available for the bank account.

There is a PDF extractor but afaict, that one only deals with transactions of securities. What I believe
is missing is handling of money deposits and monthly account fees. This hole can be closed with
this CVS import template.

Granted, the security transactions are not fully modeled with this importer. A 100% complete
solution would be an extractor for the monthly "Ihr Finanzstatus" documents. They can have multiple
sections and pages.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/csv-config.json
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/csv-config.json
@@ -567,4 +567,28 @@
             "label": "Gestrichen Zeit"
         }
     ]
+},
+{
+    "label": "Targobank Banktransaktionen",
+    "delimiter": ";",
+    "encoding": "ISO-8859-1",
+    "isFirstLineHeader": false,
+    "skipLines": 0,
+    "target": "account-transaction",
+    "columns": [
+        {
+            "field": "date",
+            "format": "dd.MM.yyyy",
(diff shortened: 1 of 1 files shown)


Nirus2000:
Hello @hporten ,

It is possible to import these via CSV. Every PP user can create this individually and save it as a template.
We are no longer pursuing the addition of templates in the CSV importer. There are simply too many changes, so we are focusing on PDF documents for permanent import. For example, `Gratisbroker Orders `no longer exists, but is still kept as a template to serve as a template for users.

If you provide us with PDF debug files, we would be happy to extend the Targobank importer.

We will merge... as template... 

Best regards,
Alex

hporten:
Thanks for the merge, @Nirus2000 . I had already realized that the CSV way is not en vogue anymore. And I would not mind start to learn how to implement a PDF extractor myself. Alas, in the case of Targobank there simply is no easy way: the bank transactions view supports export via CSV or printing only. And the customer mailbox does not contain the classic bank statements either. What they issue instead is a monthly "Finanzstatus" document. It contains the bank transfers and more. But no fees and likely no tax info.  So one will have to extend TargobankPDFExtractor even further to parse three different kind of documents in one go.

I made a start by uploading the extracted text here: https://github.com/portfolio-performance/portfolio/issues/5131

Issue #5069, opened 2025-10-11
[CSV] Import does not work due to rounding errors on exchange rates


wouter-in2facts:
**Describe the bug**
The gross amount is rounded to 2 decimals, creating rounding errors during the validation.
Example during a dividend transaction, see attachment.

Relevant data:
Value = 95.26
FX Rate = 1.3795
Those are the only 2 inputs.
The Gross Amount is calculated (in a separate processor preparing the csv for PP) by dividing the Value by the FX Rate, which equals to 69.0540050743023.
However the application only sees "69.05". 69.05 * 1.3795 = 95.254475.
Which is why the application says:
`Imported gross value (CAD 95.25) and calculated gross value (CAD 95.26) do not match: Jun 20, 2025 Dividend CAD 95.26 iShares MSCI Emerging Markets Min Vol Factor ETF `

Note that the Value (95.26) is correct, that is what the account is using.
If the gross amount was also at least 4 decimals, this would not be an issue:
69.0540 * 1.3795 = 95.259993 which would round to 95.26 which would be accepted

[example.csv](https://github.com/user-attachments/files/22860622/example.csv)

**To Reproduce**
1. Create a CSV-file with the following data (see attachment).
Note: the number of shares is left blank. This is due to the original source (extract from wealthsimple) not giving this information
2. Import CSV
3. An error will occur stating 
`Imported gross value (CAD 95.25) and calculated gross value (CAD 95.26) do not match: Jun 20, 2025 Dividend CAD 95.26 iShares MSCI Emerging Markets Min Vol Factor ETF `

**Expected behavior**
The transaction should be accepted if the number of decimals of the gross amount was 4 or more

**Screenshots**

![Image](https://github.com/user-attachments/assets/3ffcde3b-ee75-44ac-811f-988077e53d2a)

**Desktop (please complete the following information):**
 - OS: win32, x86_64
 - Version 0.80.3

PR #5130, opened 2025-11-08
Fix CSV import rounding errors with exchange rates (#5069)

funnym0nk3y:
This commit addresses issue #5069 where CSV imports fail validation due to rounding errors when dealing with foreign exchange rates.

Changes made:

1. BaseCSVExtractor.java:
   - Replaced Math.round(BigDecimal.doubleValue()) with BigDecimal.setScale(0, RoundingMode.HALF_UP).longValue()
   - This avoids precision loss from converting BigDecimal to double
   - Applies to both extractGrossAmount() and createGrossValueIfNecessary()

2. CheckForexGrossValueAction.java:
   - Added tolerance of ±2 (±0.02 in currency) to validation
   - This accounts for unavoidable rounding when Money values are limited to 2 decimal places
   - Prevents false positives while still catching actual errors

Root cause: Money stores amounts as longs with 2 decimal precision (factor of 100). When gross amounts are calculated with exchange rates, rounding to 2 decimals can cause small discrepancies that fail strict equality checks.

Example from issue:
- Value: 95.26 CAD, FX Rate: 1.3795
- Calculated: 95.26 / 1.3795 = 69.0540... → rounded to 69.05
- Recalculated: 69.05 × 1.3795 = 95.254... → 95.25 ≠ 95.26

The fix improves calculation precision and adds reasonable tolerance.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/BaseCSVExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/BaseCSVExtractor.java
@@ -205,7 +205,8 @@ import name.abuchen.portfolio.money.Money;
 
         Money forex = Money.of(currencyCode, Math.abs(grossAmount.longValue()));
         BigDecimal grossAmountConverted = exchangeRate.multiply(BigDecimal.valueOf(grossAmount));
-        Money converted = Money.of(amount.getCurrencyCode(), Math.round(grossAmountConverted.doubleValue()));
+        Money converted = Money.of(amount.getCurrencyCode(),
+                        grossAmountConverted.setScale(0, RoundingMode.HALF_UP).longValue());
 
         return Optional.of(new Unit(Unit.Type.GROSS_VALUE, converted, forex, exchangeRate));
     }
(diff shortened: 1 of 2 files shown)


buchen:
I'll have a look - I also need to rebase because I just happen to touch that close (deriving the exchange rate if it is not present).

And a test case helps to avoid breaking this feature in the future.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/BaseCSVExtractor.java
```diff
@@ -205,7 +205,8 @@ protected Optional<Unit> extractGrossAmount(String[] rawValues, Map<String, Colu
 
         Money forex = Money.of(currencyCode, Math.abs(grossAmount.longValue()));
         BigDecimal grossAmountConverted = exchangeRate.multiply(BigDecimal.valueOf(grossAmount));
-        Money converted = Money.of(amount.getCurrencyCode(), Math.round(grossAmountConverted.doubleValue()));
+        Money converted = Money.of(amount.getCurrencyCode(),
+                        grossAmountConverted.setScale(0, RoundingMode.HALF_UP).longValue());
```
buchen, line 209: I am not sure what difference this make w.r.t. rounding. The double value is already rounded. If it is done first on the BigDecimal, that should not make a difference mathematically.

buchen:
Thanks for the proposal. In the end, I did use a different approach than proposed by Cloude - see 39042f0af3962dcf94c7944af3a9c2a5c591ac26 - unfortunately, I miss on character in the issue number in the commit message so it does not show up here.

PR #5132, opened 2025-11-09
Refactor: Trades view to use only TradeElements

georgemac-labs:
@buchen I realised that I forgot the refactoring [you suggested](https://github.com/portfolio-performance/portfolio/pull/5106#issuecomment-3472107208).

Have attempted it here. See what you think.

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/views/TradesTableViewerTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/views/TradesTableViewerTest.java
@@ -5,12 +5,17 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.number.IsCloseTo.closeTo;
+import static org.junit.Assert.assertTrue;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


georgemac-labs:
@buchen I started to use sub-classifications in my taxonomies and noticed that my Trades implementation handles them in a weird way, and differently from other views (it flattens the tree and shows all classifications, with weird ordering). I'd like to fix that. Could you let me know if you envisage merging this PR? Then I would know which code to base the improvements on.

buchen:
Sorry, I was super busy with other stuff out side PP.

About this change. I somehow was thinking that reducing to to only use TradeElement that it simplifies the code. But the opposite happened. There are 200 more lines of code.

Bildschirmfoto 2025-12-19 um 10 53 30

I am also not sure about some things, for example the input now has a setTableInput method. As a consuming class, I now need to know to convert it there. This is error prone because easy to forget.

Therefore I tend to propose to close this pull request. It does not simplify the code.


> I started to use sub-classifications in my taxonomies and noticed that my Trades implementation handles them in a weird way, and differently from other views

Can you elaborate what you mean by this?
Is it fair to say that this is independent from this pull request a problem?

georgemac-labs:
@buchen haha yes I noticed it was more code in the end. No worries, we can leave it.

Yes, the taxonomy issue is independent from this PR. The Trades view just doesn't display taxonomies the same way as other views if you have sub-classifications. It's not really broken, just inconsistent. So I'd like to fix it but wouldn't call it a serious or urgent bug.

It's a busy time of year, but I will submit a few more PRs soon, including a fix for that. Schöne Feiertage.

PR #5140, opened 2025-11-11
Feature: add search filter to Security Performance view

georgemac-labs:
Dear maintainers, for me, it would be useful to have the search/filter functionality in the Security Performance view, too. Here's a PR for that. I have endeavoured to match the UI of other views that offer search. See what you think.


buchen:
Thanks. merged. I did a small change to re-arrange the position of the search box and make it all the same size

PR #5148, opened 2025-11-16
Support bond buy/sell transactions in PDFs from Deutsche Bank

hporten:
Due to native support of bonds and percentage-based price quotes, the typical workarounds are applied:

  - The number of shares is divied by 100
  - Accrued interests are handled as taxes


hporten:
I will further extend the bond support. Closing the pull request for the time being.

PR #5152, opened 2025-11-18
Fix small grammar error in translation

Lotendan:
"Historique" is a masculine noun.
ref: https://www.larousse.fr/dictionnaires/francais/historique/40092

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_fr.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_fr.properties
@@ -340,7 +340,7 @@ ColumnDataSeries = S\u00E9rie de donn\u00E9es
 
 ColumnDate = Date
 
-ColumnDateFirstHistoricalQuote = Premier historique (Date)
+ColumnDateFirstHistoricalQuote = Premi\u00E8re cotation historique (Date)
 
 ColumnDateFirstHistoricalQuote_MenuLabel = Date de premi\u00E8re cotation historique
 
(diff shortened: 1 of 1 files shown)


wahlbrink:
Isn't this string a short form of "Date of _last historical_ quote" (see following string in the file)? So the noun is "quote" ≙ "cotation" (f).

buchen:
@Lotendan

The column shows the date of the latest historical price.

Lotendan:
I understand we are referring to a "cotation historique". But using the feminine together with "historique" alone simply does not work. "Cotation" as a noun is a feminine. "Historique" as a noun is a masculine.

In any case there is a discrepancy because we also have this string that is used in the same view (list of instruments):
https://github.com/portfolio-performance/portfolio/blob/3a86259a45289da6f596ab33adb5d855c89883d4/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_fr.properties#L343

So either we have:

`Premier historique (Date)` and `Dernier historique (Date)`
or
`Première historique (Date)` and `Dernière historique (Date)`
But not a mix of both ;) So in any case we need to fix something.

Otherwise how about we use "cotation" instead of "historique", which looks more natural to me?
`Première cotation (Date)` and `Dernière cotation (Date)`?

buchen:
Thanks for pointing out the inconsistency. We should fix this.

Quick as background: because some APIs differentiate between "historical" prices (end of day prices) and the "latest" price (the current price which may or may have a delay while trading is ongoing), Portfolio Performance is storing prices also in two places: a list for the historical prices, and one price as the latest price.

* ColumnDateFirstHistoricalQuote
* ColumnLatestHistoricalDate

--> looks at the list of historical prices and returns the date of the first and last

* ColumnLatestDate

--> looks at both the historical prices and the lastest prices and determines what is the most recent date (it could be the case that the API with the latest ist actually not returning any data so the historical prices are actually newer).

--> that is the reason why ColumnLatestHistoricalDate and ColumnLatestDate can differ

Long story short, that is the reason why also the labels expose some of this information.

Maybe the full label makes the most sense then (the user can abbreviate if he wants to) (I do not speak French):

Première cotation historique (Date)
Dernière cotation historique (Date)

Lotendan:
Thanks for the comments. Really helpful.
I have applied the change based on your inputs.

PR #5154, opened 2025-11-19
Support import of bonds from Deutsche Bank PDFs

hporten:
These two commits add support for:

- Recognize percentage-quoted prices and accrued interest of a bond transaction
- Purchases labeled as "Zeichnung" (happened in a case of a "Festpreisgeschäft" on a newly issued bonds)

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java
@@ -43,7 +43,7 @@ public class DeutscheBankPDFExtractor extends AbstractPDFExtractor
 
     private void addBuySellTransaction()
     {
-        var type = new DocumentType("Abrechnung: (Kauf|Verkauf) von Wertpapieren");
+        var type = new DocumentType("Abrechnung: (Kauf|Verkauf|Zeichnung) von Wertpapieren");
         this.addDocumentTyp(type);
 
         var pdfTransaction = new Transaction<BuySellEntry>();
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
@@ -956,6 +956,39 @@ public class DeutscheBankPDFExtractorTest
     }
 
     @Test
+    public void testWertpapierKauf11()
+    {
+        var extractor = new DeutscheBankPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        // "Festpreisgeschäft" titled as "Zeichnung"
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Kauf11.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(1L));
+        assertThat(countBuySell(results), is(1L));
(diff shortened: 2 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java
```diff
@@ -616,7 +642,14 @@ private <T extends Transaction<?>> void addTaxesSectionsTransaction(T transactio
...
                         .section("creditableWithHoldingTax", "currency").optional() //
                         .match("^Anrechenbare ausl.ndische Quellensteuer (?<creditableWithHoldingTax>[\\.,\\d]+) (?<currency>[A-Z]{3})$") //
-                        .assign((t, v) -> processWithHoldingTaxEntries(t, v, "creditableWithHoldingTax", type));
+                        .assign((t, v) -> processWithHoldingTaxEntries(t, v, "creditableWithHoldingTax", type))
+
+                        // @formatter:off
+                        // Zinsen für 158 Zinstage USD 150,86
+                        // @formatter:on
+                        .section("tax", "currency").optional() //
+                        .match("^Zinsen f.r \\d+ Zinstage (?<currency>[A-Z]{3}) (?<tax>[\\.,\\d]+)$") //
```
Nirus2000, line 651: Das ist falsch und auch keine Steuern, @buchen 

Bitte revert... ich werde das die Tage korrigieren.
Wir behandeln diese Punkte in anderen PDF-Importern...

buchen: @Nirus2000 Good catch. With #5161 I pushed a pull request to revert the recording as taxes.

I tried to find out how the other importers are doing this. I found ING Diba and Baader. Both are recording the Stückzinsen only in the note of the transaction. Is this how it is handled? (I Know bonds are not support in PP)

hporten: I followed the advise of recording a "false tax" that is given in https://help.portfolio-performance.info/en/getting-started/manage-portfolio/bonds/  But if there is a nicer solution, that would be great. And if @Nirus2000 could give a hint, I would be happy to adjust the handling myself.

Nirus2000: Hello @hporten 

> @Nirus2000 Good catch. With #5161 I pushed a pull request to revert the recording as taxes.
> 
> I tried to find out how the other importers are doing this. I found ING Diba and Baader. Both are recording the Stückzinsen only in the note of the transaction. Is this how it is handled? (I Know bonds are not support in PP)

Yes, since these are all just workarounds for the actual posting, we have decided to only record the accrued interest as a note. Everyone can then decide for themselves how these should be posted later (interest, dividende, taxes ...).

---

> I followed the advise of recording a "false tax" that is given in https://help.portfolio-performance.info/en/getting-started/manage-portfolio/bonds/ But if there is a nicer solution, that would be great. And if @Nirus2000 could give a hint, I would be happy to adjust the handling myself.

We would stick with the current strategy of leaving the accrued interest as a note during automatic import.
Each user can then decide for themselves how these accrued interest payments are recorded using a manual transaction.
This information is also stored in the PP Help section “Recording the interest payment.” We therefore leave this option up to the users themselves.

Would you please correct this PR accordingly?

Regards
Alex

hporten: @Nirus2000, thanks for the explanation. I will adapt the patch accordingly once done with 1-2 open issues.

PR #5156, opened 2025-11-21
Accept trailing whitespace in bank statement from Deutsche Bank

hporten:
The statement year got extracted via a pattern that has strictly been anchored to the end of line so far. This is now relaxed to accept trailing whitespace.

QUESTION: is there an easier way? Like an option to strip trailing whitespace altogether? This could be a convenient
pre-processing.


buchen:
> QUESTION: is there an easier way? Like an option to strip trailing whitespace altogether? This could be a convenient
pre-processing.

You mean on the text before passing it to any extractors?
@Nirus2000 do you have an opinion?

To be honest, I am not sure how often that is the case. But there is some basic sanitization in the PDFInputFile class.

PR #5157, opened 2025-11-21
Use SVG images (here: the gear icon often found top right)

buchen:
Eclipse has added support for SVG images. This should support nicer icons when the screen is scaled at different zoom levels.

This **draft** change is exchanging the configuration icon - the gear in the top right.


buchen:
@Nirus2000 does this make a difference on Windows?

On macOS, I cannot see any visible difference at all (of course, it would make it simpler to have only one icon file in the source).

Nirus2000:
i didn't see any differences...
grafik

grafik

buchen:
@Nirus2000 
> i didn't see any differences... [on Windows]

Thanks for the feedback. Apparently, 16x16 and 32x32 sizes is all that is used. I would have thought that intermediate zoom levels would have shown it. Then this change is nice (less icon files) but also nice to have.

PR #5186, opened 2025-12-02
Introduced an IgnoredItem class to avoid false import errors.

hporten:
Background: In addition to bank statements, some banks issue
   additional dedicated documents with more information about
   buy/sell transactions or dividend payments. We expect to
   see those extra documents and will skip such items when
   found in the list of a bank statement. Concrete examples:
   See the skipTransaction flag used in extractors for Deutsche Bank,
   Ebase and Onvista.

Observed behavior: when importing a bank statement from e.g. Deutsche Bank
   that has an ignorable entry (like a divident payment) among other
   relevant entries, the divident item is ignored silently. In case
   the item list happens to consist of nothing but such entries,
   the import fails with an error:

       "Unknown or unsupported transaction type"

   This lead to a futile debugging attempt: the file type as well as
   the transaction type are known and supported.

Expected behavior: an import should succeed without errors independant
   of how many ignorable (but known and valid) items are found in a
   document.

The responsible code is

 1.) DeutscheBankPDFExtractor.java:

    if (type.getCurrentContext().getBoolean("skipTransaction"))
        return null;

 2.) AbstractPDFExtractor.java:

      List<Item> items = parseDocumentTypes(documentTypes, filename, text);
      if (items.isEmpty())
      {
         errors.add(new UnsupportedOperationException(
                    MessageFormat.format(Messages.PDFdbMsgCannotDetermineFileType, getLabel(), filename)));
      }

This patch introduces a IgnoredItem class that will be returned instead of null. That way, the returned list will not become empty anymore in case the parsing is successful. This approach is similar to the "null object" design pattern.

This patch goes way behind just silencing this error: the items will bubble up to the ReviewExtractedItems wizard page. They will be shown with a yellow warning sign denoting that the item will be ignored upon import. This is meant to provide more transparency to the user. But this advanced behavior is not that important.

A more minimal patch would not show those items to the user. The items in question could be filtered out much earlier in the process. By use of a stripped down IgnoredItem class or some other means that will avoid flagging an empty list as an error.

Todo: patch Ebase and Onvista extractors as well.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
@@ -2900,56 +2900,6 @@ public class DeutscheBankPDFExtractorTest
         new AssertImportActions().check(results, "EUR");
 
         // assert transaction
-        assertThat(results, hasItem(deposit(hasDate("2021-10-21"), hasAmount("EUR", 1000.00), //
-                        hasSource("GiroKontoauszug09.txt"), hasNote("Übertrag (Überweisung) von Max Mustermann"))));
-    }
-
-    @Test
-    public void testGiroKontoauszug10()
-    {
-        var extractor = new DeutscheBankPDFExtractor(new Client());
-
-        List<Exception> errors = new ArrayList<>();
-
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "GiroKontoauszug10.txt"), errors);
-
-        assertThat(errors, empty());
-        assertThat(countSecurities(results), is(0L));
-        assertThat(countBuySell(results), is(0L));
-        assertThat(countAccountTransactions(results), is(2L));
-        assertThat(countAccountTransfers(results), is(0L));
-        assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(results.size(), is(2));
-        new AssertImportActions().check(results, "EUR");
-
-        // assert transaction
-        assertThat(results, hasItem(deposit(hasDate("2025-10-15"), hasAmount("EUR", 400.00), //
-                        hasSource("GiroKontoauszug10.txt"), hasNote("Überweisung von Dr. kQEbfBPDq ZgltrGG wBPgFcQwn"))));
-
-        // assert transaction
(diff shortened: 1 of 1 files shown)


hporten:
This pull request is meant to serve as a base for a discussion rather than a fixed implementation. It works for me but "all roads that lead to Rome", i. e. different implementations that have the same net effect (no more import errors)
would be equally good.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -1035,6 +1035,10 @@ private void checkEntries(List<ExtractedEntry> entries)
                 entry.addStatus(new ImportAction.Status(Code.ERROR, entry.getItem().getFailureMessage()));
                 allErrors.add(new IOException(entry.getItem().getFailureMessage() + ": " + entry.getItem().toString())); //$NON-NLS-1$
             }
+            else if (entry.getItem().isIgnored())
+            {
+                entry.addStatus(new ImportAction.Status(Code.WARNING, "Ignored")); // TODO: translate message
```
buchen: With the type WARNING, the transaction is not imported by default.
However, the user can choose to import it nevertheless by choosing "import" from the context menu.
(Background: it is used if a duplicate is detected. But the user can still choose to import).

Either we mark this here as ERROR or we introduce a new category (say IGNORED).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/Extractor.java
```diff
@@ -676,6 +681,71 @@ public Status apply(ImportAction action, Context context)
         }
     }
 
+    static class IgnoredItem extends Item
```
buchen: Should we add a "reason" label to the item?
The the importer could say *why* it ignored that item.
And this is what we could display as message in the dialog.

hporten:
> I propose two things:
> 
>     * add a reason to the Ignored item

Yes! I already contemplated adding that. The import review UI (as it is now) would not show the reason anywhere but via the tooltip but that would be a good start.

>     * use the code "ERROR" (or introduce "IGNORED")

Okay. The reuse of WARNING was indeed a hack. I will work on a new iteration of the patch within the next days.

hporten:
I will soon close this pull request and suggest a new one. Big renamings and messed up merge conflicts with master make this necessary.

PR #5200, opened 2025-12-07
add local-sensitive viewer comparator to data series selection

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4529

Hello, on some locations there is a case sensitive sorting, on some location not. And more than a case sensitive sorting, I understand it is local sorting for special letter.
I reused what is done in ColumnViewerSorter, https://github.com/portfolio-performance/portfolio/commit/f61269ac8954f50fb36133c49dff75cce8d87e22, I understand the goal is to use `TextUtil.compare` in the different `ViewerComparator`.

the location where the locale-sensitive sorting is now applied too : 

- when selecting data series to plot on graph
- when selecting accounts for creating grouped account and add new accounts to them

**Expected behavior :** 
Capture d’écran 2025-12-07 131410
OK

**Before :**  different behavior elsewhere :
Capture d’écran 2025-12-07 132645

Capture d’écran 2025-12-07 131459

**After :** 
Capture d’écran 2025-12-07 131746
Capture d’écran 2025-12-07 131813


buchen:
Thanks for the change.

I just wonder if we should have a reusable ViewerComparator instance instead? That could reduce the code duplication.

buchen:
I refactored the change to combine the viewer comparator into one helper class and merged it

PR #5201, opened 2025-12-07
IBFlex import zero value option expiry transactions

georgemac-labs:
- Amend IBFlex import to include option expirations transactions,
  to avoid expired "zombie" positions accumulating in the portfolio
- Add test, with examples of OTM and ITM (assignment) transactions

Explanation / discussion of this PR: https://forum.portfolio-performance.info/t/interactive-brokers/276/365

Note: The real examples I have are generally shorts. I could do with examples of expiring profitable long options trades, to ensure those are handled correctly.


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/testIBFlexStatementFile02.xml
```diff
@@ -5,7 +5,6 @@
...
-<Trade currency="USD" symbol="ORCL  170915P00050000" exchange="--" isin="" cusip="" conid="268828466" description="ORCL 15SEP17 50.0 P" buySell="BUY" tradeDate="20170915" reportDate="20170915" underlyingConid="272800" underlyingSymbol="ORCL" putCall="P" transactionType="BookTrade" quantity="1" tradePrice="0" tradeMoney="0" ibCommission="0" ibCommissionCurrency="USD" assetCategory="OPT" accountId="U1234567" acctAlias="" model="" fxRateToBase="0.83701" securityID="" securityIDType="" issuer="" multiplier="100" strike="50" expiry="20170915" principalAdjustFactor="" tradeID="1908991474" tradeTime="162000" settleDateTarget="20170918" proceeds="-0" taxes="0" netCash="0" closePrice="1.26" openCloseIndicator="C" notes="A" cost="49.5" fifoPnlRealized="0" fxPnl="0" mtmPnl="126" origTradePrice="0" origTradeDate="" origTradeID="" origOrderID="0" clearingFirmID="" orderReference="" volatilityOrderLink="" exchOrderId="N/A" extExecID="N/A" orderTime="" openDateTime="" holdingPeriodDateTime="" whenRealized="" whenReopened="" levelOfDetail="EXECUTION" changeInPrice="0" changeInQuantity="0" orderType="" isAPIOrder="N" />
```
lmb, line 8: Why does this have to be removed?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -485,6 +485,22 @@ private class IBFlexStatementExtractorResult
                             ASSETKEY_WARRANTS).contains(element.getAttribute("assetCategory")))
                 return;
 
+            // An option expiration (both OTM and ITM) is represented as a BookTrade closing transaction
+            // with zero net cash value. OTM options are marked as "Ep" (Expired) in the notes,
+            // while ITM options are marked as "A" (Assignment) when they expire in the money.
+            // Notes can be semicolon-separated (e.g., "C;Ep").
+            String notes = element.getAttribute("notes");
+            boolean hasExpirationIndicator = notes != null && 
+                            (Arrays.asList("Ep", "A").contains(notes) ||
```
lmb, line 494: Doesn't notes.split(";") also work for the case where no delimiter is present?

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
```diff
@@ -3367,4 +3311,57 @@ public void testIBFlexStatementFile25() throws IOException
...
+
+        var results = extractor.extract(Collections.singletonList(tempFile), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(2L));
+        assertThat(countBuySell(results), is(2L));
+        assertThat(countAccountTransactions(results), is(0L));
+
+        // check 1st buy sell transaction (OTM option expiration - notes="Ep")
+        BuySellEntryItem item = results.stream() //
```
lmb, line 3335: Check the test above, there are some newer helpers which make writing (and reading) these tests a bit less tedious.

pfalcon:
This patch is over-engineering. If your broker says there was a transaction for the amount of 0, it means there was a transaction for the amount of 0. (They may be wrong too, as anything at all may be wrong, but then you'll need to review and adjust on case by case basis; by default, they' right). There's no need to test whether it's option or not, obscure codes, etc. So, the right patch is:

```
-            if (portfolioTransaction.getPortfolioTransaction().getCurrencyCode() != null && portfolioTransaction.getPortfolioTransaction().getAmount() == 0)
-            {
-                item.setFailureMessage(Messages.MsgErrorTransactionTypeNotSupported);
-            }
-            else if (Messages.MsgErrorOrderCancellationUnsupported.equals(portfolioTransaction.getPortfolioTransaction().getNote()))
+            if (Messages.MsgErrorOrderCancellationUnsupported.equals(portfolioTransaction.getPortfolioTransaction().getNote()))
```

(I.e. removing ungroundedly added confusing conditions, vs adding even more confusing conditions.)

georgemac-labs:
Thank you for the comments, but I will be changing the approach anyway, following discussions in the forum

PR #5206, opened 2025-12-08
IBFlex: import multiple accounts from a single XML

lmb:
Allow importing multiple accounts from a single XML file by pre-populating Extractor.Item.accountPrimary, portfolioPrimary, etc.

Introduces a heuristic to be able to find reference accounts in different currencies.

See the commit messages for details.


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -124,7 +124,7 @@ private static DateTimeFormatter createFormatter(String pattern)
      */
     public IBFlexStatementExtractor(Client client)
     {
-        this.client = new Client();
+        this.client = client;
```
buchen, line 127: Good catch. Thanks for the fix.

I realized now it was actually reported on the original commit that introduced this problem:
https://github.com/portfolio-performance/portfolio/commit/fbaf29a68c9514f0cd318469f94d8b18d8645211#diff-7aa193362f68c7e51a5624dfc2481ee1179984f8449df3994a387c171a382880

It applies only to the test case (for historical reasons, the test cases use a slightly different method signature than when the dialog is exporting data).

buchen:
merged

PR #5208, opened 2025-12-09
Fix discreet mode for grouped account list

mierin12:
Issue : https://forum.portfolio-performance.info/t/group-accounts-does-not-update/38300

Hello, I can reproduce the issue mentionned above : when on the Grouped account view, if you switch the Discreet Mode, the values of the top table does not immediatly follow the change. It applies if you switch view then go back to grouped account view.
I understand that the change of discreet mode calls `notifyModelUpdated()` and therefore this method is added. 
I think `notifyModelUpdated `is now called also from other actions such as add/delete an element to a grouped account, so in the second commit I am assuming it can be simplified.


buchen:
LGTM.

I believe the notifyModelUpdated is also needed if other data is updated - say that the historical prices are updated in the background.

PR #5223, opened 2025-12-16
Add average payments per time unit

gamue:
This is my first PR to this project, please bear with me when not fulfilling all requirements and guidelines yet.
It's a functionality I've missed couple of times already and thought it's a good first feature to contribute.

## Description
This PR adds the option to display average payments per month, quarter or year for tracking incoming cashflow inside the Payments-tab. This functionality already exists in the "Payments per month"-widget on the dashboard, but not inside the bigger Payments-table-views

No tests have been added as contribution-guide states that there are no UI tests needed

Localisation added for German and English, will add remaining languages after feedback.

## How do I test this
Go to Performance -> Payments, nside inside the table-views for Months, Quarters and Years you'll find the option to show the average. Once activated a new column will show up behind the last sum-colum

image


Nirus2000:
## Problem

The average columns in `PaymentsPerYearMatrixTab` and `PaymentsPerQuarterMatrixTab` calculated incorrect values by rounding up partial time periods. This caused artificially low averages, particularly noticeable when users had incomplete years or quarters of data.
  
### Calculation Example
  
Consider a user with **13 months of dividend data**, receiving **€100 per month** (total: **€1,300**):
  
**Before this fix:**
```
Average per Year    = €1,300 / 2 years     = €650/year   ❌ WRONG
Average per Quarter = €1,300 / 5 quarters  = €260/quarter ❌ WRONG
```
  
The old code rounded 13 months up to "2 complete years" and "5 complete quarters", resulting in incorrectly low averages.
  
**After this fix:**
```
Average per Year    = €1,300 / (13/12 years)    = €1,200/year   ✓ CORRECT
Average per Quarter = €1,300 / (13/3 quarters)  = €300/quarter  ✓ CORRECT
```
| Tab | Old Calculation | Old Result | Correct Result |
|-----|----------------|------------|----------------|
| **Year** | 1300 / **2** years (rounded up) | **650** ❌ | **1200** ✓ |
| **Quarter** | 1300 / **5** quarters (rounded up) | **260** ❌ | **300** ✓ |
| **Month** | 1300 / 13 months | **100** ✓ | **100** ✓ |

gamue:
thanks a lot, much appreciate the feedback and commit!
Great to know that there are actually unit-tests possible for this as well. I did some manual testing with different files, but doing it in Dec hasn't covert that, should change dates or do you suggest a better way to test this manually?

For calculating the average I was thinking about using the "amount of columns" as divider as this is the logic which I understood is also used when using the "monthly payments"-widget on a dashboard and so rounded up, not havign a strong preference and think it can argued for both versions.
Right now the average inside the widget is taking "started months until today" as base. (Using the kommer-example and added a 100-interest payment in Nov 2025 looks like:

image

Let me know what I should do to finalize this PR or how I can support the review process

Nirus2000:
I found the dashboard widget logic. Look at InvestmentHeatmapWidget.java lines 186-190:

```
  model.getRows().forEach(row -> {
      long average = Math.round(row.getDataSubList(0, 12).stream().filter(Objects::nonNull)
                      .mapToLong(l -> l == null ? 0L : l.longValue()).average().getAsDouble());
      row.addData(average);
  });
```
The dashboard widget does NOT round up the number of periods. It uses .average() which calculates the mathematical average over non-null months.

I think... all perfect 👍🏻

Nirus2000:
Hello @gamue 
> Let me know what I should do to finalize this PR or how I can support the review process

Completion of translations via Deepl <--- THX !!

gamue:
done :) 

The file pt_BR could be reduced and in general just needs the texts/keys which are different to "normal" Portuegese, all i18n-frameworks I know handle this correctly. But as texts are currently duplicated I kept it this way. Not sure if that's something you've already thought about doing, but could reduce the maintenance effort.

Besides that smaller things which I noticed and was wondering about:

1. Do you have plans to use junit5?
2. I noticed in some tests hamcrest is used, is there some guidance to use this? Personally like assertj much more

Nirus2000:
Hello @gamue 

> The file pt_BR could be reduced and in general just needs the texts/keys which are different to "normal" Portuegese, all i18n-frameworks I know handle this correctly.

We need these for our [POEditor project](https://github.com/Nirus2000/portfolio/blob/master/CONTRIBUTING.md#translations).
We do this in and export. --> See [Sync translations with POEditor.com](https://github.com/portfolio-performance/portfolio-releng)

---

> 1. Do you have plans to use junit5?
> 2. I noticed in some tests hamcrest is used, is there some guidance to use this? Personally like assertj much more

Yes, that could be done, but which tests should we cover with Junit5?
@buchen can certainly say more about that as the head of the department.

---


Then let's take the final step.
We still need to correct the column order.
This means that the average is currently displayed first, followed by the sum, as is the case with the “reverse column order.” 
We need to reverse this for both.

grafik
...
grafik

gamue:
> We still need to correct the column order.
> This means that the average is displayed first, followed by the total, as is the case with the “reverse column order.”

Do you want to have the order "Average + Sum" in "normal" view and "Sum + Average" in reversed view, always "Sum + Average" or always "Average + Sum"?
I personally don't have a preference, so can do it either or


> Yes, that could be done, but which tests should we cover with Junit5?

all of them :D junit 4 is in maintenance mode since a longer time, afaik there's a nice way to have both in parallel during migration. But also don't know i that's important, just noticed that I haven't seen junit4 in use since many years

Nirus2000:
> Do you want to have the order "Average + Sum" in "normal" view and "Sum + Average" in reversed view, always "Sum + Average" or always "Average + Sum"?
> I personally don't have a preference, so can do it either or

[Total] [Avg] ...

and ...

... [Avg] [Total]

gamue:
order adjusted:

normal view:
image

reversed view
image

buchen:
thanks for your contribution. Merged and released.

PR #5228, opened 2025-12-17
Add new Debitum Investments PDF-Importer

gamue:
Debitum is a P2P-platform and offers the option to get account-statements as pdfs, unfortunately this isn't very detailed, however provides data to track the performance and payments on a regular basis.

I understand that this extractor is a little bit special as it's creating transactions based on a summary instead of a bank statement, but think it's reducing the maintenance work to enter them manually. 
The workflow is the following:
- Create a monthly account statement per pdf
- The pdf is sent into the mail of the user
- The pdf will be imported into PP
- PP creates transactions to summarize account changes
  - Sum of all deposits during the time range.
  - Sum of all withdrawings during the time range
  - Gains during the time range: Besides interests and withheld taxes, they have regular bonus-promotions. This amount is listed on it's own inside the report, however the withheld tax for this isn't and so I decided to sum the three values together. I personally see those promo-payments as gains/interests and so think it's a fair assessment.

In case there are no withdrawings or deposits during the time range, no transaction is created for them. This is what test-case 2 is about.

It's my first dive into the pdf-extractors and was quite a challenge to get around them, as the nature of this differs and the examples I checked felt too complex (e.g. Baader) for what I need. Happy to adjust the code and get feedback around it


Nirus2000:
Hello @gamue 
thx for this Pull-Request... Please take a look at the structure of other importers such as “SuresseDirektBankPDFExtractor.” 
This structure is easier to maintain and is standard. Adapt the source formatting and variable names, then we can quickly move this into the master :-)

Regards
Alex

gamue:
Thanks a lot, looking at `SuresseDirektBankPDFExtractor` was definitely helpful. I previously checked `BondoraCapitalPDFExtractor` and `EstateGuruPDFExtractor`. I refactored it a little bit, however the main-transaction one still needs some adjustments as it's failing when I combine the remaining section and chain the matches. Will check this later again.

Now it makes much more sense to me, when having just a single assign inside the call :)

Nirus2000:
> Thanks a lot, looking at `SuresseDirektBankPDFExtractor` was definitely helpful. I previously checked `BondoraCapitalPDFExtractor` and `EstateGuruPDFExtractor`. I refactored it a little bit, however the main-transaction one still needs some adjustments as it's failing when I combine the remaining section and chain the matches. Will check this later again.
> 
> Not it makes much more sense to me, when having just a single assign inside the call :)

If you look at the source code for SuresseDirektBankPDFExtractor for DebitumPDFExtractor... it looks completely different, doesn't it? So refactor, formatting, cleaning, renaming it as I already mentioned. 😄

Please only address one problem per pull request. Mixing several changes independently of each other is bad. It causes us to lose track of the overview and history. Please remove it. 👍🏻

gamue:
missed that you're using `//` inside the classes to manipulate the code-formatter and was searching for some formatting issue inside eclipse previously. But think now they are quite similar.

Nirus2000:
Hello @gamue 
I implemented the desired changes myself and completed the importer.

Regards
Alex

Nirus2000:
Add new threat
https://forum.portfolio-performance.info/t/pdf-import-from-debitum-investments/38442/

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/debituminvestments/DebitumPDFExtractorTest.java
```diff
@@ -31,8 +35,9 @@ public class DebitumPDFExtractorTest
     @Test
     public void testAccountStatement01()
     {
-        var extractor = new DebitumPDFExtractor(new Client());
-        var errors = new ArrayList<Exception>();
+        var extractor = new DebitumInvestmentsPDFExtractor(new Client());
```
gamue, line 38: as this is very common in all classes, and some of the tests are very long, I was thinking that it's a better style to extract the init of `extractor` and `errors` into an own method and annotate it with `BeforeEach`. Not sure if you'd like that though ;)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DebitumInvestmentsPDFExtractor.java
```diff
@@ -0,0 +1,142 @@
...
+                                        .match("^Report date to \\(inclusive\\): (?<date>[\\d]{4}\\-[\\d]{2}\\-[\\d]{2})[\\s]*$") //
+                                        .assign((ctx, v) -> ctx.put("date", v.get("date"))));
+        this.addDocumentTyp(type);
+
+
+        // @formatter:off
+        // DEPOSITS during reporting period: 4170.0
+        // WITHDRAWALS during reporting period: 0.0
+        // @formatter:on
+        var depositRemovalBlock = new Block("^(DEPOSITS|WITHDRAWALS) during reporting period: [\\.\\d]+[\\s]*$");
```
gamue, line 55: ah good point in merging those blocks, thanks

gamue (commented):
thanks for the changes, learnt quite some things on the code-base and coding-style/-preferences. Feel free to leave some comments on the lines you don't like in the future, so it's easier to address them :)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/debituminvestments/DebitumPDFExtractorTest.java
Nirus2000: I don't think... we need to update over 9000 tests, and the benefit is very small.
They are just tests... nothing more. They must be clearly conclusive

Issue #4063, opened 2024-06-08
Original size in Security Chart offsets the Market Trend vs Purchase value areas


mierin12:
**Describe the bug**
In Security Chart, when Market Trends vs Purchase value is used and a negative and positive area are visible, the option "Original size" (accelerator key 0), the pos/neg area are offset from the Market Trend.
The effect is sometimes not very visible, sometimes well visible.

**To Reproduce**
1. Go to 'All Securities'
2. Click on 'Chart'
3. Tick 'Market Trend vs Purchase Value'
4. Select a period where both a negative and positive area are visible
5. Use Original Size (0)
6. See error

**Expected behavior**
No offset

**Screenshots**
From Kommer file :
Before using "Original Size":
![Capture d’écran 2024-06-08 155722](https://github.com/portfolio-performance/portfolio/assets/160436107/12eeaefd-3436-43e7-83e1-3d007f35896e)
After using "Original Size":
![Capture d’écran 2024-06-08 155354](https://github.com/portfolio-performance/portfolio/assets/160436107/0791f5f3-6569-41bc-b581-7bed9c0f0044)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version 0.68.4

Issue #2490, opened 2021-11-01
Reseting views leads to graph error


funnym0nk3y:
**Describe the bug**
Zooming an moving the graph in securities view (below list of each security) and resetting it with 0 leads to a corrupt graph like: 

![grafik](https://user-images.githubusercontent.com/41870754/139662153-425d0f4b-6f11-40ef-9f86-77b5cb4e15f7.png)

**Desktop (please complete the following information):**
 - OS: Win10
 - Version: 0.55

mierin12:
Will be fixed in the next revision by https://github.com/portfolio-performance/portfolio/commit/a61304679f4d37ea4eabfa902b03b812c584d4d5

PR #5244, opened 2025-12-21
fix Original Size action for SecuritiesChart

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/2490
Closes https://github.com/portfolio-performance/portfolio/issues/4063

Hello, on the Securities Chart, when you hit the '0' accelerator key or trigger the 'reset to Original size' action, the graph becomes incorrect when there are the positive/négative area on it.
What I understand : 
* SecuritiesChart is a TimelineChart and therefore the adjustRange of Timelinechart is used for those actions
* However In the updateChart of SecuritiesCharts, dedicated ranges are used for yAxis2nd and yAxis3rd, so if adjustRange is called after, the particular settings is lost.

So what I propose is to call updateChart at the end of the adjustRange method of timelinechart. But this means changing from private to public. I think this works. Otherwise, at first I was thinking we could get an override adjustRange method for SecuritiesCharts, but this was not so straightforward.

~~While I write this, I note that I did not check the behaviour when multiple securities are selected, I will check.~~
-> OK, no problem seen with multi securities selection.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/PriceTimelineChart.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/chart/PriceTimelineChart.java
@@ -1,0 +1,78 @@
+package name.abuchen.portfolio.ui.util.chart;
+
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swtchart.IAxis;
+import org.eclipse.swtchart.IAxis.Position;
+import org.eclipse.swtchart.LineStyle;
+import org.eclipse.swtchart.Range;
+
+import name.abuchen.portfolio.ui.util.format.AxisTickPercentNumberFormat;
+
+public class PriceTimelineChart extends TimelineChart
+{
+    private Double firstQuote;
+
+    public PriceTimelineChart(Composite parent)
+    {
+        super(parent);
+
+        // 2nd y axis
+        int axisId = getAxisSet().createYAxis();
+        IAxis y2Axis = getAxisSet().getYAxis(axisId);
+        y2Axis.getTitle().setVisible(false);
+        y2Axis.getTick().setVisible(false);
+        y2Axis.getGrid().setStyle(LineStyle.NONE);
+        y2Axis.setPosition(Position.Primary);
+
+        // 3rd y axis (percentage)
+        int axisId3rd = getAxisSet().createYAxis();
+        IAxis y3Axis = getAxisSet().getYAxis(axisId3rd);
+        y3Axis.getTitle().setVisible(false);
+        y3Axis.getTick().setVisible(false);
+        y3Axis.getTick().setFormat(new AxisTickPercentNumberFormat("+#.##%;-#.##%")); //$NON-NLS-1$
+        y3Axis.getGrid().setStyle(LineStyle.NONE);
+        y3Axis.setPosition(Position.Primary);
+    }
+
+    @Override
+    public void adjustRange()
+    {
+        // custom implementation because the 2nd and 3rd y axis must follow
+        // exactly the first one
+        try
+        {
+            setRedraw(false);
+
+            getAxisSet().adjustRange();
+
+            var yAxis1st = getAxisSet().getYAxis(0);
+            var yAxis2nd = getAxisSet().getYAxis(1);
+            var yAxis3rd = getAxisSet().getYAxis(2);
+
+            if (firstQuote != null)
+            {
+                yAxis2nd.setRange(new Range(yAxis1st.getRange().lower - firstQuote,
+                                yAxis1st.getRange().upper - firstQuote));
+
+                if (firstQuote != 0)
+                {
+                    yAxis3rd.setRange(new Range(yAxis1st.getRange().lower / firstQuote - 1,
+                                    yAxis1st.getRange().upper / firstQuote - 1));
+                }
+            }
+
+            ChartUtil.addYMargins(this, 0.08);
+
+        }
+        finally
+        {
+            setRedraw(true);
+        }
+    }
+
+    public void setFirstQuote(Double firstQuote)
+    {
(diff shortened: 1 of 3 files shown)


buchen:
I am not sure about the circular dependency between SecuritiesChart and TimelineChart. The former uses the latter, not the other way around. The "util.chart" package should not depend on views.

However, I can follow your analysis. The 2nd and 3rd axis cannot be freely set, but must "inherit" the range from the first y axis. The generic adjustRange implementation resets each axis independently.

I could see two ways to address this issue:
a) The SecuritiesChart creates its own subclass of the TimelineChart and that subclass overrides the adjustRange method. And we could also move the y2Axis and y3Axis from the TimelineChart into this class (to my understanding that already "leaked" from the SecuritiesChart into the TimelineChart).
b) we somehow make the MovePlotKeyListener configurable so that the SecuritiesChart can hook into the "0" accelerator key instead.

I was playing around a little. Maybe something along these lines (although I have to admit the "firstQuote" is a little bit awkward because we would have to set it with each update method)

@mierin12 what do you think?

```java
package name.abuchen.portfolio.ui.views.securitychart;

import org.eclipse.swt.widgets.Composite;
import org.eclipse.swtchart.IAxis;
import org.eclipse.swtchart.IAxis.Position;
import org.eclipse.swtchart.LineStyle;
import org.eclipse.swtchart.Range;

import name.abuchen.portfolio.ui.util.chart.ChartUtil;
import name.abuchen.portfolio.ui.util.chart.TimelineChart;
import name.abuchen.portfolio.ui.util.format.AxisTickPercentNumberFormat;

public class PriceTimelineChart extends TimelineChart
{
    private Double firstQuote;

    public PriceTimelineChart(Composite parent)
    {
        super(parent);

        // 2nd y axis
        int axisId = getAxisSet().createYAxis();
        IAxis y2Axis = getAxisSet().getYAxis(axisId);
        y2Axis.getTitle().setVisible(false);
        y2Axis.getTick().setVisible(false);
        y2Axis.getGrid().setStyle(LineStyle.NONE);
        y2Axis.setPosition(Position.Primary);

        // 3rd y axis (percentage)
        int axisId3rd = getAxisSet().createYAxis();
        IAxis y3Axis = getAxisSet().getYAxis(axisId3rd);
        y3Axis.getTitle().setVisible(false);
        y3Axis.getTick().setVisible(false);
        y3Axis.getTick().setFormat(new AxisTickPercentNumberFormat("+#.##%;-#.##%")); //$NON-NLS-1$
        y3Axis.getGrid().setStyle(LineStyle.NONE);
        y3Axis.setPosition(Position.Primary);
    }

    @Override
    public void adjustRange()
    {
        // custom implementation because the 2nd and 3rd y axis must follow
        // exactly the first one
        try
        {
            setRedraw(false);

            getAxisSet().adjustRange();

            var yAxis1st = getAxisSet().getYAxis(0);
            var yAxis2nd = getAxisSet().getYAxis(1);
            var yAxis3rd = getAxisSet().getYAxis(2);

            if (firstQuote != null)
            {
                yAxis2nd.setRange(new Range(yAxis1st.getRange().lower - firstQuote,
                                yAxis1st.getRange().upper - firstQuote));

                if (firstQuote != 0)
                {
                    yAxis3rd.setRange(new Range(yAxis1st.getRange().lower / firstQuote - 1,
                                    yAxis1st.getRange().upper / firstQuote - 1));
                }
            }

            ChartUtil.addYMargins(this, 0.08);

        }
        finally
        {
            setRedraw(true);
        }
    }

    public void setFirstQuote(Double firstQuote)
    {
        this.firstQuote = firstQuote;
    }

}
```

mierin12:
Hello @buchen, thanks for the review, yes your method is much better, the way a) that you did is better I think. In way b), it is not only MovePlotKeyListener that should be made configurable but also ChartContextMenu.

I will update the PR accordingly with your PriceTimelineChart, but I believe you have already the fix on your side. I put PriceTimelineChart in name.abuchen.portfolio.ui.util.chart instead.

All of this is also fixing the same king of bug: if you have the percentage axis shown in the security chart, the '0' short cut will mess up the axis each time.

mierin12:
> And we could also move the y2Axis and y3Axis from the TimelineChart into this class (to my understanding that already "leaked" from the SecuritiesChart into the TimelineChart).

I understand that too. The second and third axis were added in https://github.com/portfolio-performance/portfolio/commit/9613b4f80c805ce4b834c07054f41398183d510f and https://github.com/portfolio-performance/portfolio/commit/c62f1b4f3d6d84232787144cadecc0801e1c1fc4 each time for SecuritiesChart purpose.

I did not find any call to `chart.getAxisSet().getYAxis(1) `other than in SecuritiesChart.
However, we do have call to `chart.getAxisSet().getYAxis(2) `in ChartToolsManager and CrossHairTool, so we might get null exception ? I did not manage to trigger them yet but who knows.

buchen:
> I put PriceTimelineChart in name.abuchen.portfolio.ui.util.chart instead.

I took the liberty to move it to ui.views.securitychart package. If we refactor the SecuritiesChart (it is almost 2000) lines, we can use that package for more specific classes (like the SharesHeldSeries already in that package). My thinking was that the PriceTimelineChart is specific to the SecuritiesChart - it is about the historical prices.

> However, we do have call to chart.getAxisSet().getYAxis(2) in ChartToolsManager and CrossHairTool, so we might get null exception ? I did not manage to trigger them yet but who knows.

Good catch. I added one commit that at least does a check if the axis exists.

I think it is okay. Maybe a "better" way would be for the measurement tool to inspect the available axes, determine the "primary" and "secondary or percentage" ones, and, if available, remember the values (in the Spot in ChartToolsManager) and draw the labels (in the CrossHairTool). But SWTchart has no options to attach additional information to the axis. Therefore, I am not sure if the effort is worth it.

BTW, I have seen that there is a new SWTCHart version 1.1 available. I have updated the target platform.

PR #5247, opened 2025-12-23
Fix for Issue 5242 - Import von flatex.at PDFs nach Formatänderung durch Flatex nicht mehr möglich

arnonm:


Nirus2000:
This pull request does not work as it stands.
Firstly, tests are missing for the other pull requests, and secondly, the implementation is faulty. I had already pointed out to you that you should look at the architecture of PP and first carry out “one” pull request cleanly. It doesn't help us if you start pull requests that are incomplete—they just create more work. You haven't responded to changes in other pull requests either, and instead have opened more incomplete ones.
Please think more carefully about what you want to contribute and take your cue from other contributors.
Read through the contributing rules...

arnonm:
Alexander
I am puzzled by your comment. I have responded and followed up on every
pull request. I don't think that I have left any open. However, with your
comment - I will go back and check.

This one also seems very complete - one example, one change to the parser
and a test.
The CI failed due to an issue on the main branch, not due to this pull
request, which I see that you fixed since I submitted this, and now passes.

So again I am puzzled by the comment

Issue #3520, opened 2023-08-24
PP freezes, when changing allocation percentages in taxonomy


mknudsen:
**Describe the bug**

**To Reproduce**

1. Go to a Taxonomy
2. Click on an Allocation percentage
3. Fiddle with the percentage a bit
4. PP freezes, uses 100% CPU

**Expected behavior**

PP does not freeze but keeps on working

**Screenshots**

See also my screenrecording:

https://github.com/portfolio-performance/portfolio/assets/564654/b64148f3-665d-4d7a-9368-da1e80371612

**Desktop (please complete the following information):**
 - OS: Mac OS Ventura on M1 V 13.5.1 (22G90)
 - Version Portfolio Performance
   - Version: 0.65.1 (Aug 2023)
   - Platform: macosx, aarch64
   - Java: 17.0.5+8-LTS, Azul Systems, Inc.

Nirus2000:
Hello @mknudsen 
Can you please create a DUMP file?

How this works you can find here in this post. (Translate with deepl.com)
https://forum.portfolio-performance.info/t/import-hangt-sich-auf-seit-update/23106/40

Regards
Alex

mknudsen:
Here is the dump file 
[freeze.txt](https://github.com/portfolio-performance/portfolio/files/12446569/freeze.txt)

buchen:
I update the title - I do not think it is specific to changing the allocation percentages, but the framework has issues when switching back and forth from edit mode. I have experienced this before occasionally, but never dared to dive into the SWT source code.

buchen:
Note to myself:
first test if the issue is reproducible with the latest SWT integration build.
if yes, then we would need a reproducible example for the SWT team

buchen:
eclipse-platform/eclipse.platform.ui#1069

ReessKennedy:
+1 and this also looks like related discussion: https://forum.portfolio-performance.info/t/application-hangs-when-editing-transaction-in-line/22840 ... wondering if this is just on Ventura and if I could solve by moving to Sonoma? I am only seeeing Ventura users mention this.

AlexanderLill:
I have created a video to reproduce, see here: https://github.com/portfolio-performance/portfolio/issues/3818#issuecomment-2568424986

georgemac-labs:
[...]

PR #5252, opened 2025-12-26
Add double-click edit toggle and i18n improvements

liuhaoXD:
This PR introduces several usability and localization enhancements:

1. Added 'Enable Double Click To Edit' toggle in program settings
   to allow users to disable double-click editing behavior. This
   provides a workaround for users affected by the rapid-click
   issue described in the linked issue #3520.

2. Enhanced 'Save File' dialog title with i18n support.

3. Updated macOS application menu to use 'Settings' instead of
   'Preferences' to align with Apple's naming convention changes
   introduced in macOS Ventura (13.0) and later versions.

4. Refined multiple Chinese (zh) translation strings to improve
   clarity and consistency with standard terminology.


buchen:
Hi @liuhaoXD, the changes make sense. I cherry picked them into separate commits and merged them into master. Next time, please separate different features/changes into separate pull requests. That makes it much easier to review.

there is one question - see the code comment - where I do not understand the intention. Maybe you can comment (and then create a new pull request with the change).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/editor/ClientInput.java
```diff
@@ -323,10 +324,10 @@ private File pickFile(Shell shell, String extension, String fileNameProposal)
 
         // if an extension is given, make sure the file name proposal has the
         // right extension in the save as dialog
-        if (extension != null && !fileNameProposal.endsWith('.' + extension))
+        if (extension != null && fileNameProposal.endsWith('.' + extension))
         {
             int p = fileNameProposal.lastIndexOf('.');
-            fileNameProposal = (p > 0 ? fileNameProposal.substring(0, p + 1) : fileNameProposal + '.') + extension;
+            fileNameProposal = (p > 0 ? fileNameProposal.substring(0, p + 1) : fileNameProposal + '.');
```
buchen, line 330: What is the idea of this change?

Originally, the code wants to make sure that if a file has an extension (say: "my-file.xml") but the file is to be saved with a different extension (say a password-protected "my-file.portfolio"), then the extensions is fixed to have the correct one. At least in my testing, this works.

The proposal now applies a change only if the extension is already the expected one.

liuhaoXD: Thanks for your kindly remind. I will pay attention to splitting MRs and commits properly in the future.

I made this change to address a specific case I encountered: 
The original code here checks if the `fileNameProposal` contains the target file extension (.portfolio). If not, it automatically appends the extension to `fileNameProposal`. 

Logically speaking, this design seems sound, but I ran into the following issue during usage: 
The `Save File` dialog component will automatically add extension to given filename, which results in the saved filename ending up as `.portfolio.portfolio`, here is the screenshot:
image

PR #5256, opened 2025-12-28
Adjust styling of Bondora-PDF-tests

gamue:
While browsing through the code I've seen that this class had a mixture of old and new coding styles, so modernized it that all tests are looking the same


buchen:
Hi @gamue, the change looks good to me. Merged.

In general, I believe there are a lot more tests of the "old" style. In an ideal world, we never have to touch the tests again. But if we refactor the tests, we should do this like you did it (thanks for that!): in a separate commit that only changes the test itself. If a commit changes feature code and test at the same time, one never knows if the functionality accidentally changes.

Similarly, my thinking on feature code is similar. If you change the line, then it makes sense to update to the latest Java features (for example use the keyword "var"). But if the line is not touched, then it is often better to not touch that code. The smaller the commit, the easier

PR #5260, opened 2025-12-29
Minor refactorings inside csv-classes

gamue:
I read through the CSV-classes (exporter and importer) and while doing this noted that this seems to be older code, so thought I give it some smaller touch-ups
- In most cases changed to `var` according to code-guidelines
- Modernized switch-statements. Either rewrote them to have a return or bundled connected types with comma

There are no functional changes and everything works as before, I haven't touched any test-classes to display this.

Thought while going through the code, I think it'll be better to offload csv-parsing to existing libs (e.g. `de.siegmar:fastcsv`) to reduce complexity and size of the code-base while focussing on the main business logic. However as the code already exists might be an unnecessary bigger refactoring.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractor.java
@@ -2,7 +2,6 @@ package name.abuchen.portfolio.datatransfer.csv;
 
 import java.text.MessageFormat;
 import java.text.ParseException;
-import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -17,13 +16,11 @@ import name.abuchen.portfolio.datatransfer.csv.CSVImporter.DateField;
 import name.abuchen.portfolio.datatransfer.csv.CSVImporter.EnumField;
 import name.abuchen.portfolio.datatransfer.csv.CSVImporter.Field;
 import name.abuchen.portfolio.datatransfer.csv.CSVImporter.ISINField;
-import name.abuchen.portfolio.model.Account;
 import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.AccountTransaction.Type;
 import name.abuchen.portfolio.model.AccountTransferEntry;
 import name.abuchen.portfolio.model.BuySellEntry;
 import name.abuchen.portfolio.model.Client;
-import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.PortfolioTransaction;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.model.Transaction.Unit;
@@ -35,7 +32,7 @@ import name.abuchen.portfolio.money.Money;
     {
         super(client, Messages.CSVDefAccountTransactions);
 
-        List<Field> fields = getFields();
+        var fields = getFields();
         fields.add(new DateField("date", Messages.CSVColumn_Date)); //$NON-NLS-1$
         fields.add(new Field("time", Messages.CSVColumn_Time).setOptional(true)); //$NON-NLS-1$
         fields.add(new ISINField("isin", Messages.CSVColumn_ISIN).setOptional(true)); //$NON-NLS-1$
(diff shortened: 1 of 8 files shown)


buchen:
The changes look good to me (and, yes, the code has not seen significant changes in the past years).
I'll merge this change.

 > I think it'll be better to offload csv-parsing to existing libs

At the moment, the project is using [commons-csv](https://commons.apache.org/proper/commons-csv/). I looked at FastCSV and it looks good (particularly the "no dependency" thing, although I have to say I did not have issues with commons-csv in the past). So it could make sense to switch and I would entertain a pull request.

What I do not understand at the moment is what we could additionally offload to FastCSV. The [CSVImporter#processStream](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java#L918-L976) is parsing the CSV. The rest of the code is concerned with converting the strings to values according to configuration, deriving transactions from the values, deriving missing or derived properties such as exchange rates, validating mandatory and optional parameters, etc. Where do you see the potential?

Having said that, I believe there is room for improvement in the area of the CSV import. For example, I think it is counterintuitive for the user to separate between account and portfolio transactions. I think that should be one category (transactions) which internally understands what to propose. Then I think the handling of the "templates" is awkward. All to often, I first want to test if the import is successful before saving the configuration as a template - but I have to do the saving on the first page before forgetting. There are more usability things - I'd love to see if transactions are successful already on the first page. etc.

gamue:
thanks! 
yes, maybe you're right and not so much can be offloaded, will need to check more detailed. Think I might have been confused by the amount of classes and boilerplate-code.
You're right, I was also puzzled by the difference between AccountTransaction and PortfolioTransaction, besides this I had the following thoughts about the code:

- The package contains importers, exporters and config classes. An additional package-layer could improve readability (`csv.import` & `csv.export` might be enough for the start). 
- Some classes like `CSVConfig` or `CSVImporter` also contain internal classes, which I personally don't like that much and extracting them into own classes should also improve readability imo. Maybe they should be in an own `csv.model`-package
- Almost all classes do have a comment `/* package */` somewhere, which I haven't really understood. Should those be `protected`?
- Not only related to the csv-code: I'm quite a fan of using lombok for boiler-plate code like getter, setter, constructor, hashcode, equals and all bigger projects I worked on in the last 15y have used this. Was wondering if introducing that to PP  would be an option? Would think it could reduce the code-base to make it easier readable and would expect most engineers are used to it

buchen:
> The package contains importers, exporters and config classes.

"import" as a package name is forbidden but "in" and "out" could work as package names.

In general, I am not a fan of super fine granular packages. I understand there is a "overwhelming" for new developers, but Claude will talk users through this. In this case, I see the argument as import and export are quite different. Feel free to make a pull request here.

> Some classes like CSVConfig or CSVImporter also contain internal classes, which I personally don't like that much and extracting them into own classes should also improve readability imo. Maybe they should be in an own csv.model-package

Personally, I prefer private classes where it is connected to the internals of a class. I do not want to see them anywhere unless I am working/editing with the class. Take CSVConfig: the "Key" interface is really only about readability of the constants (could also be KEY_FIELD instead of Key.FIELD). CSVColumnConfig probably would be a ```record``` today. But because it is private, I do not get it proposed, I see it with the red square when opening, etc. If CSVColumnConfig is in a separate model package, it must be public and visible everywhere. Of course, I fully understand that this a matter of taste and there could be place in the code where it is not the best choice.

In the same thinking I also like nested packages. If find  ```Transaction.Unit.Type``` readable. Of course, one could have classes Transaction, TransactionUnit, and TransactionsUnitType.

> Almost all classes do have a comment /* package */ somewhere, which I haven't really understood. Should those be protected?

I am not sure what you mean with "should be protected"? ```Protected``` is not a valid classifier for classes. If the class has no classifier, it is "package protected = visible to classes in the same package". To remind myself that this was an explicit decision, I usually add the "/* package */" comment.

> Was wondering if introducing [lombok] to PP would be an option?

To be honest, not really. I haven't looked at lombok for years and was surprised to see the project that active. But I do not want to add another build dependency. When I generate code, I tend to check it in (e.g. protobuf). I guess the LLMs also can handle lombok, but they generate much of the stuff also. I don't want to wait to update Eclipse for Lombok. I'd need much more convincing... ;-)

gamue:
thanks for the feedback and answers, I get the point about personal preferences and reducing dependencies, totally makes sense. Just saying that my preference is to have small classes and methods, if something can be extracted or replaced with a simple annotation I usually go for that ;) 
AI simplifies a lot, that's true, but I'm not confinced everyone is using it or having a license, so would still argue that having a smaller code-base improves maintainability.

About lombok: IntelliJ comes with out-of-the-box support since v2020.3 which might already show that it's commonly used. But yes, would be another setup-step in Eclipse (install plugin). Not sure if you're having some "regular contributor chats" and could exchange opinions on that there, but in the end it's on you and another preference topic :)

buchen:
> Just saying that my preference is to have small classes and methods, if something can be extracted or replaced with a simple annotation I usually go for that 

I understand that. Any maybe with your next feature, the classes are smaller because you created them smaller... ;-)

PR #5261, opened 2025-12-29
Factor out a getStatusImage(Status.Code) convenience function.

hporten:
We had the same logic implemented in two separate places that risk getting out of sync in case of new codes being introduced.

Note: the new implementation is simpler but also stricter than before. On unknown status codes it will throw an
exception. But all current statuses should be covered right now. So future coding errors will be caught earlier.

(This is a preparation for a bigger change being worked on)

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
@@ -192,6 +192,17 @@ public class ReviewExtractedItemsPage extends AbstractWizardPage implements Impo
         return allEntries;
     }
 
+    private Images getStatusImage(Code code)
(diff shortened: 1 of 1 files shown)


buchen:
The change is good, of course.

> This is a preparation for a bigger change being worked on

Do you want to share what you have in mind?

Because I was also thinking about enhancing the the status code or adding a new "category" to the extracted item. For example, if the transaction cannot be imported because it is not supported, then ERROR is too hard. Or if I have pseudo transactions created as part of the IBFlex import, then I need another category then WARNING (which is currently used to indicate duplicates).

hporten:
> Because I was also thinking about enhancing the the status code or adding a new "category" to the extracted item.

Exactly this is what I needed as well to solve the "skipItem" problem reported in another pull request earlier. While working on this I also noticed the PDFMsgFinTechGroup_DoNotProcess_MissingExchangeRateIfInForex message that you just touched.

I will push another proposal in a bit. Otherwise we will have a nasty merge commit.

Issue #4471, opened 2025-01-13
Cannot create a new file


dmitry-kabanov:
**Describe the bug**

**To Reproduce**
1. Click "Create a New File" on the Welcome screen or choose `File -> New -> File`
2. See that nothing happens.

**Expected behavior**
A dialog window for creating a new portfolio file should appear on the screen.

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 0.73.0

buchen:
@dmitry-kabanov @TBdt38 

I have trouble reproducing the problem. I see the "new file wizard" open up when clicking the menu or the link on the welcome screen.

Are there any exceptions / entries in the error log? can you share?

Go to **Help -> Show error log**.

buchen:
I might have an idea.

Depending on your local locale, we cannot determine the default currency. And that results in a NullPointerException in the current version. It is already fixed af9618b953a355ab3a44e93d4a17d7ef858de3b1 but not yet published.

To verify, can you share what properties you have in "About Portfolio Performance" -> "Installationdetails" for these two entries:

```
user.country: DE
user.language: de
```

TBdt38:
> [@dmitry-kabanov](https://github.com/dmitry-kabanov) [@TBdt38](https://github.com/TBdt38)
> 
> I have trouble reproducing the problem. I see the "new file wizard" open up when clicking the menu or the link on the welcome screen.
> 
> Are there any exceptions / entries in the error log? can you share?
> 
> Go to **Help -> Show error log**.

Yes i get an error in the diary,
```
Fri Jan 17 20:22:27 CET 2025
Command 'name.abuchen.portfolio.ui.command.newclient' failed

org.eclipse.core.commands.ExecutionException: Error executing 'name.abuchen.portfolio.ui.command.newclient': java.lang.IllegalArgumentException

	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.execute(HandlerServiceHandler.java:179)

	at org.eclipse.core.commands.Command.executeWithChecks(Command.java:488)

	at org.eclipse.core.commands.ParameterizedCommand.executeWithChecks(ParameterizedCommand.java:485)

	at org.eclipse.e4.core.commands.internal.HandlerServiceImpl.executeHandler(HandlerServiceImpl.java:204)

	at org.eclipse.e4.ui.workbench.renderers.swt.HandledContributionItem.executeItem(HandledContributionItem.java:444)

[...]

PR #5262, opened 2025-12-29
fix currency initialisation for locale with region instead of country

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/4471
Issue https://forum.portfolio-performance.info/t/command-name-abuchen-portfolio-ui-command-newclient-failed/38488

Hello,

The previous fix about this issue (cannot create new file) dealt with empty country, however it seems there are still some particular cases leading to the same issue :

Locale can contain both ISO 3166 country code (2 letters) or UN M.49 area (three digit); while `java.util.Currency.getInstance `only accepts ISO 3166 local, so en_001 (English-World) or en_150 (English-Europe) locale were throwing IllegalArgumentException and no new file could be created.

`Locale.getISOCountries `returns the list of the ISO 3166 countries, so what should be accepted by `java.util.Currency.getInstance`, so we can check if the locale country is in there or not.
For good measure, I have added a try catch too, but I assume it is now redundant with the getISOCountries fix.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/money/CurrencyUnit.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/money/CurrencyUnit.java
@@ -11,6 +11,7 @@ import java.util.MissingResourceException;
 import java.util.NavigableMap;
 import java.util.Objects;
 import java.util.ResourceBundle;
+import java.util.Set;
 import java.util.TreeMap;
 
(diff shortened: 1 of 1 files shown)


buchen:
> For good measure, I have added a try catch too, but I assume it is now redundant with the getISOCountries fix.

Thanks... I should have added that a long time ago. I'll add a commit with some test cases.

PR #5263, opened 2025-12-30
Introduced the concept of skipped items

hporten:
The second attempt at introducing a new import item state for items that cannot be imported but should
not be fully hidden from the user either.

Items with issues (e. g. missing details about fees) previously got skipped entirely. This lacks transparency
for the user. But worse, if a document contains nothing but skipped items the document import lead to
a not-supported-document-type error. With this change, a null-object-style item is introduced. It carries
a "reason" message that gets displayed to the user as a tooltip on a "no entry" icon.

The importers for Deutsche Bank, Ebase and Onvista make use of the new SkippedItem class. The 
 PDFMsgErrorDoNotProcessMissingExchangeRateIfInForex error emitted by the Fintech Group importer
would be another candidate for converting.

The earlier pull request was https://github.com/portfolio-performance/portfolio/pull/5186 Meanwhile, I
also experimented with just having a "skipped" flag in the Item base class. This meant less code. But suddenly,
lots of test failures showed that counting of transaction items becomes more complex. Especially if related
security items also survive the extracted lists.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/deutschebank/DeutscheBankPDFExtractorTest.java
@@ -2138,8 +2138,7 @@ public class DeutscheBankPDFExtractorTest
         assertThat(skipped.getSkipReason(), is(Messages.PDFSkipMissingDetails));
         assertThat(skipped.getDate(), is(LocalDateTime.parse("2020-12-31T00:00")));
         assertThat(skipped.getAmount(), is(Money.of("EUR", Values.Amount.factorize(13.47))));
-        assertThat(skipped.getTypeInformation(), is("Removal"));
-        // assertThat(skipped.getSource(), is("GiroKontoauszug02.txt"));
+        assertThat(skipped.getTypeInformation(), is(AccountTransaction.Type.REMOVAL.toString()));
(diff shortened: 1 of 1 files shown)


hporten:
Here is how the import review dialog looks like with the change applied:
Screenshot_20251230_190900

buchen:
The change looks good to me.

Before merging, I would like to @Nirus2000 to also have a look at it.

After that, I would like to squash a couple of commits (we do not need three for the icons). But we can leave the other commits separate. (To be honest, I did not know if I can force push into the pull request. And even then, for the review, it is clearer what changes I did on top)

About the icons: We recently switched over to using SVG images. Then it is not necessary to provide PNG images in two resolutions anymore. I will update the contributing guidelines accordingly. 

The change from PNG to SVG is an "ongoing" activity - on most platforms and resolutions, there is not really a visible change. Therefore I started with converting a couple icons in the last release to see if I get feedback.

hporten:
> After that, I would like to squash a couple of commits (we do not need three for the icons).

Sure. I merely left the icon commits separate to show traces of fix-up work to reviewers. Last time I tried to update an existing pull request, things went really wrong as I lacked the permissions to fully reset all changes.

> The change from PNG to SVG is an "ongoing" activity - on most platforms and resolutions,

Saw that. Did not want to pace ahead. Thanks for updating the contributor guidelines.

buchen:
merged.

Nirus2000:
In memory of https://github.com/portfolio-performance/portfolio/pull/5512, who simply pushed this pull request into the master branch unfinished.

PR #5267, opened 2025-12-31
sort alphabetically accounts and portofolios in the transaction tables

mierin12:
for inline transactions editing

Issue : https://forum.portfolio-performance.info/t/konten-oder-depotliste-in-auswahlmenus-nicht-alphabetisch-sortiert/38499/3

Hello, this time it is for the inline transaction edition where a combolist of accounts is available and was unsorted.

I was wondering if we need the inactive accounts there ? If not, then I think another way would be to call `client.getActiveAccounts` and `client.getActivePortfolios` which are already sorted. I do not really use those combo list a lot myself, so no strong opinion. Maybe the answer is different for this inline edition combolist of accounts and for the list of available reference accounts from yesterday ?

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/TransactionOwnerListEditingSupport.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/TransactionOwnerListEditingSupport.java
@@ -136,11 +136,13 @@ public class TransactionOwnerListEditingSupport extends ColumnEditingSupport
         {
             String ownerCurrencyCode = account.getCurrencyCode();
             comboBoxItems = client.getAccounts().stream().filter(a -> !a.equals(skipTransfer))
-                            .filter(a -> a.getCurrencyCode().equals(ownerCurrencyCode)).collect(Collectors.toList());
+                            .filter(a -> a.getCurrencyCode().equals(ownerCurrencyCode)).sorted(new Account.ByName())
+                            .collect(Collectors.toList());
         }
         else if (ownerToEdit instanceof Portfolio)
         {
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/TransactionOwnerListEditingSupport.java
```diff
@@ -136,11 +136,13 @@ public final void prepareEditor(Object element)
         {
             String ownerCurrencyCode = account.getCurrencyCode();
             comboBoxItems = client.getAccounts().stream().filter(a -> !a.equals(skipTransfer))
-                            .filter(a -> a.getCurrencyCode().equals(ownerCurrencyCode)).collect(Collectors.toList());
+                            .filter(a -> a.getCurrencyCode().equals(ownerCurrencyCode)).sorted(new Account.ByName())
+                            .collect(Collectors.toList());
```
gamue, line 140:
```suggestion
                            .toList();
```
should be possible to use the newer version. same in L146

buchen: I believe in this case it using toList is okay.

But keep in mind that ```toList``` creates an immutable list while ```.collect(Collectors.toList()``` creates a mutable list. It is a subtle change. I missed it at least once and I had to make a ad hoc bug fix release... :-(

To get rid of the annoying proposals in the IDE, I now use ```.collect(toMutableList())``` from my CollectorsUtils class. This way I remember that I want a mutable list. I am not sure if there is a better way.

Anyway, I will merge this change, because this is about sorting the list.

buchen:
merged.

PR #5280, opened 2025-12-31
Refactorings of csv exporter

gamue:
This is a follow-up of the previous csv-refactorings. I'm currently browsing through the codebase and as this classes/package puzzled me a little bit I thought i try to make it simpler. Couple of those changes are personal preference so open to discussions and change-requests :) 

- the csv-package contains importers/extractors, exporters and config classes which are quite some classes. Having an own `csv.exporter` package clusters the classes better. Those 3 are also independent while the extractor classes and their configs share an interface or inheritance. I tend towards moving the importer/extractor classes into their own package as well, but that would be a follow-up PR to keep this small and easier to review.
- `CSVExporter` had own methods for exporting `AccountTransaction` and `PortfolioTransaction` which shared a lot of code, I extracted code into smaller sub-methods and merged them into a single-method
- Adjusted naming of variable to prevent one-char variables and using more descriptive names
- Split exporting security-prices export into smaller methods to make it easier to read and understand
- Thought: Every class has usually one export use-case but the CSVExporter has two: exporting the transactions and the security-prices, I was wondering if it makes sense to split this class into multiple ones, to comply to "one class per export-case"

I'm not sure why gh detects the CSVExporter as delete & created, I moved it via refactoring-menu. Maybe because of he amount of lines changed. ~~However I did the changes using multiple commits, so it's easier to review (commit-wise) and in case something isn't liked, can be reverted on it's own.~~


Nirus2000:
Hello @gamue 

I'll borrow the words of @buchen ...

> First, it would be super helpful to squash these many commits into one. That makes it much easier to review. Additionally, it is always helpful if you rebase your contributions instead of merging the upstream master into your branch. Can you be so kind and force push the squashed changes into this pull request?

https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#contributing-code

Regards
Alex

gamue:
Hi @Nirus2000 ,
sure can do, if that's easier for you, but don't really understand the background. The branch is quite up-to-date and personally think reviewing commit-wise is easier for refactorings. When using squash-merge there's also no real difference imo.

Nirus2000:
> Hi @Nirus2000 ,
> sure can do, if that's easier for you, but don't really understand the background. The branch is quite up-to-date and personally think reviewing commit-wise is easier for refactorings. When using squash-merge there's also no real difference imo.

Why "easier to review":
When you squash multiple commits into one, the reviewer sees:

- Single unified diff in the "Files changed" tab
- Final state only - no intermediate refactoring steps that get revised later
- Less noise - no "fix typo", "oops forgot this", "revert previous change" commits
- GitHub's UI is optimized for reviewing single commits, not commit-by-commit history

Nirus2000:
Hello @buchen and @gamue 
We might want to reconsider whether the AktienfreundeNetExporter is still needed. I tested it in early 2025 and it no longer worked. We should verify if this exporter is obsolete and can potentially be removed from the codebase. Has anyone successfully used it recently, or should we deprecate/remove it?

Regards
Alex

gamue:
> Why "easier to review":

that's not reflecting my experience, happy to chat about it but might be too much for this thread. Actually working with improvement- or step-commits together with squash-merge since many years at multiple teamsand companies.

- You can configure if you see a "single unified diff" or "commit by commit" inside the files-changed-tab
- Once changes are ok for the reviewer, the "viewed" checkbox can be clicked. With this it's easier to focus on follow-up changes

> We might want to reconsider whether the AktienfreundeNetExporter is still needed.

I've never used it and don't have knowledge around it. When you say you've tested it and it doesn't work, then I agree with rather removing it. 



P.S:: happy new year :)

buchen:
I have rebased and merged the change (tiny reformatting done).

I can't see the individual commits anymore, but because the CSVExporter was both moved and refactored, separate commits could have been easier. Unfortunately, there is no "one size fits all rule". In the end it was also not a problem. The other pull request has some 15 commits that all change the same file. And a merge commit which makes it hard.


@Nirus2000 writes:

> We might want to reconsider whether the AktienfreundeNetExporter is still needed. I tested it in early 2025 and it no longer worked.

Thanks for the hint. I tested it: the import on Aktienfreunde (or Rentablo) supports the default CSV exports from PP:

Bildschirmfoto 2026-01-02 um 19 16 28

PR #4081, opened 2024-06-16
update translation : 'removal' becomes 'withdrawal'

mierin12:
Issue : https://github.com/portfolio-performance/portfolio/issues/3585#issuecomment-1789561488

(no change on the internal variables names)

PR #5283, opened 2026-01-01
Less Java, more English in Securities context menu.

hporten:
The texts sounded a bit odd. The underlying setter function shone through a bit.

The menu items will be in line with the "Duplicate" and "Delete" options this way.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
@@ -2678,21 +2678,21 @@ SecurityMenuSearch4Securities = Search for instruments...
 
 SecurityMenuSell = Sell
 
-SecurityMenuSetMultipleSecurityActive = Set {0} securities active
+SecurityMenuSetMultipleSecurityActive = Activate {0} securities
 
-SecurityMenuSetMultipleSecurityActiveConfirm = Do you really want to set {0} securities active?
+SecurityMenuSetMultipleSecurityActiveConfirm = Do you really want to activate {0} securities?
 
-SecurityMenuSetMultipleSecurityInactive = Set {0} securities inactive
+SecurityMenuSetMultipleSecurityInactive = Deactivate {0} securities
 
-SecurityMenuSetMultipleSecurityInactiveConfirm = Do you really want to set {0} securities inactive?
+SecurityMenuSetMultipleSecurityInactiveConfirm = Do you really want to deactivate {0} securities?
 
-SecurityMenuSetSingleSecurityActive = Set security active
+SecurityMenuSetSingleSecurityActive = Activate security
 
-SecurityMenuSetSingleSecurityActiveConfirm = Do you really want to set the security ''{0}'' active?
+SecurityMenuSetSingleSecurityActiveConfirm = Do you really want to activate the security ''{0}''?
 
-SecurityMenuSetSingleSecurityInactive = Set security inactive
+SecurityMenuSetSingleSecurityInactive = Deactivate security
 
-SecurityMenuSetSingleSecurityInactiveConfirm = Do you really want to set the security ''{0}'' inactive?
+SecurityMenuSetSingleSecurityInactiveConfirm = Do you really want to deactivate the security ''{0}''?
 
 SecurityMenuStockSplit = Stock split...
 


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
```diff
@@ -2678,21 +2678,21 @@ SecurityMenuSearch4Securities = Search for instruments...
 
 SecurityMenuSell = Sell
 
-SecurityMenuSetMultipleSecurityActive = Set {0} securities active
+SecurityMenuSetMultipleSecurityActive = Activate {0} securities
```
pfalcon, line 2681: I apologize, but is English your native language? For full disclosure, mine's not, but even I can detect the fact that "activate security" is not English at all. Maybe Java indeed. You can activate or deactivate a switch, but not a security. However, you can set security active or not.

hporten:
English is not my native language either. So please excuse any wrong judgements I made. But I am very confident that "Set X securities active" is not a good style for an UI label.

Whether the term "active" is a good one is another matter that can be discussed application-wide. Enable/disable would be an alternative. 

To hear another opinion, I asked one of those fashionable tools that has been trained on lots of English text:

> *Check the grammar of "Set 11 securities active"*
> .
> The phrase "Set 11 securities active" is grammatically incorrect. A better phrasing would be either "Set 11 securities to active" or "Activate 11 securities." These alternatives are clearer and more standard in English.
> Correct Options
> 
>    "Set 11 securities to active": This indicates that you are changing the status of the securities.
>    "Activate 11 securities": This clearly states the action of enabling the securities.
> 
> These revisions clarify your intent and improve the overall grammatical structure.

The last alternative is preferable due to the briefness recommended for menu items.  And as a next step, I intended to simply all of these three items and bring them in line with Edit etc.:

    Activate security -> Activate
    Duplicate security -> Duplicate
    Delete security -> Delete

buchen:
I am also not an English native speaker...

Claude code says:
> More natural English - "Activate" and "Deactivate" are standard action verbs, while "Set ... active" is awkward and sounds like programming code (Java setter methods)

ChatGPT tells me:
> the new wording is better and more idiomatic.
>
> “Set {0} securities active”
> Grammatically correct, but sounds mechanical and non-native. A native speaker would expect a verb that directly expresses the action.

Let me see if I can talk to a native speaker. However, at the moment I tend to accept this change

pfalcon:
> However, at the moment I tend to accept this change

It's funny how changes which fix obvious miswordings got ignored for years: https://github.com/portfolio-performance/portfolio/pull/4081 , while changes bringing in confusion are quick to be picked up. Even funnier that random word permutators are used to decide what's "better". Ok, for your amusement, let's permute some more:

---

> What does "activate security" mean?

"Activate security" generally means to turn on or enable a protective system—whether digital or physical—so that it can begin monitoring for and responding to threats.
...

> Oh no, dudes who want to "activate" security deal with "security" as stock, bond, etc. So, what does it mean?

I see—you’re talking about Financial Securities (stocks, bonds, etc.). In that world, when someone talks about "activating" a security, they are usually referring to one of three specific professional actions.

It isn't a common phrase for a casual investor, but here is what it means to the "dudes" in finance:
1. Activating a "Lending" Program (Most Likely)
...
2. "Activating" an Exotic Option (The "Trigger")
...
3. Activating Your Trading Account
...

> None of the above, and you clearly don't get it. For example, if you activate bomb, it's going to explode soon. If you activate alarm, and don't leave in time, alarm will go off on you, and police comes to arrest you. If you activate switch, and you don't know what the switch does, somebody may be hurt (e.g. electrocuted or run over). So, what does "activate security" mean?

In the world of finance, if someone says they are going to "activate a security," they are using the word in its active, dangerous, or mechanical sense—just like your bomb or alarm examples.

While "security" usually means a boring stock or bond, "activating" it changes it from a piece of paper into a live financial weapon. Here is what that looks like in the real world:

1. The "Trap Door" (Knock-In Options)
Some securities are "dormant." They sit there doing nothing until the stock price hits a specific "trigger" or "barrier" price.

...

2. "Arming" the Collateral (Security Interest)
In a big loan deal, "security" refers to the assets (like a factory or a fleet of cars) that the bank can take if you don't pay.

...

3. The "Self-Destruct" (Margin Calls)
If you trade with borrowed money, your broker holds your stocks as "security."

...

hporten:
The context to evaluate terms in is PP itself. It deals with securities but it also brings in the concept of "active"/"inactive" items. Hence one will not necessarily have luck consulting a financial dictionary. Classic English dictionaries state synonyms like "turn on", "set in motion", etc. The authors of the software chose to base the software attribute on that concept. The switch example you cited above fits perfectly actually. So I still do not see any issue.

Alternative words like "archive" may be in order for better general clarity. But I did not want to open that can of worms. To get avoid getting stuck like https://github.com/portfolio-performance/portfolio/pull/4081. ;-) Which I already stated my support for btw. And more voices would help to raise attention to the fact that many users do indeed see the need.

buchen:
Independent of this change but triggered by this discussion, I have merged #4081

mierin12:
Hello, thanks for the 4081 merge, one comment about it, sorry I comment here for visibility but we can continue on 4081, one side effect is I noticed : previously exported csv with "removal" will not be matched anymore during csv importation as withdrawal is expected by the importer now, maybe there is a way to allow both to be matched ?
Capture d’écran 2026-01-03 102354

georgemac-labs:
Hi all, I am a native speaker, so I'll offer my two pence (GBX):

- **activate** has a bit of a different connotation to **set as active**. When I activate something, it suggests it's going to actively start doing something ("activate the defences!"). Whereas "set as active" is about an attribute / a status – like being in the active roster of a sports team.
- But, ironically, I would use **deactivate** as the reverse operation to "set as active". It's less cumbersome and has no confusing connotation.
- And I'd say "set **as** active" is preferred for a boolean state. "Set to" is used for values, like setting the volume to 11, or warp to 9.
- That said, I have considered active, inactive, deactivate, hidden, archive, retired ... I am not satisfied with any of them. This feature cannot be unambiguously explained in a couple of words; nor can I see strong parallels to familiar terms in other apps.
- My vote would probably be for **archive(d)/unarchive**, as those concepts are also now commonplace, and I think the feature fits well enough. Unlike hidden, it is somewhat intuitive that a security in the archive also doesn't get quote updates. Unarchive is probably not in any formal dictionary, but it is easily understood, and that's what matters.
- **retired** is not terrible, but not great either. In non-technical English, it's mostly used in a literal way (for people), not metaphorically. It can be used for superseded equipment. It doesn't feel like a great fit for securities.
- **active/inactive** are also alright, but I find them a little less intuitive

Incidentally, it appears there may be inconsistencies in current terminology? E.g. LabelPaymentsConsolidateRetired uses "Retired" instead of "Inactive". If somebody's working on this, I'd suggest bringing that into line, too.

Feel free to @ me for any questions related to English translations.

adeorato:
> Hi all, I am a native speaker, so I'll offer my two pence (GBX):
> 
>     * **activate** has a bit of a different connotation to **set as active**. When I activate something, it suggests it's going to actively start doing something ("activate the defences!"). Whereas "set as active" is about an attribute / a status – like being in the active roster of a sports team.
> 
>     * But, ironically, I would use **deactivate** as the reverse operation to "set as active". It's less cumbersome and has no confusing connotation.

Thanks for the input. Would enable/disable best describe the actual function? Unfortunately, the interface name would then differ from the code.
As you correctly pointed out, nothing is activated here; rather, certain functions are enabled for this security.
Archived sounds good, but the default status “not archived” sounds strange (at least when translated into my native language).

However, I agree with the issue creator that no CRUD words (create, update, delete, set, etc.) should be used, but rather terms that describe what is to be done rather than how.

pfalcon:
@hporten : 

> The context to evaluate terms in is PP itself. It deals with securities but it also brings in the concept of "active"/"inactive" items. Hence one will not necessarily have luck consulting a financial dictionary.

That's essentially my concern - you're advantaged by having read the proper description of the menu items ("set active/inactive"), and having some experience with PP. But then in a strange twist, you decided that it was written in "Java", and wanted to reword it. I'm not sure if you tried (hard enough) to look at the result from novice user's perspective. And the LLM transcript above tries to show the issues - the phrase "activate security (as in: financial security)" doesn't have a meaning in English, it's essentially a nonsensical phrase. Outside of PP context (and novice users don't have it!), one would need to apply imagination to assign meaning to it. You can read LLM's hallucinations on what it might mean, and the transcript also presents IMHO valid human point on it - "activating things you don't know about can be dangerous". So, yeah, IMHO this change makes these menu items "scary", with expected outcome that people don't use them, then over the years, rare but regular questions on forums, etc. regarding what they mean. YMMV

Btw, there was an attempt to *discuss* terminology improvements before going for them: https://github.com/portfolio-performance/portfolio/issues/3585 . It didn't catch much attention, with the core issue brought up there - funky "Removal" instead of standard "Withdrawal" taking more than 2 years to resolve (thanks @buchen !). But IMHO it's still better approach than applying patchy changes on spot.

georgemac-labs:
> the phrase "activate security (as in: financial security)" doesn't have a meaning in English, it's essentially a nonsensical phrase

"Activate security" is not nonsensical, it's fine. It's just a little less appropriate than "set as active".

> "Activate security" generally means to turn on or enable a protective system—whether digital or physical—so that it can begin monitoring for and responding to threats.

It can mean that, but security is a polyseme – one word; multiple meanings. In those cases, we always infer meaning from context. When you use PP, you have already seen "Security" everywhere to refer to your investments. I don't believe a user is going to suddenly think of the other meaning.

The secondary problem here is that "security" itself is a bit old-school! Normal vocab for Graham, Buffett, Munger... but **(financial) instrument** is used more nowadays. And something like "stock" would be _much_ more intuitive, but unfortunately doesn't have the right scope.

Coming back to the original question, I overlooked enable/disable. That's also OK, but not great – for me it's similar to activate, it fits better for something that works dynamically, like a firewall or spellchecker. I like archive/unarchive because to me securities feel more like objects – like emails.

buchen:
@georgemac-labs writes:
> but **(financial) instrument** is used more nowadays.

If I could go back in time, I would also call the class "Instrument". I already was thinking to introduce an Instrument super class once I need to separate behavior - say when introducing better support for bonds. Anyway... back to the original conversation.

> My vote would probably be for **archive(d)/unarchive**, as those concepts are also now commonplace, and I think the feature fits well enough. Unlike hidden, it is somewhat intuitive that a security in the archive also doesn't get quote updates. Unarchive is probably not in any formal dictionary, but it is easily understood, and that's what matters.

My immediate reaction was: I like "archive". We need to keep this instrument around for historical purposes - archive conveys that message. **I am just wondering: does "archive" somehow imply that the instrument/transactions are immutable?** Because that is (currently) not the case. Once can edit transactions of archived instruments, one could add new ones, etc. It is more about getting them out of the way for normal creation.

georgemac-labs:
> does "archive" somehow imply that the instrument/transactions are immutable

Not strongly, but it could give that impression. However, for me, you could say that about most of the terms: inactive, retired, disabled...

Hidden is the only one without that problem. The downside there is it doesn't imply different treatment (e.g. no quote updates).

As I said initially: I don't see any perfect term. The user will always need an explanation to fully understand what this feature does. It's just a tradeoff question.

buchen:
If we were to rename the thing to "archiving", what would we name the filters:

Bildschirmfoto 2026-01-06 um 09 08 14

"Only active instruments" --> "Only non-archived instruments" sounds strange. Maybe using here "active" is still okay?

In German, the equivalent would be "Nur aktive Instrumente" und "Nur archivierte Instrumente".

hporten:
> If we were to rename the thing to "archiving", what would we name the filters: [...] Maybe using here "active" is still okay?

Good question! I did not think that far when throwing "archive" into the ring. I frankly did not even ever use that menu.

Some [Oracle software](https://docs.oracle.com/en/cloud/saas/marketing/eloqua-user/Help/ProgramBuilder/Tasks/ArchivingPrograms.htm) uses "_Not Archived_" lists. But that sounds a bit clunky when prefixed by "Only". So "active" sounds quite okay.

hporten:
Fwiw, someone raised the same question here: https://ux.stackexchange.com/questions/72545/a-better-name-for-non-archived-entries   "Live" was a term I had on my mind earlier. But it did not convince me compared to "active", yet.

hporten:
Last comment for the day (in UTC+0): the prefix "Only" could be entirely superfluous. But I do not want to open yet another can of worms!

kimmerin:
What about "enabled/disabled"? That term is commonly used in applications like "enabled menu items" or "disabled features".

PR #5289, opened 2026-01-02
Some more API docs

hporten:
Added a bit of documentation here and there while I was studying the extraction parser code.


name.abuchen.portfolio/src/name/abuchen/portfolio/model/Transaction.java
```diff
@@ -485,6 +485,9 @@ else if (unit.getForex() != null
                         }));
     }
 
+    /**
+     * Returns the XXXXXXXXXXXXXXXXXXXXXXXXXXXXXxx
+     */
```
buchen, line 507: ????

hporten: > ????

The reminder to myself apparently was not big enough. Fixed with a new commit.

Issue #5288, opened 2026-01-02
Taxonomy name with at sign (@) causes issue in Statement of Assets in gear icon settings


Alfons1Qto12:
**Describe the bug**
If name of a taxonomy contains an at sign (@), the list of taxonomies behind the gear icon in Statement of Assets does not display the name of the taxonomy correctly. 

**To Reproduce**
1. Create a taxonomy that contains at least one at sign (@), e.g. "Country@Region" or "Peter@Mary@John".
2. Go to Reports Statement of Assets
3. Click on the gear icon in the upper right corner which allows to select Taxonomies and/or Columns.
4. Look at the list of taxonomies and realise that the name of the taxonomy is not displayed correctly.

**Expected behavior**
Name of taxonomy should be displayed correctly (e.g. "Country@Region" or "Peter@Mary@John")

**Observed behavior**
Name of taxonomy is missing the last at sign and the subsequent characters (e.g. "Country" instead of "Country@Region" or "Peter@Mary" instead of "Peter@Mary@John").

(Other special characters seem to work as intended).

**Desktop**
 - OS: Linux
 - Version 0.81.0

mierin12:
Fixed in [release 0.81.1](https://github.com/portfolio-performance/portfolio/releases/tag/0.81.1) by https://github.com/portfolio-performance/portfolio/commit/8212919f9d9bb3de79567320d6d9e99abdf4143a

Alfons1Qto12:
Thanks a lot for fixing this. I can confirm that it works as it should in 0.81.1 on Linux.

PR #5290, opened 2026-01-03
fix taxonomy name with '@' in MenuManager action

mierin12:
Issue : https://github.com/portfolio-performance/portfolio/issues/5288

Hello, I can reproduce the issue of #5288. The proposition of fix is to use the idea of the stackoverflow thread : 
https://stackoverflow.com/questions/77472261/how-to-keep-character-in-swt-menu-item-text
Feels a bit strange, but on Windows it works and I do not notice visual impact from the added tab.
This adds the tab by default, maybe best to add only if there is the @ character ?

With : 
Capture d’écran 2026-01-03 144834

**Before :**

**After :**

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/StatementOfAssetsViewer.java
@@ -1123,7 +1123,7 @@ public class StatementOfAssetsViewer
 
         for (final Taxonomy t : client.getTaxonomies())
         {
-            Action action = new SimpleAction(TextUtil.tooltip(t.getName()), a -> {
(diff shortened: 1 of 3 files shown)


buchen:
Thanks @mierin12 for researching a solution.

Two thoughts:

* This applies to more cases: whenever the user can enter text and that text shows up in a context menu. For example, a filter with an @ will also produce the same problem.
* By always adding the tab, we now always create an empty accelerator for every item regardless if there is an @ in the text. If we use this implementation, I think we should add it only if there is actually a '@' in the text.

While looking through the code, I noticed that the ```DropDown``` does not have the problem. It is extending ```ContributionItem``` and the ```fill``` method has full control of how to create the menu item. It would not need to do the accelerator "magic" (which we anyway do not use).

Let me explore that solution as well.

buchen:
Closing this pull request as I believe that #5293 is a cleaner (but more verbose) fix.

PR #5291, opened 2026-01-03
Improvement: IBFlex handle currency unit mismatches

georgemac-labs:
When using the IBFlex importer, some of my transactions will not import.

Problem:
- Some currencies have major and minor units, such as UK Sterling, with GBX (pence) and GBP (pound, equal to 100 GBX)
- It seems that Interactive Brokers flex queries do not always report the unit that the security is quoted in
- For example, EQQQ on LSE is quoted in GBX, but transactions in the IBKR export are in GBP

Until now, the importer flagged these cases as a currency mismatch and they were not imported. But there is no real mismatch – they just require a fixed-rate conversion between major and minor unit.

This PR attempts to do that, and complete the import. It works correctly with my personal data.

Some minor logic changes were needed, to detect mismatches, and ensure correct handling when GROSS_AMOUNT is set.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
@@ -58,6 +58,7 @@ import name.abuchen.portfolio.money.CurrencyUnit;
 import name.abuchen.portfolio.money.ExchangeRate;
 import name.abuchen.portfolio.money.Money;
 import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.money.impl.FixedExchangeRateProvider;
 import name.abuchen.portfolio.online.QuoteFeed;
 import name.abuchen.portfolio.online.impl.YahooFinanceQuoteFeed;
 import name.abuchen.portfolio.util.Pair;
@@ -611,6 +612,9 @@ public class IBFlexStatementExtractor implements Extractor
             
             portfolioTransaction.setDate(extractDate(element));
 
+            // Set security before amount so that setAmount can detect currency mismatches in all cases
+            portfolioTransaction.setSecurity(this.getOrCreateSecurity(element, true));
+
             // @formatter:off
             // Set amount and check if the element contains the "netCash"
             // attribute. If the element contains only the "cost" attribute, the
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -621,14 +625,14 @@ private class IBFlexStatementExtractorResult
                 Money amount = Money.of(asCurrencyCode(element.getAttribute("currency")), asAmount(element.getAttribute("netCash")));
 
                 setAmount(element, portfolioTransaction.getPortfolioTransaction(), amount);
-                setAmount(element, portfolioTransaction.getAccountTransaction(), amount);
+                portfolioTransaction.getAccountTransaction().setMonetaryAmount(amount);
```
buchen, line 628: I understand this change is not directly related to your change. But, yes, the account transaction should not have the forex gross values therefore it is correct to not call the method.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -938,6 +940,43 @@ private BigDecimal getExchangeRate(Element element, String fromCurrency, String
...
+
+        /**
+         * Returns the exchange rate for currency pairs with a fixed relationship
+         * (e.g. GBX/GBP) using FixedExchangeRateProvider. Handles both directions.
+         *
+         * @param fromCurrency The source currency
+         * @param toCurrency The target currency
+         * @return The exchange rate, or null if not a known fixed-rate pair
+         */
+        private BigDecimal getUnitExchangeRate(String fromCurrency, String toCurrency)
```
buchen, line 959: I propose to rename this method to something like getWellKnownFixedExchangeRate. It is actually not getting the unit exchange rate, but attempts to lookup a fixed exchange rate.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -938,6 +940,43 @@ private BigDecimal getExchangeRate(Element element, String fromCurrency, String
...
+        {
+            var provider = new FixedExchangeRateProvider();
```
buchen, line 961: This will create the provider each and every time - potentially per transaction - therefore I think it makes sense to cache that.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: If I remember correctly, this did seem to be necessary. Without it, I was hitting MsgCheckTransactionMustNotHaveGrossAmount.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: I'm aware, but it's fast on my machine so I kept it simple  :)

georgemac-labs:
@buchen thanks, LMK if you want me to make any changes

buchen:
merged

PR #5293, opened 2026-01-03
Fix '@' not shown in menu items

buchen:
The default ActionContributionItem interprets an @ sign as accelerator separator. Therefore any text that includes an @ will be split and displayed separated with a tab character.

This change introduces a MenuContribution with a simplified implementation for the contribution item. It can only be used to contribute items to the menu.

Issue: #5288
Issue: #5290

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/BookmarkMenu.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/BookmarkMenu.java
@@ -12,6 +12,7 @@ import name.abuchen.portfolio.model.ClientSettings;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.ui.Messages;
 import name.abuchen.portfolio.ui.editor.PortfolioPart;
+import name.abuchen.portfolio.ui.util.action.MenuContribution;
 import name.abuchen.portfolio.ui.views.settings.SettingsView;
 import name.abuchen.portfolio.util.TextUtil;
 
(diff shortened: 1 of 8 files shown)


mierin12:
Hello @buchen, thanks, I did not expect this would be so large. I checked for the taxonomy and clientfiltermenu, for me this works. I think one other location should use the new MenuContribution : `BasicDataSeriesConfigurator`
```
    public void configMenuAboutToShow(IMenuManager manager)
    {
        for (final DataSeries series : selectedSeries)
        {
            Action action = new SimpleAction(series.getLabel(), a -> doDeleteSeries(series));
            action.setChecked(true);
            manager.add(action);
        }

```



this should fix both the @ and & escape in accounts names at this location.

For `ReviewExtractedItemsPage `and `FindQuoteProviderDialog `I am not sure where to look to check the behaviour.
I did not follow why the LabelOnly has to be updated from extending the SimpleAction to MenuContribution, some of them could get a label with a @ ?

buchen:
> I did not follow why the LabelOnly has to be updated from extending the SimpleAction to MenuContribution, some of them could get a label with a @ ?

Not all, but some labels can include customer text and therefore the @ sign. For example, the context menu of a widget on the dashboard. And because ```LabelOnly``` was almost exclusively used for menu entries, the easiest first attempt was to change the super class from ```Action``` to ```ContributionItem```. Only the ```FindQuoteProviderDialog``` was using Actions and therefore I left there a private class as before. And ```SecurityPriceUpdateView``` was adding a read-only image to the toolbar.

buchen:
merged.

PR #5294, opened 2026-01-03
Support Deutsche Bank dividends going to forex account while security is held in base currency 

hporten:
Scenario:

 1. An US-security originally purchased in EUR and held in an EUR-based deposit
 2. Dividend payment is now supposed to go into new USD-based clearing account

Import error:

    AssertionError (Exchange rate of gross value is missing (transaction currency USD and
    security currency EUR) Dec 10, 2025 Dividend USD 18.62 BWX TECHNOLOGIES INC.RG.SH. DL -,01

The exchange rate could be extracted from the document. But why? No currency conversion is required in this case.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java
@@ -423,7 +423,18 @@ public class DeutscheBankPDFExtractor extends AbstractPDFExtractor
                                                                 var gross = rate.convert(rate.getBaseCurrency(),
                                                                                 fxGross);
 
-                                                                checkAndSetGrossUnit(gross, fxGross, t, type.getCurrentContext());
+                                                                // @formatter:off
+                                                                // Even though this is a forex transaction, the security may be
+                                                                // held in the native currency. In that case, the GROSS_VALUE
+                                                                // unit needs to be set. In the foreign currency that is.
+                                                                // @formatter:on
+                                                                if (t.getSecurity().getCurrencyCode()
+                                                                                .equals(t.getCurrencyCode()))
+                                                                    checkAndSetGrossUnit(gross, fxGross, t,
+                                                                                    type.getCurrentContext());
+                                                                else
+                                                                    checkAndSetGrossUnit(fxGross, gross, t,
+                                                                                    type.getCurrentContext());
                                                             }
                                                         }))
 


georgemac-labs:
@hporten looks very similar to the issue I'm trying to resolve for Interactive Brokers with #5292.

My understanding is: the currency conversion _is_ required, because otherwise some performance calculations for that security will be broken. They need to know the value of the dividend in the base currency.

So the importer needs a contemporaneous exchange rate – in this case EUR/USD for 2025-12-10 – and must calculate and store the base currency amount in GROSS_VALUE.

hporten:
@georgemac-labs  Thanks for the quick input. I was also reminded about the Interaktive Brokers issue. Which made me wonder about a regression somewhere deeper in the framework.

But okay. Your explanation makes sense. Especially since I had received dividends in EUR for the longest time. It was the creation of a USD sub-account that suddenly made the payment take another route. A statistic covering all of that is indeed not trivial.

I will try to make the import work.

buchen:
@hporten Indeed, all transactions related to a security (instrument) must have currency conversion to the currency of the instrument. If the instrument is EUR and the dividends only in USD, then the importer complains.

I admit that this an "arbitrary" requirement. It was a simplified assumption and it is in the code base forever. For example [here](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/model/PortfolioTransaction.java#L177-L183) I am assuming and never check that the forex unit is in the currency of the security. 

Of course, it does not have to be that way. We could also admit transactions in other currencies and the convert "on the fly" if needed. The hard part is to find and fix all those places. I fear that the test coverage might not be good enough. I just prompted Claude to find such places - not sure that is helpful.

If we were to start, I would start with the dividend transaction (= account transaction).

buchen:
To keep it somewhere, this is what Claude thinks we need to look at:


### Core Validation - This Enforces the Restriction

  CheckCurrenciesAction.java:130-179 - This is the "gatekeeper" that enforces your current requirement:

  // If transaction currency != security currency, REQUIRES:
  // 1. GROSS_VALUE unit MUST exist
  // 2. Its forex MUST be in security currency
  // 3. ALL other units' forex MUST be in security currency

```
  String forex = grossValue.get().getForex() != null ? grossValue.get().getForex().getCurrencyCode() : null;
  if (!securityCurrency.equals(forex))
      return new Status(Status.Code.ERROR, MessageFormat.format(Messages.MsgCheckGrossValueUnitForexMismatch,
                      forex, securityCurrency));
```

  This is the primary place you'll need to change to relax the restriction.

  ### PortfolioTransaction - Multiple Currency Methods

  Beyond your example, there are two more methods with the same assumption:

  PortfolioTransaction.java:112-134 - getMonetaryAmount(CurrencyConverter)
```
  else if (getSecurity().getCurrencyCode().equals(converter.getTermCurrency()))
  {
      Optional<Unit> grossValue = getUnit(Unit.Type.GROSS_VALUE);
      BigDecimal exchangeRate = grossValue.isPresent() ? grossValue.get().getExchangeRate()
                      : converter.getRate(getDateTime(), getCurrencyCode()).getValue();
      // Uses exchange rate WITHOUT verifying forex currency matches
  }
```
  PortfolioTransaction.java:211-229 - getGrossPricePerShare(CurrencyConverter)
```
  // Comment shows awareness: "Because the gross value is calculated with taxes 
  // (which might be in transaction currency and not in security currency)"
  long grossPrice = BigDecimal.valueOf(getGrossValue(converter).getAmount())...
```

### Performance Calculations - Capital Gains

  CapitalGainsCalculation.java:131-150, 306-342 - Forex gain calculations assume security currency:

```
  if (!termCurrency.equals(t.getSecurity().getCurrencyCode()))
  {
      // Assumes security currency IS the forex currency for gains
      CurrencyConverter convert2forex = converter.with(t.getSecurity().getCurrencyCode());
      Money forex = convert2forex.convert(item.date, Money.of(termCurrency, start));
      Money back = forex.with(converter.at(t.getDateTime()));
      forexGain = back.getAmount() - start;
  }
```

### UI Transaction Models - Reading & Writing

  AbstractSecurityTransactionModel.java:118-190 - The transaction editor:

  Reading (line 118-142):
```
  case GROSS_VALUE:
      this.grossValue = unit.getForex().getAmount();  // <-- ASSUMES SECURITY CURRENCY
      break;
  case FEE:
      if (unit.getForex() != null)
          this.forexFees += unit.getForex().getAmount();  // <-- ASSUMES SECURITY CURRENCY
```
  Writing (line 169-190)
```
  if (forexFees != 0)
      transaction.addUnit(new Transaction.Unit(Transaction.Unit.Type.FEE,
          Money.of(getTransactionCurrencyCode(), ...),
          Money.of(getSecurityCurrencyCode(), forexFees),  // <-- CREATES IN SECURITY CURRENCY
          exchangeRate));
```
  Similar patterns in AccountTransactionModel.java:245-265.

 ### Import/Export Infrastructure

  CSV Importer (BaseCSVExtractor.java:228-239):
```
  var forex = Money.of(transaction.getSecurity().getCurrencyCode(),  // <-- SECURITY CURRENCY
      Math.round(exchangeRate.multiply(BigDecimal.valueOf(grossValue.getAmount())).doubleValue()));
  transaction.addUnit(new Unit(Unit.Type.GROSS_VALUE, grossValue, forex, exchangeRate));
```
  Data Repair (ClientFactory.java:1585-1629 - fixGrossValueUnit):
```  Money updatedGrossValue = Money.of(grossValueUnit.getForex().getCurrencyCode(),  // <-- USES EXISTING FOREX CURRENCY
      BigDecimal.valueOf(calculatedGrossValue.getAmount())
          .divide(grossValueUnit.getExchangeRate(), Values.MC)...);
```

  PDF Extractors (ExtractorUtils.java:599-606 - fixGrossValueUnit):
```
    var caculatedGrossValue = Money.of(grossUnit.getForex().getCurrencyCode(),  // <-- USES EXISTING
      BigDecimal.valueOf(expectedGross.getAmount())
          .divide(grossUnit.getExchangeRate(), Values.MC)...);
```

### UI Display Code

  SecuritiesChart.java:639-652, 1583-1589 - Charts assume forex = security currency for display:
```
  Optional<Unit> grossValue = t.getUnit(Unit.Type.GROSS_VALUE);
  long gross = grossValue.isPresent() ? grossValue.get().getForex().getAmount() : t.getGrossValueAmount();
  String currency = grossValue.isPresent() ? grossValue.get().getForex().getCurrencyCode() : t.getCurrencyCode();
```

buchen:
@hporten I guess I can cherry-pick 4b37499762884c4321e788e203027ad861025ef1 regardless of this discussion here? (allowing dividends to be in an arbitrary other currency is a big one, but trailing spaces could be fixed anyway).

hporten:
> @hporten I guess I can cherry-pick [4b37499](https://github.com/portfolio-performance/portfolio/commit/4b37499762884c4321e788e203027ad861025ef1) regardless of this discussion here? 

Sure. But the commit currently references the Dividende11.txt test input that is part of this commit.

How about you still give me this weekend to find the real cause of the issue?

buchen:
> How about you still give me this weekend to find the real cause of the issue?

Sure.

But I am not sure what you mean by "real root cause"?

My understanding is: Clearly PP currently does not support a scenario where the instrument/security is configured in EUR (because it was traded on a European exchange) and the dividends are paid in USD directly on a USD account. In such a case, the PDF does not contain an exchange rate (why would have one - it is USD paid to an USD account), but PP requires there to be an exchange rage (because it expects a gross value in with an exchange rate of USD -> EUR). The importer cannot create a transaction that PP will accept. (Obviously there are edge cases: some documents include local taxes paid in EUR and yet again include an exchange rate).

In my mind, the solution is for PP to relax the requirement that the dividend payments of an instrument must either be in the currency of the instrument or must contain the exchange rate to the currency of the instrument. At runtime, we can always use the currency conversion if needed. The only thing is that this is not a trivial implementation, because it is not so easy to say which places in the code assume the currency.

hporten:
> My understanding is: Clearly PP currently does not support a scenario where the instrument/security is configured in
> EUR (because it was traded on a European exchange) and the dividends are paid in USD directly on a USD account. In
> such a case, the PDF does not contain an exchange rate (why would have one - it is USD paid to an USD account),

Surprisingly, there *is* an exchange rate still stated on the document. Maybe for the calculation of the creditable witholding tax at the bottom. The error message lead us on a wrong track. Turns out the code already parses the exchange rate and even tries to set the GROSS_VALUE unit. That gets rejected with an exception because of the currency mixup . That real error was invisible as optionalOneOf() swallows exceptions silently. 

The fix is simpler than anticipated. I will push it here once the patch got cleaned up a bit.

hporten:
@buchen @georgemac-labs The pull request is ready for review.

The commits can still be squashed, of course. I left them separate to a better understanding for the fixing process. Especially, since 9c7e1d81eec2ae5a3e850bac2a32dfb42dcdf56c is not strictly required to make the import work. Nevertheless, it fixes intermediate values that may puzzle anyone debugging the code. And after changes to fixGrossValue() they may even escape to the outside.

PR #5295, opened 2026-01-03
Fixed a seemingly conflicting documentation of the gross value.

hporten:
The first overload said "including taxes", the second "before taxes are applied". This appears contradictory. At least due to the fact that the extra costs can be added or subtracted depending on the type of transaction. Therfore, we just sync the documentation.


buchen:
Thanks @hporten. That is an inconsistency. And the "before taxes and fees are applied" is the right wording from my point of view. 

Why did you make mark the pr as draft?

hporten:
> Why did you make mark the pr as draft?

Due to muscle memory. Sorry for the mistake. There is another draft pull request of mine that I was working on in parallel.

Thanks for the review&merge!

PR #5296, opened 2026-01-04
Modify Bank Norwegian PDF-Importer, add corrections

ZfT2:
Some extensions / corrections:

- use value date
- switches deposit/removal: "Bezahlung von "  in account statement is deposit (foreign account to Bank Norwegian account), "Zahlung an " is removal (Bank Norwegian to foreign account)
- added notes

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NordaxBankABPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NordaxBankABPDFExtractor.java
@@ -1,5 +1,7 @@
 package name.abuchen.portfolio.datatransfer.pdf;
 
+import static name.abuchen.portfolio.util.TextUtil.concatenate;
+
 import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Block;
 import name.abuchen.portfolio.datatransfer.pdf.PDFParser.DocumentType;
 import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Transaction;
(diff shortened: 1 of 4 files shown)


gamue:
Thanks for the contribution and changes, I'm a little bit puzzled about the transaction-types being wrong as the discussion inside #5233 tell it otherwise. But I guess you're having an own account there and checked it with your own files?

ZfT2:
> Thanks for the contribution and changes, I'm a little bit puzzled about the transaction-types being wrong as the discussion inside #5233 tell it otherwise. But I guess you're having an own account there and checked it with your own files?

Yes, I have an own account there and so I'm very sure that it is correct  :)
The comment [here ](https://github.com/portfolio-performance/portfolio/issues/5233#issuecomment-3674581128) tells the same ;)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NordaxBankABPDFExtractor.java
```diff
@@ -119,19 +121,30 @@ private void addAccountStatementTransaction()
...
+                            t.setNote(v.get("note"));
+                        })
+
+                        .section("note2").optional() //
+                        .find("^[\\d]{2}\\.[\\d]{2}\\.[\\d]{4} [\\d]{2}\\.[\\d]{2}\\.[\\d]{4} .*$")
+                        .match("^.*$") //
+                        .match("^(?<note2>.*)$") //
+                        .assign((t, v) -> {
+                            if (!v.get("note2").startsWith("Bank Norwegian, "))
+                                t.setNote(concatenate(t.getNote(), v.get("note2"), " "));
```
gamue, line 147: (suggestion) might be nicer to replace multiple spaces with a single one and use trim

```suggestion
                                t.setNote(concatenate(t.getNote(), v.get("note2"), " ").replace("//s+"," ").trim());
```

ZfT2: hmm I generally agree but what if there are really multiple spaces in the user purpose? I think the importer should not modify the original purpose.
But we can maybe check if the original purpose (already) ends with a space and if, then don't add another one..

gamue: agree and as stated, just a minor suggestion. Think as you'll be an user of the importer yourself, you also have requirements and handy features more in mind ;)

buchen: I would also think that a ```trim``` is enough (which would avoid the duplicate space).

buchen:
FXI, I created #5311 that adds the trimming to the concatenate.

ZfT2:
> FXI, I created #5311 that adds the trimming to the concatenate.

Sounds good, so I will adjust the tests if #5311 is merged.

buchen:
@ZfT2 FYI, I have rebased the pull request on top of the changes to concatenate and force pushed it into the pull request.

PR #5297, opened 2026-01-04
Support both old 'removal' and new 'withdrawal' label in CSV import

buchen:
Issue: #4081

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
@@ -56,6 +56,7 @@ import org.apache.commons.csv.CSVRecord;
 import name.abuchen.portfolio.Messages;
 import name.abuchen.portfolio.PortfolioLog;
 import name.abuchen.portfolio.datatransfer.Extractor.Item;
+import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.util.Isin;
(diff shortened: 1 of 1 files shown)


buchen:
@mierin12 writes [here](https://github.com/portfolio-performance/portfolio/pull/5283#issuecomment-3706932449):

> Hello, thanks for the 4081 merge, one comment about it, sorry I comment here for visibility but we can continue on 4081, one side effect is I noticed : previously exported csv with "removal" will not be matched anymore during csv importation as withdrawal is expected by the importer now, maybe there is a way to allow both to be matched ?

The matching supports regular expressions and therefore both parameters could be made available. However, the default mapping is automatically derived from the enumeration. I have now drafted one idea how to fix this. It is super ugly, but we could keep it in the code for a year or so.

buchen:
This is then how it looks like:

Bildschirmfoto 2026-01-04 um 09 42 19

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVImporter.java
```diff
@@ -657,7 +658,17 @@ public EnumMapFormat(Class<M> enumType)
         {
             enumMap = new EnumMap<>(enumType);
             for (M element : enumType.getEnumConstants())
-                enumMap.put(element, element.toString());
+            {
+                if (element instanceof AccountTransaction.Type && element == AccountTransaction.Type.REMOVAL
+                                && "en".equals(Locale.getDefault().getLanguage()))
```
buchen: I am wondering if that is the right check. English is the fallback translation, but it will be used for other language as well. Therefore the test ```"en".equals(Locale.getDefault().getLanguage())``` is too limited. Maybe we really check for the label

mierin12: Hello Andreas, apologizes for the delay.
For me this works. What do you mean it will be used by other language ? I do not see the case where the English language check is too limited.
But I am also not super aware of all csv import features : while checking this, I only just now discovered that the "Mapped" column can be edited by the user.

buchen: What I mean is this: the language tag might say "fi" (=Finnish). But because we do  not have a Finnish translation, it fallsback to English. Therefore my thinking is that the check for the language tag is not good enough.

buchen:
I have also improved the column labels and added an icon to indicate that the value in the right column can be edited.

Bildschirmfoto 2026-01-10 um 09 19 56

Issue #4238, opened 2024-09-19
Create a AccountTransferItem (not a TransactionItem).


Nirus2000:
@MonkeySon writes:

> Bigbank offers overnight accounts and fixed-term accounts. This transaction was from the overnight to the fixed-term account, therefore "Interne Belastung". I replaced the Transaction Type with TRANSFER_OUT and expected a selection GUI which lets me choose accounts for the IN/OUT transaction. I intend to use two separate PP accounts for each Bigbank account.

There UI has some (limited) support to import account transfers.
As @Nirus2000 pointed out, you would need to create a AccountTransferItem (not a TransactionItem).
And there are no check for matching currencies and other things.
And there is no option to convert it to a deposit/withdrawal if one does not want it as a transfer.

So far our strategy for the "cash transfers" was:
* Create a cash deposit or removal (withdrawal) transaction

Why not a transfer transactions? Because
* the user might not want to maintain the other account in PP (let's say your regular payroll account)
* the user might import the documents from the other account/bank as well (which will create the transaction again). 

But @Nirus2000 and me already discussed that it makes sense to provide more options *after* the import:
* Right-click to convert any deposit/withdrawal to a transfer transaction
* Select two transactions (a pair of deposit and withdrawal) and "join" them to a transfer
* Run some kind of heuristic after the import to suggest a transfer (account at the same bank, same amount in a transaction).

For this particular change, I propose to create a deposit/withdrawal transaction instead.

_Originally posted by @buchen in https://github.com/portfolio-performance/portfolio/issues/3894#issuecomment-2036206552_

PR #5302, opened 2026-01-04
Add new Austrian Anadi PDF-Importer

MK-42:
For importing account-statements from Austrian Anadi Bank AG.

Please note that this is my first contribution to this repository. Feel free to point out any misuse of the best-practices!

I'm not really sure how to handle the simultanous transfers/deposits from fixed deposit accounts while making an interest booking at the same time. I'm now parsing the same line twice, once for the deposit and once for the interest. That does not really feel correct.

Looking forward for your feedback

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
@@ -3,8 +3,10 @@ package name.abuchen.portfolio.datatransfer.pdf.austriananadibank;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.deposit;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasAmount;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasDate;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasFees;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasGrossValue;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasNote;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasShares;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasSource;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasTaxes;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.interest;
@@ -14,6 +16,7 @@ import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countAc
 import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countBuySell;
 import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countItemsWithFailureMessage;
 import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countSecurities;
+import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countSkippedItems;
 import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -47,17 +50,21 @@ public class AustrianAnadiBankPDFExtractorTest
         assertThat(countAccountTransactions(results), is(3L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
         assertThat(results.size(), is(3));
         new AssertImportActions().check(results, "EUR");
 
         // assert transaction
         assertThat(results, hasItem(deposit(hasDate("2025-05-14"), hasAmount("EUR", 2500.00), //
-                        hasSource("Kontoauszug01.txt"))));
+                        hasSource("Kontoauszug01.txt"), hasNote("IBAN: TT40 1705 1256 3324 8502"))));
+
+        // assert transaction
         assertThat(results, hasItem(removal(hasDate("2025-05-15"), hasAmount("EUR", 2500.00), //
-                        hasSource("Kontoauszug01.txt"), //
-                        hasNote("ONLINE-FESTGELD / KONTO 12345555555"))));
+                        hasSource("Kontoauszug01.txt"), hasNote("ONLINE-FESTGELD / KONTO 12345555555"))));
+
+        // assert transaction
         assertThat(results, hasItem(deposit(hasDate("2025-05-19"), hasAmount("EUR", 8320.00), //
-                        hasSource("Kontoauszug01.txt"))));
+                        hasSource("Kontoauszug01.txt"), hasNote("IBAN: ur98 0202 5368 2910 8017"))));
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
```diff
@@ -0,0 +1,177 @@
...
+        assertThat(results.size(), is(3));
+        new AssertImportActions().check(results, "EUR");
+
+        // assert transaction
+        assertThat(results, hasItem(deposit( //
+                        hasDate("2025-05-14"), hasShares(0), //
+                        hasSource("Kontoauszug01.txt"), //
+                        hasNote(null), //
+                        hasAmount("EUR", 2500.00), hasGrossValue("EUR", 2500.00), //
+                        hasTaxes("EUR", (0.00)), hasFees("EUR", 0.00))));
```
gamue: usually just a interest-type is having fees & taxes, so you can skip those checks. Compare it for example with other test cases like the `SuresseDirektBankPDFExtractorTest`

your taxes are also having unnecessary braces, might be copy&paste

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
```diff
@@ -0,0 +1,177 @@
...
+                        hasSource("Kontoauszug02.txt"), //
+                        hasNote(null), //
+                        hasAmount("EUR", 2000.00), hasGrossValue("EUR", 2000.00), //
+                        hasTaxes("EUR", (0.00)), hasFees("EUR", 0.00))));
+        assertThat(results, hasItem(interest( //
+                        hasDate("2025-12-31"), hasShares(0), //
+                        hasSource("Kontoauszug02.txt"), //
+                        hasNote("Habenzinsen"), //
+                        hasAmount("EUR", 123.91), hasGrossValue("EUR", 165.21), //
+                        hasTaxes("EUR", (41.30)), hasFees("EUR", 0.00))));
```
gamue:
```suggestion
                        hasTaxes("EUR", 41.30), hasFees("EUR", 0.00))));
```
the braces aren't necessary

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AustrianAnadiBankPDFExtractor.java
```diff
@@ -0,0 +1,252 @@
...
+                                                .attributes("date", "amount", "account", "interest", "tax") //
+                                                .match("^(?i)(?<date>[\\d]{1,2}\\.[\\d]{1,2}) ABRECHNUNG ZU KONTO (?<account>[\\d]+) [\\d]{1,2}\\.[\\d]{1,2} [\\.,\\d]+$") //
+                                                .match("^(?i)VERANLAGUNGSBETRAG[^\\w]*(?<amount>[\\.,\\d]+)$") //
+                                                .find("^Habenzinsen$") //
+                                                .match("^AB [\\d]{4}-[\\d]{2}-[\\d]{2} .* (?<interest>[\\.,\\d]+)$") //
+                                                .match("^(?i)KESt[^\\w]*(?<tax>[\\.,\\d]+)-$")
+                                                .assign((t, v) -> { //
+                                                        t.setDateTime(asDate(v.get("date") + "." + v.get("year")));
+                                                        t.setCurrencyCode(v.get("currency"));
+                                                        t.setType(AccountTransaction.Type.DEPOSIT);
```
gamue: that's already set in L65

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AustrianAnadiBankPDFExtractor.java
```diff
@@ -0,0 +1,252 @@
...
+        addSubAccountInterestTransaction();
+    }
+
+    @Override
+    public String getLabel()
+    {
+        return "Austrian Anadi Bank AG";
+    }
+
+    private void addDepositRemoveAccountTransaction()
```
gamue: you're also setting Interest in this method, would suggest addAccountStatementTransactions as method name

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AustrianAnadiBankPDFExtractor.java
```diff
@@ -0,0 +1,252 @@
...
+                                                .attributes("date", "amount", "account", "interest", "tax") //
+                                                .match("^(?i)(?<date>[\\d]{1,2}\\.[\\d]{1,2}) ABRECHNUNG ZU KONTO (?<account>[\\d]+) [\\d]{1,2}\\.[\\d]{1,2} [\\.,\\d]+$") //
+                                                .match("^(?i)VERANLAGUNGSBETRAG[^\\w]*(?<amount>[\\.,\\d]+)$") //
+                                                .find("^Habenzinsen$") //
+                                                .match("^AB [\\d]{4}-[\\d]{2}-[\\d]{2} .* (?<interest>[\\.,\\d]+)$") //
+                                                .match("^(?i)KESt[^\\w]*(?<tax>[\\.,\\d]+)-$")
+                                                .assign((t, v) -> { //
+                                                        t.setDateTime(asDate(v.get("date") + "." + v.get("year")));
+                                                        t.setCurrencyCode(v.get("currency"));
+                                                        t.setType(AccountTransaction.Type.INTEREST);
```
gamue: as you're just creating interest events you can set this in L211 directly

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
MK-42: Thanks. that was indeed copy&paste

MK-42:
Thanks for your input. I have adressed your suggestions in a followup commit

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/austriananadibank/AustrianAnadiBankPDFExtractorTest.java
```diff
@@ -0,0 +1,177 @@
...
+                        hasNote("Habenzinsen ABRECHNUNG ZU KONTO 00123456678"), //
+                        hasAmount("EUR", 22.75), hasGrossValue("EUR", 30.33), //
+                        hasTaxes("EUR", (7.58)), hasFees("EUR", 0.00))));
+        assertThat(results, hasItem(removal( //
+                        hasDate("2025-11-24"), hasShares(0), //
+                        hasSource("Kontoauszug03.txt"), //
+                        hasNote(null), //
+                        hasAmount("EUR", 2784.00), hasGrossValue("EUR", 2784.00), //
+                        hasTaxes("EUR", (0.00)), hasFees("EUR", 0.00))));
+    }
```
Nirus2000, line 156: Hello,...
 great... that looks better already.

These are all normal transactions that do not involve gross amounts, taxes, or fees, so all tests need to be corrected. The formatting of the source code also needs to be revised.
Please take another look at https://github.com/portfolio-performance/portfolio/blob/78a72d6dfc56a9743252f49fc7ca9aa6eecdd2a2/name.abuchen.portfolio. tests/src/name/abuchen/portfolio/datatransfer/pdf/scalablecapital/ScalableCapitalPDFExtractorTest.java#L1267-L1324.

Only interest transactions have gross, net, taxes, and fees.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AustrianAnadiBankPDFExtractor.java
```diff
@@ -0,0 +1,248 @@
...
+                                                }) //
+                        ) //
+
+                        .wrap(t -> {
+                            if (t.getCurrencyCode() != null && t.getAmount() != 0)
+                                return new TransactionItem(t);
+                                
+                            return null;
+                        });
+    }
```
Nirus2000, line 228: This division of bookings will always cause problems in the future.
Please take another look at the source code for  https://github.com/portfolio-performance/portfolio/blob/78a72d6dfc56a9743252f49fc7ca9aa6eecdd2a2/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor. java#L377-L507 and it is best to adopt it. 
The taxes for the interest are then via https://github.com/portfolio-performance/portfolio/blob/78a72d6dfc56a9743252f49fc7ca9aa6eecdd2a2/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.

MK-42:
@Nirus2000 Thanks for your feedback and pointing me to the scalable importer as a reference. I tried to address your comments. Unfortunately, I rebased my local repo on the current master and had to force-push. I kept the commit history to make reviewing easier - will squash those before merging if you like.

I had to use negative-lookaheads in the regexes for deposit/removal to not match the transactions when a sub-account "Festgeldkonto" is closed. I could not find an example for that in the codebase so i wasn't able to adapt to standards. Are negative-lookaheads okay, or should we match that stuff in another way?

Would it be possible to create new accounts when a "Festgeldkonto" is opened like the creation of securities? In that case I would make a transfer-transaction rather than a removal/deposit when opening a "Festgeldkonto". Or don't we do that in portfolio-performance?

Additionally I'm not really proud of matching the sub-account-closure block twice, once for the deposit, once for the interest. is there a better way to do that?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AustrianAnadiBankPDFExtractor.java
MK-42: I wasn't too happy with that solution either.... I tried with negative lookaheads now..

MK-42:
Remark: see edit below, this is probably obsolete...

I'm second guessing the interest transactions when fixed-deposit-accounts are closed.
there is no account statement for the fixed-deposit-accounts that shows the interest booking there, but the interest is directly booked on the main account when the fixed-deposit-account expires.
might the interest calculation show wrong percentages due to that? i'm currently "removing" the money from the main account (and expect the user to create the fixed-deposit-account as a seperate account).
the interest is booked on the main account while the money was not in there during the duration of the fixed-deposit-account. will that lead to wrong calculations in the reports?

would it be better to not book the retrieval/deposit when fixed-deposit-accounts are opened/closed? we would then show wrong balances if a user wants to know how much is on the main account "Tagesgeldkonto".

Edit: I found #4238 from which I would assume the current handling in this PR is state of the art for portfolio performance.

MK-42:
@Nirus2000 I would be happy to integrate your next round of feedback to this PR.

I have some more import changes staged in my local repo but would like to understand your way of doing things before creating additional PRs to reduce review load on your side

Nirus2000:
Adding:

https://forum.portfolio-performance.info/t/pdf-importer-von-austrian-anadi-bank-ag/38774

PR #5303, opened 2026-01-04
Improvement: IBFlex stricter security matching

### LINKED ISSUE #4700: Import IBKR, multiple currencies - the same security, cash
**Describe the bug**
The IBKR XML report contains both GOOG (USD) and ABEC (EUR), NVDA (USD) and NVD (EUR). I imported the file twice - First for EUR transactions. And second for USD transactions. During the first import, securities ABEC and NVD are created.

During the second import, securities GOOG and NVDA are not imported, because it is assumed that they are duplicates. The related transactions are not imported also.

In the import window, there is a good error message, but it is hidden because there are multiple warnings (like "transaction X in EUR/USD will not be imported to account USD/EUR"). Good, the positions are open and it was easy to notice that number of shares is wrong.

**Expected behavior**
During the second import there is appropriate error message.
Manually I am able to duplicated securities ABEC and VND.
Then I adjusted currency and symbol:
```
ABEC (EUR) to GOOG (USD)
NVD (EUR) to NVDA (USD)
```
After this manual fix, all transactions was imported properly.

**Screenshots**
Here extract from IBKR XML file:
```
<SecurityInfo currency="EUR" assetCategory="STK" subCategory="COMMON" symbol="ABEC" description="ALPHABET INC-CL C" conid="(digits 1)" securityID="US02079K1079" securityIDType="ISIN" cusip="" isin="US02079K1079" figi="(letters 1)" listingExchange="IBIS" underlyingConid="" underlyingSymbol="ABEC" underlyingSecurityID="" underlyingListingExchange="" issuer="" issuerCountryCode="US" multiplier="1" strike="" expiry="" putCall="" principalAdjustFactor="" maturity="" issueDate="" underlyingCategory="" settlementPolicyMethod="" code="" serialNumber="" deliveryType="" commodityType="" fineness="0.0" weight="0.0" />
<SecurityInfo currency="USD" assetCategory="STK" subCategory="COMMON" symbol="GOOG" description="ALPHABET INC-CL C" conid="(digits 2)" securityID="US02079K1079" securityIDType="ISIN" cusip="02079K107" isin="US02079K1079" figi="(letters 2)" listingExchange="NASDAQ" underlyingConid="" underlyingSymbol="GOOG" underlyingSecurityID="" underlyingListingExchange="" issuer="" issuerCountryCode="US" multiplier="1" strike="" expiry="" putCall="" principalAdjustFactor="" maturity="" issueDate="" underlyingCategory="" settlementPolicyMethod="" code="" serialNumber="" deliveryType="" commodityType="" fineness="0.0" weight="0.0" />

<SecurityInfo currency="EUR" assetCategory="STK" subCategory="COMMON" symbol="NVD" description="NVIDIA CORP" conid="(digits 3)" securityID="US67066G1040" securityIDType="ISIN" cusip="" isin="US67066G1040" figi="(letters 3)" listingExchange="IBIS" underlyingConid="" underlyingSymbol="NVD" underlyingSecurityID="" underlyingListingExchange="" issuer="" issuerCountryCode="US" multiplier="1" strike="" expiry="" putCall="" principalAdjustFactor="" maturity="" issueDate="" underlyingCategory="" settlementPolicyMethod="" code="" serialNumber="" deliveryType="" commodityType="" fineness="0.0" weight="0.0" />
<SecurityInfo currency="USD" assetCategory="STK" subCategory="COMMON" symbol="NVDA" description="NVIDIA CORP" conid="(digits 4)" securityID="US67066G1040" securityIDType="ISIN" cusip="67066G104" isin="US67066G1040" figi="(letters 4)" listingExchange="NASDAQ" underlyingConid="" underlyingSymbol="NVDA" underlyingSecurityID="" underlyingListingExchange="" issuer="" issuerCountryCode="US" multiplier="1" strike="" expiry="" putCall="" principalAdjustFactor="" maturity="" issueDate="" underlyingCategory="" settlementPolicyMethod="" code="" serialNumber="" deliveryType="" commodityType="" fineness="0.0" weight="0.0" />
```

 - OS: Linux, Debian 12
 - Portfolio Performance Version 0.76.1

georgemac-labs:
**Note: depends on #5292**

When importing from IBKR, I have had some issues with transactions being matched to the wrong securities.

The importer currently starts with CONID, which is good, as it’s closest to universal and unambiguous. However, if that fails, it falls back to ISIN, and then symbol, which are ambiguous identifiers. The following scenarios are considered a match:
- Same ISIN, even if there’s a currency mismatch (this looks like a different trading line of the security e.g. EQQQ.F vs EQQQ.L)
- Same symbol, even if there’s an ISIN mismatch (this looks like a completely different instrument!)
- Same symbol, even if there’s a currency mismatch (could be either of the above)

This PR tightens the rules as follows:
- If an ISIN is stored for a security, it must match the import transaction. If it doesn’t, it’s not a match, and a new security should be created.
- If the transaction is a trade, currency must match the stored security. If it doesn’t, it’s not a match, and a new security should be created. Caveat: this rule does not apply for dividends, as those are not always paid in the listing currency.
- The currency checks build on previous PRs[^1] and recognise that the currency does match even if there is a mix of major and minor units (important because e.g. IB Flex reports can use GBP values for GBX listed securities)

Note: I amended the existing unit test, because it contained a trade and a dividend payment where the symbols matched but the CONIDs did not. I traced the origin of the test data and it seems the dividend payment is copied from a different security: https://github.com/portfolio-performance/portfolio/pull/812#issuecomment-340785116, so it doesn't tell us anything about real IBKR data. I only have my personal data to go on, but it's almost 9000 records, and I have zero cases where a CashTransaction has a different CONID from the associated Trade. Thus: I'm working with the assumption that they should always match.

These rules make sense to me, but I may have overlooked some facts and – as always – I’m happy to hear feedback.

Closes: #4700

[^1]: #5291 
#5292


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -1199,11 +1220,13 @@ public void addError(Exception e)
          * Looks up a Security in the model or creates a new one if it does not yet exist.
          * It uses the IB ContractID (conID) for the WKN, tries to degrade if conID or ISIN are not available.
          *
-         * @param element   The XML element containing information about the security.
-         * @param doCreate  A flag indicating whether to create a new Security if not found.
-         * @return          The found or created Security object.
+         * @param element              The XML element containing information about the security.
+         * @param doCreate             A flag indicating whether to create a new Security if not found.
+         * @param strictCurrencyMatch  If true (for trades), only match on ISIN if currency also matches.
+         *                             If false (for dividends), allow ISIN match regardless of currency.
```
buchen, line 1226: What is the rationale between separating the instrument matching whether it is a trade (purchase, sale) or a dividend transactions?

Wouldn't we get mismatches? Say a purchase creates a new security (strict match) but the corresponding dividend than is matched to the previous instrument (because it is not strictly matching).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -1276,18 +1300,36 @@ private Security getOrCreateSecurity(Element element, boolean doCreate)
                     return security;
 
                 if (!isin.isEmpty() && isin.equals(security.getIsin()))
-                    if (currency.equals(security.getCurrencyCode()))
+                    if (isCurrencyCompatible(currency, security.getCurrencyCode()))
```
buchen, line 1303: I get that point (and I understand this has nothing to do with the strict mode). I am just wondering if we take the use of the FixedEchangeRateProvider a little to far. The user could still have a DEM instrument, but I am not sure if he wants to use that instead of the creating a new EUR instrument. Maybe GBX and GBP is a special case because apparently it used interchangeably.

buchen (commented):
let's wait for the other pull request to conclude - this one is independent but sits on top of the other pull request.

The checks wether CONID or ISIN conflict if the ticker is present make sense to me.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: I will need to look at this, but I'd say the data modelling is fundamentally a bit off. GBP and GBX are not two currencies – they are two units of the same, like Euros and Euro cents. Their relationship is quite different from DEM and EUR. But I didn't even consider opening that can of worms!

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: Trades are treated differently because the currency should always match. If it doesn't, you can be sure that should be a different security in PP.

With dividends, you don't know, because they are sometimes not paid in the listing currency. Currency mismatch is not evidence for creating a security.

And yes, you can have a mismatch. Real-world example: you have EQQQ.F (EUR) and EQQQ.L (GBP) and both pay a dividend in USD. You don't have the CONIDs stored in the WKD field. Both have the same ISIN. Currency doesn't match, but that proves nothing. Where do you assign the dividend? Currently impossible to say.

You could try to solve this using the CashTransaction's listingExchange attribute, but I think that would be more complexity, a pain to maintain (mapping table with exchange codes) and still imperfect – and all that for an edge-case.

I ignored this case, because it's not easy to get into this situation. You need to:
- Own the same instrument in multiple currencies AND
- Receive a third-currency dividend AND
- Not have the CONIDs saved

... although the joke is – that's exactly what happened to me.

Really, if you're doing non-trivial stuff with IBKR, you want CONIDs saved for all securities. And I think the IBFlex importer is getting more powerful and accurate, so in the future that should be the norm. I only ended up with securities without CONIDs due to importer issues.

buchen: Fair. I am wondering if the user will understand the subtle difference between matching trades and dividends. I would think it is easier if it is one algorithm - and if it mixes it up, fix the meta data once and for all.

And if a new instrument is created, the user could still choose to apply them to an existing instrument:
 
Bildschirmfoto 2026-01-05 um 20 22 00

georgemac-labs: Coming back to this now the parent commit has been merged...

I don't really agree on this point because I don't think the user is thinking in depth about the logic. And I think this approach will produce the right result more often than the alternative. However, it's not a hill I would die on, and it makes the code is more complex. Don't mind dropping it.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: ... I would say this has been clarified in the meantime? Solution for GBP/GBX et al in IBFlex is in https://github.com/portfolio-performance/portfolio/commit/8319a8659beb5b15ea64678fd5e18ac40c77540d

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
georgemac-labs: Here's an alternative proposal, please let me know if you agree and I will implement.

- Match on:
  - conid (no currency check — dividends can be in a non-base currency)
  else
  - ISIN (disambiguate by currency if multiple match)
  else
  - symbol (disambiguate by currency if multiple match)
- If the chosen identifier matches more than one security after currency disambiguation, use the first match and emit a warning that there is ambiguity
- Same rules for all transaction types — no dividend/trade distinction

Nirus2000:
The improvement — stricter security matching rules for IBFlex imports — is well-motivated and addresses real-world mismatches (ISIN with currency mismatch, symbol with ISIN conflict). @buchen acknowledged the CONID/ISIN conflict checks as sensible.
However, the PR stalled for two reasons:

Open design question: @buchen suggested preferring a single unified matching algorithm over separate logic for trades vs. dividends, to avoid confusing users. This was never addressed.
Dependency on #5292: The PR sits on top of that one and was explicitly put on hold until it concluded. It is unclear whether #5292 has since been merged and whether this PR was ever rebased.

If someone wants to revive this, please open a new pull request that:

Rebases cleanly on the current master
Addresses the unified-algorithm concern raised by @buchen
Revisits the dividend vs. trade matching distinction with a clear rationale or simplification

Thank you to @georgemac-labs for the thorough analysis of the matching edge cases and for taking the time to contribute. The problem is real, the approach is sound — it just needs someone to pick it up and carry it over the finish line.

georgemac-labs:
@Nirus2000 I find this frustrating. I put a good amount of work into this PR and I think I responded to all the feedback. It is not inactive because of me, but because I'm waiting for an answer to my [proposal](https://github.com/portfolio-performance/portfolio/pull/5303#discussion_r3236360241), which was indeed for a different matching algorithm, based on @buchen's comments.

Please point me to the question you believe was unanswered.

PR #5309, opened 2026-01-05
IBflex Import: Steuergutschrift hinzugefügt

quapla:
Es werden jetzt auch Steuergutschriften berücksichtigt (if abfrage einfegügt)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractor.java
```diff
@@ -1252,6 +1258,8 @@ public List<Item> extract(SecurityCache securityCache, Extractor.InputFile input
     @Override
     public void postProcessing(List<Item> items)
```
buchen, line 1344: This method has now no implementation. If that is on purpose (I see failing tests), then it should be removed in the code. Deleted code should not be in the file (it can be retrieved from the git history if needed).

buchen (changes requested):
This change has
- no test cases for the tax refund
- includes commented out code
- and the unit test fails

PR #5311, opened 2026-01-05
When concatenating strings, do trim the individual strings before

buchen:
Issue: #5296


buchen:
@Nirus2000 what do you think? Always trim? It appears that this is almost always done already and would avoid specific code in #5296

Nirus2000:
Sure... LGTM 👍🏻 

But please keep in mind that we will then have to remove trim() from all importers everywhere in the notes so that we don't have to check or clean everything twice or three times!!!
That's a lot of work.

ZfT2:
> Sure... LGTM 👍🏻
> 
> But please keep in mind that we will then have to remove trim() from all importers everywhere in the notes so that we don't have to check or clean everything twice or three times!!! That's a lot of work.

But I think that could be done time after time, wenn  the one or other importer is modified anyway?

buchen:
@Nirus2000 writes
> But please keep in mind that we will then have to remove trim() from all importers everywhere in the notes so that we don't have to check or clean everything twice or three times!!!

Claude Code says:
> The agent found 109 redundant trim calls across 33 files.

PR #5313, opened 2026-01-06
Refactor assertions inside DeutscheBankPDFExtractorTest

gamue:
As I created a new test yesterday in another PR and noted that this class is having an outdated style, I thought it's a good task for Claude to update the format ;) 

Tests worked locally, build is failing with a 403 while setting up maven. I don't have permissions to retrigger, but should be fine


Nirus2000:
Dont do THIS !! It takes a lot more effort to check all these changes again, and the added value is zero.
Please leave the old tests as they are.

Issue #2707, opened 2022-02-23
Colors in graph are too similar


MichaelMure:
**Is your feature request related to a problem? Please describe.**

I find the time series hard to read because of the colors that are very similar. Here is an example where the generated palette is some variation of blue/violet/orange. Figuring out how which line match with the legend becomes a game of guessing which color match the most.

![image](https://user-images.githubusercontent.com/294669/155295609-bd19163d-fcf6-40fc-b030-cc9cb8717f79.png)

**Describe the solution you'd like**

A different algorithm to generate the palette would help a lot. There is some suggestion over here: https://softwareengineering.stackexchange.com/questions/198065/what-algorithms-are-there-for-picking-colors-for-plot-lines-on-graphs

It looks like the golden ratio algorithm would work quite well.

**Describe alternatives you've considered**

Another possibility (or rather, a complementary solution) would be to have some cue about which series is the closest to the mouse cursor in the tooltip window showing up when clicking on the graph.

![image](https://user-images.githubusercontent.com/294669/155296756-7425979a-3c8c-4cb6-a900-3054b72efa0a.png)

funnym0nk3y:
You have to keep in mind that the colors could (and in my opinion should) also represent the taxonomy of the asset. 

![grafik](https://user-images.githubusercontent.com/41870754/156569716-94e6761a-7eed-47b8-903b-c76bc9dc5198.png)

flywire:
> You have to keep in mind that the colors could (and in my opinion should) also represent the taxonomy of the asset.

I don't agree. Unless displayed colours can be distinguished from each other the chart is useless. It would be useful to click on legend item(s) and have the line change to bold.

Issue #1878, opened 2020-12-18
Black theme


chrisaut:
I have an OLED screen. Dark theme is really nice, but I think it would look even better if the background was true instead of the grey. There seem to be different shades of grey for the sidebar and other panels. Just make them all black. Would probably give a bit of a bloomberg terminal feel even if the panels aren't so distinguishable from each other.

I think nothing else would have to change compared to dark mode.

buchen:
The CSS files (with limitations, of course, as we talk about SWT) are in ```n.a.portfolio.ui/css```. I would be happy to add an "extra dark" theme or darken the existing theme (I don't know what other users say).

PR #5315, opened 2026-01-06
replace arror icon file with dynamically render

liuhaoXD:


liuhaoXD:
This is a preparatory PR.

I noticed that several issues related to customizing themes (like https://github.com/portfolio-performance/portfolio/issues/1878,  https://github.com/portfolio-performance/portfolio/issues/2707) so I intend to handle them in a unified manner. 

Given that the change scope could be quite extensive to support customize theme, I first switched the up/down-related arrow icons from static files to dynamic rendering, which will facilitate the color switching of arrows during theme switching.

The changes made in this PR should theoretically have no impact on existing functionalities.

PR #5318, opened 2026-01-06
Adapt AT-Format for Renault-Bank-Direkt PDF Importer

MK-42:
Correctly handle multiple whitespace characters in interest transactions

I have received multiple statements with more than one whitespace / tab character in the interest lines


buchen:
Thanks. One character change. Safeguarded by a test case.

PR #5324, opened 2026-01-09
Updated FlatEx Degiro to support Jan 26 format of purchase transactions

buchen:
Issue: https://forum.portfolio-performance.info/t/pdf-import-von-flatex-at/10776/84

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
@@ -10,6 +10,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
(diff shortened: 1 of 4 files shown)


buchen:
@Nirus2000 - I found it more intuitive adding a new entry for ```* Fremde Spesen``` instead of complicating the regular expression further to match both. But the argument could go both ways.

buchen:
Obviously I have to rework this one...

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
```diff
@@ -206,14 +207,23 @@ private void parseContext(String filename, DocumentContext context, String[] lin
         private int maxSize = -1;
         private Transaction<?> transaction;
 
+        /**
+         * Creates a block that represents the whole file.
+         */
+        public Block()
```
buchen, line 213: Proposal: We allow to create blocks that represent the whole file. That makes sense if the document spans the whole file anyway. In the current case, the "marker" for the blocks has changed (it is not Auftragsdatum anymore). But this is really an arbitrary change because that term was never matched anyway to retrieve meaningful data.

buchen:
@Nirus2000, I know you are enjoying your time in the snow. I'll take the risk to include this change in today's release. If need be, we revert.

PR #5325, opened 2026-01-10
Add Support for Volkswagen Bank

gamue:
- Audi Bank is already supported, which is a subsidiary of the Volkswagen Financial Services AG. To support multiple banks of this company, I've renamed the classes to Volkswagen Financial Services
  - Note: A new customer of one bank only counts as such, when he's not having an account in the other.
- Adding support for Volkswagen Bank, which uses the same pdf-format

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AudiBankPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AudiBankPDFExtractor.java
@@ -23,7 +23,7 @@ public class AudiBankPDFExtractor extends AbstractPDFExtractor
     @Override
     public String getLabel()
     {
-        return "Volkswagen Financial Services (Volkswagen Bank, Audi Bank)";
+        return "Audi Bank";
     }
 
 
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/VolkswagenFinancialServicesPDFExtractor.java
```diff
@@ -9,21 +9,22 @@
 import name.abuchen.portfolio.money.Money;
 
 @SuppressWarnings("nls")
-public class AudiBankPDFExtractor extends AbstractPDFExtractor
+public class VolkswagenFinancialServicesPDFExtractor extends AbstractPDFExtractor
```
Nirus2000: Hello,

Unfortunately, we don't do that. As our experience has shown, banks and brokers connect and disconnect from time to time, and then we have to keep dragging these changes along with us.

We prevent this by applying the contributing rules.
https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#pdf-importers
(...) **Each bank/broker has its own extractor class.**

Please create a separate importer. Thanks.. 👍🏻

Nirus2000 (changes requested):

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/VolkswagenFinancialServicesPDFExtractor.java
gamue: fair point, however would still like to add the "Echtzeit" regex-adjustment to the AudiBank one. As said they use the same format and are tightly coupled so guess the format-addition is valid for both. ok?

buchen:
I split this into two commits: one for Volkswagen, and one for Audi Bank

gamue:
Thanks @buchen , I've thought that the PR will be squash-merged so haven't cleaned up the commit history, sorry about this. For me squash-commits are totally fine and like that with them the PR is linked directly inside the commit history

PR #5332, opened 2026-01-12
Improve Wir Bank PDF extractor for French documents

otbinary:
Adjusted PDF extractor for Wir Bank and VIAC to correctly import some additional French documents that were rejected before due to differences in wording and spacing.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
@@ -1194,6 +1194,42 @@ public class WirBankPDFExtractorTest
     }
 
     @Test
+    public void testInterest06()
+    {
+        var client = new Client();
+
+        var extractor = new WirBankPDFExtractor(client);
+
+        List<Exception> errors = new ArrayList<>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Zins06_Francais_2019.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(0L));
+        assertThat(countBuySell(results), is(0L));
+        assertThat(countAccountTransactions(results), is(1L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(results.size(), is(1));
+        new AssertImportActions().check(results, "CHF");
+
+        // check interest transaction
+        var transaction = (AccountTransaction) results.stream().filter(TransactionItem.class::isInstance).findFirst()
+                        .orElseThrow(IllegalArgumentException::new).getSubject();
+
+        assertThat(transaction.getType(), is(AccountTransaction.Type.INTEREST));
+
+        assertThat(transaction.getDateTime(), is(LocalDateTime.parse("2019-11-30T00:00")));
+        assertThat(transaction.getSource(), is("Zins06_Francais_2019.txt"));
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
```diff
@@ -1429,6 +1492,42 @@ public void testCreditNote03()
...
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(results.size(), is(1));
+        new AssertImportActions().check(results, "CHF");
+
+        // check deposit transaction
+        var transaction = (AccountTransaction) results.stream().filter(TransactionItem.class::isInstance).findFirst()
+                        .orElseThrow(IllegalArgumentException::new).getSubject();
+
+        assertThat(transaction.getType(), is(AccountTransaction.Type.DEPOSIT));
```
gamue, line 1519: you could use the same style for assertion as in L1379

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/WirBankPDFExtractor.java
```diff
@@ -44,12 +44,12 @@ public String getLabel()
 
     private void addDepositTransaction()
     {
-        DocumentType type = new DocumentType("(Einzahlung|Deposit) 3a");
+        DocumentType type = new DocumentType("(Einzahlung|Deposit|Avis de versement) 3a");
```
gamue, line 47: (optional) Its a good habit to adjust code that you're touching and update the style. For example in more modern extractors `var` is used

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
otbinary: That would be possible, though I preferred to maintain the same assertion pattern as in the existing `testCreditNote` test cases to facilitate comparing them and spotting the differences when reading the code.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/WirBankPDFExtractor.java
otbinary: I generally prefer to avoid refactoring code and doing other unrelated changes as part of small fixes and improvements.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
gamue: this is the old assertion style which is still used in older test-cases when writing new ones please update it to the newer one like above, as this is easier to read

otbinary: Thank you for your feedback. At this point, I prefer not to invest additional time in this. Please feel free to make any changes you see fit, reject the pull request, or use the code as is.

buchen: I converted it to the new unit test matchers. It makes for much more readable tests (but at the same time there is limited value in converting the old tests - ideally we never touch them again anyway).

buchen:
merged.

PR #5334, opened 2026-01-14
Add assertion for countSkippedItems to extractors

gamue:
- Added `countSkippedItems` to pdf extractor tests if they have the newer style
- Methods and classes in the old style haven't been adjusted, e.g. `DegiroPDFExtractorTest` is not having any method in the new style
- The code was done with the support of AI-tools, I manually reviewed them

@Nirus2000 hope that's as requested :)


buchen:
Generally makes sense for consistency reasons. Although the test also checks for the total number of items. This implies 0 for skipped items.

Merged. I have added an organized imports for all of them to sort them alphabetically.

PR #5339, opened 2026-01-15
Update Postbank: Advance lump-sum tax with zero tax due

buchen:
Create SkippedItems for documents where the advance lump-sum tax (Vorabpauschale) is fully covered by the tax exemption allowance, resulting in zero tax.

Issue: https://forum.portfolio-performance.info/t/pdf-import-von-postbank/7234/96


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostbankPDFExtractor.java
```diff
@@ -568,21 +568,17 @@ private void addAdvanceTaxTransaction()
...
-                        .section("date") //
-                        .match("^Belastung mit Wert (?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}).*$") //
-                        .assign((t, v) -> t.setDateTime(asDate(v.get("date"))))
-
-                        // @formatter:off
-                        // Belastung mit Wert 02.01.2024 26,41 EUR
-                        // @formatter:on
-                        .section("currency", "amount") //
-                        .match("^Belastung mit Wert [\\d]{2}\\.[\\d]{2}\\.[\\d]{4} (?<amount>[\\.,\\d]+) (?<currency>[A-Z]{3})$") //
+                        .section("date", "currency", "amount").optional() //
```
buchen, line 571: I pulled these two sections into one. The reason is that if I need to mark this optional (and the payable amount is optional), then I want to be sure if it is matched, then both date and amount are matched. Otherwise we would have to check for both date and amount further down below. Of course, one could also argue that the regular expression is identical and therefore either both match or both do not match.

PR #5340, opened 2026-01-16
IBFlex: support additional date and time formats

lmb:
Interactive Brokers let's customers configure the date and time format of flex queries. Users have reported NPEs on the forum because their existing flex queries use a format that is not recognised by the importer since my changes to correct exchange rates.

Extend the supported date formats. Note that we drop time zone information when converting to LocalDateTime since the rest of the codebase doesn't seem to be TZ aware.

We currently only support semicolons as delimiters, with one exception for a historical test case.


buchen:
merged. thanks.

One of these days I will reformat the IBFlex formatter to allow use of auto-formatting (after the pr sitting are done).

PR #5347, opened 2026-01-17
dedicated submenu for TWR reporting period choice for TreeMap color

mierin12:
Issue https://forum.portfolio-performance.info/t/treemap-farbe-nach-twr/38652

Hello,

It seems the selection of TWR for color use in TreeMap is not clear , I think adding a sub menu makes it clearer :

**Before**

**After**
2026-01-17 15_12_36-Portfolio Performance


buchen:
@mierin12 - makes sense. I wonder though if we then really need the "more..." submenu. The idea was to limit the entries in the top level menus, however, that is not relevant anymore.

PR #5350, opened 2026-01-18
Document head of Money class.

hporten:
Some Javadoc for the Money class. This is after I tried to determine what exactly is stored inside.

Please review thoroughly.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/money/Money.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/money/Money.java
@@ -3,6 +3,17 @@ package name.abuchen.portfolio.money;
 import java.text.MessageFormat;
 import java.util.Objects;
 
+/**
+ * A monetary amount, such as 'EUR 2,500.50', combining a currency with an
+ * amount.
+ * <p>
+ * Monetary values are stored as long values in minor currency units (e.g.,
+ * cents for EUR/USD).
+ * <p>
+ * To avoid overflow of long integers, particularly when mathematically
+ * combining different values such as monetary amounts with prices or shares, it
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/money/Money.java
```diff
@@ -1,8 +1,24 @@
 package name.abuchen.portfolio.money;
 
+import java.math.BigDecimal;
```
buchen: I am not sure if we really want to add new imports just for Java docs...

name.abuchen.portfolio/src/name/abuchen/portfolio/money/Money.java
```diff
@@ -1,8 +1,24 @@
...
+
+/**
+ * A monetary amount, such as 'EUR 2,500.50'.
+ * <p>
+ * Monetary values are stored with a {@link Values#precision() precision} of 2
+ * decimal digits. In order to handle them with integer types, this class thus
+ * stores the amount multiplied with a {@link Values#factor() factor} of 100.
+ * <p>
+ * To avoid rounding errors and other precision errors, it is recommended to
+ * manipulate values with the help of the {@link BigDecimal} class.
```
buchen: Which is correct, but I think the main problem occurs are buffer overflow. The amount has two decimal digits, shares have 8 decimal digits. If multiplied, there is a chance for a buffer overflow for (relatively) small monetary amounts already. Therefore my "heuristic" is: whenever I mathematically combine values of different types, I tend to use BigDecimal.

hporten:
@buchen Thanks for the various enhancements. It is much clearer now and will help me and other users of the class.

PR #5352, opened 2026-01-19
IBFlex: Import Flex Queries via HTTP API

### LINKED ISSUE #5246: IB Flex: automatically fetch flex queries via web api
I've just discovered that IB allows accessing Flex Queries via a fairly simple HTTP API: https://www.interactivebrokers.com/campus/ibkr-api-page/flex-web-service/

I think it would be possible to use this to implement automatic transaction updates from IB. It could work similarly to how exchange rate fetchers work:

- In config, add query ID and secret token
- PP periodically fetches transactions, at most once a day

Open questions:
- The max. duration for a single export is 365 days. There would need to be some place to store / retrieve the last fetch date.
- The secret token expires after max one year. There would need to be a UI to surface this to the user.

lmb:
This is an implementation of the Flex Web Service API to automatically fetch Flex Queries. Please see the individual commit descriptions for details.

Disclosure: I used Claude Code to draft this PR and have done manual code review on it. I think the code is OK, maybe a bit verbose. Please let me know if I've missed something or if I can somehow remove some code.

Fixes: https://github.com/portfolio-performance/portfolio/issues/5246


lmb:
@buchen is this something you would consider merging? Happy to try and slim this down more if it is too much to review.

buchen:
Sorry @lmb it was just not top on my mind (and a lot of other pull requests have run up).
I think there is nothing in principle against this. The only thing is that I avoid storing secrets in the XML file itself. Here I see it is attached to as a property.

Besides that, I haven't looked in detail but hope to be able to do so over public holidays.

lmb:
@buchen reworked this to use a new preference dialog to store tokens + query cutoff times. Also improved cancellation of the web fetch.

buchen:
@lmb 

I should not promise anymore when to look over changes... I only now had the chance to do so

I did rename IBFlexConfigurationDialog to IBFlexConfiguration because I felt it has nothing to do with dialogs anymore but is a class for managing the configuration.

Please check #5651

Issue #5362, opened 2026-01-22
Support different theming for positive/negative colors


buchen:
As a user, I want to choose between different color schemes representing positive/negative values, for example using a blue/orange color scheme.

Context:

* There are two themes: dark and light. The "positive/negative" theme is orthogonal: We need a 'green/red' or 'blue/orange' theme that might be different per dark/light theme.
* We have colors and icons (for example the up and down arrows are currently green and red respectively)

There have been some pull requests created: #4739 #5315

buchen:
@liuhaoXD I created this item - linking to your two pull requests - to discuss the best way forward.

What do you think of this idea:

In the CSS (dark and light) we define additional "ValueColorSchema"s:

```css
ValueColorSchema-greenred {
	positive-background: #02F500;
	negative-background: #FE2B05;
	up-arrow: "icons/greenred/up_arrow.svg";
	down-arrow: "icons/greenred/down_arrow.svg";
}

ValueColorSchema-blueorange {
	positive-background: #02F500;
	negative-background: #FE2B05;
	up-arrow: "icons/blueorange/up_arrow.svg";
	down-arrow: "icons/blueorange/down_arrow.svg";
}
```

This allows us to tweak the colors for the dark and light themes. It could even use different icons.

Then we create - similar to the ```Colors.Theme``` class - a couple classes that get the values from the CSS injected (similar to the ```ColorsThemeCSSHandler```).

In the preferences, the user can pick one of the pre-defined scenarios, say "green/red" or "blue/orange". This preference property assigns the chosen ValueColorSchema to a constant (similar to the ```Colors.theme```). 

The usage could look like this:

```java
@Override
public Color getForeground(Object element)
{
        return t.getType().isLiquidation()
               ? ValueColorScheme.current().negativeForeground()
               : ValueColorScheme.current().positiveForeground();
}
```

or

```java
@Override
public Image getImage(Object element)
{
    Money money = valueProvider.apply(element);
    if (money == null || money.isZero())
        return null;
    return money.getAmount() >= 0
              ? ValueColorScheme.current().upArrow().image()
              : ValueColorScheme.current().downArrow().image();
}
```

Let me know what you think?

liuhaoXD:
@buchen I closed these two PRs because they were developed in a rush, and on further reflection, I realized the thinking behind them was not thorough enough. This issue turns out to be more complex when explored in depth, here is a perfect place to have a full discussion about it.

[...]

PR #5363, opened 2026-01-23
Support different theming for positive/negative colors

buchen:
See discussion in #5362


buchen:
@liuhaoXD can you have a look and comment on the code?

Open:
* when switching the color scheme, it is not applied to existing tables. One has to change the view
* no dynamic rendering of the up and down arrows
* theme 'automatic' not renamed to 'sync with OS' (separate pull request)

buchen:
@liuhaoXD any feedback - and possibly colors for another schema more fitting for Asia?

PR #5369, opened 2026-01-25
fix pie chart when title is shared between subtaxonomy

mierin12:
Hello, sorry I did not catch this issue :
https://forum.portfolio-performance.info/t/strobo-wenn-pp-in-kuchendiagramm-startet-und-kurse-aktualisiert/38671/7

I can reproduce when the security is shared between several sub-taxonomy, I did not have this case when checking..
I am now checking against kommer/Industries taxonomy which clearly show the issue

**Before**
Capture d’écran 2026-01-25 150855

**After**
Capture d’écran 2026-01-25 151155
I think this was the previous behavior ?

What do you think ? It is bit trial and error, I am not sure why it works, I guess uuid is unique per security, but not per slice in the case the security weight is shared among several folder. I hope this is not introducing other issues.

Similar issue for flare, I did not find yet the fix.

### DIFF
--- a/name.abuchen.portfolio.ui/META-INF/html/pie.html
+++ b/name.abuchen.portfolio.ui/META-INF/html/pie.html
@@ -86,7 +86,7 @@ var thePieChart = {
         }
 
         this.path = pathGroup.selectAll("path.pie")
-            .data(this.piedata, function(d) { return d.data.uuid; });
+            .data(this.piedata);
 
         // Enter: create new paths
         this.path.enter()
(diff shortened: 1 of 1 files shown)


buchen:
As I am on the road at the moment, I will revert the two commits and release a version (only as online update).
Then we have more time to properly do the changes. I was trusting Claude Code too much...

mierin12:
Yes ok, sorry I should have catched it, but I did not look at this case of shared security. 
While trying to find a fix for flare, I am noticing a second unwanted behavior for flare chart : if you select an investment node (not a category) and then untoggle "investment vehicles", the flare chart goes crazy.

buchen:
> Yes ok, sorry I should have catched it, but I did not look at this case of shared security.

Ah, come on, I did not ask you to do a formal test. I was too enthusiastic with the change where I dod not really understand the code...

PR #5451, opened 2026-02-09
add moving average Cost Method in Payment view

mierin12:
for Closed Trades and Sum tabs

Hello, 

The proposition is to add the possibility to select in the Payments view, the gross profit from Closed Trades either from FIFO either from moving average method. So this applies to the Closed Trades tab, but also the Sum tab.

2026-02-09 19_31_30-Portfolio Performance


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/payments/PaymentsViewInput.java
```diff
@@ -146,6 +175,7 @@ public void writeToPreferences(IPreferenceStore preferences, Client client)
         preferences.setValue(KEY_MODE, mode.name());
         preferences.setValue(KEY_USE_GROSS_VALUE, useGrossValue);
         preferences.setValue(KEY_USE_CONSOLIDATE_RETIRED, useConsolidateRetired);
+        preferences.setValue(KEY_COST_METHOD, costMethod.name());
```
buchen, line 177: Tiny remark (in case users ask): if the user navigates from a widget here, then the cost method is always FIFO. And then leaving this view, it will overwrite preferences that the user picked previously. That is already the case with the "consolidate retired instruments" option as well - nothing new - but just in case users are asking.

I have no good easy fix. Unless we also disable all those options (which does not make sense here because they change the calculation). There is no good indicator for "currently you are editing but I will not remember" either.

buchen (commented):
rebased and merged.

Issue #5453, opened 2026-02-10
Ease handling of bonds: percentage quotes


hporten:
**Is your feature request related to a problem? Please describe.**
Right now, bonds-related transactions can only be entered via a couple of workarounds described in a [howto](https://help.portfolio-performance.info/en/getting-started/manage-portfolio/bonds/) and still have some limitations making them not so well supported. Several forum discussions show prior requests.

**Describe the solution you'd like**
To keep things manageable, this issue is not about full-blown support for bonds. Older discussions also reveal concerns about specializing on particular types of securities. After all, PP should remain
generic and allow for support of all kinds of financial products.

So what could be done to minimize the number of workarounds and remain generic? A possible start would be the support of percentage-based quotes. After all, that is a very fundamental principle described in e.g. the  [Market Value](https://www.deutsche-boerse.com/dbg-en/about-us/contact/glossary/glossary-article/Market-value-243052) glossary entry by Deutsche Börse:
 
> Depending on the type of security involved, the market value is expressed either as a unit quotation or a percentage quotation:

So how about we introduce the concept of unit quotation (German: Stücknotierung) vs. percentage quotation without even talking about bonds anywhere?

I have experimenting with this through some patches (link will follow as a comment). In order to get a feel for the number of places in the code will need to be adapted in the end. 

**Describe alternatives you've considered**
The documented workarounds require quite some manual labor (i. e. break the convenient process of simply importing of PDFs) and skew the purchase/sales units.

For example, the howto suggests to book a 5000 EUR bonds with a price of 100% via 50 shares x 100 EUR. This division of the nominal value will have to be applied consequently and will affect several places in the UI. "50 shares" is simply not reflecting a nominal value of 5000 very well.

**Additional context**
It would be very convenient to have e.g. the Buy dialog look like this working proof-of-concept:
<img width="646" height="535" alt="Image" src="https://github.com/user-attachments/assets/8f88ba11-8707-457c-afb7-87369105da67" />

hporten:
Here is draft of partial support for percentage-based quotes: https://github.com/portfolio-performance/portfolio/pull/5454

buchen:
> To keep things manageable, this issue is not about full-blown support for bonds. Older discussions also reveal concerns about specializing on particular types of securities. 

@hporten  Which instrument - other than bonds - would use percentage-based pricing?

Looking at the pull request, there are only two places where the calculation changes: PortfolioTransaction and SecurityPosition. That looks manageable.

[...]

PR #5454, opened 2026-02-10
Introduction of percentage-based price quotation

hporten:
This draft of patches introduces a Security.percentageQuote boolean with setter and getter methods. The quotation type can be configured in the Security dialog or will be set automatically when a percentage-based security is recognized on imports. This has been tried out for the case of Deutsche Bank documents.

A few places in the UI and the backend calculation have been adapted. There are likely going to be several statistics that still need adjustement.

Here is how the Buy dialog could look like:

buy_percent

For now, backward compatibility will be kept by leaving old securities entirely untouched. More sophisticated approaches that would would enable a gradual migration of legacy data are thinkable.

The individual commits from the experiment have been kept. For a better understanding of how things developed. Cleanups and refactoring will lead to a smaller delta.

Closes: #5453

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/SecurityPosition.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/SecurityPosition.java
@@ -101,6 +101,8 @@ public class SecurityPosition
                         .multiply(BigDecimal.valueOf(price.getValue()), Values.MC)
                         .movePointLeft(Values.Quote.precisionDeltaToMoney()) //
                         .setScale(0, RoundingMode.HALF_UP).longValue();
+        if (investment instanceof Security security && security.isPercentageQuoted())
+            marketValue /= 100;
         return Money.of(investment.getCurrencyCode(), marketValue);
     }
 


Nirus2000:
Mir sagt die isPercentageQuoted irgendwie nicht so richtig zu... es ist Mehrarbeit bei der Importerpflege und man müsste alle Importer wo der "~Workaround~" ist nachpflegen und noch eine Prüfung hinzufügen.

buchen:
> es ist Mehrarbeit bei der Importerpflege und man müsste alle Importer wo der "Workaround" ist nachpflegen und noch eine Prüfung hinzufügen.

Don't the importer already have a workaround today?

This pull request maintains some kind of backwards compatibility:
* if it is marked as bond, it uses the value as is
* If it is still the old workaround, it divides the value by 100.

We could also decide - once bonds are supported - to not support the "divide by 100" workaround anymore.

At the moment it is still open how to support bonds. I am not a big fan of a "percentageQuoted" flag. See #5453

hporten:
> Mir sagt die isPercentageQuoted irgendwie nicht so richtig zu... es ist Mehrarbeit bei der
> Importerpflege und man müsste alle Importer wo der "~Workaround~" ist nachpflegen
> und noch eine Prüfung hinzufügen.

I am confident, that the workaround can easily be hidden inside of a single call:
  
  getOrCreateBond(v)

call. That one would either create a brand new Bond object or apply the old-style workaround of legacy Security items. It will not be mandatory to port all importers at once. But those that will be adapted, will have simpler code than before.

PS: Yes, all fine with dropping the explicit percentage-flag and moving to an implicit calculation based on the type.

Nirus2000:
All I see here is that the attempt to implement this ends up in chaos and failure, just like the implementation of skippedItem.
I  memory of #5512 ...

PR #5463, opened 2026-02-14
add filter by account in the Trades view

### LINKED ISSUE #1597: Add filter to the Trade view

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/1597 for the account part
Issue https://forum.portfolio-performance.info/t/filter-trades-view-by-account/38850
Issue https://forum.portfolio-performance.info/t/weitere-filter-bei-trades-unter-dem-reiter-berichte-fur-einzelne-depots/21955

Hello this is a proposition to add to the Trades view a filter by account.
If I am not mistaken, this works for the general use of the view. The change to the getPorfolio method is to get the correct logo in the portfolio column, as LogoManager does not recognize ReadOnlyPortfolio from a filtered client.

What I am not sure is what to do when we get to the Trades view from the widgets. Currently the widgets comes into the view with a list of trades already collected. 

- What should then happen with such a preselected trades when the account filter is changed on the view ? Should it reset the preselected trades to honor the new account selection ? Should it further filter the preselected trades in case a "smaller" account is selected ? Currently is has no effect, so can be a strange behavior for the user.

- To be noted that if a filter account is already set on the view and the widget is collecting trades for another account, the activated view will be empty. But this is already the case with the other filtering option (if the view is set as Open Trades only, while the widget looks for Closed Trades -> empty view until Closed Trades is manually selected). 
Instead of setting the input already calculated, should the widget only gives the reference of the filter to apply both for account and Closed Trades ?

I am not sure of the behavior you want to see for this widget-to-view compatibility.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
@@ -33,6 +33,7 @@ import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Text;
 
+import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.model.Taxonomy;
 import name.abuchen.portfolio.money.CurrencyConverter;
@@ -52,6 +53,7 @@ import name.abuchen.portfolio.ui.UIConstants;
 import name.abuchen.portfolio.ui.editor.AbstractFinanceView;
 import name.abuchen.portfolio.ui.selection.SecuritySelection;
 import name.abuchen.portfolio.ui.selection.SelectionService;
+import name.abuchen.portfolio.ui.util.ClientFilterDropDown;
 import name.abuchen.portfolio.ui.util.ContextMenu;
 import name.abuchen.portfolio.ui.util.DropDown;
 import name.abuchen.portfolio.ui.util.LabelOnly;
@@ -173,6 +175,7 @@ public final class TradeDetailsView extends AbstractFinanceView
 
             update();
             updateFilterButtonImage(theMenu);
+            updateClientFilterEnablement();
         }
     }
 
@@ -181,6 +184,7 @@ public final class TradeDetailsView extends AbstractFinanceView
         private final Input preselectedInput;
         private final boolean useSecCurrency;
         private final CurrencyConverter jobConverter;
+        private final Client jobFilteredClient;
         private final boolean onlyOpen;
         private final boolean onlyClosed;
         private final boolean onlyProfitable;
@@ -190,14 +194,16 @@ public final class TradeDetailsView extends AbstractFinanceView
         private final boolean hideTotalsAtTheTopJob;
         private final boolean hideTotalsAtTheBottomJob;
 
-        UpdateTradesJob(Input preselectedInput, boolean useSecCurrency, CurrencyConverter converter, boolean onlyOpen,
-                        boolean onlyClosed, boolean onlyProfitable, boolean onlyLossMaking, Pattern filterPattern,
-                        Taxonomy taxonomy, boolean hideTotalsAtTheTop, boolean hideTotalsAtTheBottom)
+        UpdateTradesJob(Input preselectedInput, boolean useSecCurrency, CurrencyConverter converter,
+                        Client filteredClient, boolean onlyOpen, boolean onlyClosed, boolean onlyProfitable,
+                        boolean onlyLossMaking, Pattern filterPattern, Taxonomy taxonomy, boolean hideTotalsAtTheTop,
+                        boolean hideTotalsAtTheBottom)
         {
             super(Messages.LabelTrades);
             this.preselectedInput = preselectedInput;
             this.useSecCurrency = useSecCurrency;
             this.jobConverter = converter;
+            this.jobFilteredClient = filteredClient;
             this.onlyOpen = onlyOpen;
             this.onlyClosed = onlyClosed;
             this.onlyProfitable = onlyProfitable;
@@ -217,7 +223,7 @@ public final class TradeDetailsView extends AbstractFinanceView
             try
             {
                 Input data = preselectedInput != null ? preselectedInput
-                                : collectAllTrades(useSecCurrency, jobConverter);
+                                : collectAllTrades(useSecCurrency, jobConverter, jobFilteredClient);
 
                 if (monitor != null && monitor.isCanceled())
                     return Status.CANCEL_STATUS;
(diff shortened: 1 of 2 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
```diff
@@ -805,7 +811,8 @@ private Input collectAllTrades(boolean useSecCurrency, CurrencyConverter current
                 CurrencyConverter converterToUse = useSecCurrency && s.getCurrencyCode() != null
                                 ? currentConverter.with(s.getCurrencyCode())
                                 : currentConverter;
-                var collector = new TradeCollector(getClient(), converterToUse);
+                var collector = new TradeCollector(clientFilterDropDown.getSelectedFilter().filter(getClient()),
+                                converterToUse);
```
buchen: Claude says:

collectAllTrades() calls clientFilterDropDown.getSelectedFilter().filter(getClient()) inside getClient().getSecurities().forEach(...). PortfolioClientFilter.filter() builds a complete ReadOnlyClient — every portfolio, account and a copy of every transaction — so with an active filter that whole copy is rebuilt once per security on every refresh (search keystroke, filter toggle, model change). The existing dashboard code does it correctly: AbstractTradesWidget:110 hoists Client filteredClient = clientFilter.filter(getClient()) above the loop.

Not sure how "big" this performance thing is, but I will change it. Posting here only so you know what was changed.

buchen: And one more thing: if we filter the client, we also need to put FILTERED_CLIENT in the context so that the pane below also has the filtered view.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
```diff
@@ -400,6 +402,10 @@ protected void addButtons(ToolBarManager toolBarManager)
 
         addFilterButton(toolBarManager);
 
+        this.clientFilterDropDown = new ClientFilterDropDown(getClient(), getPreferenceStore(),
+                        TradeDetailsView.class.getSimpleName(), filter -> notifyModelUpdated());
+        toolBarManager.add(clientFilterDropDown);
```
buchen, line 412: If the trades are passed in from the outside from a widget, this menu does nothing. I will disable the dropdown in this case.

buchen (commented):
I have applied the changes mentioned and will merge.

buchen:
It turns out that the TransactionPair also needs unwrapping. Will have a go at it tomorrow.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
mierin12: Ok I understand the out of loop advise as per AbstractTradesWidget. 
FYI, I think I initially went with something else : instead of filtering in collectAllTrades, i guess it should be possible to filter the trades later, around line 225 : from all collected trades from the whole portfolio, if the trade belong to a portfolio which is within the clientfilter, then they are kept. This meant no need for unwrapping, but i do not remember why I went with the current way instead. Maybe to (imperfectly) mimic the AbstractTradesWidget way, considering we should have more confidence in TradeCollector to do the trade filtering from any client, rather than adding a new filter logic.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
mierin12: Ok I agree it is correct to disable it, because to be exact, in this case the menu did almost nothing : nothing visually, but it was still being remembered when the file was closed then reopened. So it was possible, after getting the trades from the widget, to change the menu selection without changing anything visually on the table, but the last selection would still be the one saved. it was a bit unconsistent.

(I guess another choice would be that a new menu selection, when made after a widget hop on, overrides this widget selection)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/trades/TradeDetailsView.java
buchen: > instead of filtering in collectAllTrades, i guess it should be possible to filter the trades later

In a way yes, but if there are transfers between investment accounts, then the trade could include them. And the filter would actually understand them as boundary.

Now one can argue what is right (one could argue that the history in the other account is not the core filtering criteria) on the other hand this is how all other views/filters work.

buchen: Now the remaining problem is that the transaction pane is showing the actual transactions and not according to the filter. In the example below, I filter by one investment account and have an outbound transfer to another account. The trade is closed (if the filter is applied) but the transaction list shows all transactions.

Bildschirmfoto 2026-07-15 um 06 24 04

It is not easy to change: If I want to show the filtered transactions, then it is not easy to make the editable. It is unclear which are derived and which are not.

The statement of asset view has the same behavior so I think it is okay here too.

I have pushed the changes into the PR if you want to have a look. I hope to create a release tomorrow so by then I will pick this up.

mierin12: I will have a look late this evening. 

I remember the discussion about editable vs non editable transactions list, I agree it is a complicated topic larger than this PR. At least the securities graph is correctly impacted.

mierin12:
@buchen  seems perfect !

mierin12:
one minor thing I am noticing, the Note column is empty when filtering by a single Portfolio, and not empty when filtering by the same Portfolio + its reference account. 

More accurately : Delivery Inbound/outbound 's Notes appear, but Buy/Sell's Notes do not appear in the case of a single portfolio filter.

mierin12:
> one minor thing I am noticing, the Note column is empty when filtering by a single Portfolio, and not empty when filtering by the same Portfolio + its reference account.

I think I found why : 
A client filter of a single portfolio will be created by `PortfolioClientFilter`, the readonlyportolio and its transactions are created line 201 through `adaptPortfolioTransactions`, which then will fall into line 248 as there will be not account cross owner (single portfolio filter), so the buy is converted into a delivery inbound using `convertTo` line 581, and inside `convertTo` there is no         `clone.setNote(t.getNote());` (that is present in the `recreateBuySell` however)

If I add it, i see the note column filled.

The convertTo for AccountTransaction is also without `clone.setNote(t.getNote());`

buchen:
> I think I found why 

Good catch. I will add the ```setNote``` calls. The note is not editable in the Trade view anyway. Otherwise we would have to think how to "route" that back to transaction that is editable.

mierin12:
Thanks again for picking this up, I this it is a good addition.

PR #5464, opened 2026-02-14
Fix Renault Bank Direkt PDF importer

MK-42:
Deposits and Removals now also have two whitespace characters.

I wasn't sure if refactoring that importer to the current state of the art is desired - therefore I went with the least-impact approach. Also taxes and interests are imported as two transactions. should we merge these two?


Nirus2000:
Hello @MK-42 
I had to revise the Renault PDF importer because, on the one hand, it no longer complied with importer standards and, on the other hand, it contained too many errors.

See #5467 

Regards
Alex

MK-42:
Hi Alex, that's what I was thinking myself. But I wasn't sure if refactoring that would be desired from your side. Thanks for taking care and integrating the additional test case.

Nirus2000:
Hello @MK-42 
Sure... I'd be happy to. I'm still trying to clean up the last few issues... 

If you want... or @gamue 

- Bank11PDFExtractor 
- JTDirektbankPDFExtractor

also needs refactoring like https://github.com/portfolio-performance/portfolio/pull/5467.

Regards
Alex

PR #5465, opened 2026-02-14
Improve open bookmark menu

wahlbrink:
Improve menu items to open bookmarks
  - Use attribute name instead of url for attribute bookmarks in context menus
  - Add support to open attribute bookmarks for multiple securities
  - Add indicator if attribute bookmark is only available for some selected securities
  - Change limit for number of securities to hard limit for more clear behavior (I prefer such clear behavior, but it is maybe a question of personal preference)
  - Add support to open bookmark for multiple securities in Security Performance view

Open_Bookmark_ContextMenu_marked

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/PortfolioListView.java
@@ -158,7 +158,7 @@ public class PortfolioListView extends AbstractFinanceView implements Modificati
             manager.add(new Separator());
 
             Portfolio portfolio = (Portfolio) portfolios.getStructuredSelection().getFirstElement();
-            new SecurityContextMenu(PortfolioListView.this).menuAboutToShow(manager, null, portfolio);
+            new SecurityContextMenu(PortfolioListView.this).menuAboutToShow(manager, List.of(), portfolio);
         }));
     }
 
@@ -318,7 +318,7 @@ public class PortfolioListView extends AbstractFinanceView implements Modificati
         if (portfolio == null)
             return;
 
-        new SecurityContextMenu(this).menuAboutToShow(manager, null, portfolio);
+        new SecurityContextMenu(this).menuAboutToShow(manager, List.of(), portfolio);
 
         if (!portfolio.isRetired())
         {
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/BookmarkMenu.java
```diff
@@ -40,6 +40,11 @@ public BookmarkMenu(PortfolioPart editor, List<Security> securities)
 
     private void addDefaultPages()
     {
+        if (securities.size() > 15)
+        {
+            return;
+        }
```
buchen, line 46: BTW, was there previously simply no limit to the number of browser tabs opened?
I am trying to wrap my head around what limitation this check here would introduce?

wahlbrink: The previous behavior was: The user can select any number of securities, but the menu item opens the bookmark only for the first 10 securities, the other securities are silently omitted (.limit(10) in the stream).

buchen: Thanks, I see it now.
Today, only the first 10 tabs are opened.
With this change, the bookmark menu is not shown at all if there are more than 15 instruments.
I think that is totally fine - too many clicks to clean up the browser anyway.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/Security.java
```diff
@@ -796,13 +796,19 @@ public void setAttributes(Attributes attributes)
...
 
         // extract bookmarks from attributes
-
-        getAttributes().getAllValues().filter(v -> v instanceof Bookmark).forEach(v -> bookmarks.add((Bookmark) v));
+        Attributes attributes = getAttributes();
+        client.getSettings().getAttributeTypes().filter((t) -> t.getType() == Bookmark.class).forEachOrdered((t) -> {
+            Bookmark bookmark = (Bookmark) attributes.get(t);
+            if (bookmark != null)
+            {
+                bookmarks.add(new Bookmark(t.getName(), bookmark.getPattern()));
```
buchen, line 809: I think we should **not** always overwrite the name of the bookmark.

**If** the user has given a custom name, then we should use that name.
**If** the user has not given a custom name, then the attribute type name is ok.

In the Bookmark Menu within the addDefaultPages, there is another algorithm implemented to consolidate the links. I think there it is okay to only show the attribute type name (in the case of multi-selection of securities).

Image

wahlbrink: I didn't know about this option (markdown input format) until just now. Great!

PR #5479, opened 2026-02-17
PDF extractor for 3a account documents from Neon Switzerland AG

manuelseeger:
PDF importer for neon Switzerland AG (3a accounts). 

Neon Switzerland AG partners with Hypothekarbank Lenzburg AG for their regular investment offering. However, their 3a accounts are a separate product held by Simply3a and managed by Lienhardt & Partner Privatbank Zürich AG.

The PDF statements for 3a accounts from Neon thus cannot be extracted by HypothekarbankLenzburgAGPDFExtractor. 

3a from neon is new as of 2026 thus only deposit, buy/sell, and fee documents are available so far.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NeonSwitzerlandAGPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NeonSwitzerlandAGPDFExtractor.java
@@ -52,11 +52,7 @@ public class NeonSwitzerlandAGPDFExtractor extends AbstractPDFExtractor
                                         .match("^Currency (?<currency>[A-Z]{3})$") //
                                         .assign((ctx, v) -> ctx.put("currency", asCurrencyCode(v.get("currency")))));
 
-<<<<<<< HEAD
-        this.addDocumentType(type);
-=======
         this.addDocumentTyp(type);
->>>>>>> 917d862fecc4058dd824ccee5e9f6ba756a58fe0
 
         addFundBuySellBlock(type);
         addDepositBlock(type);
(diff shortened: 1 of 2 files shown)


manuelseeger:
This includes code for deriving the ISIN from the NSIN: 
https://github.com/portfolio-performance/portfolio/pull/5479/changes#diff-15cde55d6eed844471e651396e97649abf8bdec9ace9e5d9126be4a7954e17f7R269

I kept this in the extractor to keep this PR to PDF import. But I feel like the ISIN from NSIN code could also go to util.Isin, which currently contains only checking valid check digits. 
https://github.com/portfolio-performance/portfolio/blob/6a4e08f511fd1b0f00a6919258bf3b9441a794e2/name.abuchen.portfolio/src/name/abuchen/portfolio/util/Isin.java#L7

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NeonSwitzerlandAGPDFExtractor.java
```diff
@@ -0,0 +1,296 @@
...
+                        // Secur.Nr. 1/039,462,806 Secur. Cur CHF
+                        // @formatter:on
+                        .section("name", "valor", "currency") //
+                        .match("^(?!Unit |Rate )(?<name>[A-Za-z].*)$") //
+                        .match("^Secur\\.Nr\\. 1\\/(?<valor>[\\d,]+) Secur\\. Cur (?<currency>[A-Z]{3})$") //
+                        .assign((t, v) -> {
+                            // NEON exclusively offers Swisscanto (CH)
+                            // securities:
+                            // https://www.neon-free.ch/en/strategien
+                            v.put("isin", toIsin("CH", v.get("valor")));
```
Nirus2000: You assume that all numbers are always preceded by the country code “CH.” But what if it's a different country code? This looks to me like a very individual importer created specifically for this case.

manuelseeger: So Neon exclusively offers Swiss securities which is why I went this way. 

I am a bit wary using the name because honestly what's in the PDFs are terrible names and matching on name seems a lot more brittle than match on the security number. 

What's the consensus of writing the valoren number into the WKN field? I didn't go this route because WKN is a German term even though they both are NSINs. That way the securities could reliably be matched using the security number and users could use proper names for their PP entries.

Nirus2000: The WKN is just a designation. We also use it as a valoren number. You can use it...
have a look into contributing rules

manuelseeger:
Messed up my branch trying to undo a merge with master and rebase instead. Will reopen with a new branch. 

Sorry for the mess.

PR #5502, opened 2026-02-21
Updated FinTechGroupBank PDF Extractor to support parsing of dividend…

buchen:
… ex-date

Issue: #5439


Nirus2000:
No, I don't want that.

buchen:
Please let me know what you want to change.

All 9 pull request apply the parsing of the ex-date to existing banks that actually have the date in the text. That should complete the support for importing ex-date from bank documents. Each is creating a new section - or should it be wrapped into the section parsing the date? Change the order of the Junit matcher expressions?

buchen:
Merged.

The development around ex-date is independent from the development work of SkippedItems. The latter one does not stop the former one. PP is developed incrementally and continuously.

There were no comments to the change itself. For future improvements, let's open new pull requests.

PR #5503, opened 2026-02-21
Updated Weberbank PDF Extractor to support parsing of dividend ex-date

buchen:
Issue: #5439


Nirus2000:
No, I don't want that.

buchen:
Merged.

The development around ex-date is independent from the development work of SkippedItems. The latter one does not stop the former one. PP is developed incrementally and continuously.

There were no comments to the change itself. For future improvements, let's open new pull requests.

PR #5506, opened 2026-02-21
Updated RaiffeisenBankgruppe PDF Extractor to support parsing of divi…

buchen:
…dend ex-date

Issue: #5439


Nirus2000:
No, I don't want that.

PR #5508, opened 2026-02-21
Updated GladbacherBankAG PDF Extractor to support parsing of dividend…

buchen:
… ex-date

Issue: #5439


Nirus2000:
No, I don't want that.

PR #5509, opened 2026-02-21
Updated DKB PDF Extractor to support parsing of dividend ex-date

buchen:
Issue: #5439


Nirus2000:
No, I don't want that.

PR #5511, opened 2026-02-21
Allow editing the ex-date in the account transaction dialog

buchen:
Issue: #5439


Nirus2000:
In memory of #5512

ZfT2:
Copied from https://github.com/portfolio-performance/portfolio/pull/5439:


> What is missing: Update the dialog to edit the ex-date. Right now, one can add the ex-date column to the transaction lists, but not edit the date in the dialog. I did not have time. The main problem is that the date picker does not allow to select no date. The "checkbox" is a good solution. I did not have to time to implement it. If we do, we should also pre-select the date from the divvy diary information (if available).

I will try do do that. @buchen Do you have a hint for me how to use the divvy diary information? Somewhere related in the code I can look into it?

buchen:
> Do you have a hint for me how to use the divvy diary information? Somewhere related in the code I can look into it?

The information is stored as ```SecurityEvent``` with the sub-class ```DividendEvent```.
I think you should be able to say ```Security.getEvents()``` to read them.
The class has payment date and ex-date.

I think with the matching of the payment date (= what the user enters) to the payment date in the security event (=what you find in the data), we have to be lenient. The payment date in security event is when the company pays, but it might take a couple days into the bank pays (Wertstellung). But I would hope a heuristic such as this could work: if there is a Dividend event with a pay date within 3 days, then I propose this as the ex-date.

ZfT2:
One more question: I'm trying to implement the check box and handling. As you said, the date picker does not allow null values. So if setting the value to null, there are NullPointers in other classes.

So for the logic: If we show the ex date in the edit dialog: Should we say, if there is an ex date present which is equal to the transaction date, we hide the ex date input field? The user has then the option to enable the  ex date picker with a checkbox.
Also the same behaviour if the user creates a new transaction. 
If there is an ex date different to the transaction date present, we show the ex date picker (and the checked checkbox) already when the dialog opens?

buchen:
> So for the logic: If we show the ex date in the edit dialog: Should we say, if there is an ex date present which is equal to the transaction date, we hide the ex date input field? The user has then the option to enable the ex date picker with a checkbox.

Good question. The behavior with the date picker is not so easy. I think we must make sure the model in the backend is correct.

What do you think about the following behavior?

When creating a new transaction:
* If the user creates a new dividend, the checkbox is off, the ex-date not shown, the ex-date is null (in the model)
* If the user the user sets the checkbox, the ex-date is shown and pre-filled with the transaction date (it cannot be null) and has the value in the backend
* If the user unselects the checkbox, the field is hidden and the ex-date is set to null in the model
* add a validator that ensures the ex-date is not later than the payment date
* (enhancement) instead of the transaction date, pre-fill it with the date from the DividendEvent if available
* (enhancement) if there is a DividendEvent that matches the payment date, then activate automatically
* (enhancement) if the user changes the transaction date, the ex-date is updated as well
  * if it is now after the transaction date, it is set at least to the transaction date
  * if the new transaction date matches a DividendEvent, then set it to the ex-date of that event

When editing:
* if the transaction contains a ex-date, then show the ex-date out of the box

ZfT2:
Sounds good.
I implemented the changes but for now without checking the dividend events. Will try to do that later when I find time :)

Here is the feature branch:
[ZfT2:portfolio:feature/ex-date/edit-transaction-dialog](https://github.com/ZfT2/portfolio/tree/feature/ex-date/edit-transaction-dialog)
Maybe you want to have a look already :)

buchen:
@ZfT2 I have picked up your changes from the branch, merged it into this changes, did some enhancements, and and force pushed it here.

Can you give it a test?

There are a couple things missing:
* if the transaction type has a security only optionally, it is still possible to set a ex-date
* the ex-date is not filled from the SecurityEvent

However, I think it is "good enough" to be released.

ZfT2:
@buchen looks good for me and seems that you implemented the security event lookup already :) 
Only thing missing is the update of the ex date if the date is changed. But I'm not sure if we really want that?

buchen:
> Only thing missing is the update of the ex date if the date is changed. But I'm not sure if we really want that?

Good point. I'll add that. Let's see what the feedback from users will be.

Issue #5517, opened 2026-02-23
Skipped Items - PDF import: Tracking new usage of Skipped Item


buchen:
The PDF importer support returning a skipped items for transactions that can be successfully extracted from documents but are not supported by the domain model of Portfolio Performance. One example are taxation of retained (undistributed) income where not payment is recorded and the transaction has a value of zero. A skipped item may link to an incomplete transaction.

---

### Related Work

**Completed**
- #5263 (initial implementation of skipped items)

---

### (Potentially) use skipped items for other importer

Check if other, existing PDF importer could make use of the skipped item (instead of return no item at all).
Here is a draft list of importer that have checks for transactions that are not supported at the moment:

- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java:685
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGPDFExtractor.java:350
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java:1093
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DADATBankenhausPDFExtractor.java:745
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DZBankGruppePDFExtractor.java:653
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DegiroPDFExtractor.java:361
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DekaBankPDFExtractor.java:438
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java:692
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/Direkt1822BankPDFExtractor.java:380
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DkbPDFExtractor.java:839
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/EasyBankAGPDFExtractor.java:835
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ErsteBankPDFExtractor.java:1009
[...]

Issue #5518, opened 2026-02-23
Skipped Items - PDF import: Tracking Missing asserts


Nirus2000:
The PDF importer support returning a skipped items for transactions that can be successfully extracted from documents but are not supported by the domain model of Portfolio Performance. One example are taxation of retained (undistributed) income where not payment is recorded and the transaction has a value of zero. A skipped item may link to an incomplete transaction.

---

### Related Work

**Completed**
- #5263 (initial implementation of skipped items)

---

Special ultra thanks to @gamue for pull request https://github.com/portfolio-performance/portfolio/pull/5334 ❤️ 

---

### Missing `assertThat(countSkippedItems(results), is(0L));`

The following test methods are missing the assertion:

```java
assertThat(countSkippedItems(results), is(0L));
```

### Checklist

- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/BaaderBankPDFExtractorTest.java:5811
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/ComputershareExtractorTest.java:36
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:35
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:99
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:163
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:211
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:275
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGExtractorTest.java:318
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/DegiroPDFExtractorTest.java:63
- [ ] name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/DegiroPDFExtractorTest.java:93
[...]

PR #5519, opened 2026-02-23
Adding countSkippedItems test to Deutsche Bank, FinTech, Onvista, Que…

buchen:
…strade

Issue: #5517


buchen:
I have added more countSkippedItems items tests for four importer.
Does it rather make sense to split them by bank?

buchen:
Refers to #5518

Nirus2000:
> Does it rather make sense to split them by bank?

No... add it... it makes no sens... That's mindless mass import.

Issue #4459, opened 2025-01-05
Why "Transfer" transaction has "amount" value?


pfalcon:
Intro: PP's trade matching code has a number of issues. I'd like to enumerate them in one ticket, but even approaching that, there're other issues "on the outskirts". Trying to collect them in one ticket would be problematic, so trying to submit them separately.

When entering "Transfer" transaction for a security holding, there's a dialog with field for number of shares, which is expected. But there's also field to enter per-share quote and/or total amount. What's the purpose of this amount?

To answer that questions properly, it first would make sense to answer questions of "what's the purpose of transfer transaction" and "what is mental model of it"? For example, my mental model was that it's equivalent to Delivery Outbound + Delivery Inbound combo (which are in turn equivalent of Sell/Buy). Then as DO/DI have "amount" values, then maybe it makes some sense to have it for Transfer too.

I was astonished to see that people have completely different expectations of Transfer conceptual model: https://github.com/portfolio-performance/portfolio/issues/4446 . There, the expectation is that "Transfer" is kind of "symlink", which preserves attributes of the original transaction it "points" to, like date and amount. If purpose of "Transfer" is to preserve original amount / cost basis, then it clearly shouldn't have another amount associated with itself.

pfalcon:
While the description tries to approach the question in top-down manner ("let's (re)think what's the purpose of Transfer transaction and its associated amount"), what caused submitting this ticket is actually bottom-up approach of reviewing how PP's code handle the TRANSFER_IN/TRANSFER_OUT transaction types:

CapitalGainsCalculation: Transfer's amount is clearly not used
CostCalculation: Transfer's amount is clearly not used
TradeCollector: This is the most correct trade matching implementation, so less trivial code to analyze than above 2, but seems that Transfer's amount is not used

SharesHeldCalculation, DeltaCalculation: Transfers are altogether ignored, as expected

IRRCalculation: Suddenly, Transfer's amount is used. But how? On TRANSFER_IN, the amount is added with negative sign to IRR calc array, then on TRANSFER_OUT, which happens on the same day, it's added with positive sign, which leads to arrays like:
```
[2024-01-01, 2024-07-01, 2024-07-01, 2024-12-31]
[-101.0, -500.0, 500.0, 200.0]
```
with those same-day negative/positive values expected to cancel each other, and indeed, they do.

[...]

PR #5523, opened 2026-02-25
Send users to documentation in English when available

andrericardo:
Use more specific documentation page when it exists, otherwise fallback to automatically translated German.

image


mierin12:
Good idea.

Andreas, I would like to take the opportunity of this PR to suggest that those reference/link to the manual, in particular English manual, might be also useful in more locations and not only for quote feed ?

I was thinking for example in the 'Transfer' dialog wizard, to be clear on what is the impact on the quote used, or maybe for stock split which are not trivial transaction. 
Or some '?' icon with tooltip. 
It was a little bit discussed here but not too much
https://github.com/portfolio-performance/portfolio/issues/4459

buchen:
> Or some '?' icon with tooltip.

I remember the Eclipse dialogs have some kind of functionality to add (?) there.
Let's start with one dialog to see the user experience.

PR #5525, opened 2026-02-26
Proposal PDF Skipped Items Check JUnit4

Morpheus1w3:


buchen:
I guess you closed it because it was on an older master.. right? Or something because of the StructureValidator? I'd be interested to learn

PR #4664, opened 2025-04-19
Allow collapsing taxonomies on Statement of Assets

JGKle:
Implement a new (optional) Collapse/Expand column on the Statement of Assets page, allowing you to collapse each taxonomy. Collapsed/expanded state is persisted to the GUI settings. 

Closes #4420

Peek 2025-04-19 08-56

PR #5528, opened 2026-02-27
Collapsible Statement of Assets page

### LINKED ISSUE #4420: Collapsible sections/headers on Statement of Assets
It would be very useful if the section/group headers on the Statement of Assets page could be collapsed (i.e. by clicking on the bold lines). Often there are some categories where you want the full detail (i.e. equities), but there are others that consume lots of valuable screen real estate but knowing every individual holding isn't always necessary, making it necessary to scroll around.

For example:
* Groups where values don't fluctuate much, & thus seeing the total is enough for most purposes (i.e. cash, bond ladders, illiquid investments like real estate, etc).
* Holdings with very small values could be grouped together & then collapsed, again saving space for more important holdings to fit on one screen. Supersedes [this](https://github.com/portfolio-performance/portfolio/issues/3268) and [this](https://github.com/portfolio-performance/portfolio/issues/2045), both of which were attempts to address this problem.

JGKle:
Allow taxonomies to be collapsed on the Statement of Assets page.

Reimplements [PR #4664](https://github.com/portfolio-performance/portfolio/pull/4664) using TreeViewer.

Closes https://github.com/portfolio-performance/portfolio/issues/4420

Peek 2026-02-26 19-26

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TreeViewerCSVExporter.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TreeViewerCSVExporter.java
@@ -16,7 +16,9 @@ import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.swt.widgets.Shell;
 
 import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.ui.util.viewers.OptionLabelProvider;
 import name.abuchen.portfolio.ui.util.viewers.SharesLabelProvider;
+import name.abuchen.portfolio.ui.util.viewers.ShowHideColumnHelper;
 import name.abuchen.portfolio.util.TextUtil;
 
 public class TreeViewerCSVExporter extends AbstractCSVExporter
@@ -86,6 +88,19 @@ public class TreeViewerCSVExporter extends AbstractCSVExporter
                     {
                         Long value = labelProvider.getValue(element);
                         return value != null ? Values.Share.format(value) : null;
+                    }
+                };
+            }
+            else if (blp instanceof OptionLabelProvider<?> labelProvider)
+            {
+                var option = viewer.getTree().getColumn(ii).getData(ShowHideColumnHelper.OPTIONS_KEY);
+                labels[ii] = new LabelProvider()
+                {
+                    @SuppressWarnings("unchecked")
+                    @Override
+                    public String getText(Object element)
+                    {
+                        return ((OptionLabelProvider<Object>) labelProvider).getText(element, option);
                     }
                 };
             }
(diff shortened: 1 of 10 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/TreeViewerCSVExporter.java
```diff
@@ -16,11 +16,15 @@
...
 
 import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.ui.util.viewers.Column;
+import name.abuchen.portfolio.ui.util.viewers.OptionLabelProvider;
 import name.abuchen.portfolio.ui.util.viewers.SharesLabelProvider;
 import name.abuchen.portfolio.util.TextUtil;
 
 public class TreeViewerCSVExporter extends AbstractCSVExporter
 {
+    private static final String OPTION_KEY = Column.class.getName() + "_OPTION"; //$NON-NLS-1$
```
buchen: Instead of duplicating the definition here, we should make the original constant visible.

buchen (commented):
Thanks for the contribution.

I cannot remember why the CSV exporter for tables and trees look so different. It could be that the children of a collapsed element might not exist but should be exported.

What I noticed:
*  the CSV export now adds an additional column to reflect the nesting - but in the statement view that does not make sense
* the range column in the CSV exports contains the class name as label. This is not a regression. It was a bug before as well.

But that doesn't have to be part of this pull request

JGKle:
Thank you for the quick merge! 💯 😄 🙏

buchen:
I am thinking about reverting this change for the time being.

The feedback is:
* significantly slower on Windows 10/11 to resize columns
* random collapsing rows under "master data" (we are looking only at the taxonomy change, not a model change)
* behavior of scroll position when updating historical prices (now jumps to the top all the time)
* first column now with unusable screen real estate (all cells reserve space for the arrows even if not displayed)
* contrast of the test of the first column changes (really strange)
* tool tips not shown for column header (unclear - at least on macOS they are displayed)
* maybe related: an "Unknown column layout data" error shows up on Linux: https://forum.portfolio-performance.info/t/nach-update-auf-0-40-1-fehler-unknown-column-layout-data/5470/6

This list within the first 24 hours tells me there are too many annoying things. It looks like it needs some more polishing and maybe an "experimental" flag to roll it out incrementally.

https://forum.portfolio-performance.info/t/wie-einklappbare-seite-vermogensaufstellung-deaktivieren/38997

JGKle:
My [original implementation](https://github.com/portfolio-performance/portfolio/pull/4664) was far less invasive: just an optional additional column with a collapser that disables drawing of the undesired rows - no wide-reaching refactor. But it was rejected. Can we just use that instead?

I've been using it locally on both Windows and Linux for nearly a year. Just took the time to redo it this way because it's quite cumbersome to keep rebasing & doing custom builds after each new release.

buchen:
> My https://github.com/portfolio-performance/portfolio/pull/4664 was far less invasive: just an optional additional column with a collapser that disables drawing of the undesired rows - no wide-reaching refactor. But it was rejected. Can we just use that instead?

Yes. Maybe that is the better choice.

JGKle:
Cool. Hopefully if this one is being reverted, that can be merged at the same time :)

JGKle:
Could we reopen+merge https://github.com/portfolio-performance/portfolio/pull/4664 please?

PR #5530, opened 2026-02-27
Improvement: switch theme automatically

georgemac-labs:
- Add automatic theme switching while the app is running when the theme preference is set to System
- Use polling on macOS and SWT.Settings handling for non-macOS platforms

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/OSXStartupAddon.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/OSXStartupAddon.java
@@ -1,10 +1,16 @@
 package name.abuchen.portfolio.ui.addons;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.annotation.PostConstruct;
+import jakarta.annotation.PreDestroy;
 import jakarta.inject.Inject;
 
 import org.eclipse.core.commands.Command;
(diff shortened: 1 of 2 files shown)


georgemac-labs:
Hi maintainers, I use auto light/dark mode on macOS and noticed PP did not switch when other apps did, so this is my attempt to address that.

Unfortunately, the cross-platform approach didn't work for me – I could only get it to work with a macOS specific implementation.

However, I left the other approach in. If you think the PR makes sense in principle, maybe somebody could check it on Linux/Windows?

buchen:
Just a couple notes:

https://github.com/eclipse-chemclipse/chemclipse/pull/1972 detects it based on background color - might be better than having the AppleInferfaceStyle hard coded (the text might change).

https://github.com/dbeaver/dbeaver/issues/21600 it is make sure that the right theme is picked at the start (not sure if that works for PP at the moment).

And I found the feature request to ask SWT to support theme change on macOS: https://github.com/eclipse-platform/eclipse.platform.swt/issues/1027

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/OSXStartupAddon.java
```diff
@@ -98,6 +120,42 @@ public void widgetSelected(SelectionEvent event)
...
     }
 
+    @PostConstruct
+    public void monitorSystemTheme(Display display)
+    {
+        if (!"Mac OS X".equals(System.getProperty("os.name"))) //$NON-NLS-1$ //$NON-NLS-2$
+            return;
+
+        this.display = display;
+        Boolean initialSystemDarkTheme = detectMacOSDarkMode();
```
buchen, line 130: This is on the UI thread. The other parts it is on the background thread.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/OSXStartupAddon.java
```diff
@@ -98,6 +120,42 @@ public void widgetSelected(SelectionEvent event)
...
+        this.lastSystemDarkTheme = initialSystemDarkTheme != null ? initialSystemDarkTheme.booleanValue()
+                        : UIConstants.Theme.DARK.equals(themeEngineManager.getEngineForDisplay(display).getActiveTheme().getId());
```
buchen, line 132: not sure if getActiveTheme is always non-null. Let's be defensive

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/OSXStartupAddon.java
```diff
@@ -98,6 +120,42 @@ public void widgetSelected(SelectionEvent event)
...
+            {
+                if (OSXStartupAddon.this.display == null || OSXStartupAddon.this.display.isDisposed())
+                    return;
+
+                OSXStartupAddon.this.display.timerExec(THEME_POLL_INTERVAL_MS, this);
+                pollSystemThemeAsync();
+            }
+        };
+
+        this.display.timerExec(THEME_POLL_INTERVAL_MS, themePoller);
```
buchen, line 147: Why have to mechanisms? Could we ScheduledExecutorService.scheduleWithFixedDelay() instead?

buchen:
To be honest, I am still not sure to run a check every 2 seconds only because some users have configured macOS to dynamically change. Therefore, I think we should guard this behind an experimental flag (see Experiments class).

Then I think it is not helpful for future changes, that the code is not split into different classes - ThemeAddon for Linux and Windows, OSXStartupAddon for macOS. Can we refactor this have one class - maybe even a separate TrackSystemThemeChangeAddon or something that has the logic in one place.

georgemac-labs:
No objection to the feedback, but then I'd probably deprioritise this for now. It's only a very minor irritation and I thought it would be quick fix, but evidently it's not  :)   Maybe somebody else will feel like picking it up.

Issue #5397, opened 2026-01-28
[PDF] Scalable: tax refunds not shown correctly in Sell-events


gamue:
### Affected Bank

Scalable

### Extracted Text from PDF

```shell
PDFBox Version: 3.0.6
Portfolio Performance Version: 0.81.4
System: win32 | x86_64 | 21.0.5+11-LTS | Azul Systems, Inc.
-----------------------------------------
Scalable Capital Bank GmbH • Seitzstraße 8e • 80538 München • Deutschland 
HoVy zZrElRq
LT hAyXauxN 3 
23615 rzOGqoWImWiz RLemA 29.12.2025
Deutschland Seite 1 / 2
Wertpapierabrechnung 
für Kundenauftrag 
Typ LIMIT Order RgWCVXnhH0pG3Xr
Ausführung 29.12.2025 12:38:52 Geschäft 12439pyBKDv34037
Ausführungsplatz EIX Lagerland Deutschland
Depot 6505481522 Verwahrart Girosammelverwahrung
Typ Wertpapier Anzahl Kurs Betrag
Verkauf Hershey 70,00 Stk. 155,68 EUR 10.897,60 EUR
US4278661081
Steuern +366,75 EUR
Gutschrift 11.264,35 EUR
Der Betrag wird mit dem Verrechnungskonto gn86339910422583872441 (Valuta: 31.12.2025) verrechnet. 
Bitte überprüfen Sie die Informationen auf Richtigkeit und melden Sie etwaige Einwände unverzüglich bei uns.
Verwenden Sie dafür den Menüpunkt Support im Kundenbereich. 
Scalable Capital Bank GmbH HRB 217778 Geschäftsführer: Aufsichtsrat: Seite 
Seitzstraße 8e Amtsgericht München Florian Prucker, Martin Krebs, Patrick Olson (Vorsitzender) 
80538 München USt.-Id. Nr.: DE300434774 Dirk Franzmeyer 1 / 2
Wertpapierabrechnung
Verkauf 70,00 Stk. Hershey 
Ermittlung steuerrelevanter Erträge
Typ Betrag
Verlust -1.390,50 EUR
Verrechnung mit vorherigen Gewinnen +1.390,50 EUR
Zu versteuern 0,00 EUR
Erstattete Steuern 
Typ Betrag
Kapitalertragsteuer -347,63 EUR
Solidaritätszuschlag -19,12 EUR
Erstattete Steuern 366,75 EUR
[...]

PR #5390, opened 2026-01-26
Modify Scalable PDF-impoter to support additional formats

gamue:
Closes #5356
Closes #5392
Closes #5419 
Closes #5397 

### Knockout
- Thought adding "KnockOut" to the test-file-name makes it easier to see special cases, let me know if that's not wanted. I understand that this is not common, but saw it in some tests (e.g. Wir-Bank, Tiger Broker, degiro)
- Replaced the `optionalOneOf`-block with a single-section as not necessary 
- The `firstRelevantLine`-Block was puzzling me again, was almost using an empty `Block()` again :D  Main issue is that "Knock Out" is coming before "Berechtigtes Wertpapier" and Scalable seems to place the keywords on different places, so guess using the Page-keyword like in the other methods makes sense and simplifies the regex.

### Vorabpauschale
- Add additional test-case from https://github.com/portfolio-performance/portfolio/issues/5392
- Also added document with amount=0 to check that proper error message is returned like in the other extractors.
- Added additional test-case from forum: https://forum.portfolio-performance.info/t/pdf-import-von-scalable-capital-bank/33088/105

PR #5533, opened 2026-02-28
Modify PDFExtractor to support skippedItems (Part 1)

### LINKED ISSUE #5517: Skipped Items - PDF import: Tracking new usage of Skipped Item

gamue:
Closes #5517 

Especially in the AccountStatement-methods that work with very generic regexs like `"^[\\d]{2} [\\p{L}]{3,4}([\\.]{1})?.*$"` it's not possible to change the `return null`, otherwise there will be lots of `SkippedItem`. I think best to keep it as it. wdyt?

- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java:685
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/CreditSuisseAGPDFExtractor.java:350
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/MLPBankingAGPDFExtractor.java:501
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/N26BankAGPDFExtractor.java:68
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NIBCBankPDFExtractor.java:231
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostfinancePDFExtractor.java:985
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/QuirinBankAGPDFExtractor.java:721
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java:904
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SBrokerPDFExtractor.java:1166
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java:767
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SwissquotePDFExtractor.java:644
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java:987
- [x] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/WealthsimpleInvestmentsIncPDFExtractor.java:169

--- 
The following ones will be handled in a follow-up PR:
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java:1093
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DADATBankenhausPDFExtractor.java:745
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DZBankGruppePDFExtractor.java:653
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DegiroPDFExtractor.java:361
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DekaBankPDFExtractor.java:438
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DeutscheBankPDFExtractor.java:692
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/Direkt1822BankPDFExtractor.java:380
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DkbPDFExtractor.java:839
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/EasyBankAGPDFExtractor.java:835
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ErsteBankPDFExtractor.java:1009
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FinTechGroupBankPDFExtractor.java:841
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/LimeTradingCorpPDFExtractor.java:115
- [ ] name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TigerBrokersPteLtdPDFExtractor.java:194

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java
@@ -717,7 +717,7 @@ public class ScalableCapitalPDFExtractor extends AbstractPDFExtractor
                             var item = new TransactionItem(t);
 
                             if (t.getCurrencyCode() != null && t.getAmount() == 0)
-                                ctx.markAsFailure(Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
+                                return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
 
                             return item;
                         });
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/scalablecapital/ScalableCapitalPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/scalablecapital/ScalableCapitalPDFExtractorTest.java
@@ -21,6 +21,7 @@ import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.purchase;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.removal;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.sale;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.security;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.skippedItem;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.taxRefund;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.taxes;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.withFailureMessage;
@@ -2213,10 +2214,10 @@ public class ScalableCapitalPDFExtractorTest
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
         assertThat(countBuySell(results), is(0L));
-        assertThat(countAccountTransactions(results), is(1L));
+        assertThat(countAccountTransactions(results), is(0L));
         assertThat(countAccountTransfers(results), is(0L));
-        assertThat(countItemsWithFailureMessage(results), is(1L));
-        assertThat(countSkippedItems(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(1L));
         assertThat(results.size(), is(2));
         new AssertImportActions().check(results, "EUR");
 
@@ -2226,15 +2227,8 @@ public class ScalableCapitalPDFExtractorTest
                         hasName("Scalable MSCI AC Wld Xtrackers"), //
                         hasCurrencyCode("EUR"))));
 
-        // check taxes transaction
-        assertThat(results, hasItem(withFailureMessage( //
-                        Messages.MsgErrorTransactionTypeNotSupportedOrRequired, //
-                        taxes( //
-                                        hasDate("2026-01-02"), hasShares(430), //
-                                        hasSource("Vorabpauschale05.txt"), //
-                                        hasNote(null), //
-                                        hasAmount("EUR", 0.00), hasGrossValue("EUR", 0.00), //
-                                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00)))));
+        // check skipped item
+        assertThat(results, hasItem(skippedItem(Messages.MsgErrorTransactionTypeNotSupportedOrRequired)));
     }
 
     @Test


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/raiffeisenbankgruppe/RaiffeisenbankgruppePDFExtractorTest.java
```diff
@@ -1656,8 +1656,8 @@ public void testKontoauszug01()
         assertThat(countAccountTransactions(results), is(17L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(17));
+        assertThat(countSkippedItems(results), is(18L));
```
gamue: the skipped count in this tests are quite high, let's recheck that

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/postfinance/PostfinancePDFExtractorTest.java
```diff
@@ -2782,8 +2782,8 @@ public void testKontoauszug05()
         assertThat(countAccountTransactions(results), is(27L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(27));
+        assertThat(countSkippedItems(results), is(97L));
```
gamue: same here

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
```diff
@@ -1967,8 +1967,8 @@ public void testKontoauszug06()
         assertThat(countAccountTransactions(results), is(20L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(1L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(20));
+        assertThat(countSkippedItems(results), is(44L));
```
gamue: recheck

buchen:
Hi @gamue, I propose not add more banks to this pull request. Instead, let's start a new pull request for the other banks. This change is not a simple "do a refactoring without semantic change". It needs more review, and then it is easier to do this on smaller diffs.

Are the banks here in the pull request don from your point of view? Let me know when I should take a look at the pull request

gamue:
> I propose not add more banks to this pull request. Instead, let's start a new pull request for the other banks. This change is not a simple "do a refactoring without semantic change". It needs more review, and then it is easier to do this on smaller diffs.

Agree, and actually had the same idea :) Was already reviewing it more in depth on my own. Think the current state should be fine for review, just a few notes/thoughts:
- The Comdirect is one that we might want to keep as null, but couldn't really understand which is the skipped-item in the tests I've checked
- Might be worth defining what should be done when `return null` was checked and is fine, to differ between this and unchecked ones. I left a comment at a few. Especially the TradeRepublic one is hard as this uses very generic regex and would result in many SkipItems
- **EDIT:** Was rechecking something and noticed that in some cases that are currently result in an failure, it might be better to use skippedItem instead. As an example I've adjusted the Vorabpauschale-Import at Scalable, when there's 0 Euro. Looks like `return null` might not be the only option to check

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java
```diff
@@ -754,9 +754,11 @@ private void addTaxLostAdjustmentTransaction(Map<String, String> context, Docume
                         })
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
-                            return null;
+                                return item;
+
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 761: @gamue 
I am reviewing the changes to the PDF importer (and started with Scalable - one step at a time).

I am not sure about the side effects of this change.
The block pattern is a simple ```^Erstattete Steuern[\\s]*$``` and then there is only one matching, but optional element. There is no test that hits the skipped item.

You introduced this TaxLostAdjustmentTransaction. What was the reason to mark it optional() in the first place?

If we convert this to a SkippedItem, then - by definition - PP will stop attempting to try other blocks. My gut feeling is we should return a skipped item only if we reliably know that we parsed a transaction. But maybe you have example cases?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java
```diff
@@ -717,7 +717,7 @@ private void addAdvanceTaxTransaction()
                             var item = new TransactionItem(t);
 
                             if (t.getCurrencyCode() != null && t.getAmount() == 0)
-                                ctx.markAsFailure(Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
+                                return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 720: This is exactly the kind of transactions we want to mark as skipped.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ScalableCapitalPDFExtractor.java
gamue: I digged a little bit deeper as I couldn't remember that change and actually Nirus did this change (https://github.com/portfolio-performance/portfolio/pull/5390/changes/bde411bfd6c202950a561a69f0aabee01d2163e1 ) background is that back then there's been multiple Scalable change-requests and they got bundled in this change. This particular was https://github.com/portfolio-performance/portfolio/issues/5397

buchen: Thanks for the research. I tend to be conservative here: keeping it as it is does not make it worse. Once we have an example, we can update the section. I would think that involves removing the optional(). I will also collect these discussions in the main issue.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/comdirect/ComdirectPDFExtractorTest.java
```diff
@@ -999,8 +999,8 @@ public void testWertpapierVerkauf01()
         assertThat(countAccountTransactions(results), is(0L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(2));
+        assertThat(countSkippedItems(results), is(1L));
+        assertThat(results.size(), is(3));
```
buchen, line 1003: @gamue Looking at the document "Verkauf01.txt", I do not see the reason to expect an additional SkippedItem here. It looks to me like a normal "sale" which, yes, does not pay taxes because it is applied to the "Freistellungsauftrag". But as I user, I would not understand to why a skipped item shows up in the import.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/comdirect/ComdirectPDFExtractorTest.java
```diff
@@ -1199,8 +1199,8 @@ public void testWertpapierVerkaufMitSteuerbehandlung01()
...
```
buchen, line 1203: Also here. The document is successfully parsed. It contains the sale transaction with the right taxes. There should be no additional skipped item here.

The more I think about, I believe the check for 0 amount is in a block `addSellWithNegativeAmountTransaction` which is creating a fees transaction if the sale amount is negative. But it should silently continue in case there is no such transaction. Therefore this is not a case to return a SkippedItem.

I do not have time now to check the other test cases.

@gamue @Nirus2000 
What is your opinion?

buchen: Ok, I did some more research and I am convinced now that we must **not** put a SkippedItem in the `addSellWithNegativeAmountTransaction` block.

The logic is this:
* The purchase order is parsed in `addBuySellTransaction`
* If it is a sale but - due to fees - the customer actually has to pay the bank to sell (fees > proceeds), then the transaction is marked as "negative" ([see L267](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java#L267)) and parsing of fees is skipped ([see L2039](https://github.com/portfolio-performance/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java#L2039))
* The `addSellWithNegativeAmountTransaction` is parsing the same section again - now creating a separate fee transaction. Because the optional section looks for "Zu Ihren Lasten..." it will produce only a valid fee transaction if there is an amount to pay.
* In the case of a payable amount for a sale, there are two transactions created.

Clever.

I will add a comment for our future selves.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/creditsuisseag/CreditSuisseAGExtractorTest.java
```diff
@@ -181,8 +181,8 @@ public void testWertpapierVerkauf01()
         assertThat(countAccountTransactions(results), is(0L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(2));
+        assertThat(countSkippedItems(results), is(1L));
+        assertThat(results.size(), is(3));
```
buchen, line 185: Again, looking at the original document and the code, we should not return a skipped item here.

We have here a "two pass" parsing logic: First, the purchase or sale, then - optionally - a fee refund (when purchasing via the Internet). One could argue why this is a separate transaction and not offset with the fees of the original transaction, but that is the way it is implemented. We do not need to inform the user about skipping a non-existing fee refund.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/MLPBankingAGPDFExtractor.java
```diff
@@ -498,9 +499,11 @@ private void addTaxLostAdjustmentTransaction(Map<String, String> context, Docume
                         })
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
-                            return null;
+                                return item;
+
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen: Same picture here: if, and only if, the document contains a tax refund due to a realized loss, we need to create an additional transaction. The root cause being that PP cannot add a tax refund directly to a sale transaction. But it does not make sense to inform the user if we "skip" a non-existing tax refund. In this case, returning null is the right thing.

The more importer I look at, the more I am convinced that the skipped item is used only in a few places. That is good. It means that most transactions are supported after all :-)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/N26BankAGPDFExtractor.java
```diff
@@ -65,10 +66,11 @@ private void addAccountStatementDepositRemovalTransaction()
                         })
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
+                                return item;
 
-                            return null;
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen: Here we could argue to add a skipped item. The sections are mandatory, so it must match a value. And if the value is zero, then it could be skipped. But there is no example at all. And it is unclear to me if entries with 0 would be added to the statement at all. Maybe if we have an account statement (Kontoauszug) where the user has no deposits anymore?

Again, without any more examples, it is a hard call. Therefore I propose to be conservative and just return null.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/MLPBankingAGPDFExtractor.java
buchen, line 506: Same picture here: if, and only if, the document contains a tax refund due to a realized loss, we need to create an additional transaction. The root cause being that PP cannot add a tax refund directly to a sale transaction. But it does not make sense to inform the user if we "skip" a non-existing tax refund. In this case, returning null is the right thing.

The more importer I look at, the more I am convinced that the skipped item is used only in a few places. That is good. It means that most transactions are supported after all :-)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/N26BankAGPDFExtractor.java
buchen, line 73: Here we could argue to add a skipped item. The sections are mandatory, so it must match a value. And if the value is zero, then it could be skipped. But there is no example at all. And it is unclear to me if entries with 0 would be added to the statement at all. Maybe if we have an account statement (Kontoauszug) where the user has no deposits anymore?

Again, without any more examples, it is a hard call. Therefore I propose to be conservative and just return null.

gamue: > Therefore I propose to be conservative and just return null.

yes, makes sense

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/NIBCBankPDFExtractor.java
```diff
@@ -228,9 +229,10 @@ private void addTaxLostAdjustmentTransaction(Map<String, String> context, Docume
                 .assign((t, v) -> t.setNote(trim(v.get("note"))))
 
                 .wrap(t -> {
+                    var item = new TransactionItem(t);
                     if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                        return new TransactionItem(t);
-                    return null;
+                        return item;
+                    return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 235: This code is not executed at all. There is not event an example for the tax refund, there is no example when this is not imported. Most likely it is again part of an existing document if, and only if, there is a tax refund. Therefore reporting a skipped transaction is probably irritating to the user.

gamue: if not needed maybe remove it?

buchen: Well, it is pattern that we see in other importer. Therefore I think there is a good chance that this actually works (and we just miss the test case). We should keep it.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PostfinancePDFExtractor.java
```diff
@@ -984,6 +984,8 @@ else if (note.contains("KAUF/DIENSTLEISTUNG"))
                         .wrap(t -> {
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
                                 return new TransactionItem(t);
+
+                            // nothing matched, so might be one of the others
```
buchen, line 988: I agree. The block pattern for removals, deposits, fees, and interest is identical. Each block is testing each line. There is nothing meaningful to skip.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/raiffeisenbankgruppe/RaiffeisenbankgruppePDFExtractorTest.java
```diff
@@ -1852,8 +1852,8 @@ public void testKontoauszug02()
         assertThat(countAccountTransactions(results), is(31L));
         assertThat(countAccountTransfers(results), is(0L));
         assertThat(countItemsWithFailureMessage(results), is(0L));
-        assertThat(countSkippedItems(results), is(0L));
-        assertThat(results.size(), is(31));
+        assertThat(countSkippedItems(results), is(1L));
+        assertThat(results.size(), is(32));
```
buchen, line 1856: Looking at the document, I count only 31 transactions. The test is correctly identifying all 31 one. We should not add a skipped item for a transaction that is not in the document. I think here the parsing logic for "Abschluss" is distributed across multiple blocks. One of them matches the transaction - but not every flavor is in the document. It is okay to return null

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -995,9 +997,10 @@ private void addAccountStatementTransactions()
                         })
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
-                            return null;
+                                return item;
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 1003: See above the comment on the test case. The "Abschluss" section has a lot of variations. There are multiple blocks attempting to parse this. If one of them does not match, it does not mean we have found a transaction that PP does not support, i.e., needs to be skipped.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SBrokerPDFExtractor.java
```diff
@@ -1790,9 +1790,10 @@ private void addTaxReturnBlock(DocumentType type)
                                                         .assign((t, v) -> t.setNote(trim(v.get("note")))))
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
-                            return null;
+                                return item;
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 1796: Not a skip.

In case the document does not contain a tax refund due to realized loss, nothing to skip.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SwissquotePDFExtractor.java
```diff
@@ -641,9 +641,11 @@ private void addAccountStatementTransaction()
                         })
 
                         .wrap(t -> {
+                            var item = new TransactionItem(t);
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                                return new TransactionItem(t);
-                            return null;
+                                return item;
+
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 648: This block looks at Depotgebühren. There is no test case for skipping at all. The section is not marked as optional. I see this as "defensive coding" but I am not sure that branch will ever run.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/SwissquotePDFExtractor.java
```diff
@@ -674,9 +676,11 @@ private void addAccountStatementTransaction()
...
```
buchen, line 683: This block looks at Sollzinsen. There is no test case for skipping at all. The section is not marked as optional. I see this as "defensive coding" but I am not sure that branch will ever run.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
```diff
@@ -2686,6 +2686,8 @@ private void addAccountStatementTransaction_Format02()
                         .wrap(t -> {
                             if (t.getCurrencyCode() != null && t.getAmount() != 0)
                                 return new TransactionItem(t);
+
+                            // nothing matched, so might be one of the others
```
buchen, line 2690: Agree. The account statement parser uses "multi pass" approach, e.g., going over the same entries multiple times. A "null" does not indicate that the transaction is not supported.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/WealthsimpleInvestmentsIncPDFExtractor.java
```diff
@@ -166,9 +166,11 @@ private void addDepotStatementTransaction()
                 })
 
                 .wrap(t -> {
+                            var item = new TransactionItem(t);
                     if (t.getCurrencyCode() != null && t.getAmount() != 0)
-                        return new TransactionItem(t);
-                    return null;
+                                return item;
+
+                            return new SkippedItem(item, Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
```
buchen, line 173: Again, no test case. The section is not marked optional.

buchen:
I am closing this pull request. I have reviewed all places. I have included some skipped items, but most of the locations are not valid items for skipped items. I have added comments in the document. But mainly would use the original Github issue for tracking.

Thank you very much for creating the pull request.

PR #5534, opened 2026-02-28
Improvement: IBFlex allow re-marking "OK" entry for import

georgemac-labs:
I noticed in the IBFlex process you can right-click to mark an "OK" transaction as "Do not import" but you cannot then revert that flag! That seems like a gap to me.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
@@ -659,10 +659,15 @@ public class ReviewExtractedItemsPage extends AbstractWizardPage implements Impo
             // marked as to be imported by the user
             atLeastOneImported = atLeastOneImported || entry.isImported();
 
-            // an entry will not be imported if it marked as not to be
-            // imported *or* if it has a WARNING code (e.g. is a duplicate)
+            // show "Do Import" only for entries the user can actually import
+            // (not dependency-blocked, not ERROR/SKIP) that aren't currently
+            // marked for import
+            boolean canImport = entry.getMaxCode() == Code.OK || entry.getMaxCode() == Code.WARNING;
+            boolean notDependencyBlocked = entry.getSecurityDependency() == null
+                            || entry.getSecurityDependency().isImported()
+                            || entry.getSecurityOverride() != null;
             atLeastOneNotImported = atLeastOneNotImported
-                            || (!entry.isImported() && (entry.getMaxCode() == Code.WARNING));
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -659,10 +659,12 @@ private void showContextMenu(IMenuManager manager)
...
 
-            // an entry will not be imported if it marked as not to be
-            // imported *or* if it has a WARNING code (e.g. is a duplicate)
+            // an entry can be marked for import again if warnings may be
+            // overridden or if the user explicitly disabled an otherwise valid
+            // entry
             atLeastOneNotImported = atLeastOneNotImported
-                            || (!entry.isImported() && (entry.getMaxCode() == Code.WARNING));
+                            || (!entry.isImported() && (entry.getMaxCode() == Code.WARNING
+                                            || (entry.getMaxCode() == Code.OK && entry.isExplicitlyNotImported())));
```
buchen: Good catch. But couldn't we simplify even further:

If it the item
a) is currently not imported AND
b) it could be imported (e.g. the maxcode is WARNING or OK)
then the user can change the decision. 
If it is FAILURE, of course, the user cannot make that decision. Same is true for SKIP (= transactions not supported by PP)

```java
            atLeastOneNotImported = atLeastOneNotImported || (!entry.isImported()
                            && (entry.getMaxCode() == Code.WARNING || entry.getMaxCode() == Code.OK));
```

georgemac-labs: I don't think so, because user choice is not the only reason that entry.isImported() can be false – it includes a dependency check.

So if you decide based on maxCode, the app might allow the user to explicitly mark an entry to be imported, but then it won't be. I think that's a less good UI.

georgemac-labs: Incidentally, I would say this area is slightly muddled. There are two questions: whether the entry CAN be imported and whether the user WANTS it imported.

The concept seems to be that "can" is represented by maxCode and the icon.

However, a piece of "can" has ended up mixed up with "want" in the code (`isImported()`), which is presumably what misled you into thinking the code could be simpler. And it's leaked into the UI, too, using strikethrough for both user choice AND for entries blocked by failed dependencies.

georgemac-labs: @buchen did I address your comment? Anything further to do to get this merged?

PR #5549, opened 2026-03-04
add Top 5 / Flop 5 securities widget

mierin12:
Closes https://github.com/portfolio-performance/portfolio/issues/656
Issue https://forum.portfolio-performance.info/t/gewinner-verlierer-liste/27079
Issue https://forum.portfolio-performance.info/t/auswertung-performance-top-contributors-in-desktop-anwendung/27812

Hello,
This is a proposition to add a Top/Flop widge with TWR, TWR annualized, IRR or Abs Perf as performance metrics. I have taken some structures from the recent dividend widget and a bit of Claude help.

Examples from kommer file : 
Capture d’écran 2026-03-04 204422

I have added a "subtitle" showing if the selection option is Winner or Loser and the performance metric chosen, I think it gives a clearer view than if it is on the tooltip only.
I have one issue with this subtitle : if the widget is created on a small column, the long text does not appear fully. It does appear fully on several lines on widget update. 
Another thing, maybe a detail, it might give the feeling that a right click on this bold subtitle it can open the context menu to change the widget options, but only a right click on the title can do that (in general, would it make sense to extend this right click action to the whole widget, and not only the titles ? Not for chart widget, but for all the other ?). 

I was unsure if the 5 limit should be hardcoded or user selectable (do they want a Top 3 ?).

I was also unsure of the name to give. I like Top Flop (and it is still in the code file name) but I was a bit afraid of the accuracy of the translation of such short words. And maybe it would be too informal ? So in the end I followed the "Gewinner-/Verlierer-Liste" found in the forum, but as always with naming I am never sure of myself.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_de.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages_de.properties
@@ -1923,7 +1923,7 @@ LabelWeeksAgo = {0,choice,0#vor 0 Wochen|1#vor 1 Woche|1<vor {0,number} Wochen}
 
 LabelWinner = Gewinner
 
-LabelWinnerLoserList = Gewinner-/Verlierer-Liste 
+LabelWinnerLoserList = Gewinner-/Verlierer-Liste
 
 LabelWithQualifier = {0} ({1})
 


buchen:
Nice. Let me have a look at it - hopefully over the weekend - I have a similar widget in the mobile application but never came around to porting it to the desktop application. Ideally we use the same concept (and keys) then one can define on the desktop and use on the mobile.

buchen:
Not exactly the same thing, but I ported the widget implementation from the mobile application back to the desktop. That allows to create the same widget on the desktop and then view it on the mobile application as well.

Bildschirmfoto 2026-04-10 um 08 17 12

mierin12:
Thanks !
Do you see the postive/negative colors in dark theme ? They only appear on light theme for me. Testing again my proposal it seems it also did not appear in dark theme.

Issue #5548, opened 2026-03-04
[PDF] PDF Import for Oberbank


iAppDeveloper88:
### Affected Bank

Oberbank

### Extracted Text from PDF

```shell
PDFBox Version: 3.0.6
Portfolio Performance Version: 0.81.5
System: macosx | aarch64 | 21.0.5+11-LTS | Azul Systems, Inc.
-----------------------------------------
Depotumsätze
Depotbezeichnung Mustermann Max
Bezeichnung Depotguthaben
Depotnummer 111-1111.11
Ihre Auswahl   
Zeitraum 01.01.2010 bis 04.03.2026
Betrag -
ISIN/WKN/Valor Alle
Währung Alle
Wertpapiertyp Alle
Geschäftsart Alle
Bezeichnung/ISIN Menge Kurs Endbetrag Schluss-/ Abrechnung 
KONTO Kassatag Datum 
Dividende
VISA Inc. 10 0,6700 4,11 EUR 10.02.2026 730071
Reg. Shares Class A DL 
-,0001 Stück USD 222222222 02.03.2026 02.03.2026
US92826C8394
Dividende
Apple Inc. 24 0,2600 3,79 EUR 09.02.2026 402816
Registered Shares o.N. Stück USD 222222222 12.02.2026 12.02.2026
US0378331005
KEST-Belastung
3 Banken Mensch&Umwelt 
Aktienf.(R) 142 0,0000 0,00 EUR 02.02.2026 352431
Miteigentumsanteile - 
Thesaurierend Stück EUR 222222222 03.02.2026 03.02.2026
AT0000A23YE9
Dividende
Salesforce Inc. 12 0,4160 3,10 EUR 18.12.2025 550001
Registered Shares DL -,001 Stück USD 222222222 08.01.2026 08.01.2026
US79466L3024
Oberbank AG, Rechtsform: Aktiengesellschaft, Sitz: Linz, Firmenbuch Nr. FN 79063 w, Landesgericht Linz 
 
Seite 1/36
Bezeichnung/ISIN Menge Kurs Endbetrag Schluss-/ Abrechnung 
KONTO Kassatag Datum 
Kauf
3 Banken Aktienfonds-
Selektion 2 26,9297 49,94 EUR 05.08.2021 882406
Miteigentumsanteile - 
Thesaurierend Stück EUR 222222222 09.08.2021 05.08.2021
AT0000784830
Kauf
3 Banken Aktienfonds-
[...]

PR #5559, opened 2026-03-08
PDF Extractor for Oberbank documents

iAppDeveloper88:
Started working on issue #5548
So far buy and sell documents are working. I would appreciate a short feedback if the implementation is OK before i continue with the other documents.
This is just a draft, so I did not check formatting and comments too much.


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/oberbank/OberbankPDFExtractorTest.java
```diff
@@ -0,0 +1,187 @@
...
+        assertThat(countBuySell(results), is(1L));
+        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
+        assertThat(results.size(), is(2));
+        new AssertImportActions().check(results, "EUR");
+
+        // check security
+        var security = results.stream().filter(SecurityItem.class::isInstance).findFirst()
```
gamue: this is the old assertion style, please use the newer one. See https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md#pdf-importers for an example. You can also check other (recent) tests-methods e.g. in Scalable

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+    private void addBuySellTransaction()
+    {
+        final var type = new DocumentType("(Wertpapier-Abrechnu\\s*n\\s*g\\s*" //
+                        + "(Kauf" //
+                        + "|Verkauf))");
+        this.addDocumentTyp(type);
+
+        var pdfTransaction = new Transaction<BuySellEntry>();
+
+        // var firstRelevantLine = new
```
gamue: please remove not necessary comments

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+        pdfTransaction //
+
+                        .subject(() -> {
+                            var portfolioTransaction = new BuySellEntry();
+                            portfolioTransaction.setType(PortfolioTransaction.Type.BUY);
+                            return portfolioTransaction;
+                        })
+
+                        // Is type --> "Verkauf" change from BUY to SELL
+                        .section("type").optional() //
```
gamue: is that really `optional`?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+                            if ("Verkauf".equals(v.get("type")))
+                                t.setType(PortfolioTransaction.Type.SELL);
+                        })
+
+                        .oneOf( //
+                        // @formatter:off
+                                        // Wertpapiernummer Bezeichnung Nominale/Stück
+                                        // CA09228F1036 BlackBerry Ltd. Zugang Stk .              14,00
+                                        // Registered Shares o.N.
+                                        // Kurs 19,098 EUR Kurswert EUR              267,37
```
gamue: think this line isn't necessary for the security, or?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("isin", "name", "nameContinued", "local", "shares") //
+                                                        .find("^Wertpapiernummer Bezeichnung Nominale/St.ck$") //
+                                                        .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<name>.*) (Zugang|Abgang) (?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
```
gamue: maybe add Abgang to the comment above as well as example

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+
+                        .oneOf( //
+                        // @formatter:off
+                                        // Wertpapiernummer Bezeichnung Nominale/Stück
+                                        // CA09228F1036 BlackBerry Ltd. Zugang Stk .              14,00
+                                        // Registered Shares o.N.
+                                        // Kurs 19,098 EUR Kurswert EUR              267,37
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("isin", "name", "nameContinued", "local", "shares") //
```
gamue: please see other extractors (like `ScalableCapitalPDFExtractor`). It's important to have the sections split in the same way, even that adds code-duplication. `shares `should be it's own section

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,204 @@
...
+                        // @formatter:off
+                        // Auftrags-Nr. 999999-28.01.2021
+                        // @formatter:on
+                        .section("note").optional() //
+                        .match("^(?<note>Auftrags-Nr\\. \\d+)-[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}$") //
+                        .assign((t, v) -> t.setNote(trim(v.get("note"))))
+
+                        // @formatter:off
+                        // Kupon 4,55 % jährlich Stückzinsen f. 166 Tage EUR              165,55
+
```
gamue, line 222: unnecessary blank line

gamue (commented):
Thanks for contributing. I added a few comments from what I know that should be adjusted.

Personally think it'll be fine to merge the new extractor step-by-step, however I'm not a team-member, so not my decision ;)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
iAppDeveloper88: I had it like that first but I changed it as it looked so wrong ignoring shares in this regex just have to the exact same one again for it. But I understand that its important for consistency.

iAppDeveloper88:
I think it would make sense to merge a first version as soon as all the single transaction documents are done. I don't think it will take that long to finish that anyway.

iAppDeveloper88:
I just noticed that there is already a DreiBankenEDVPDFExtractor. Oberbank is part of the "3-Banken-Gruppe", but this PDFExtractor is only covering BKS Bank. 
Should the logic for Oberbank be included in DreiBankenEDVPDFExtractor instead of creating a new PDFExtractor as the documents are very similar (but not identical)?

gamue:
> Die „3 Banken Gruppe“ besteht mit Oberbank, BTV Vier Länder Bank AG und BKS Bank aus drei selbstständigen und unabhängigen Regionalbanken. Die 3 Banken sind im Verbund in allen österreichischen Bundesländern und im angrenzenden Ausland aktiv.

They sound to be independent. There are similar cases already like Audi-Bank and Volkswagen-Bank, which have connections and use similar patterns. however they might split in the future or similar, so better to keep it as own classes imo.
Maybe even rename the other one to BksBank when you say it's just for this atm

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        // Is type --> "Verkauf" change from BUY to SELL
+                        // Is type --> "Rücknahme Fonds" change from BUY to SELL
+                        .section("type") //
+                        .match("^Wertpapier-Abrechnu\\s*n\\s*g\\s+" //
+                                        + "(?<type>(Kauf" //
+                                        + "|Verkauf" //
+                                        + "|Ausgabe Fonds" //
+                                        + "|R.cknahme Fonds" //
+                                        + "))$") //
```
Nirus2000:
```suggestion
                        .match("^Wertpapier\\-.*" //
                                        + "(?<type>(Kauf" //
                                        + "|Verkauf" //
                                        + "|Ausgabe Fonds" //
                                        + "|R.cknahme Fonds" //
                                        + "))$") //
```

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                                        // AT000B127337 Oberbank AG Zugang EUR            8.000,00
+                                        // Nachr. Anleihe 2023-2031
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("isin", "name", "nameContinued") //
+                                                        .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<name>.*) (Zugang|Abgang) [A-Z]{3}\\s+[\\.,\\d]+$") //
+                                                        .match("^(?<nameContinued>.*)$") //
+                                                        .assign((t, v) -> {
+                                                            t.setSecurity(getOrCreateSecurity(v));
+                                                        }))
```
Nirus2000, line 120: Missing currency to create a security ... see getOrCreateSecurity function

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        // @formatter:off
+                                        // CA09228F1036 BlackBerry Ltd. Zugang Stk .              14,00
+                                        // AT0000730007 ANDRITZ AG Abgang Stk.               95,00
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("local", "shares") //
+                                                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .* (Zugang|Abgang) (?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
+                                                        .assign((t, v) -> {
+                                                            t.setShares(asShares(v.get("shares")));
+                                                        }),
```
Nirus2000: no "local" ... for what?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                                                            t.setSecurity(getOrCreateSecurity(v));
+                                                        }))
+                        .oneOf( //
+                        // @formatter:off
+                                        // CA09228F1036 BlackBerry Ltd. Zugang Stk .              14,00
+                                        // AT0000730007 ANDRITZ AG Abgang Stk.               95,00
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("local", "shares") //
+                                                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .* (Zugang|Abgang) (?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
```
Nirus2000:
```suggestion
                                                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .* (Zugang|Abgang) (?<local>Stk)[\\s]*\\.[\\s]+(?<shares>[\\.,\\d]+)$") //
```

Nirus2000: Use everywhere ... change \\s* to [\\s]*

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        .section("note1", "note2", "note3").optional() //
+                        .match("^Kupon [\\.,\\d]+ % .* (?<note1>St.ckzinsen .* [\\d]+ Tage).* (?<note3>[A-Z]{3})\\s+(?<note2>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setNote(concatenate(t.getNote(), v.get("note1"), " | "));
+                            t.setNote(concatenate(t.getNote(), v.get("note2"), ": "));
+                            t.setNote(concatenate(t.getNote(), v.get("note3"), " "));
+                        })
+
+                        .conclude(ExtractorUtils.fixGrossValueBuySell())
+
```
Nirus2000, line 207: Why? I didnt see any calculation fixes...

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+        addTaxesSectionsTransaction(pdfTransaction, type);
+        addFeesSectionsTransaction(pdfTransaction, type);
+    }
+
+    private void addDividendTransaction()
+    {
+        final var type = new DocumentType("(Wertpapier-Abrechnu\\s*n\\s*g\\s+" //
+                        + "(Dividende" //
+                        + "|Ausschüttung))");
+        this.addDocumentTyp(type);
```
Nirus2000: See regex https://github.com/portfolio-performance/portfolio/pull/5559/changes#r2917777787

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+
+        var pdfTransaction = new Transaction<AccountTransaction>();
+
+        var firstRelevantLine = new Block("^Wertpapier-Abrechnu\\s*n\\s*g\\s+" + "(Dividende" + "|Ausschüttung)$");
```
Nirus2000: See regex https://github.com/portfolio-performance/portfolio/pull/5559/changes#r2917777787

Nirus2000: "Ausschüttung" to "Aussch.ttung"

see contibuting rules

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+        pdfTransaction //
+
+                        .subject(() -> {
+                            var accountTransaction = new AccountTransaction();
+                            accountTransaction.setType(AccountTransaction.Type.DIVIDENDS);
+                            return accountTransaction;
+                        })
+
+                        // TODO: Ausschüttung / Thesaurierung needs to be
+                        // handled
```
Nirus2000: ToDO what?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        })
+
+                        // @formatter:off
+                        // US92826C8394 VISA Inc. Stk .              10,00
+                        // @formatter:on
+                        .section("local", "shares") //
+                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .*(?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setShares(asShares(v.get("shares")));
+                        })
```
Nirus2000: See https://github.com/portfolio-performance/portfolio/pull/5559/changes#r2917785382

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                            t.setShares(asShares(v.get("shares")));
+                        })
+
+                        // @formatter:off
+                        // Wertpapierrechnung Wert 02.03.2026 EUR                4,11
+                        // SVK Sammelverwahrung Wert 16.06.2025 EUR              147,03
+                        // @formatter:on
+                        .section("date") //
+                        .match("^(Wertpapierrechnung|SVK Sammelverwahrung) Wert (?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}) (?<currency>[A-Z]{3})\\s+(?<amount>[\\.,\\d]+)$") //
+                        .assign((t, v) -> t.setDateTime(asDate(v.get("date"))))
```
Nirus2000, line 319: attribute "currency" and "amount" ? CopyPaste fault... :-P

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+
+                        // @formatter:off
+                        // @formatter:off
+                        // Ertrag 0,67 USD
+                        // 15 % QUSt a 1,1797 v. 26.02.2026 EUR                4,11
+                        // @formatter:on
+                        .section("termCurrency", "fxGross", "exchangeRate", "baseCurrency").optional()
+                        .match("^Ertrag (?<fxGross>[\\.,\\d]+) (?<termCurrency>[\\w]{3})$") //
+                        .match("^.* a (?<exchangeRate>[\\.,\\d]+) v. [\\d]{2}\\.[\\d]{2}\\.[\\d]{4} (?<baseCurrency>[\\w]{3}) ([\\s]+)?[\\.,\\d]+$") //
```
Nirus2000: Change everywhere if more than 1 blank...
` ([\\s]+)? ` to `[\\s]{1,}`

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                            ExtrExchangeRate rate = asExchangeRate(v);
+                            type.getCurrentContext().putType(rate);
+
+                            Money fxGross = Money.of(rate.getTermCurrency(), asAmount(v.get("fxGross")));
+                            Money gross = rate.convert(rate.getBaseCurrency(), fxGross);
+
+                            checkAndSetGrossUnit(gross, fxGross, t, type.getCurrentContext());
+                        })
+
+                        .conclude(ExtractorUtils.fixGrossValueA())
```
Nirus2000, line 367: for what to fix?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+            // second: convert to line spans
+            var spans = new ArrayList<LineSpan>();
+            for (var ii = 0; ii < blockStarts.size(); ii++)
+            {
+                int startLine = blockStarts.get(ii);
+                var endLine = ii + 1 < blockStarts.size() ? blockStarts.get(ii + 1) - 1 : lines.length - 1;
+                spans.add(new LineSpan(startLine, endLine));
+            }
+            return spans;
+        };
```
Nirus2000: Nope... look at other importers to fix this problem.
See .rangeAs or in the document class

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        // Registered Shares o.N.
+                        // IE00BYZK4552 iShsIV-Automation&Robot.U.ETF Abgang Stk.              123,00
+                        // Registered Shares o.N.
+                        // @formatter:on
+                        .section("isin", "name", "nameContinued") //
+                        .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<name>.*) (Zugang|Abgang) Stk\\s*\\.\\s+[\\.,\\d]+$") //
+                        .match("^(?<nameContinued>.*)$") //
+                        .assign((t, v) -> {
+                            t.setSecurity(getOrCreateSecurity(v));
+                        })
```
Nirus2000, line 419: Missing currency

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        // Wertpapiernummer Bezeichnung Nominale/Stück
+                        // IE00BYZK4552 iShsIV-Automation&Robot.U.ETF Zugang Stk.              123,00
+                        // Wertpapiernummer Bezeichnung Nominale/Stück
+                        // IE00BYZK4552 iShsIV-Automation&Robot.U.ETF Abgang Stk.              123,00
+                        // @formatter:on
+                        .section("local", "shares") //
+                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .* (Zugang|Abgang) (?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setShares(asShares(v.get("shares")));
+                        })
```
Nirus2000: "local" again

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                        .match("^[A-Z]{2}[A-Z0-9]{9}[0-9] .* (Zugang|Abgang) (?<local>Stk)\\s*\\.\\s+(?<shares>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setShares(asShares(v.get("shares")));
+                        })
+
+                        // @formatter:off
+                        // Schlusstag 05.02.2021
+                        // @formatter:on
+                        .section("date").match("^Schlusstag (?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4})$") //
+                        .assign((t, v) -> t.setDateTime(asDate(v.get("date"))))
```
Nirus2000, line 438: formatting

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankPDFExtractor.java
```diff
@@ -0,0 +1,470 @@
...
+                            t.setCurrencyCode(asCurrencyCode(v.get("currency")));
+                            t.setAmount(asAmount(v.get("amount")));
+                        })
+
+                        // @formatter:off
+                        // Auftrags-Nr. 999999-28.01.2021
+                        // @formatter:on
+                        .section("note").optional() //
+                        .match("^(?<note>Auftrags-Nr\\. \\d+)-[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}$") //
+                        .assign((t, v) -> t.setNote(trim(v.get("note"))))
```
Nirus2000: trim() for what?

Nirus2000 (commented):
Hello,
my comments to this importer
rename the importer to OberbankAG.....

iAppDeveloper88:
Thank you for your comments @Nirus2000. I resolved most of them already; to some of them i added a question. I will rename the importer after i fixed all issues from your review.
The TODO is because i have some more dividend documents (for "Thesaurierende und Ausschüttende Fonds/ETFs"). It wont be there in my final pull request

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
buchen: To add to the comment by @Nirus2000:

If the document has only one type of transaction (here for example "Freier Erhalt" and "Freie Lieferung"), then you do not need a block pattern at all. You could also say that the block is the whole document (blocks are helpful if you have multiple transaction in one document).

I jaustet tested it by removing the block splitting strategy completely.

```java
var firstRelevantLine = new Block();
```

Can you confirm that the Oberbank Freie Erhalt/Lieferung document only contain a Freie Erhalt oder Frei Lieferung respectively?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,657 @@
...
+                        // @formatter:on
+                        .section("currency", "amount") //
+                        .match("^(Wertpapierrechnung|SVK Sammelverwahrung) Wert [\\d]{2}\\.[\\d]{2}\\.[\\d]{4} (?<currency>[A-Z]{3})[\\s]+\\-?(?<amount>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setCurrencyCode(asCurrencyCode(v.get("currency")));
+                            t.setAmount(asAmount(v.get("amount")));
+                        })
+
+                        // @formatter:off
+                        // @formatter:off
```
buchen, line 346: duplicate

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,657 @@
...
+                        // @formatter:on
+                        .section("date") //
+                        .match("^Schlusstag (?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4})$") //
+                        .assign((t, v) -> t.setDateTime(asDate(v.get("date"))))
+
+                        // @formatter:off
+                        //   steuerlicher Anschaffungswert:                                                 1.199,76 EUR
+                        // @formatter:on
+                        .section("currency", "amount")
+                        .match("^[\\s]*Anschaffungskurswert:[\\s]+(?<amount>[\\.,\\d]+) (?<currency>[A-Z]{3})$") //
```
buchen, line 482: I think the sample (in the comment) is wrong.
It has to be "Anschaffungskurswert", right?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,657 @@
...
+                        documentContext -> documentContext //
+                                        .section("transaction") //
+                                        .match("^(?<transaction>(Split))$") //
+                                        .assign((ctx, v) -> ctx.put("transaction", v.get("transaction"))));
+        this.addDocumentTyp(type);
+
+        var pdfTransaction = new Transaction<PortfolioTransaction>();
+
+        var firstRelevantLine = new Block("^Durchf.hrungsanzeig[\\s]*e[\\s]+" //
+                        + "(Einbuchung)$");
```
buchen, line 520: This does not seem to fit together:
The document looks for "Kapitalmaßnahme" but later for "Einbuchung".

If this is wanted, add the reason as a comment.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
buchen: I agree. It only works because all examples happen to have EUR as currency and EUR is the default currency for instruments.

I propose to add a ```.match``` on the "Kurs 136,9" a couple of lines down to pick up the currency.

buchen (commented):
Hey @iAppDeveloper88,

thanks for the contribution. And sorry for only looking at it now. The bigger the change, the harder to find a time slot...

It looks good to me. I have added a couple of comments that should be easy to fix. I think the main question is if "Freier Erhalt" are documents that have only one transaction. Then we simplify the code and help our future selves...

iAppDeveloper88:
Thank you @buchen. I will look into resolving your comments either this week or next one.
Yes, the "Freier Erhalt" document is for a single transaction. The document has 3 pages and each of these pages start with "Durchfürhungsanzeige" (written with a weird blank). Information from multiple pages is needed for this single transaction.

iAppDeveloper88:
Better late than never, I pushed the fixes regarding your comments

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+                        .match("^Wertpapier\\-.*" //
+                                        + "(?<type>(Kauf" //
+                                        + "|Verkauf" //
+                                        + "|Ausgabe Fonds" //
+                                        + "|Ausgabe Fonds aus Dauerauftrag" //
+                                        + "|R.cknahme Fonds" //
+                                        + "))$") //
+                        .assign((t, v) -> {
+                            if ("Verkauf".equals(v.get("type")) || "Rücknahme Fonds".equals(v.get("type")))
+                                t.setType(PortfolioTransaction.Type.SELL);
```
kimmerin, line 84: Just my two cents: Instead of matching against everything that is possible, why not make this section `.optional` and only put the elements into the regex that actually define a `Type.SELL`. That way you only need to put new elements into the regex and don't need to copy/paste them as additional condition into the if-statement afterwards (and you can omit the whole if-statement altogether).

Personally I think that is better readable and maintainable.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+                                        // CA09228F1036 BlackBerry Ltd. Zugang Stk .              14,00
+                                        // Registered Shares o.N.
+                                        // AT0000730007 ANDRITZ AG Abgang Stk.               95,00
+                                        // AKTIEN O.N.
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("isin", "name", "nameContinued") //
+                                                        .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<name>.*) (Zugang|Abgang) Stk[\\s]*\\.[\\s]+[\\.,\\d]+$") //
+                                                        .match("^(?<nameContinued>.*)$") //
+                                                        .match("^Kurs[\\s]+[\\.,\\d]+ (?<currency>[A-Z]{3})[\\s]+Kurswert[\\s]+[A-Z]{3}[\\s]+[\\.,\\d]+")
```
kimmerin, line 105: Can you add the corrsponding example text into the sample block above?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+                                                        }),
+                                        // @formatter:off
+                                        // AT000B127337 Oberbank AG Zugang EUR            8.000,00
+                                        // Nachr. Anleihe 2023-2031
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("isin", "name", "nameContinued") //
+                                                        .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<name>.*) (Zugang|Abgang) [A-Z]{3}[\\s]+[\\.,\\d]+$") //
+                                                        .match("^(?<nameContinued>.*)$") //
+                                                        .match("^Kurs[\\s]+[\\.,\\d]+ (?<currency>[A-Z]{3})[\\s]+Kurswert[\\s]+[A-Z]{3}[\\s]+[\\.,\\d]+")
```
kimmerin, line 117: Same here concerning the sample line

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+                                                        .attributes("currency", "amount") //
+                                                        .match("^(Wertpapierrechnung|SVK Sammelverwahrung) Wert [\\d]{2}\\.[\\d]{2}\\.[\\d]{4} (?<currency>[A-Z]{3})[\\s]+(?<amount>[\\.,\\d]+)$") //
+                                                        .assign((t, v) -> {
+                                                            t.setCurrencyCode(asCurrencyCode(v.get("currency")));
+                                                            t.setAmount(asAmount(v.get("amount")));
+                                                        }),
+
+                                        // @formatter:off
+                                        // Verwahrart, Positionsdaten EUR            8.068,87
+                                        // SVK Sammelverwahrung Wert 09.08.2023
```
kimmerin, line 170: That line doesn't seem to be relevant in the match-part below.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+
+                        // @formatter:off
+                        // Kupon 4,55 % jährlich Stückzinsen f. 166 Tage EUR              165,55
+                        // @formatter:on
+                        .section("note2", "note3", "note4").optional() //
+                        .match("^Kupon [\\.,\\d]+ % .* (?<note2>St.ckzinsen .* [\\d]+ Tage).* (?<note4>[A-Z]{3})[\\s]+(?<note3>[\\.,\\d]+)$") //
+                        .assign((t, v) -> {
+                            t.setNote(concatenate(t.getNote(), v.get("note2"), " | "));
+                            t.setNote(concatenate(t.getNote(), v.get("note3"), ": "));
+                            t.setNote(concatenate(t.getNote(), v.get("note4"), " "));
```
kimmerin, line 205: The reason why there is a dot instead of `ü` in the regex is because the PDF-to-text-conversion sometimes mangels umlauts and you end up e.g. with `St}ckzinsen`. Since "Stückzinsen" is the only value that is possible, leave it out of the matcher-group and add it as a fixed text in the assign-block: `t.setNote(concatenate(t.getNote(), "Stückzinsen " + v.get("note2"), " | "))`

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/OberbankAGPDFExtractor.java
```diff
@@ -0,0 +1,619 @@
...
+
+                        // @formatter:off
+                        // Is type --> "Einbuchung" change from DELIVERY_INBOUND to DELIVERY_OUTBOUND
+                        // @formatter:on
+                        .section("type").optional() //
+                        .match("^Durchf.hrungsanzeig[\\s]*e[\\s]+" //
+                                        + "(?<type>(Freier Erhalt" //
+                                        + "|Freie Lieferung))$") //
+                        .assign((t, v) -> {
+                            if ("Freie Lieferung".equals(v.get("type")))
```
kimmerin, line 404: Same here as proposed before (here you already have made the section optional). Just put in the elements that define this transaction to be an outbound delivery and avoid the if.

kimmerin (changes requested):

stoeggich:
Hi @iAppDeveloper88,

I went through the current state of the branch and ran the regexes against the test documents in the PR. One blocker, a few review comments that look still open, and some conventions from .claude/rules/pdfextractors.md.

Blocker: testWertpapierKauf02 cannot pass

OberbankAGPDFExtractor.java L113–L117

```
section -> section //
                .attributes("isin", "name", "nameContinued") //
                .match("^(?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) (?<n>.*) (Zugang|Abgang) [A-Z]{3}[\\s]+[\\.,\\d]+$") //
                .match("^(?<nameContinued>.*)$") //
                .match("^Kurs[\\s]+[\\.,\\d]+ (?<currency>[A-Z]{3})[\\s]+Kurswert[\\s]+[A-Z]{3}[\\s]+[\\.,\\d]+")
```

This is the bond alternative, but bonds are quoted in percent — Kauf02.txt L13 reads:

Kurs 98,3 PROZ Kurswert EUR            7.864,00

PROZ does not match [A-Z]{3}[\s]+, so the third pattern never matches, the sub-section fails, and since the first alternative also fails (no Zugang Stk .), the whole oneOf throws — errors is not empty.

Suggestion: take the currency from Kurswert instead, that works for both shapes and makes the extra alternative unnecessary for this purpose:
```

.match("^Kurs[\\s]+[\\.,\\d]+ .* Kurswert[\\s]+(?<currency>[A-Z]{3})[\\s]+[\\.,\\d]+$") //
The currency is still not passed to getOrCreateSecurity()
```

OberbankAGPDFExtractor.java L102, L114

```
.attributes("isin", "name", "nameContinued") //
```

PDFParser.Section.extractAttributes() only extracts groups that are declared in attributes(...). currency is captured in the pattern but never lands in ParsedData, so getOrCreateSecurity(v) still falls back to the client's base currency. The .match currently only acts as an additional mandatory pattern. Please add "currency" to both attribute lists.

Same thing is still missing entirely in two other places:

L414–L419 — addDeliveryInOutBoundTransaction(), no currency at all. FreierErhalt01.txt has Anschaffungskurswert: 1.199,76 EUR a few lines down.
L499–L503 — addNonImportableTransaction(), no currency at all. Split01.txt has steuerlicher Anschaffungswert: 1.298,55 EUR.
Ertrag is the per-share rate, not the gross amount

OberbankAGPDFExtractor.java L345–L347

```
.match("^Ertrag (?<fxGross>[\\.,\\d]+) (?<termCurrency>[\\w]{3})$") //
```

In Dividende01.txt this matches Ertrag 0,67 USD, which is the dividend per share — the gross is Ertrag USD 6,70 one line above. In Dividende04.txt the same line is Ertrag 0 USD, so fxGross becomes zero. Both tests only pass because conclude(ExtractorUtils.fixGrossValueA()) recalculates the forex unit afterwards from the transaction's own gross value.

Please read the gross from the Ertrag <CUR> <amount> line:

```
.match("^Ertrag (?<termCurrency>[A-Z]{3})[\\s]+(?<fxGross>[\\.,\\d]+)$") //
```

Once that is fixed, hasForexGrossValue("USD", 22.83 * 0.9976) in OberbankAGPDFExtractorTest.java L531 should be re-checked against the actual document value.

documentContext section is mandatory but the document type is broader

OberbankAGPDFExtractor.java L471–L482

```
final var type = new DocumentType("(Durchf.hrungsanzeig[\\s]*e[\\s]+" //
                + "(Kapitalmaßnahme))", //
                documentContext -> documentContext //
                                .section("transaction") //
                                .match("^(?<transaction>(Split))$") //
                                .assign((ctx, v) -> ctx.put("transaction", v.get("transaction"))));
```

The document type accepts every Kapitalmaßnahme, but the context section is not optional() and only knows Split. Any other corporate action makes the whole document fail. As a consequence the else branch in L507–L508 (MsgErrorTransactionTypeNotSupportedOrRequired) is unreachable today.

Also still open from @buchen's review: the document type looks for Kapitalmaßnahme while the block at L481–L482 looks for Einbuchung — please add the reason as a comment.

Open review comments

1. L74–L85 — @kimmerin's suggestion: make the type section .optional(), put only the values that define a Type.SELL into the regex and drop the if. Same pattern at L399–L406 (there the section is already optional, but the if is still there).
2. L201 — (?<note2>St.ckzinsen .* [\d]+ Tage): the . is there because PDFBox may mangle the umlaut, so the mangled text ends up in the note. Move it out of the group and add it as fixed text: t.setNote(concatenate(t.getNote(), "Stückzinsen " + v.get("note2"), " | ")).
3. L95–L100 and L109–L112 — the @formatter:off sample blocks don't contain the Kurs … line that the section matches.
4. L168–L171 — the sample block contains SVK Sammelverwahrung Wert 09.08.2023, which is not part of the match below.
Conventions

subject() — pdfextractors.md requires the single-argument convenience constructor. Currently the verbose form is used four times: L66–L70, L235–L239, L390–L394, L488–L492.

```
.subject(() -> new BuySellEntry(PortfolioTransaction.Type.BUY))
.subject(() -> new AccountTransaction(AccountTransaction.Type.DIVIDENDS))
.subject(() -> new PortfolioTransaction(PortfolioTransaction.Type.DELIVERY_INBOUND))
```

Anchors — .match() always needs ^…$. Missing $ at L105 and L117.

Special characters must be matched by . — PDFBox output varies:

L472 Kapitalmaßnahme → Kapitalma.nahme
L612 Kupon jährlich KESt → Kupon j.hrlich KESt

Currency class — the table in pdfextractors.md says [A-Z]{3}, [\w]{3} is used at L269, L282, L346, L347, L592, L602, L612.

Wrong comment — L397:

```
// Is type --> "Einbuchung" change from DELIVERY_INBOUND to DELIVERY_OUTBOUND
```

The section below checks for Freie Lieferung. Also, this is explanatory prose, so the surrounding @formatter:off / @formatter:on at L396/L398 are noise and can go.

Comments shredded by the formatter — L139–L142 and L304–L307:

```
// Percentage
// quotation,
// workaround for
// bonds
```

Should be one line: // Percentage quotation, workaround for bonds.

Redundant capture groups — L472 (Kapitalmaßnahme), L475 (Split), L482 (Einbuchung).

Blank line right after the class opening brace — OberbankAGPDFExtractor.java L21–L22 and OberbankAGPDFExtractorTest.java L52–L53.

Test

Share counts must be double literals, never integers — OberbankAGPDFExtractorTest.java L81, L115, L219, L253, L321, L355, L388, L422, L456, L526, L560, L594, L631:

```
hasShares(14),      // → hasShares(14.00)
```

Stray blank lines before the closing brace of the test method at L461, L496, L599.

Questions

1. L369 / Dividende03.txt — the document ends with Summe EUR -0,00, so the extractor creates a TAX_REFUND of 0,00 EUR. Wouldn't the canonical SkippedItem guard in wrap() with MsgErrorTransactionTypeNotSupportedOrRequired be the better fit here?
5. L443–L448 / FreierErhalt01.txt — the document lists Nebenkosten: 7,25 EUR in addition to Anschaffungskurswert: 1.199,76 EUR. Should those be booked as Unit.Type.FEE (amount 1.207,01 / gross 1.199,76 / fee 7,25), or is the acquisition value meant to be net of them
6. L537–L553 / L571–L616 — the fee sections accept an optional sign (\\-?), the tax sections require a leading \\-. In Dividende04.txt the KESt lines carry no sign at all. Would it make sense to use (\\-)? consistently and evaluate the sign in assign()?
7. Test document names: Dividende03.txt and Dividende04.txt are Ausschüttungsgleicher Ertrag (taxes / tax refund), Dividende05.txt is a Zinszahlung. Should those be renamed to something like AusschuettungsgleicherErtrag01/02.txt and Zinszahlung01.txt?

PR #5560, opened 2026-03-10
Add claude-rules around pdf-extractors

gamue:
Thought it makes sense to create a dedicated Claude rules-file for pdf-extractors which is used when the files get adjust there. It can also be used as reference like the `Contributing.md` and is a little bit more detailed with the learnings I had in during the changes I did.

Used https://github.com/portfolio-performance/portfolio/pull/5561 for testing this rule

### DIFF
--- a/.claude/rules/pdfextractors.md
+++ b/.claude/rules/pdfextractors.md
@@ -1,0 +1,67 @@
+---
+paths:
+ - "name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/*"
+ - "name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/**"
+---
+
+PDF importers extract transactions from bank and broker PDF statements. 
+
+- Each bank/broker has its own extractor class inside `name.abuchen.portfolio/src/name/abuchen/datatransfer/pdf/`
+- Tests are located at `name.abuchen.portfolio.tests/src/name/abuchen/datatransfer/pdf/` within a folder for each bank, containing all connected test-exports.
+
+# Implementation
+
+- `AbstractPDFExtractor` is the base class for every pdf-extractor, containing utility methods
+- `ExtractorUtils` offers additional utility methods, if the ones from the base-class aren't enough.
+- Use `TextUtil` when it's required to manipulate text
+- When implementing a new extractor or adding new parsing logic method, check `BaaderBankPDFExtractor` for reference and use the same pattern
+  - Split logic into methods like `addBuySellTransaction()`, `addDividendTransaction()` or `addInterestTransaction()`
+  - When extracting information from pdf-documents, create multiple `section`-Blocks. 
+  - Consistency in coding style between all extractors is more important than nice code.
+  - Add comment-blocks with `@formatter:off` and `@formatter:on` before each section-block showing the format that is handled there.
+
+
+# Testing
+Test methods should be in following structure
+
+```
+@Test
+public void testWertpapierKauf01()
+{
+    var extractor = new ScalableCapitalPDFExtractor(new Client());
+
+    List<Exception> errors = new ArrayList<>();
+
+    var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Kauf01.txt"), errors);
+
+    assertThat(errors, empty());
+    assertThat(countSecurities(results), is(1L));
+    assertThat(countBuySell(results), is(1L));
+    assertThat(countAccountTransactions(results), is(0L));
+    assertThat(countAccountTransfers(results), is(0L));
+    assertThat(countItemsWithFailureMessage(results), is(0L));
+    assertThat(countSkippedItems(results), is(0L));
+    assertThat(results.size(), is(2));
+    new AssertImportActions().check(results, "EUR");
+
+    // check security
+    assertThat(results, hasItem(security( //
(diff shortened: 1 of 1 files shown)


gamue:
@buchen maybe you have an opinion on this? The linked PR shows that it worked well and in case older tests should be updated (maybe step by step) this could help. When going through the exDate tests and needing to adjust multiple different formats I was wondering if that would be something useful. However main reason is that it should help when creating new ones

.claude/rules/pdfextractors.md
```diff
@@ -0,0 +1,67 @@
+---
+paths:
+ - "name.abuchen.portfolio/src/name/abuchen/datatransfer/pdf/*"
+ - "name.abuchen.portfolio.tests/src/name/abuchen/datatransfer/pdf/*"
```
buchen: I wonder how this works currently - it is missing the "portfolio" segment in the path.
And it probably needs a ** because the tests are distributed into sub-packages.


```
paths:
 - "name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/*"
 - "name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/**"
```

.claude/rules/pdfextractors.md
```diff
@@ -0,0 +1,67 @@
...
+    new AssertImportActions().check(results, "EUR");
+
+    // check security
+    assertThat(results, hasItem(security( //
+                    hasIsin("IE0008T6IUX0"), hasWkn(null), hasTicker(null), //
+                    hasName("Vngrd Fds-ESG Dv.As-Pc Al ETF"), //
+                    hasCurrencyCode("EUR"))));
+
+    // check dividends transaction
+    assertThat(results, hasItem(purchase( //
```
buchen: comment and test to do not fit together - dividend vs. purchase

Could it confuse Claude?

.claude/rules/pdfextractors.md
```diff
@@ -0,0 +1,67 @@
...
+                    hasNote("Ord.-Nr.: SCALsin78vS5CYz"), //
+                    hasAmount("EUR", 19.49), hasGrossValue("EUR", 18.50), //
+                    hasTaxes("EUR", 0.00), hasFees("EUR", 0.99))));
+
+}
+```
+
+- Each method has a starting `assertThat`-Block checking the counts. All 8 assertions need to be present.
+- Use `//` to enforce line-breaks when checking securities and transactions
+- All dates should be checked with hours, minutes and seconds.
```
buchen: Maybe better:

> Include time (hours, minutes, seconds) in `hasDate` when the source document provides it.

Not all sources provide the time. Then we do not want it to add "T00:00" all the time. Not need to check for that.

buchen (changes requested):
Hi @gamue,

it sure makes sense to instruct Claude code better. So far I am very hesitant to add a lot to these files - the tools seem to be able to discover more and more each month. I am even thinking about further reducing the CLAUDE.md to make sure not to pollute the context with things it can discover anyway. But this one is very short. It is worth a try.

About the change, I think we can merge it and see what benefits it brings. I noticed a couple things that we should change before merging.

.claude/rules/pdfextractors.md
gamue: good one, this is actually a copy & paste from the `CONTRIBUTING.md`. Will fix it there too

.claude/rules/pdfextractors.md
gamue: I actually got the feedback to add `T00:00` into all tests when creating tests, that's why I added it there. Also when checking some example tests (e.g. `testPeriodenauszug `in `BaaderBankPDFExtractorTest`; `testKontoauszug `in `TradeRepublicPDFExtractorTest`) this is consistent with the current way, even no time is given and would generally not be necessary.

.claude/rules/pdfextractors.md
gamue: I think it doesn't matter if it's * or ** in this pattern, at least when asking to adjust the `WirBankPDFExtractorTest` it said that it's loading this file. Will do some more experiments around it

Reference: https://code.claude.com/docs/en/memory#path-specific-rules

.claude/rules/pdfextractors.md
buchen: The "old" style

```java
assertThat(entry.getPortfolioTransaction().getDateTime(),
           is(LocalDateTime.parse("2020-05-05T00:00")));
```

definitely needs it.

But the ```hasDate``` actually has the logic to "upgrade" a plain date to the DateTime as needed. And I think it reduces the noise on the tests. 

```java
    public static Matcher<Transaction> hasDate(String dateString)
    {
        LocalDateTime expectecd = dateString.contains("T") //$NON-NLS-1$
                        ? LocalDateTime.parse(dateString)
                        : LocalDate.parse(dateString).atStartOfDay();

        return new PropertyMatcher<>("date", //$NON-NLS-1$
                        expectecd, //
                        Transaction::getDateTime);
    }
```

.claude/rules/pdfextractors.md
buchen: not sure about the * vs ** but without the "portfolio" in the path, I do not see the rule activated in Claude Code

buchen:
I let it randomly check the BISON importer only to tell me there is a duplicate (and therefore irrelevant) assert. It's crazy what they can do. I'll merge. Let's learn.

PR #5563, opened 2026-03-12
ScorePriority: fixed error message about invalid CUSIP length.

hporten:
1.) Corrected placeholder value for error message
2.) Do what the comment says and reject all lengths other than 9
3.) Fixed typo in comment.

It is questionable to match CUSIPS with "(?<wkn>.*)" btw. Particularly as the dot also matches whitespace that later has to be dropped with trim() again. \w would be better but without more test cases we have to guess whether it needs to be surrounded with \s.

PS: 1-2 more pull requests for similar checks will follow.


hporten:
Fwiw: I do not have an account with ScorePriority. I merely ran across this check as part of the review for possible uses of SkippedItem (https://github.com/portfolio-performance/portfolio/issues/5517).

PR #5564, opened 2026-03-12
Dropped unreachable failure code.

hporten:
The wkn attribute is matched with \w{9} or .{9} and therefore cannot have a length different from 9.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/LimeTradingCorpPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/LimeTradingCorpPDFExtractor.java
@@ -101,12 +101,7 @@ public class LimeTradingCorpPDFExtractor extends AbstractPDFExtractor
                             t.setShares(asShares(v.get("shares")));
                             t.setAmount(asAmount(v.get("amount")));
                             t.setCurrencyCode(asCurrencyCode(v.get("currency")));
-
-                            // if CUSIP lenght != 9
-                            if (v.get("wkn").length() != 9)
-                                v.markAsFailure("CUSIP is maybe incorrect. " + t.getPortfolioTransaction().getDateTime() + " " + t.getPortfolioTransaction().getSecurity());
-                            else
-                                t.setSecurity(getOrCreateSecurity(v));
+                            t.setSecurity(getOrCreateSecurity(v));
                         })
 
                         .wrap(BuySellEntryItem::new));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/LimeTradingCorpPDFExtractor.java
```diff
@@ -155,12 +150,7 @@ private void addAccountStatementTransaction()
                                                             t.setShares(asShares(v.get("shares")));
                                                             t.setAmount(asAmount(v.get("amount")) - asAmount(v.get("tax")));
                                                             t.setCurrencyCode(asCurrencyCode(v.get("currency")));
-
-                                                            // if CUSIP lenght != 9
-                                                            if (v.get("wkn").length() != 9)
-                                                                v.markAsFailure("CUSIP is maybe incorrect. " + t.getDateTime() + " " + t.getSecurity());
-                                                            else
-                                                                t.setSecurity(getOrCreateSecurity(v));
+                                                            t.setSecurity(getOrCreateSecurity(v));
```
buchen, line 153: I agree that the length check is not needed here. The string must be exactly 9 characters to match in the first place.

Here is it matching with a dot (.) that matches any character and [later](https://github.com/portfolio-performance/portfolio/blob/67a02c57ca13b9b6d84dcc4b06c5a7304f6116e8/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/AbstractPDFExtractor.java#L287) the string is trimmed to lookup up an instrument with this CUSIP. So it could be an invalid CUSIP. However, your commit does not change this behavior.

hporten: Yeah, the whitespace inclusion is not pretty. As mentioned in https://github.com/portfolio-performance/portfolio/pull/5563, the usage of \w would be a more precise. But in light of unknown surroundings of whitespace, the patterns would need to be embedded in \s* on both sides.

Other [PP code](https://github.com/portfolio-performance/portfolio/blob/67a02c57ca13b9b6d84dcc4b06c5a7304f6116e8/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComputersharePDFExtractor.java#L73) uses (?<wkn>[A-Z0-9]{9})$ for matching CUSIPs. But according to the [Wikipedia article](https://en.wikipedia.org/wiki/CUSIP) '*', '@' and '#' are allowed as well. A [Stackoverflow answer](https://stackoverflow.com/questions/5901805/need-help-with-regular-expression-for-nine-digit-alphanumeric-with-minimum-one-s) therefore recommends this pattern:

    ^[0-9]{3}[a-zA-Z0-9]{2}[a-zA-Z0-9*@#]{3}[0-9]$

But given that the current change is only about failures vs. skips only, I did not dare to dive deeper. A more sophisticated matching would be up for someone with more example documents in hand. And due to the thumb-down that the pull request got already, I do not mind postponing the the whole change entirely.

buchen:
merged

The WKN can only be 9 characters long. The additional if statement is not needed. The change to the regular expressions could be a separate pull request (or whenever the importer needs to be touched)

PR #5567, opened 2026-03-13
Add option to show last transaction date in accounts

### LINKED ISSUE #5556: Add "last transaction date" to account overview
I guess like most others, I'm doing regular (usually monthly) updates of the values and adding transactions to the different accounts. It'll be handy to see which ones of the accounts already been updated by seeing the "last transaction date" inside the overview

Inside the `Stammdaten -> Konten` overview, it's possible to show a `last transaction` column, similar to last date in stocks.

- The column can be activated/deactivated via the settings
- It's possible to sort the column by clicking on it

gamue:
Closes #5556

Added functionality to the account-screen to show the last transaction date. This helps with keeping your balances up-to-date and potentially identify for which account you might need to recheck or add the transactions e.g. when having mutliple accounts that pay dividends or interests on a monthly basis, you can now see where you haven't added them yet.

image

Notes: 
- I haven't adjusted the translations files yet as I first wanted to get the code and feature right. Also wondering about naming
- I thought the UI code was little bit hard to read, so I extracted it into smaller methods. Hope that's fine
- Unsure how testing works, I manually checked it (sorting; having accounts without transactions). and added unit tests for the main functionality


buchen:
Thanks for the contribution.

I have updated the commit:
* Extracted the column into a separate class -> that allows for caching the date -> particularly if the column is sorted, the value will be read again and again
* Make it generic to the TransactionOwner interface -> now also used in the PortfolioListView
* Added the translations (it is good to not have them in the initial pull request - thanks for that). It is fairly easy to add them later
* I changed the refactoring (extracting of the creation of columns) in such a way that I ordered the methods in the same way the columns used to be created. That makes the diff a little bit more readable.

gamue:
Thanks a lot for the changes, anything missing which I should take care of?

I implemented the changes similar to the `defineAccountBalanceColumn`-method, when talking about caching maybe you want to check this too as `.getCurrentAmount(LocalDateTime.now().with(LocalTime.MAX)))` looks like it's also iterating through all transactions

buchen:
> defineAccountBalanceColumn

I guess so. I added this option only recently. The idea was to have the caching locally within the column implementation which is much easier to oversee the trying to do machine on the AbstractFinanceView level. I also think there are many more columns that could profit. But Rome wasn't built in one day...

The change is good. It just took to long for the build to complete and we were then out of the house. Will merge now.

gamue:
great, thanks a lot!

PR #5568, opened 2026-03-14
add bold to treemap label

mierin12:
Hello,

This is a proposition to add bold to the security name in TreeMap.
In the tooltips of Donut and Pie Chart they are already in bold, so I thought it looks more consistent.

**Before :**
Capture d’écran 2026-03-14 112758

**After :**
Capture d’écran 2026-03-14 112930

I assumed there would always be two lines of label at least, so `setFont(defaultFont)` is always called in the different loops, so no need to set it again after the loops. but maybe it would have been safer ?

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/TaxonomyNodeRenderer.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/taxonomy/TaxonomyNodeRenderer.java
@@ -7,8 +7,11 @@ import java.util.Map;
 import java.util.function.DoubleFunction;
 import java.util.function.Function;
 
+import org.eclipse.jface.resource.FontDescriptor;
 import org.eclipse.jface.resource.LocalResourceManager;
+import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.GC;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
@@ -105,6 +108,8 @@ import name.abuchen.portfolio.util.ColorConversion;
     protected Map<String, Color> hex2color = new HashMap<>();
     protected Function<String, Color> colorFactory = color -> new Color(ColorConversion.hex2RGB(color));
 
+    private Font boldFont;
+
     public TaxonomyNodeRenderer(TaxonomyModel model, LocalResourceManager resources)
     {
         this.model = model;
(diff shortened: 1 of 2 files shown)


buchen:
Nice addition. But I think we should not create the bold font in every drawRectangle function. I think it is a fairly expensive operation.

> I assumed there would always be two lines of label at least, so setFont(defaultFont) is always called in the different loops, so no need to set it again after the loops. but maybe it would have been safer ?

I think it is good practice to leave the graphics context as it was received, i.e. set back the default font.

mierin12:
I have updated to try to take into the comment, and have boldFont created outside of drawRectangle.
However 
- I am now unsure where/when boldFont should be disposed,
- In the end setBoldFont is called as often as drawRectangle in TreeMapViewer, so still expensive
- If drawrectangle can now fail if setBoldFont is not called before (boldFont can be null)

buchen:
> I am now unsure where/when boldFont should be disposed,

I See that that the renderer already has a LocalResourceManager. I think you can just use that. It will take care of cleaning this up.

```
FontDescriptor defaultFont = JFaceResources.getDefaultFontDescriptor();
FontDescriptor boldDescriptor = defaultFont.setStyle(SWT.BOLD);
Font boldFont = resourceManager.createFont(boldDescriptor);
label.setFont(boldFont);
```

> In the end setBoldFont is called as often as drawRectangle in TreeMapViewer, so still expensive

I think creating the bold font is expensive. Setting it to the graphics context I do not think so much.


> If drawrectangle can now fail if setBoldFont is not called before (boldFont can be null)

I think we need to create it in the constructor before the first call to the drawRectangle method.

Or maybe lazily in the first drawRectangle.
Maybe it actually has to be done here because the font might first need to be styled, e.g., the size from the CSS applied.

mierin12:
OK I think I understand now :)  I have added one new commit.
I think in TreeMapLegend there is no other way to create it several times, as it is used to size the rectangles themselves (while for the real treeMap the rectangle size is not dependent on the label size. The bold font can impact if the label will be drawn horizontaly or vertically, and is authorized anyway to "disappear" outside the rectangle).

>  Maybe it actually has to be done here because the font might first need to be styled, e.g., the size from the CSS applied.

yes i think you are right :
`FontDescriptor defaultFontDescriptor = JFaceResources.getDefaultFontDescriptor();`
is not taking into account a different size than the default size. So if I set in the setting a smaller or larger font size, the bold name will not be impacted : 
Capture d’écran 2026-03-30 à 21 50 44

buchen:
Maybe we have to take the current font (from the graphics context), get the descriptor, then ask the LocalResourceManager to create one based on this.

mierin12:
So i think I am getting closer : it is now back in the drawRectangle method, but only created if non already existing. And created through LocalResourceManage which should manage the disposal.

buchen:
> is not taking into account a different size than the default size

I changed that now to use the default font from the graphics context. That font seems to be styled according to the CSS style sheets. At least I was able to change the font size and the bold font was changing accordingly.

I will merge this pull request. If it does not work for me, please create a fix.

mierin12:
Thanks, it works for me.

PR #5577, opened 2026-03-18
Add missing NON-NLS

KarlBelow:
Just some minor clean-up of ugly warnings.


buchen:
To be honest, I wasn't sure if I want to translate these labels. The text is visible in the UI, but only in the log window when importing classifications. This is not regular usage. The warnings were always the reminder to do so. But, I haven't done it for the last 9 months. Therefore I think it is okay to exclude the warnings. They create noise.

PR #5580, opened 2026-03-19
Fix: Prevent unwanted split of the app into two windows

### LINKED ISSUE #5514: Second window in the background (MacOS)
**Describe the bug**
PP launches an additional window at startup in the background on macOS. The window stays in the background, has no content and when I close it, then the entire app exits.

**To Reproduce**
1. Start PP on macOS.
2. The window will appear in the background

**Expected behavior**
There should be no additional window.

**Screenshots**
Image

 - Version: 0.81.5 (February 2026)
 - Platform: macosx, aarch64
 - Java: 21.0.5+11-LTS, Azul Systems, Inc.

georgemac-labs:
Attempt to fix #5514.

As far as I can tell, the ability to detach is just a nuisance here, it serves no purpose?

Sorry if I overlooked something.

### DIFF
--- a/name.abuchen.portfolio.bootstrap/Application.e4xmi
+++ b/name.abuchen.portfolio.bootstrap/Application.e4xmi
@@ -260,6 +260,7 @@
     <tags>removeOnHide</tags>
     <tags>Editor</tags>
     <tags>Cloneable</tags>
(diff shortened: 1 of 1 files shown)


buchen:
I like it. I never had the time to check

Issue #5513, opened 2026-02-23
Ex-Date Support: Tracking Open Work


buchen:
Portfolio Performance is adding support for the **ex-date** (record date) on dividend transactions. The ex-date is the cutoff date on which a shareholder must hold a security to be entitled to receive the dividend. Making this date explicit in PP enables more accurate performance calculations (in , better data attribution, and richer import/export fidelity — particularly in combination with external data sources like DivvyDiary.

---

### Related Work

**Completed / In Progress:**
- #5439 — Introduce ex-date field to the PP data model (ZfT2, draft)

---

### Open Features

- [ ] **PDF Importer: enhancing PDF importer to fill-in ex-date**
- #5502 — FinTechGroupBank PDF extractor: parse ex-date
- #5503 — Weberbank PDF extractor: parse ex-date
- #5504 — SantanderConsumerBank PDF extractor: parse ex-date
- #5505 — SBroker PDF extractor: parse ex-date
- #5507 — NIBCBank PDF extractor: parse ex-date
- #5508 — GladbacherBankAG PDF extractor: parse ex-date
- #5509 — DKB PDF extractor: parse ex-date
- #5510 — DZBankGruppe PDF extractor: parse ex-date
- [ ] **UI: Adapt transaction dialogs** to expose and allow editing of the ex-date field. The dialog should make the date optional (the current date widget requires that user inputs a date), should prefill the date with the ex-date from the SecurityEvent data (if available).
- [ ] **Retrofit:** During PDF import, if the ex-date is absent from the document but available in security events for the security and period, automatically populate the ex-date
- [ ] **Performance calculation:** Accrue dividends at the time of the ex-date rather than the payment date, to correctly attribute returns to the holding period
- [ ] **Backfill tool:** Provide a function that scans existing transactions and attempts to determine and set the ex-date for historical dividend entries. Add to the File -> Tools menu.
- [ ] **In-place editing:** in-place editing of the ex-date should be available for all account transactions that are linked to an instrument (not only dividends)

gamue:
[...]

PR #5590, opened 2026-03-24
Ex-Date Support: Backfill tool

ZfT2:
From: https://github.com/portfolio-performance/portfolio/issues/5513

> Backfill tool: Provide a function that scans existing transactions and attempts to determine and set the ex-date for historical dividend entries. Add to the File -> Tools menu.


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/tools/BackfillExDatesHandler.java
```diff
@@ -0,0 +1,352 @@
...
+    @CanExecute
+    boolean isVisible(@Named(IServiceConstants.ACTIVE_PART) MPart part)
+    {
+        return MenuHelper.isClientPartActive(part);
+    }
+
+    @Execute
+    public void execute(@Named(IServiceConstants.ACTIVE_PART) MPart part,
+                    @Named(IServiceConstants.ACTIVE_SHELL) Shell shell,
+                    @Optional @Preference(UIConstants.Preferences.DIVVYDIARY_API_KEY) String divvyDiaryApiKey)
```
buchen: The API Key should be injected into the DivvyDiaryDividendFeed automatically. The Pref2Env (preference to environment) does that.

Did it not work? Or was ist just not obvious code to manually inject that here?

name.abuchen.portfolio.bootstrap/Application.e4xmi
```diff
@@ -52,6 +52,7 @@
         </children>
         <children xsi:type="menu:Menu" xmi:id="_zEQasOIVEe6pEIPYaDGzWA" elementId="name.abuchen.portfolio.bootstrap.menu.tools" label="%command.tools.name">
           <children xsi:type="menu:HandledMenuItem" xmi:id="_B_nS4BfWEeOders_POVQHQ" elementId="name.abuchen.portfolio.ui.menu.file.consistencychecks" label="%command.consistencyChecks.name" tooltip="%command.consistencyChecks.tooltip" command="_D9tYMBiPEeOR1rNFCzC82A"/>
+          <children xsi:type="menu:HandledMenuItem" xmi:id="_BackfillExDatesMenuItem" elementId="name.abuchen.portfolio.ui.menu.tools.backfillExDates" label="%command.backfillExDates.name" command="_BackfillExDatesCommand"/>
```
buchen, line 55: Interesting ```xmi:id``` values.
For me, the editor is always generating EMF-generated UUIDs. (Open the file with "Eclipse Model Editor").
I guess this is not a problem just unusual

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/tools/BackfillExDatesHandler.java
```diff
@@ -0,0 +1,352 @@
...
+            for (AccountTransaction transaction : account.getTransactions())
+            {
+                if (transaction.getType() != AccountTransaction.Type.DIVIDENDS)
+                    continue;
+
+                if (transaction.getExDate() != null)
+                    continue;
+
+                var security = transaction.getSecurity();
+                if (security == null || Strings.isNullOrEmpty(security.getIsin()))
```
buchen: Why exclude securities that have no ISIN?

Yes, most likely we would not match anything because we have no security events and because DivvyDiary uses the ISIN as identifier. But maybe the user has removed the ISIN. And maybe we add other sources of dividend events in the future. I propose to just skip this limitation.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/tools/BackfillExDatesHandler.java
```diff
@@ -0,0 +1,352 @@
...
+
+            var security = pending.transaction().getSecurity();
+            monitor.subTask(security.getName());
+
+            var events = security2events.get(security);
+            if (events == null)
+            {
+                try
+                {
+                    events = feed.getDividendPayments(security);
```
buchen: Do we really have to query the dividend events online?

If the user has configured an DivvyDiary API key, then the events should also be part of the "events" that are stored with a Security. We could run the update job.

buchen (dismissed):
Thanks for working to complete the ex-date support. :-)

Let me understand why you chose to have an online request for the dividend events. Shouldn't they be available locally already?

And if so, should we then also change the AccountTransactionDialog?

At the moment, I tend towards the already downloaded events, but argue for your solution.

PR #5592, opened 2026-03-24
Ex-Date Support: Lookup during PDF import

ZfT2:
From https://github.com/portfolio-performance/portfolio/issues/5513

> Retrofit: During PDF import, if the ex-date is absent from the document but available in security events for the security and period, automatically populate the ex-date

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/dialogs/transactions/AccountTransactionDialog.java
@@ -9,7 +9,6 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.MessageFormat;
 import java.time.LocalDate;
-import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -43,6 +42,7 @@ import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.model.SecurityEvent;
+import name.abuchen.portfolio.model.SecurityEvent.DividendEvent;
 import name.abuchen.portfolio.money.CurrencyConverter;
 import name.abuchen.portfolio.money.CurrencyConverterImpl;
 import name.abuchen.portfolio.money.ExchangeRateProviderFactory;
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/wizards/datatransfer/ReviewExtractedItemsPage.java
```diff
@@ -488,6 +489,20 @@ public String getText(ExtractedEntry entry)
...
+            @Override
+            public String getText(ExtractedEntry entry)
+            {
+                LocalDate exDate = getExDate(entry.getItem());
+                return exDate != null ? Values.Date.format(exDate) : null;
+            }
+        });
+        ColumnViewerSorter.create(entry -> getExDate(((ExtractedEntry) entry).getItem())).attachTo(viewer, column);
+        layout.setColumnData(column.getColumn(), new ColumnPixelData(80, true));
+
```
buchen: I think the ex date is not so important to actually show it here in the review table.
The table is not showing all information anyway - think of fees and taxes.
The idea is to quickly understand what is being important and not completeness.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/ImportPDFHandler.java
```diff
@@ -47,6 +61,10 @@
 
 public class ImportPDFHandler
 {
+    private static final long EX_DATE_PAYMENT_DATE_TOLERANCE_DAYS = 5;
+
+    private static final Map<Security, List<DividendEvent>> securityEventsMap = new HashMap<>();
```
buchen: I do not think that the performance gain is enough to add additional state to the handler. The handler is managed by Eclipse and we have no control whatsoever when this map is cleared, if it is used across multiple clients, etc.

ZfT2:
Thanks, I will check the issues (and also the other PR) hopefully next week.

ZfT2:
OK, I moved the exDate logic to the ImportController and adjusted the JUnit Tests.
Also, I removed the ex-date from the preview page again and deleted the tests.


Regarding:
> As commented in the other pull request, I think it should not do a online check but use the data already in the model.

Maybe I'm doing something wrong, but when I tested it, the events list is empty. 
To do so, I used two different dividend PDF files for diverend securities from Flatex, commented out the exDate extration in the Fintechgroup PDF importer (to avoid that the ex date from the PDF is already used) and then debugged it in eclipse. I only get events with the online call.
Do yo have an idea, what might be the issue here?

PR #5595, opened 2026-03-25
Allow different tickers for quote providers

### LINKED ISSUE #5594: [improvement] allow to use different ticker symbols for historical and current quote providers
There are cases where historical prices are available from one exchange, and the current price from another exchange. But currently both quote providers for historical and current prices are using the same ticker symbol. Hence, it is not possible to configure different ticker symbols for historical and current prices.  
And also: changing the ticker symbol for one quote provider, changes also the ticker symbol for the other quote provider, what is not obvious for the user.

In class Security there is currently just one field tickerSymbol, which is used by both quote providers.  
We could add another field latestTickerSymbol, which is used by the quote provider for current price, when set. When not set, we default to tickerSymbol.

**Additional context**  
See here for an example: [PP forum](https://forum.portfolio-performance.info/t/neuer-kurslieferant-direkt-von-portfolio-performance/32800/18)

KarlBelow:
Closes #5594

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/ProtobufWriterTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/model/ProtobufWriterTest.java
@@ -34,6 +34,7 @@ public class ProtobufWriterTest
                         { "/issues/Issue672CapitalGainsIfSecurityIsTransferred.xml" },
                         { "/issues/IssueCurrencyGainsRoundingError.xml" }, //
                         { "/scenarios/protobuf.xml" }, //
+                        { "/scenarios/protobuf70.xml" }, //
                         { "/scenarios/account_performance_tax_refund.xml" }, //
                         { "/scenarios/classification_test_case.xml" }, //
                         { "/scenarios/currency_sample.xml" }, //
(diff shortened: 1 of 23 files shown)


buchen:
Thanks for the pull request. I agree we should allow to use alternative ticker symbols for the latest price feed.

Here a couple initial remarks on the change:
* the pull request does not change which ticker symbol is used to retrieve the prices. The QuoteFeed implementation is never using the latest ticker symbol to retrieve prices. I haven't checked in detail, but I am not even sure that the preview in the dialog is using it (it depends on how it constructs the "template" Security)
* I think there is an argument to be made to add this as a "top level attribute" to the "Security". I have stopped adding top level properties that do not apply to all instruments to avoid polluting the class with attributes. Therefore other QuoteFeeds actually use the SecurityProperty to store additional information with the instrument. Did you consider this approach? Why did you choose the class-level attribute?
* The getExchanges signature now contains the ticker. And then has calls where it is explicitly null. I think we need a different signature. However, I am not sure which one. getExchangesForLatestPrice seems to duplicate unnecessarily. Do you have ideas? What alternatives did you consider?

buchen:
see #5601

buchen:
Replaced by #5601

PR #5597, opened 2026-03-25
TradeRepublic: Support parsing Ex-Date

### LINKED ISSUE #5596: [PDF] Add ExDate support for TradeRepublic
### Affected Bank

TradeRepublic

### Extracted Text from PDF

```shell
xx
```

### Issue Description

connected to #5513

gamue:
Closes #5596

- unfortunately the format of some other rows got formatted as well, however as I'm using the included code-format I think this is fine. Think when reviewing activate "ignore whitespaces"-option
- The Dividende17 test-file has a messed up date, so not adjusted this
- The test structure looks a little bit odd to me, some languages have their own files and it's also mixed into the "DividendeXX"-files, maybe worth to split them in a follow-up


buchen:
I removed the additional formatting changes. I know the formatting is not consistent. At the moment I tend to prefer that the changes only include things that actually changed. (Sometime in the golden future, I reformat all code and then enforce strict formatting. But at the moment it is not worth it to invalidate all pull requests, pollute the change history, etc.)

Issue #3289, opened 2023-04-16
Allow to override ticker symbol used for accessing quote feed


pfalcon:
**Is your feature request related to a problem? Please describe.**

There's no consistency in naming securities. Beyond a few well-established exchanges, whose tickers could be considered "standard", there's no standard and consistent naming across other exchanges and data providers. Even for "well known" tickers, the problem surfaces as soon as you go to preferred shares or similar. Of course, PortfolioPerf allows to change a ticker symbol of security. But that doesn't help if your data "belongs" to (different) 3rd parties. For example, if you import securities from a broker, and then use some service's quote feed. If broker's vs feed's symbols differ, it's a problem, because if broker's symbol changed to make the feed work, one may expect problems later, when broker's updates are imported.

**Describe the solution you'd like**

I'd like to propose to add a new field to Security model, feedTickerSymbol. If set, quote feed implementations should use that field when accessing data. If not set, standard tickerSymbol should be used as before. Then, there should be a UI to allow to set feedTickerSymbol. Note that there already seems to be adhoc UI to "override" ticker for some feeds (e.g. AlphavantageQuoteFeed has a text field for free-form override, and YahooFinanceQuoteFeed - dropdown for "per-exchange" overrides), except that it actually updates the main security symbol.

**Describe alternatives you've considered**

An alternative would be ignore problem or rely on behavior of particular importers which works around this problem. For example, IBFlex importer currently appear to use ticker matching only as fallback, and otherwise tries to do matching on some internal broker id stored in the "wkn" field. Of course, this fact wouldn't be obvious to end users (I can read the source code, so can learn it, but what about other people), relies on particular internal trait which can change any time, and still confusing to the end users regarding why they see in PortfolioPerf a different symbol than in their brokerage.

KarlBelow:
We have `feed `for historical prices and `latestFeed `for latest quote. Both feeds could require different ticker, hence we need also something like `latestFeedTickerSymbol`.

PR #5601, opened 2026-03-28
Allow to use different tickers for historical and latest quote feeds

### LINKED ISSUE #5594: [improvement] allow to use different ticker symbols for historical and current quote providers

buchen:
Closes #5594
Issue: #5595


buchen:
@KarlBelow  I experimented around and created #5601.

It addresses my comments from original pull request:
* it is actually using the latest ticker to retrieve the data
* it is uses a property instead of a new attribute (to avoid polluting the class)
* it keeps the signatures stable - the point is to pass in a the Security with the updated ticker - that removes the code checks from the feed implementations (maybe it makes sense to change the QuoteFeed signature to pass in a "SecuritySpec" which implies it is just read only data)

However, I release it is missing more parts:
* There are a number of feeds that use a freeform text field. Those do not allow to override the latest yet.
  - AlphavantageQuoteFeed
  - FinnhubQuoteFeed
  - BinanceQuoteFeed
  - BinanceFuturesUsdsMarginedQuoteFeed
  - MEXCQuoteFeed
  - BitfinexQuoteFeed
  - CoinGeckoQuoteFeed
  - KrakenQuoteFeed
  - EODHistoricalDataQuoteFeed
 I believe we should also allow to override them. Everything else will then confuse the use.

buchen:
I have refactored the change.

It now uses a checkbox to enable the "ticker override".

That works for both modes - the dropdown of exchanges and the free-style text box.

Bildschirmfoto 2026-03-29 um 09 32 14
Bildschirmfoto 2026-03-29 um 09 32 23
Bildschirmfoto 2026-03-29 um 09 32 33
Bildschirmfoto 2026-03-29 um 09 32 59

KarlBelow:
My initial thinking was like this:
 - We have unique identifiers of a security, i.e., ISIN and WKN. But it isn't guaranteed that they are always available.
 - We have a non-unique identifier, i.e., ticker symbol.
 - The ticker symbol is used to refer to the position(s) hold by my broker(s), and for retrieving historical and latest quotes.
 - In a perfect world, all brokers would use the same ticker symbols, and all quote providers would be able to provide historical and latest quotes, using the same unique ticker symbol.
 - But in real life:
    - brokers could use different ticker symbols
    - quote providers could only provide historical or latest quotes
    - quote providers could require different ticker symbols
   
This leads me to the following requirements:
 - It must be possible to keep quote feeds for historical and latest quotes independent from each other.
 - It must be possible to have independent ticker symbols to refer my positions at different brokers.
 
See #3289 which addresses also this problem.

Design sketch:
 - keep only ISIN and WKN inside Security class
 - attach the ticker symobl used by the broker to the buy/sell transaction class
 - keep all parameters of historical and latest quote feed independent and move these parameters from Security class into the respective feed class
 
What do you think about this?

name.abuchen.portfolio/src/name/abuchen/portfolio/online/QuoteFeed.java
KarlBelow, line 16:
```suggestion
    Pattern HOST_PATTERN = Pattern.compile("^https?://([^/]+)/?.*"); //$NON-NLS-1$
```

KarlBelow:
The fields are not pretty aligned in case of CoinGecko:

image

name.abuchen.portfolio/src/name/abuchen/portfolio/online/QuoteFeed.java
buchen: Not sure I understand propose. Can you elaborate?

KarlBelow: Just remove _static final_ from HOST_PATTERN to use the same scheme for all consts.

PR #5607, opened 2026-03-30
separate filter dropdowns in All Transactions view

mierin12:
and fix saved config for transaction filter criteria

Hello,

In the All Transactions view, there is one filter dropdown which consist in a transaction filter and a client filter, while in other views client filters have their own dedicated dropdown.
So this PR proposes to separate those two filters for consistency + it is easier to use smaller dropdowns. 

And : currently the transaction filter is not remembered when switching views or closing the file, this separation also fixes this. 
Unless this was intentional for this particular filter to not be remembered ?

**Before :**
Capture d’écran 2026-03-30 à 19 40 39

**After :**
Capture d’écran 2026-03-30 à 19 38 59

I was wondering if the number of transactions in the title of the view should be updated along with the filters ? It always shown the full total number of transactions, independently from the transaction or client filters.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AllTransactionsView.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AllTransactionsView.java
@@ -11,8 +11,6 @@ import java.util.List;
 
 import jakarta.inject.Inject;
 
-import org.eclipse.jface.action.IMenuListener;
-import org.eclipse.jface.action.IMenuManager;
 import org.eclipse.jface.action.Separator;
 import org.eclipse.jface.action.ToolBarManager;
 import org.eclipse.jface.dialogs.MessageDialog;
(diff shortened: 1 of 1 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/AllTransactionsView.java
```diff
@@ -51,73 +46,11 @@
...
-                            criteria -> {
-                                typeFilter = criteria;
-                                updateIcon();
-                                notifyModelUpdated();
-                            });
-
-            clientFilterMenu = new ClientFilterMenu(getClient(), getPreferenceStore());
-
-            clientFilterMenu.addListener(f -> {
-                setInformationPaneInput(null);
```
buchen, line 77: We do we need this setInfomationPane(null) not anymore?

mierin12: I am not sure to be honest, I did not manage to see a different behaviour with and without it so did not reuse it, but maybe I am missing something. Was it needed even in the current master ? 

I select a security row, the information pane is set up accordingly, then:
- If I change the account filter to an account which possess this transaction, the selection is preserved and the information pane is still related to this security, this happens with and without the setInformationPane(null).
- If I change the account filter to an account which does not possess this transaction, the selection disappear and the information pane becomes empty, this happens with and without the setInformationPane(null).

I am assuming        
 ```
table.getTableViewer().addSelectionChangedListener(
                        e -> setInformationPaneInput(e.getStructuredSelection().getFirstElement()));
```
covers everything on its own.

Also, if it is needed, shouldn't it be needed for the transaction filter too ? I do not see why there would be a different behavior between them, both filter impacts the table content, just on different criteria.

To be on the safe side, I can add it back no problem.

mierin12: (i have added it back in the second commit)

Issue #5574, opened 2026-03-17
Export ex-date of transaction as DateTime


capmlbpkel:
**Describe the bug**
Exporting transactions incl. an ex-date to CSV should export the ex-date in the same format as the date, but exports with month description.

**To Reproduce**
1. Add ex-date to transaction
2. Export transaction as CSV
3. Open export
4. Find (restricted to relevant transaction columns):

```
Date,Ex-Date
2025-07-20 00:00:00,"Jul 15, 2025"
```

**Expected behavior**
```
Date,Ex-Date
2025-07-20 00:00:00,2025-07-15 00:00:00
```

**Desktop (please complete the following information):**
 - OS: Linux
 - Version: 0.82.0

Sn1kk3r5:
you trying to export the ex-date?

capmlbpkel:
I am exporting a transaction that includes an ex-date.

Sn1kk3r5:
Just tried. works as expected. You need to have the ex-date column visible.

capmlbpkel:
@Sn1kk3r5 For me, it does not work. The column is visible. Here is a full example:

**Screenshot of transactions**

https://imgur.com/a/MdICkYi

**CSV with wrong ex-date format**

```
Date,Type,Security,Shares,Quote,Amount,Fees,Taxes,Net Transaction Value,Account,Offset Account,Note,Source,Ex-Date
2026-01-31 00:00:00,Dividend,VANGUARD FUNDS PLC VANGUARD FTS,1,,100.00,10.00,20.00,70.00,Deposit,,,,"Jan 15, 2026"
```

**JSON without any ex-date**

```
{
    "version": 1,
    "transactions": [
        {
            "type": "DIVIDEND",
            "account": "Deposit",
            "date": "2026-01-31",
            "currency": "EUR",
            "amount": 70.0,
            "shares": 1.0,
            "security": {
                "name": "VANGUARD FUNDS PLC VANGUARD FTS",
                "ticker": "VWRD.L",
                "currency": "EUR"
            },
            "units": [
                {
                    "type": "FEE",
                    "amount": 10.0
                },
                {
                    "type": "TAX",
                    "amount": 20.0
                }
            ]
        }
    ]
}
```

Sn1kk3r5:
My bad, I completely misunderstood your point.

You are talking about the format, not the existence. Copy that. Hence, reopen the case. 

[...]

PR #5615, opened 2026-04-09
Ex-Date Support: CSV Import/Export

ZfT2:
added support for ex-date field for the CSV Import/Export.

### DIFF
--- a/name.abuchen.portfolio.tests/META-INF/MANIFEST.MF
+++ b/name.abuchen.portfolio.tests/META-INF/MANIFEST.MF
@@ -19,6 +19,8 @@ Require-Bundle: org.junit,
  slf4j.api,
  net.minidev.accessors-smart,
  net.minidev.json-smart,
+ org.apache.commons.commons-io,
+ org.apache.commons.commons-codec,
  name.abuchen.portfolio.junit,
  org.mockito.mockito-core
 Automatic-Module-Name: name.abuchen.portfolio.tests
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
@@ -4,6 +4,7 @@ import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.dividend;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasAmount;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasCurrencyCode;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasDate;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasExDate;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasFees;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasForexGrossValue;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasGrossValue;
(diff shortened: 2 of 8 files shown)


gamue:
Let's link #5574 here :)

name.abuchen.portfolio.tests/META-INF/MANIFEST.MF
```diff
@@ -19,6 +19,8 @@ Require-Bundle: org.junit,
  slf4j.api,
  net.minidev.accessors-smart,
  net.minidev.json-smart,
+ org.apache.commons.commons-io,
```
gamue, line 22: what's the background of this? Wondering as the new code isn't using them

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/exporter/CSVExporterTest.java
```diff
@@ -0,0 +1,89 @@
...
+import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.util.TextUtil;
+
+@SuppressWarnings("nls")
+public class CSVExporterTest
+{
+    @Test
+    public void testExportsExDateForAccountTransactions() throws Exception
+    {
+        AccountTransaction transaction = new AccountTransaction();
```
gamue:
```suggestion
        var transaction = new AccountTransaction();
```

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/exporter/CSVExporterTest.java
```diff
@@ -0,0 +1,89 @@
...
+        finally
+        {
+            Files.deleteIfExists(file);
+        }
+    }
+
+    @Test
+    public void testLeavesExDateEmptyForPortfolioTransactions() throws Exception
+    {
+        PortfolioTransaction transaction = new PortfolioTransaction();
```
gamue:
```suggestion
        var transaction = new PortfolioTransaction();
```

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
```diff
@@ -789,4 +790,29 @@ public void testRoundingOnExchangeRates()
...
+
+    @Test
+    public void testExDateIsImported()
+    {
+        var client = new Client();
+        var extractor = new CSVAccountTransactionExtractor(client);
+
+        var errors = new ArrayList<Exception>();
+        var results = extractor.extract(1,
+                        Arrays.<String[]>asList(new String[] { "2026-03-31", "2026-03-28", "DE0007164600", "SAP.DE",
```
gamue, line 802:
```suggestion
                        List.of(new String[] { "2026-03-31", "2026-03-28", "DE0007164600", "SAP.DE",
```
haven't checked out the code, but would this also work? I think the currentl way is quite hard to read

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/exporter/CSVExporter.java
```diff
@@ -64,6 +64,7 @@ else if (transaction instanceof PortfolioTransaction portfolioTransaction)
     private void writeAccountTransaction(CSVPrinter printer, AccountTransaction accountTransaction) throws IOException
     {
         printer.print(accountTransaction.getDateTime().toString());
+        printer.print(accountTransaction.getExDate() != null ? accountTransaction.getExDate().toString() : ""); //$NON-NLS-1$
```
gamue, line 67: (personal opinion) for something like this I like stream-api with `ofNullable(..).map(..).orElse()`

gamue (commented):
looks good to me :) Thanks!

name.abuchen.portfolio.tests/META-INF/MANIFEST.MF
ZfT2: It seems to be needed for the infinitest run. Without, it fails, at least on my build environment.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/exporter/CSVExporterTest.java
ZfT2: Personally, I like the explicit typing more :) But I agree in this case var would also be fine :)

gamue: yes, in general I also like explicit typing more, but the code-base of this project prefers var and I think if the type is repeated directly like in this example it's making sense :) But yes, all about personal preferences

ZfT2: ok, I changed it to var, the secound finding also. :)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/csv/CSVAccountTransactionExtractorTest.java
ZfT2: Yours results in List< String > return type, not List< String[] > ...

PR #5617, opened 2026-04-10
Ex-Date Support: Performance Calculation

ZfT2:
From https://github.com/portfolio-performance/portfolio/issues/5513

> Performance calculation: Accrue dividends at the time of the ex-date rather than the payment date, to correctly attribute returns to the holding period

This is a first draft/attempt to apply the ex-date, if available, to dividend payments in the calculation logic... 
Feedback is welcome :)

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/ClientIndexTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/ClientIndexTest.java
@@ -22,6 +22,7 @@ import name.abuchen.portfolio.junit.PortfolioBuilder;
 import name.abuchen.portfolio.junit.SecurityBuilder;
 import name.abuchen.portfolio.junit.TestCurrencyConverter;
 import name.abuchen.portfolio.model.Account;
+import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.Client;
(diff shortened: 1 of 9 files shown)


buchen:
Thanks for starting an implementation.

This draft is using the ex-date (if available) instead of the regular date. I am unsure at the moment how to best handle the monetary values (say the totals series). Let say the reporting period ends on day X. The ex-date is before, but the booking date of the dividend is after. The valuation on day X should be identical to what I would see on my account though. If we now use the ex-date in this case, it would not be identical anymore.

Or am I thinking to complicated?

ZfT2:
Good question... I've only just started looking into the calculation logic... 

What I understand from the discussions on this topic:

https://forum.portfolio-performance.info/t/buchung-von-dividenden-am-ex-tag-oder-bei-gutschrift/1718
and
https://github.com/portfolio-performance/portfolio/issues/2173

Assumption: The idea is that the ex-date should be used for performance allocation and the payment date for liquidity/account balance. (Ideally, a separate "accrued dividends" account would be best in the long run.)

If I understand correctly, that does mean the following in the code: 

The security performance mapping is based on the ex-date (if available)

CalculationLineItem.java
SecurityPerformanceSnapshotBuilder.java

However, the ex-date must not be used in account-based aggregations (ClientIndex, ClientPerformanceSnapshot), instead, the booking or value date is used there.
Might this be the right direction? Did I identify the corect classes?

PR #5619, opened 2026-04-11
Handle negative timestamps in date parsing

### LINKED ISSUE #5618: Negative Timestamp not correctly calculated?
**Describe the bug**
via for example https://api.onvista.de/api/v1/instruments/INDEX/4359526/chart_history?idNotation=4359526&range=MAX&resolution=1D&withCurrentDay=true&withEarnings=false we can get the history of a fund. This dates back very long, so it displays negative timestamps in datetimeLast.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Edit URL of historical quotes
2. Enter the example URL provided above. Date: `$.datetimeLast[*]` , Quote: `$.last[*]`
3. Scroll down to '8.1.1970'
4. See error: the next lines date is displayed ad 22.09.1851, ans subsequent dates make even less sense: 
Image

**Expected behavior**
Calculate the date for dates before 1970

**Screenshots**
see error screenshot.

 - OS: macosx, aarch64
 - Version: 0.82.1 (März 2026)

utafrali:
Fixes #5618

Turns out some APIs like onvista use negative timestamps for pre-1970 dates, which the code wasn't handling. Added a check for negative values and a test case to keep this from breaking again.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeedTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/online/impl/GenericJSONQuoteFeedTest.java
@@ -710,6 +710,19 @@ public class GenericJSONQuoteFeedTest
     }
 
     @Test
+    public void testDateExtractionNegativeTimestamp()
+    {
+        // Date is in seconds from epoch, negative for pre-1970 dates (e.g. onvista API)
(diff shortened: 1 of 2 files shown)


buchen:
Makes sense.
I think we should also apply the same heuristic that checks whether this is a days, milliseconds, or epoch seconds.

buchen:
merged.

PR #5625, opened 2026-04-15
Add template for scalable capital csv import

mjb6:
Hi 

with this change, a user can import scalable capital csv files by using File -> import -> templates -> scalable captial, ie. without manually mapping the csv fields to the portfolio performance fields.

I created / tested it based on a personal csv file that I downloaded from the scalable webpage, which is probably incomplete.


buchen:
Thanks for the contribution!!

PR #5626, opened 2026-04-16
Combine multiple Portfolio Performance files into a newly merged file

Morpheus1w3:
## Summary
Adds support for merging multiple Portfolio Performance files into a new combined file.

## Scope
- loads multiple selected files
- merges them into a new client instance
- reuses matching securities where possible based on a conservative identifier heuristic
- remaps security references in transactions, watchlists, and investment plans
- regenerates UUIDs for imported accounts, portfolios, and transactions to avoid conflicts

## Limitations
- dashboards are intentionally ignored
- taxonomies are currently not merged
- ambiguous securities are not force-merged and may still be duplicated

## Notes for review
This feature is intentionally implemented as a "create new merged file" workflow.
It does not merge additional data into the currently opened file in place.

## Design decisions
- The merge creates a new combined client from multiple selected files.
- It does not merge into the currently opened file in place.
- Security matching is intentionally conservative to avoid incorrect merges.
- Ambiguous matches are left duplicated instead of being force-merged.

## Intentionally out of scope
- Dashboards are ignored.
- Taxonomies are currently not merged.
- Existing workspace/UI state is not preserved as part of the merge result.

## Processing flow
1. The user selects two or more files.
2. The first selected file is loaded as the base target client.
3. Each additional file is loaded as a source client.
4. Securities from the source are matched conservatively against the target.
5. Security references in transactions, watchlists, and investment plans are remapped.
6. UUIDs for imported accounts, portfolios, and transactions are regenerated.
7. The prepared source objects are appended to the target client.
8. The merged result is opened as a new portfolio file.
9. A summary dialog is shown after completion.

## UUID handling
UUIDs for imported accounts, portfolios, and transactions are regenerated after remapping to avoid collisions when combining data from multiple files.

## Security matching
Securities are only reused when the identifier match is unique on both sides.

Current heuristic prefers stronger identifiers first (e.g. ISIN/WKN/currency combinations) and only falls back to weaker matches if they remain unambiguous.
Ambiguous entries are intentionally kept separate.

## Security matching heuristic

The merge logic uses a conservative, priority-based heuristic to decide whether two securities should be reused instead of duplicated.

A match is only accepted if the corresponding key is **unique on both sides** (source and target).  
If a key is missing or ambiguous, the next weaker level may be evaluated.  
If no unique match can be found, the security is imported as a new entry.

| Priority | SecurityMatchLevel   | Required fields            | Meaning | Strength | Typical use case | Risk of false match | Notes |
|----------|----------------------|----------------------------|---------|----------|------------------|---------------------|-------|
| 1 | `ISIN_WKN_CURRENCY` | ISIN + WKN + Currency | Strongest identifier combination | Very high | Fully maintained security master data | Very low | Preferred match level whenever available |
| 2 | `ISIN_CURRENCY` | ISIN + Currency | Match by ISIN with currency as additional safeguard | High | WKN missing, but ISIN is available | Low | Useful when WKN is not maintained consistently |
| 3 | `WKN_CURRENCY` | WKN + Currency | Match by WKN with currency as additional safeguard | Medium to high | ISIN missing, but WKN and currency are available | Medium | More practical for older or locally maintained data |
| 4 | `ISIN_WKN` | ISIN + WKN | Match by two identifiers without currency | High | Currency missing, but both identifiers are present | Low to medium | Still strong if both identifiers are available |
| 5 | `ISIN_ONLY` | ISIN | Match only by ISIN | Medium | Only ISIN is available | Medium | Accepted only if unique on both sides |
| Fallback | `WKN_ONLY` | WKN | Match only by WKN | Low | Only WKN is available | Higher | Treated as a separate fallback, not as a normal primary key level |

## Matching rules

- Matching is evaluated from **highest priority to lowest priority**.
- A match is only accepted if the key occurs **exactly once in the source** and **exactly once in the target**.
- Ambiguous entries are **not force-merged**.
- If no unique match is found, the security is imported as a new security.
- `WKN_ONLY` is treated as a dedicated fallback because it is weaker and more error-prone than the other key combinations.

## Design rationale

This heuristic intentionally prioritizes **safety over aggressive deduplication**.

That means:
- correct merges are preferred over broad merges
- ambiguous securities may remain duplicated
- incomplete real-world data can still be handled through weaker fallback levels
- the logic remains traceable and reviewable


Morpheus1w3:
Second commit add a summary/statistic as below, if the merge is completed (3 times Kommer fle combined):

Popup


Stats

Morpheus1w3:
GitHub CI currently seems to fail in Tycho while resolving Eclipse mirror metadata with HTTP 403 from eclipse.org download/mirror URLs. Local build succeeds, which would be consistent with cached dependencies

Morpheus1w3:
I implemented the feature itself and verified the functional behavior locally.

What I could not reliably reproduce in my local Windows setup is the repository-specific canonical sorting step for the managed properties files. The CI failure appears to be limited to that formatting/order check.

PR #5588, opened 2026-03-23
Modify Scalable PDF Exporter to support new formats and fix bugs

gamue:
Closes #5582 
Closes #5589

# Account Statement

## bugfix: Taxes been imported as taxRefunds
```
01.01.2026 31.12.2025 Auf Kundenebene einbehaltene oder erstattete Kirchensteuer -1,10 EUR 
01.01.2026 31.12.2025 Auf Kundenebene einbehaltene oder erstattete Kapitalertragssteuer -12,31 EUR 
```
The minus indicates that the amount is removed from the account, however it's booked as taxRefund which is wrong. The title indicates "oder erstattete" and guess that's why it was implemented like this. I checked this with own documents, so sure that this change is correct

## change: don't import taxes/taxesRefunds
The taxes, like the lines above, do always belong to a sell, dividend or interest event for which there is an own document. For interests this error was already thrown, but not for taxes. I changed this and also added another document with more events.

## change: fix reading Kirchensteuer
The file from #5582 caused an exception as Kirchensteuer wasn't supported yet. This is working now, however with the changes above, it's not imported but marked as "please import other file"

## change: Support disposals
Normal disposals haven't been supported yet, so added this.

# Dividend/Buy/Sell
## bugfix: wrong amount
Fixed the amount as requested in the forum: https://forum.portfolio-performance.info/t/pdf-import-von-scalable-capital-bank/33088/140

We already had one case in the tests which had DE with . and from the document it's not clear to me if that should have been thousands as well or not. The original case is #5356 and was confirmed that thousands are correct, so adjusted the test.

# Other
- When merging please use squash&merge so there's a reference to the PR, which is nice. It also allowes adjusting the commit message to add gh issue links or similar. I think this should be the default for PRs

PR #5630, opened 2026-04-19
Scalable Capital: Add support for withdrawals and mark interest as failure

buchen:
Issue: #5588

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/scalablecapital/Kontoauszug05.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/scalablecapital/Kontoauszug05.txt
@@ -1,0 +1,91 @@
+PDFBox Version: 3.0.6
+Portfolio Performance Version: 0.82.2.qualifier
+System: macosx | aarch64 | 21.0.2+13-LTS | Amazon.com Inc.
+-----------------------------------------
+Scalable Capital Bank GmbH • Seitzstraße 8e • 80538 München • Deutschland 
+MJyKbEms xEhGY rMWiZaCllbVl
+MTxtzaYCtSkHzYKqq 40 
+34435 zrxDfM Datum 18.02.2026
+Deutschland Seite 1 / 4
+Kontoauszug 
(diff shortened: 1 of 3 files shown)


buchen:
Hi @gamue,

I finally (sorry) found the time to look at the Scalable Capital change. I have merged most of the commits of the pull request.

I have extracted the remaining changes into this pull request.

Adding support for withdrawals is fine (but as the test is baked into the Kontoauszug04.txt I did not separate the change).

But I wonder about the quarterly interest payments, for example this one:

```
01.01.2026 31.12.2025 Erhaltene Zinsen +503,31 EUR 
01.01.2026 31.12.2025 Abgezogener oder erstatteter Solidaritätszuschlag auf Kundenebene -6,92 EUR 
01.01.2026 31.12.2025 Auf Kundenebene einbehaltene oder erstattete Kapitalertragssteuer -125,82 EUR 
```

It is not marked as failure with the hint to use another document. However, I wonder does Scalable has another document for this? And even if it has, it is really a problem to import here? I am for the interest on the account itself, we do not need more than this (and at least my bank would also only report this on the account statement).

I get the point on the dividend payments. There additional document makes sense for more information.

But maybe Scalable is different. What is your thinking?

gamue:
Hi @buchen ,
thanks for the review and merging the changes.

> It is not marked as failure with the hint to use another document. However, I wonder does Scalable has another document for this?

Yes, Scalable has a document for the interests they have paid quarterly in their Broker and now monthly in their new Tagesgeld-product. The document is named `Rechnungsabschluss` and there is existing import code and tests for them.

> And even if it has, it is really a problem to import here? 

That's more a conceptual question I guess. With the separate documents it's possible to link the interest/dividend payments to the corresponding taxes, merging them into a single transaction. Using the account statement, this is a little bit tricky and results in multiple transactions. I think we could theoretically support both but then users would need to remove additional transaction if they also import the `Rechnungsabschluss` documents.

gamue:
It looks like the `Rechnungsabschluss` document isn't provided monthly, but as the interest payments of the Tagesgeld-account is monthly, think it's better to keep the option to import the taxes via the account statement and not throw an error, its then up to the user which document to import.

PR #5636, opened 2026-04-20
Add Freedom24 (Freedom Finance Europe) JSON broker report importer

chefkoch:
## Summary

Adds a new importer for **Freedom24 / Freedom Finance Europe** broker report
JSON files (`.json`), exported from the Freedom24 client portal at
https://freedom24.com/cabinet?tabId=reports

The JSON format specification is documented at
https://freedom24.com/tradernet-api/broker-report (API reference only,
not required for importing).

The importer handles the following transaction types:

- **Stock trades** (buy / sell) — `trades.detailed`, `instr_type = 1`
- **Currency conversions** — `trades.detailed`, `instr_type = 6`,
  mapped to `AccountTransferEntry` between two currency accounts
- **Deposits and withdrawals** — `cash_in_outs` (preferred) or
  `cash_flows.detailed` (fallback), type `bank` only; internal
  block/unblock entries are skipped
- **Interest charges** on negative cash balances —
  `commissions.detailed`, mapped to `AccountTransaction.INTEREST_CHARGE`
- **Promo / gift shares** (DELIVERY_INBOUND) — `securities_in_outs`
  joined with `in_outs_securities.detailed` to resolve the ISIN, which
  is only present in the detailed section

### Cross-currency commission handling

Freedom24 always charges commissions in EUR, even for trades executed in
USD. Portfolio Performance rejects a `FEE` unit whose currency differs
from the portfolio transaction currency. For this reason cross-currency
commissions are emitted as a separate `AccountTransaction.FEES` with a
`GROSS_VALUE` unit carrying the exchange rate. The rate is parsed from
the embedded `comment` field of the trade using a regex pattern.

### Security lookup

Securities are matched in priority order:
1. ISIN + currency (most reliable)
2. Exact `tickerSymbol` match
3. Yahoo-format ticker match — Freedom24 uses exchange-suffixed tickers
   (e.g. `CLSK.US`), while Portfolio Performance stores the Yahoo symbol
   (`CLSK`). Both forms are tried to prevent duplicate security creation
   on re-import.

If no match is found, a new `Security` is created and a `SecurityItem`
is appended so the user can review it in the import wizard.

### Language support

Freedom24 exports are available in multiple languages. The importer has
been tested and verified against all 22 languages available in the
Freedom24 client portal: Arabic, Bulgarian, Czech, Danish, German,
Greek, English, Spanish, Estonian, French, Armenian, Italian, Kazakh,
Lithuanian, Dutch, Polish, Portuguese, Romanian, Russian, Tajik,
Ukrainian, and Chinese.

The three categories of language-dependent strings are:

| Category | Handling |
|---|---|
| Gift shares type | Exact `Set` match (`STOCK_AWARD_TYPES`) |
| Interest charge type | Prefix `Set` match (`INTEREST_CHARGE_PREFIXES`) — handles trailing currency code (EUR / USD) |
| Trade commission prefix (skipped) | Prefix `Set` match (`TRADE_COMMISSION_PREFIXES`) |

---

## Changed files

| File | Change |
|---|---|
| `name.abuchen.portfolio/.../datatransfer/freedom/Freedom24JSONExtractor.java` | New extractor |
| `name.abuchen.portfolio.ui/.../ui/handlers/ImportFreedom24Handler.java` | New UI handler |
| `name.abuchen.portfolio.bootstrap/Application.e4xmi` | Menu entry + icon URI |
| `name.abuchen.portfolio.bootstrap/OSGI-INF/l10n/bundle*.properties` | Labels in all supported languages |
| `name.abuchen.portfolio.ui/icons/logos/freedom24.svg` | Broker logo (16×16) |
| `name.abuchen.portfolio.ui/.../ui/Images.java` | Registered `FREEDOM24_LOGO` |
| `name.abuchen.portfolio.tests/.../datatransfer/freedom/Freedom24JSONExtractorTest.java` | 9 unit tests |
| `name.abuchen.portfolio.tests/.../datatransfer/freedom/*.json` | 9 JSON test fixtures |

---

## Test coverage

| Test | Transaction type | Fixture |
|---|---|---|
| `testKauf01` | BUY (cross-currency commission) | `Kauf01.json` |
| `testVerkauf01` | SELL (cross-currency commission) | `Verkauf01.json` |
| `testEinzahlung01` | Deposit | `Einzahlung01.json` |
| `testEntnahme01` | Withdrawal | `Entnahme01.json` |
| `testZinsen01` | Interest charge (DE) | `Zinsen01.json` |
| `testZinsen01_EN` | Interest charge (EN) | `Zinsen01_EN.json` |
| `testEinlieferung01` | Promo share / ISIN join (DE) | `Einlieferung01.json` |
| `testEinlieferung01_EN` | Promo share / ISIN join (EN) | `Einlieferung01_EN.json` |
| `testWaehrungskonversion01` | Currency conversion | `Waehrungskonversion01.json` |
| `testEinlieferung01_{AR–ZH}` | Gift share – 20 additional languages | `Einlieferung01_{AR–ZH}.json` |
| `testZinsen01_{AR–ZH}` | Interest charge – 20 additional languages | `Zinsen01_{AR–ZH}.json` |

---

### Icon license (blocker)
The `freedom24.svg` logo is derived from Freedom24's corporate branding.
The contributing guidelines require all icons to be
[CC0 licensed](https://creativecommons.org/publicdomain/zero/1.0/).
A replacement icon (e.g. a generic shield or custom neutral icon under
CC0) should be used unless explicit permission is obtained from Freedom24.

### No wizard pre-selection for accounts and depot
The import wizard cannot pre-select the correct account (EUR / USD) or
depot automatically because the JSON report does not contain a reliable
mapping to Portfolio Performance portfolio/account names. The user must
assign accounts manually in the wizard.

### Unsupported instrument types
`instr_type` values other than `1` (stock) and `6` (currency pair) —
such as bonds, ETFs with a different type code, or futures — are silently
ignored. No error is reported to the user.

### No test for commission skip logic
The logic that skips trade-linked commissions in `commissions.detailed`
(type prefix `"Für eine Transaktion:"` / `"For trade:"`) is not covered
by a dedicated unit test.

### No `AssertImportActions` validation in tests
The tests do not call `new AssertImportActions().check(results, "USD")`
to validate overall import action consistency. This should be added once
multi-currency import actions are better understood for this broker.

---

## Forum reference

There is an existing community thread requesting Freedom24 import support.
It is provided here as background context only and is not linked in commits:

https://forum.portfolio-performance.info/t/freedom24-import/18110


chefkoch:
added additional handling and tests for other languages from freedom24 - updated also pull request summary

chefkoch:
After some communication back and forth with the Freedom24 support I received the following statement:

> "The /cabinet?tabId=reports endpoint is an internal action only. Even if it is visible to you, we are not obligated to support it for external users. For external users, there are API methods available that provide similar functionality to the "Broker Report": https://freedom24.com/tradernet-api/broker-report JSON keys may contain information in different languages depending on the client's language. There are no plans to change this."

How should we handle this?

---

### Option 1 – PDF-based importer

| ✅ Pros | ❌ Cons |
|--------|--------|
| Simple user export, no API key needed | Less data, than JSON |
| Offline, imports possible as long as export file is archived | Multi-language string matching: complex and hard to maintain |

---

### Option 2 – Keep the current JSON importer (multi-language)

| ✅ Pros | ❌ Cons |
|--------|--------|
| Simple user export, no API key needed | Multi-language string matching: complex and hard to maintain |
| full data coverage | Internal endpoint — may break without notice |
| Offline, imports possible as long as export file is archived |  |

---

### Option 3 – Simplified JSON importer (English only)

| ✅ Pros | ❌ Cons |
|--------|--------|
| Simpler code, easier to maintain | English portal language required; users must switch manually |
| full data coverage | Internal endpoint — may break without notice |
| Offline, imports possible as long as export file is archived |  |

---

### Option 4 – API-based importer

| ✅ Pros | ❌ Cons |
|--------|--------|
| Officially supported endpoint, no language issues | Requires API key setup by the user |
|  | Online only — no imports possible after account closure |

buchen:
Hi @chefkoch,

let me first understand how users can actually get access to the JSON file. If it is an internal API, what do they have to do to get to the this JSON? If that is reasonable easy then option 3 (currently implemented) sounds reasonable to me. Requiring an API key, an additional setup, seems cumbersome.

About:

> Cross-currency commission handling

The commission handling works like this: The account transaction must be in the currency of the account. And the account transaction can either carry fees in the account currency or it can carry fees in the instrument currency. In the latter case, it must include the currency conversion. So in this case, if you have a trade in USD and the instrument is in USD and you have the account in EUR, then you can record both fees in one transaction: The EUR fees and the USD fees. No need to have two transactions.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/freedom/Freedom24JSONExtractor.java
```diff
@@ -0,0 +1,1028 @@
...
+    // "Currency exchange of trade (USD) to a currency exchange of the service plan (EUR) is 0.8494"
+    // Captured group 1 = commission-currency per trade-currency (e.g. 0.8494 EUR per USD)
+    private static final Pattern FX_RATE_PATTERN = Pattern.compile(
+                    "Currency exchange of trade \\(\\w+\\) to a currency exchange of the service plan \\(\\w+\\) is (\\d+(?:\\.\\d+)?)");
+
+    private final Client client;
+    private List<Security> allSecurities = new ArrayList<>();
+
+    /** Pre-selected depot for the import wizard, or {@code null}. */
+    private Portfolio portfolio = null;
```
buchen: I am not sure that this is used in meaningful way. Is it ever assigned? I would propose to leave it to the dialog/user to assign the right accounts.

chefkoch:
> let me first understand how users can actually get access to the JSON file. If it is an internal API, what do they have to do to get to the this JSON? If that is reasonable easy then option 3 (currently implemented) sounds reasonable to me. Requiring an API key, an additional setup, seems cumbersome.

Ok, below you can find the steps to take to get the JSON via mail from https://freedom24.com/cabinet?tabId=reports
However, this is the method, not "officially supported" by freedom according to their mail.
https://github.com/user-attachments/assets/cccb413a-9b59-4f79-93a7-06227f0dce9f

> ... then option 3 (currently implemented) sounds reasonable to me...

ok, understood: preferably no PDF and no API usage....
Using the **current** JSON implementation in multilanguage (**option 2**) or
Option 3 reducing and simplifying it to EN (or DE) only ?


> > Cross-currency commission handling
> 
> The commission handling works like this: The account transaction must be in the currency of the account. And the account transaction can either carry fees in the account currency or it can carry fees in the instrument currency. In the latter case, it must include the currency conversion. So in this case, if you have a trade in USD and the instrument is in USD and you have the account in EUR, then you can record both fees in one transaction: The EUR fees and the USD fees. No need to have two transactions.

you are correct, that portfolio can handle it that way, but
In my profile or user account at freedom I have (at least) two currency accounts: one EUR (my main account) and one USD. Before buying a USD stock I need to convert **manually** some amount from EUR to USD. This is also a separate transaction within the importer.
When buying the stock, my USD account is reduced by USD and EUR account is reduced by EUR fee.
Covering it in portfolio performance the same way, by two transactions
a)  ensures to have **correct account balances displayed in portfolio performance** and
b) prevents and skips calculation any conversion rates for the buying/fee transactions.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/freedom/Freedom24JSONExtractor.java
chefkoch: > I am not sure that this is used in meaningful way. Is it ever assigned? I would propose to leave it to the dialog/user to assign the right accounts.

yes, it's been corrected with the latest commit. Reason for this was my first (internal) attempt to assign any accounts  by name.... but the user-selection in the wizard is working fine 
user selection

PR #5637, opened 2026-04-20
Adding PostFinance PF Pay support

shochmuth:
PostFinance own PF Pay payment service uses a previously not supported syntax `PF PAY KAUF/ONLINE SHOPPING ` to describe removals of that kind.

Adding support for this removal syntax in the `PostFinancePDFExtractor`.
To isolate the concern for future changes, a new dedicated test file and test case has been added, similar to the TWINT payment service transactions.


shochmuth:
@Nirus2000 Any reason for the `do-not-merge`? Prefer an additional feature issue or an updated rebase of the merge? Will follow up, the feature itself seems very low impact to me but would remove manual adjustments after my PDF imports. thx

shochmuth:
Closing as no reason provided why it is being blocked for merge.

buchen:
I do not think the label had anything directly to do with your change.

PR #5644, opened 2026-04-25
[1/7] CSS :: Add CSS-backed core theme color infrastructure

Morpheus1w3:
This PR is part of a stacked CSS color migration series and should be reviewed in order.

Adds the core infrastructure for CSS-backed theme colors. This prepares shared UI color values for central configuration through the theme CSS files.

**[#5644 – Add CSS-backed core theme color infrastructure](https://github.com/portfolio-performance/portfolio/pull/5644)**
[#5645 – Move data series chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5645)
[#5646 – Move portfolio balance chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5646)
[#5647 – Move securities chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5647)
[#5648 – Move payments palette colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5648)
[#5649 – Move color gradient definitions to CSS](https://github.com/portfolio-performance/portfolio/pull/5649)
[#5650 – Add color architecture debug view](https://github.com/portfolio-performance/portfolio/pull/5650)

**Note:**
This series was split to make the review of each semantic color group easier.
Some later PRs may not build independently against master until the preceding PRs have been merged.

The split is artificial to some extent because shared files such as CSS files, plugin.xml, ThemeAddon and ElementProvider are touched by multiple steps. I tried to keep each PR focused and to show only the final intended change per semantic area, but minor merge follow-up may still be required after the series is reviewed.

### DIFF
--- a/name.abuchen.portfolio.ui/css/shared/light.css
+++ b/name.abuchen.portfolio.ui/css/shared/light.css
@@ -92,6 +92,8 @@ CustomColors {
 	red-foreground: #7f0000;
 	green-foreground: #377500;
 	gray-foreground: #707070;
+	
+	hyperlink: "#COLOR_LINK_FOREGROUND";
 }
 
 ValueColorScheme.standard {
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ColorsThemeCSSHandler.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/ColorsThemeCSSHandler.java
@@ -14,47 +14,45 @@ public class ColorsThemeCSSHandler implements ICSSPropertyHandler
     public boolean applyCSSProperty(Object element, String property, CSSValue value, String pseudo, CSSEngine engine)
                     throws Exception
     {
-        if (element instanceof ColorsThemeElementAdapter colorsThemeAdapter)
+        if (!(element instanceof ColorsThemeElementAdapter colorsThemeAdapter))
+            return false;
+
+        Colors.Theme theme = colorsThemeAdapter.getColorsTheme();
+
+        switch (property)
         {
-            Colors.Theme theme = colorsThemeAdapter.getColorsTheme();
-
-            switch (property)
-            {
-                case "default-foreground": //$NON-NLS-1$
-                    theme.setDefaultForeground(CSSSWTColorHelper.getRGBA(value));
-                    break;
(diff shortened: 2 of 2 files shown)


name.abuchen.portfolio.ui/css/shared/light.css
```diff
@@ -92,6 +92,8 @@ CustomColors {
 	red-foreground: #7f0000;
 	green-foreground: #377500;
 	gray-foreground: #707070;
+	
+	hyperlink: #6FC5EE;
```
buchen: The hyperlink used to use the default color of the operating system. I think it is fair to move it to the CSS file. However, we still should use the OS default. I believe something like this can be done:

```
	hyperlink: "#COLOR_LINK_FOREGROUND";
```

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/Colors.java
```diff
@@ -18,20 +18,28 @@ public final class Colors
...
+        private Color defaultForeground;
+        private Color defaultBackground;
+        private Color chipBackground;
+        private Color warningBackground;
+        private Color redBackground;
+        private Color greenBackground;
+        private Color redForeground;
+        private Color greenForeground;
+        private Color grayForeground;
+        private Color hyperlink;
```
buchen: I understand why you remove the default values. However, I know from installations where there were problems reading the CSS files. Then users were greeted with hundreds of null pointer exceptions (now: hundreds of IllegalStateExceptions).

I know it's duplicating the colors in a way, but it's a safe fallback. Therefore I would prefer to make sure here are non-null values present.

buchen (commented):
Two things:

First, let me know the "why" for this change. I understand the "what", but it helps me reviewing if I know the why.

And second, thank you very much for splitting this up into smaller pull requests. That makes it much easier to look at the individual change and discuss the content.

Morpheus1w3:
Thank you, and I am glad the split makes the review easier.

The main reason for this change is to make the color and theme behavior more explicit and consistent. Some parts were already prepared for CSS based theming, but the actual behavior was still mixed in places because Java defaults were still covering missing or incomplete CSS values. My intention was to make those cases visible and to separate clearly between values that should stay as defaults and values that are good candidates for CSS based theming.

PR #5645, opened 2026-04-25
[2/7] CSS :: Move data series chart colors to CSS

Morpheus1w3:
This PR is part of a stacked CSS color migration series and should be reviewed in order.

Moves the default data series chart colors from hard-coded Java constants to CSS-backed configuration. This keeps the existing semantic data series names while making their default colors theme-configurable.

[#5644 – Add CSS-backed core theme color infrastructure](https://github.com/portfolio-performance/portfolio/pull/5644)
**[#5645 – Move data series chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5645)**
[#5646 – Move portfolio balance chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5646)
[#5647 – Move securities chart colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5647)
[#5648 – Move payments palette colors to CSS](https://github.com/portfolio-performance/portfolio/pull/5648)
[#5649 – Move color gradient definitions to CSS](https://github.com/portfolio-performance/portfolio/pull/5649)
[#5650 – Add color architecture debug view](https://github.com/portfolio-performance/portfolio/pull/5650)

**Note:**
This series was split to make the review of each semantic color group easier.
Some later PRs may not build independently against master until the preceding PRs have been merged.

The split is artificial to some extent because shared files such as CSS files, plugin.xml, ThemeAddon and ElementProvider are touched by multiple steps. I tried to keep each PR focused and to show only the final intended change per semantic area, but minor merge follow-up may still be required after the series is reviewed.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/DataSeriesColors.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/DataSeriesColors.java
@@ -1,0 +1,286 @@
+package name.abuchen.portfolio.ui.util;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.RGBA;
+import org.eclipse.swt.widgets.Display;
+
+/**
+ * Holds the default colors for the client data series shown in the dashboard
+ * and performance charts.
+ * <p>
+ * The field initializers below are fallback values only. The authoritative
+ * colors are defined in the CSS theme files (light.css / dark.css) under the
+ * {@code DataSeries} selector and are applied via
+ * {@code name.abuchen.portfolio.ui.theme.DataSeriesColorsCSS} when the theme
+ * engine runs (see {@code ThemeAddon}). Because the theme is
+ * always applied at startup, the CSS values - not these defaults - determine
+ * what is rendered. When changing a default color, update the CSS as well.
+ */
+public final class DataSeriesColors
+{
+    private static final DataSeriesColors INSTANCE = new DataSeriesColors();
+
+    private Color totalsColor = Colors.BLACK;
+
+    private Color investedCapitalColor = Colors.getColor(235, 201, 52); // #EBC934;
+    private Color absoluteInvestedCapitalColor = Colors.getColor(235, 201, 52); // #EBC934;
+
+    private Color transferalsColor = Colors.DARK_GRAY;
+    private Color transferalsAccumulatedColor = Display.getDefault().getSystemColor(SWT.COLOR_YELLOW);
+
+    private Color taxesColor = Colors.RED;
+    private Color taxesAccumulatedColor = Colors.RED;
+
+    private Color absoluteDeltaColor = Display.getDefault().getSystemColor(SWT.COLOR_BLUE);
+    private Color absoluteDeltaAllRecordColor = Display.getDefault().getSystemColor(SWT.COLOR_BLUE);
+
+    private Color dividendsColor = Display.getDefault().getSystemColor(SWT.COLOR_DARK_MAGENTA);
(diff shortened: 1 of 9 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/theme/DataSeriesColorsElementAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.eclipse.e4.ui.css.core.dom.ElementAdapter;
+import org.eclipse.e4.ui.css.core.engine.CSSEngine;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import name.abuchen.portfolio.ui.util.DataSeriesColors;
+
+@SuppressWarnings("restriction")
+public class DataSeriesColorsElementAdapter extends ElementAdapter
```
buchen: I propose to add

```
    @Override
    public String getCSSClass()
    {
        return ValueColorScheme.current().getIdentifier();
    }
```

This way we can override the colors per value color scheme. In Asia for example, red is the color for positive development and green for negative ones. And this pr adds the CSS property "performance-positive-color". The negative color with orange does not fit there.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/DataSeriesColors.java
```diff
@@ -0,0 +1,274 @@
+package name.abuchen.portfolio.ui.util;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.graphics.Color;
+import org.eclipse.swt.graphics.RGBA;
+import org.eclipse.swt.widgets.Display;
+
+public final class DataSeriesColors
```
buchen, line 20: I haven't looked at the other pull requests in detail. If we introduce a lot more classes, we might have to think about another strategy - maybe nesting - to avoid polluting the packages.

Morpheus1w3: @buchen 
I've asked ChatGPT and the proposal is as follows:

New File Name
`FooBarCSS.java` rather than `FooBarElementAdapter.java` and `FooBarCSSHandler.java`.

### Class Design
```java
package name.abuchen.portfolio.ui.theme;

public final class FooBarCSS
{
    private FooBarCSS()
    {
    }

    public static final class ElementAdapter extends org.eclipse.e4.ui.css.core.dom.ElementAdapter
    {
        // previous content from FooBarElementAdapter
    }

    public static final class Handler implements org.eclipse.e4.ui.css.core.dom.properties.ICSSPropertyHandler
    {
        // previous content from FooBarCSSHandler
    }
}
```

The  references would change to:

```java
new FooBarCSS.ElementAdapter(target, engine)
```

And in plugin
```xml
handler="name.abuchen.portfolio.ui.theme.FooBarCSS$Handler"
```

What did you think, can I apply this proposal for existing classes and the open PR's? I'm happy to adjust my PR's, if you agree with it.

Morpheus1w3:
Added one follow up commit to nest the CSS helper classes, based on your earlier comment about package pollution.

Functional behavior should stay unchanged

buchen:
@Morpheus1w3  Sorry for taking so long for personal reasons. I didn't have too much time to look into the project. I have forced-pushed some minor changes.

There is one thing I am still thinking about. The colors for the data series are default values. If the user adds the data series to a chart, the color is actually copied into the chart configuration. The chart configuration does not differentiate between light and dark mode. With this change it depends on the current theme of the user which color is copied into the chart configuration. I am asking myself if that is confusing to the user. I mean, the default colors are better, particularly for the total series (black in dark mode does not work well), but change is relevant only when initially adding the data series. Do you have any opinion there?

By the way, for the dashboard widget that is not a problem because only the data series is stored and always the default colors are used. The user has no option to change the colors.

PR #5651, opened 2026-04-26
IBFlex: allow import from Flex Web Service

buchen:
IBKR allows retrieving preconfigured Flex Queries via a rudimentary HTTP interface. Doing so requires a query ID which identifies the Flex Query, and a secret token. The secret token can be obtained from the web interface an expires after a while.

Add a new menu item which allows importing from such a Flex Web Service. This saves users from logging in, manually selecting the query, executing it, etc.

The Flex Web Service always runs a Flex Query with a fixed date range configured when creating the query. This means that we can't choose what date range to fetch. The easiest thing for the user is to configure the query with the maximum duration (currently 365 days). The downside of this is that it populates the import wizard with a lot of duplicate items after the first import.

Filter extracted items to exclude ones we've likely already imported based on the transaction date. The user can clear the date we're using for this via the settings dialog.

Issue: #5352

[renamed config to IBFlexConfiguration; rebased to master]


buchen:
@lmb can you check this pull request? rebased, and tiny changes

buchen:
One more thing: the transactions are removed by the cutoff date. The cutoff date is the transaction date. It is parsed from the properties ```tradeDate```, ```reportDate```, and ```dateTime```.

I wonder if that filters too many transactions, for example if a transaction is corrected/late reported. The trade date might be in the past (and hence processed by a previous import). Does IBFlex differentiate between valuta and booking date?

lmb:
Just did a build + data fetch, all seems to be in order.

Re tradeDate, reportDate, dateTime Claude found the following "references" for me:

- https://www.ibkrguides.com/reportingreference/reportguide/tradesfq.htm
- https://www.ibkrguides.com/reportingreference/reportguide/cash%20transactionsfq.htm
- https://www.ibkrguides.com/reportingreference/reportguide/corporate%20actionsfq.htm

- CorporateAction, FxTransactions: reportDate, dateTime
- CashTransaction: dateTime, settleDate (undocumented), reportDate (undocumented)
- Trade: reportDate, tradeDate, tradeTime, settleTargetDate, dateTime (undocumented)

All of this is made more complicated because users can remove fields from an export. So we have some unit tests which have CashTransaction with a reportDate but without dateTime, etc. Our parsing has been so lax that we accept those as well.

Most of the time reportDate == one of the other date fields, I've found one example where it is dateTime + 1d. Trades have settleTargetDate, CashTransactions have settleDate. In short: I don't know how to detect whether a transaction has changed after the fact.

We can drop the cutoff date, and rely on the import wizard deduplicating things. That only works if that has a way of detecting "updated" transactions. Is there code for this?

lmb:
@buchen I think this fell through the cracks. What is missing to get this merged?

PR #5654, opened 2026-04-27
Dashboard: Add collapsible SectionHeader widget

andreas13xxx:
Issue: https://forum.portfolio-performance.info/t/abschnitte-im-dashboard/39179


buchen:
Thanks for the contribution. :-)

I have merged the change. You will see I did some minor changes. I added the translations, but I also renamed the widget to CollapsibleSection because I thought that is better reflecting what it is doing. If you want to reuse or update your existing file you might want to edit the XML directly because the key has changed.

PR #5665, opened 2026-05-01
Add correlation matrix view

Morpheus1w3:
> based on https://github.com/portfolio-performance/portfolio/tree/feature_correlation_matrix 

This change adds a new view to calculate and display the correlation matrix between selected data series.

The matrix is based on return data and uses only matching dates. Missing or invalid correlations are handled and shown in the UI.

Users can select data series via the existing selection mechanism (securities, accounts, portfolios, classifications).

correlation matrix


Morpheus1w3:
Second commit add a gradient color scheme.

* Only correlations with an absolute value of 0.5 or higher are highlighted.
* This keeps the matrix readable and avoids overemphasizing weak relationships.
* The colors are intended as visual guidance only. The numeric value remains the primary information.

image

Issue #5662, opened 2026-05-01
Tax calculation uses year-end closing price instead of actual purchase price


prieurcath:
**Describe the bug**
The application incorrectly calculates taxable capital gains and losses by applying the market price on December 31st instead of the actual acquisition cost of the assets. This results in inaccurate tax reports and financial summaries.

**To Reproduce**
1. Go to '..Report Calculation.'
2. Click on '.claculation select a year and any account or all...'
3. Scroll down to '..realised capital gain and pass on right to see details..'
4. See error all calculation are reflecting the buying value execpt for shares acquired the year before the nb of shares are right but the price is the last market value and not the aquired price

The system appears to use a year-end mark-to-market valuation, pulling the December 31st price for all taxable calculations regardless of when the asset was acquired.

**Expected behavior**
Gain/loss calculations should be based on the actual purchase price (or a recognized accounting method such as FIFO or average cost) recorded at the time of each transaction.
Impact:

Tax reports are inaccurate and may lead to compliance or filing issues.
Users cannot reliably track real cost basis or taxable gains.
Exported data reflects market valuation rather than historical cost.
Technical Notes for Investigation:

Search the codebase for references to year-end valuation: 31-12, closing_price, market_value, revalue, or mark_to_market.
Verify the capital gains/tax module to ensure cost basis retrieval queries transaction history (purchase_price, cost_basis, acquisition_date) rather than price feed endpoints.
Confirm that the accounting method (FIFO, AVCO, etc.) is correctly applied before calculating taxable amounts.
Check configuration files or constants that may enforce valuation_method = year_end instead of historical_cost.
Steps to Reproduce (Generic):

Import or record a security purchase in a prior tax year.
Record a sale of that security in the current tax year.
Generate a tax or capital gains report.
Compare the calculated gain/loss with the actual purchase price vs. sale price. The discrepancy will match the December 31st price difference.

**Screenshots**
I cannot as the window only appear when you pass the mouse

**Desktop (please complete the following information):**
 - OS:Window 11
 - Version: 0.83.2 (April 2026)

prieurcath:
https://github.com/portfolio-performance/portfolio/actions/runs/25289386602

https://github.com/portfolio-performance/portfolio/pull/5666

Morpheus1w3:
@prieurcath 
Just in addition to my recent PR.

[...]

Issue #3387, opened 2023-06-07
Realized capital gains calculated incorrectly


TheJiahao:
**Describe the bug**
Realized capital gains in "Calculation" view is calculated incorrectly. Purchase price of positions bought before the reporting period is its value at the beginning of the reporting period, instead of the real purchase price. That makes the feature unsuitable for tax reporting.

**To Reproduce**
1. Add empty instrument "A"
1. Create following transactions:
    2021-01-01: Buy 1 instrument "A" for 1 EUR
    2022-12-31: Sell 1 instrument "A" for 1.5 EUR
1. Create manual quote 10 EUR at 2021-12-31 for instrument "A".
1. Go to "Performance" > "Calculation"
1. Set the reporting period to year 2022 i.e. 2021-12-31 (excl.) to 2022-12-31
1. See the error: under "Realized Capital Gains", realized capital gains of instrument "A" is $1.5-10=-8.5$ EUR

**Expected behavior**
The purchase price of instruments bought before reporting period should be its real purchase value. In previous case, purchase price should be 1 EUR and realized capital gains 0.5 EUR.

**Possible causes**
- Method call inside `addCapitalGains` in `ClientPerformanceSnapshot.java`, [line 362](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/ClientPerformanceSnapshot.java#L362).
  ```java
  SecurityPerformanceSnapshot securityPerformance = SecurityPerformanceSnapshot.create(client, converter, period,
                  snapshotStart, snapshotEnd, SecurityPerformanceIndicator.CapitalGains.class);
  ```
- Method `create` in `SecurityPerformanceSnapshot`, [lines 52-83](https://github.com/buchen/portfolio/blob/master/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/SecurityPerformanceSnapshot.java#L52)

**Screenshots**
![transactions](https://github.com/buchen/portfolio/assets/42709836/872d5ad3-f612-4e8b-89a7-21eb21ed34ab)
![quotes](https://github.com/buchen/portfolio/assets/42709836/f2bd74bc-fee5-40ba-86d2-5b5e0d214c19)
![calculation](https://github.com/buchen/portfolio/assets/42709836/499e16aa-d023-4fce-b53a-bb1f10b8449e)

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 0.63.1

[...]

PR #5666, opened 2026-05-03
Fix realized capital gains calculation and add period adjustment

Morpheus1w3:
This change improves how realized capital gains are displayed in the performance calculation.
The existing period-based realized gains remain unchanged.

A new purchase-based realized gains line is added. It uses the actual historical purchase cost, based on FIFO or moving average. This helps users compare sales in the selected period against the original purchase cost.

A new adjustment line is added to keep the overall period-based performance calculation consistent.
The total portfolio value and the existing performance logic are not changed.

####Review note
The purchase-based value is built from transactions before the reporting period, so sales inside the selected period can be compared against the historical purchase cost.

The adjustment line keeps the existing period-based performance calculation consistent.

image
image


Morpheus1w3:
At name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/PerformanceCalculationWidget.java a minor Bugfix Was implemented.

For example "Cash Fremdwährungsgewinne" is basically an addition. However, it could happen that the amount is negative. So it' lead into confusion, if the calculation method is an addition, but the amount is negative and subtracted.

Sn1kk3r5:
closes: #5662

prieurcath:
Thanks for this :)

mierin12:
Hmm if this goes through we need to be extra extra clear about what are those two new values. What is the meaning of the 336.8 EUR value ?
There are two different definitions for realized/unrealized gains now. Realized gains are now related to purchase cost, but unrealized gains remains related to the start of the period ? Why not related to the purchase cost too then ?

My understanding is that it was simply not the task of Performance/Calculation to provide such kind of values. 
Performance Calculation was meant to provide information about the performance since a start date. Whatever happened before is irrelevant, as at the start everything is considered invested, both gains and principal.
https://forum.portfolio-performance.info/t/realized-capital-gains-doesnt-match-closed-trades/38737
https://github.com/portfolio-performance/portfolio/issues/3387

Maybe Trades view or Payment/Closed Trades view or the Trades Profit/Loss widget are more usable for this kind of topic. But each of them have different characteristics but we can work on them:

| View | Account filter ? | Period filter ? | FIFO and moving average ? | Before or after taxes and fees ? |
| --- | ----------- | --- |--- |--- |
| Trades | No | No | Both | Both | 
| Payment/Closed trades | Yes | Yes | FIFO only | Before only |
| Trades Profit/Loss  widget| Yes | Yes | Both | After only |

(before tax and fees = without tax and fees = pure price gains)
(after tax and fees = with tax and fees = gains reduced by tax and fees)

Another thing, I am assuming those "taxable gains" determination should be gains AFTER fees, right ? But the Performance Calculation view only provides Gains that are before tax and fees. So if you calculate your taxable gains from this view, then it means you consider that you have to pay tax on pure price gain, not yet reduced by the differents fees. So you consider you have to pay taxes on your fees.. 

All in all, PP is not made for tax calculation.

Morpheus1w3:
Existing calculation method is even misleading. Alternative proposal as below.

#### Calculation Method (Performance)

```
INITIAL
+ CAPITAL_GAINS
+ EARNINGS
- FEE
- TAX
+ FOREX
= FINAL
```

prieurcath:
Even "Portfolio Performance is focusing the performance of investments, t" . the calculation on performance when Securities have been sold the profit must be calculated on initial price paid and not on the last cotation. If fact it's the case on curent year just apprear to be different if the acquisition was made the previous year.

PR #5669, opened 2026-05-05
Add dashboard template export/import functionality

andreas13xxx:
Dashboards can now be exported as JSON template files and imported into any portfolio file. Templates contain the dashboard structure (columns, widgets, configuration) without file-specific entity references (accounts, securities, taxonomies).

- DashboardTemplateExporter: serializes dashboard to JSON, strips entity reference keys (DATA_SERIES, CLIENT_FILTER, TAXONOMY, etc.)
- DashboardTemplateImporter: deserializes JSON template, validates format, creates new dashboard with fresh UUID
- UI integration in DashboardView context menu and new dashboard dialog
- Full i18n for all 18 supported languages

Issue: https://forum.portfolio-performance.info/t/dashboard-templates/39222


andreas13xxx:
Good idea. The main advantage is that adding a new Config value forces you to decide on exportability right away (compiler-enforced), instead of hoping someone remembers to update the separate ENTITY_REFERENCE_KEYS set in the exporter.
But it breaks the separation between the core model and the exporter and needs extra effort with the configuration maps. Not a blocker, but worth noting.

Issue #5679, opened 2026-05-12
Report month names in Finnish are conjugated


thurenius:
**Describe the bug**
The month pattern format used in Portfolio DateTimeFormatter (_"MMMM"_) in non-English localizations results in e.g. Finnish using the conjugated / partitive month names. When the month is used as a column / grid label, the month names should be used as-is without conjugation. 

**To Reproduce**
1. Go to **Reports -> Performance - > Payments -> Overview** (in Finnish: Maksut -> Yleiskatsaus).
2. The formatted captions using month names are grammatically wrong.

**Expected behavior**
Displaying the correct month names without conjugation.

This can be achieved by replacing the formatter patterns:
`"MMMM" -> "LLLL"`

**Screenshots**
The conjugated name of January in Finnish using the current formatter pattern is **"tammikuuta"**, when it should be just the name of the month, aka. **"tammikuu"**.

<img width="99" height="98" alt="Image" src="https://github.com/user-attachments/assets/bd747d89-c334-4883-ae31-0f5b6ae9245b" />

**Desktop:**
 - OS: Linux
 - Version: 0.83.3.202605100919

PR #5680, opened 2026-05-12
Fix Finnish month name formatting / conjugation

thurenius:
The report month names used as captions should not be conjugated.

Issue: https://github.com/portfolio-performance/portfolio/issues/5679


Morpheus1w3:
Closes #5679

PR #5681, opened 2026-05-12
Support custom security attributes in CSV import

Morpheus1w3:
* https://forum.portfolio-performance.info/t/custom-csv-import-selbst-angelegte-attribute-wertpapiere/39271
* https://forum.portfolio-performance.info/t/wertpapier-attribute-per-csv-importieren-und-aktualisieren/5696

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
@@ -13,6 +13,7 @@ import name.abuchen.portfolio.datatransfer.csv.CSVImporter.Column;
 import name.abuchen.portfolio.datatransfer.csv.CSVImporter.DateField;
 import name.abuchen.portfolio.datatransfer.csv.CSVImporter.Field;
 import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.online.impl.YahooFinanceQuoteFeed;
 
 /* package */class CSVSecurityExtractor extends BaseCSVExtractor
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/model/ClientFactory.java
```diff
@@ -1717,6 +1720,34 @@ private static void fixNullSecurityEvents(Client client)
...
+            {
+                name = originalName + " (" + index + ")"; //$NON-NLS-1$ //$NON-NLS-2$
+                index++;
+            }
+
+            attribute.setName(name);
+            usedNames.add(name);
+        });
+    }
+
```
buchen, line 1750: I do not think we should silently rewrite user data when loading the file.

I am also not sure: when does this happen? Do users really create multiple attributes with the same name?

Maybe the import could do something different:

* show an error if importing into an attribute that is duplicated
* use the id instead (it is unique, but I am not sure if it is exposed at the moment)
* Import into the first attribute with this name

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/csv/CSVSecurityExtractor.java
```diff
@@ -72,7 +76,27 @@ void extract(List<Item> items, String[] rawValues, Map<String, Column> field2col
 
         // check if the data contains price
 
+        importAttributes(security, rawValues, field2column);
+
         getSecurityPrice(Messages.CSVColumn_DateQuote, rawValues, field2column)
                         .ifPresent(price -> items.add(new SecurityPriceItem(security, price)));
     }
+
+    private void importAttributes(Security security, String[] rawValues, Map<String, Column> field2column)
```
buchen, line 85: One more thing that caught my eye: this is editing the actual security directly. It then does not make any difference if the user cancels the import on the review page or not. But the user flow is: pick the CSV, configure the columns, review the result, press OK to accept. The users expects that when pressing cancel, the data is not written.

buchen:
Superseeded by #5857

I think the core problem to fix was for the extractors not to use the label but the already existing id. And then on top introduce a new update item that applies the changes only after the user accepted them.

PR #5689, opened 2026-05-15
Add hover toolbar for dashboard widget actions

Morpheus1w3:
This PR adds a small hover toolbar to dashboard widgets.

The toolbar makes existing widget actions directly discoverable without requiring a right-click on the widget title.

Widgets always expose the existing settings action. Widgets that already provide a detail-view action can expose that action in the same toolbar.

No widget configuration logic is changed. The toolbar only provides an additional entry point to existing actions.

image


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/earnings/EarningsChartWidget.java
```diff
@@ -121,7 +121,7 @@ public Composite createControl(Composite parent, DashboardResources resources)
 
         container.layout();
 
-        HoverButton.build(title, container, chart, chart.getPlotArea().getControl()).withListener(new HyperlinkAdapter()
+        HoverWidgetToolbar.attachViewListener(title, new HyperlinkAdapter()
```
buchen, line 124: I believe the button is now only shown if the mouse hovers over the title. Previously, it was shown when the mouse was also hovering across the plot area and the chart itself. So, just wondering if that is intentional?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/HoverWidgetToolbar.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private HoverWidgetToolbar(Control leadingControl, Menu menu)
+    {
+        this.leadingControl = leadingControl;
+        this.menu = menu;
+
+        hookMouseListeners(leadingControl);
+
+        if (leadingControl instanceof Composite)
+            hookMouseListenersRecursively((Composite) leadingControl);
```
buchen, line 45: At the moment the leading control is only a label (unless we extend the "active" area) so this is not used and the "recursively" kind of misleading.

buchen (commented):
Hi @Morpheus1w3,

I like the idea - new users sometimes struggle to find the actions.

Two thoughts:

First, I propose that we show the contact section only after hovering for 300 milliseconds or 500 milliseconds so that if I move fast around with the mouse, I don't get that visual noise of appearing buttons and removing buttons.

Second, I think the label#setData to store the listener is an indirection we do not really need. What do you think about adding adding a ```contributeToolbar``` method to the widget to make that explicit. We could use the ```IAction``` interface..

```java
@Override
public void contributeToolbar(Builder builder) {
    builder.add(new SimpleAction(Messages.MenuGoToView, Images.VIEW_SHARE,
        a -> part.activateView(view, config)));
}
```

Let me know what you think. I know the pull request is hanging around for some time now (sorry for that) so let me know if you still have the time and setup to apply some changes. If you want, I could also give it a try

buchen:
Why close the PR?

I am happy to entertain the feature (I spend time testing and reviewing the code) but not directly in the current form

PR #5677, opened 2026-05-10
Refactor lazy security performance cost API to parameterized cost met…

Morpheus1w3:
# Refactor cost and capital gains access in LazySecurityPerformanceRecord

## Summary

This refactoring replaces multiple FIFO and Moving Average specific getters in `LazySecurityPerformanceRecord` with a parameterized API based on `CostMethod`, `CostBasis`, and `CostRequest`.

The goal is to reduce duplicated logic and prepare the snapshot model for additional cost calculation methods such as LIFO without introducing more specialized getters and duplicated fields.

### Introduced

- `CostBasis`
  - `GROSS`
  - `NET`

- Parameterized access methods:
  - `getCost(CostRequest request)`
  - `getCostPerSharesHeld(CostMethod method)`
  - `getGrossCostPerSharesHeld(CostMethod method)`
  - `getCapitalGainsOnHoldings(CostMethod method)`
  - `getCapitalGainsOnHoldingsPercent(CostMethod method)`
  - `getRealizedCapitalGains(CostMethod method)`
  - `getUnrealizedCapitalGains(CostMethod method)`
  - `getTotalRateOfReturnDiv(CostMethod method)`

### Removed

Several specialized getters were removed, including:

- `getFifoCost()`
- `getMovingAverageCost()`
- `getMovingAverageCostWithoutTaxesAndFees()`
- `getCapitalGainsOnHoldingsMovingAverage()`
- `getCapitalGainsOnHoldingsMovingAveragePercent()`
- `getFifoCostPerSharesHeld()`
- `getMovingAverageCostPerSharesHeld()`
- `getGrossFifoCostPerSharesHeld()`
- `getGrossMovingAverageCostPerSharesHeld()`
- `getRealizedCapitalGainsMovingAvg()`
- `getUnrealizedCapitalGainsMovingAvg()`
- `getTotalRateOfReturnDivMovingAverage()`

## Additional Changes

- Migrated all affected UI, online integration, trade, and test usages to the new parameterized API
- Removed duplicated Moving Average NET cost handling
- Moved `CostBasis` into the model package beside `CostMethod`

## Motivation

The previous implementation duplicated logic and lazy fields for:
- FIFO
- Moving Average
- NET/GROSS variants
- capital gains calculations
- cost-per-share calculations

This refactoring consolidates the access model and creates a more extensible structure for future cost calculation methods.

| Alt                                               | Neu                                                                    |       CostMethod |        CostBasis | Bemerkung                                           |
[...]

PR #5693, opened 2026-05-17
Feature/refactor lazy record

buchen:
@Morpheus1w3 here the additional refactoring we discussed in #5677.

Can you have a look?


Morpheus1w3:
Hi @buchen,

looks good for me.

After the changes are merged, I could work on refactoring SecuritzlyPerformanceRecord, if you agree.

buchen:
> After the changes are merged, I could work on refactoring SecuritzlyPerformanceRecord, if you agree.

@Morpheus1w3  ... to do achieve what?

Morpheus1w3:
@buchen 
I've understood from LazyRecord refactoring, SecuritzlyPerformanceRecord should might be discontinued and disposed, to get just one interface via Lazy rather than two.

buchen:
I created #5694 to delete the eager record.
I just realized that maybe we also do not need BaseSecurityPerformanceRecord anymore.
But I do not have time now. In case you spot more issues, please let me know

I also added an additional test to "lock in" some of the tests. I believe they should be static (the reporting period is fixed) and should then only change if we change any algorithms.

PR #5694, opened 2026-05-17
Delete eager SecurityPerformanceRecord

buchen:
The eager SecurityPerformanceRecord / SecurityPerformanceSnapshot were kept only so existing tests would not need updating when the lazy variant was introduced.


buchen:
@Morpheus1w3

Morpheus1w3:
I don’t see further action items from my side at the moment. The direction looks consistent to me: the eager record has been removed, lazy calculation remains the implementation path, and the tests now pin the expected values where the comparator oracle disappeared.

buchen:
@Morpheus1w3 I understand all of this was preparing additional cost methods besides FIFO and moving average. What other method is on your mind and for what purpose?

Morpheus1w3:
@buchen 
Basically I don't have any fuether idea nor need so far.

The intention for the [Refactor lazy security performance cost API to parameterized](#5677) based on while I've recognised issue [[RFC] New type of Cost / Purchase Price calculation - "break even"](#4828), and while I have to much free time actually.

Nevertheless, I try to fix just the key figures "Div%/Year" followed by "Div%". See PR [Fix annual dividend yield aggregation](#5699).

https://github.com/Morpheus1w3/portfolio/blob/14e8ff6bf7f3675e2859d3a11bb96ff7b228d8a3/name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/LazySecurityPerformanceRecord.java#L151-L152

-----

While working on the dividend yield calculations, I noticed that many calculations internally still relied on rather fixed cost access patterns:
> getFifoCost() getMovingAverageCost() etc.

For the annualized dividend yield especially, I needed a cleaner and more explicit way to retrieve:
* FIFO vs moving average
* net vs gross
* and potentially future variations without multiplying APIs again

The main goal therefore was less "we need LIFO now", but more:
* reduce duplicated APIs
* make the lazy calculation model more parameterized
* and avoid additional hardcoded getter growth

mierin12:
If I may add my two cents : I find the refactor idea of using more CostMethod and TaxAndFees as parameters is helpful. LIFO may not be not a super priority (?)  but at least it will make it easier when ready to implement. I think LIFO might be relevant for US users (?)

(LIFO topic will also be UI topic, already there were discussions if the fifo/moving avg options in the menu were too complexe, if a single switch should be used to parameterized the columns etc. )

In short term, I think the higher use of TaxAndFees is great, I find it more readable. Some functions have 'gross' in their name, some don't. So does this mean when there is no 'gross' in the functions name it is always net ? Maybe not always, so TaxAndFees as parameter makes it clearer imo. A feature I would find interesting would be the Calculation with TaxAndFees.Included. maybe not to complex and quite natural if TaxAndFees as paremeter is more used.
And maybe as a end goal if TaxAndFees is more and more used in the code it will help deal with Taxonomy' performance with TaxAndFees option too.

To use CostMethod more globally will also makes it easier and natural to reach 1-to-1 functionality between both cost methods, there are still some différences : 
- Statement of Assets ' Profit Loss is FIFO only https://github.com/portfolio-performance/portfolio/pull/4796
- Payment Closed Trades are FIFO only https://github.com/portfolio-performance/portfolio/pull/5451
- the discussion about moving average trade over différent account https://github.com/portfolio-performance/portfolio/pull/5351

Morpheus1w3:
I agree. My current practical need is still the dividend yield work, not a concrete LIFO implementation. But the refactoring also makes existing FIFO / moving average differences easier to see and reduce. The examples you listed are a good explanation of why the parameterized API is useful even without adding a new cost method immediately.

PR #5696, opened 2026-05-17
simplify securities chart 

mierin12:
From https://github.com/portfolio-performance/portfolio/pull/5677,  getFifoCostPerSharesHeld and getMovingAverageCostPerSharesHeld were replaced by a single function with CostMethod parameter input.

There is no longer need for a switch which was calling the different functions


mierin12:
(done by https://github.com/portfolio-performance/portfolio/pull/5698)

PR #5699, opened 2026-05-18
Fix annual dividend yield aggregation

Morpheus1w3:
The previous implementation aggregated annual dividend yield using simplified cost-base handling.

This caused distorted values when positions changed over time, for example with:

* partial sales,
* repurchases,
* multiple dividend payments on the same date,
* and positions that already existed before the reporting period.

The calculation now evaluates dividend payments against the cost basis at the payment date and aggregates the resulting yields per year.

For dividend yield only, the cost basis is built from historical transactions before the reporting period. This avoids using the period-start valuation as dividend yield cost basis, while keeping the existing cost and capital-gains calculations unchanged.

image
image


name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/SecurityPerformanceSnapshotBuilder.java
```diff
@@ -135,29 +139,41 @@ private void extractSecurityRelatedAccountTransactions(Account account, Map<Secu
...
+                }
+                catch (ReflectiveOperationException e)
+                {
+                    throw new IllegalArgumentException(e);
+                }
+            });
+
+            CalculationLineItem item = CalculationLineItem.of(portfolio, t);
+
+            if (t.getDateTime().toLocalDate().isBefore(interval.getStart()))
```
buchen, line 168: The interval is start-exclusive. Transactions on the first day would not be picked up. I typically say:

```java
if (!t.getDateTime().toLocalDate().isAfter(interval.getStart()))
```

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/LazySecurityPerformanceRecord.java
```diff
@@ -164,9 +167,7 @@ private Money getCostMoney(CostMethod costMethod, TaxesAndFees taxesAndFees)
     }
 
     private final LazyValue<DividendCalculationResult> dividendCalculation = new LazyValue<>(() -> {
-        // ensure cost calculation is done (and has calculated
-        // moving averages)
-        costCalculation.get();
+        dividendCostCalculation.get();
```
buchen, line 170: The existing (my) code has a side effect. It is writing back into the PaymentItem as a shared state. Now we have two CostCalculation doing this. I guess it still works because a) the dividendCostCalulation is run only here and b) the values are copied into the CostCaclulation. But that makes it fragile. I think it could make sense to make it explicit (for example the costCaculation emitting a map instrument->cost)

buchen (commented):
First, I think the change should be consistent. There are more methods calculating the dividend performance - some are using now the update cost basis, some are not.

```java
    public Double getRateOfReturnPerYear()
    {
        return dividendCalculation.get().rateOfReturnPerYear();
    }

    public Double getTotalRateOfReturnDiv(CostMethod costMethod)
    {
        var costs = costCalculation.get();
        var cost = getCostMoney(costMethod, TaxesAndFees.INCLUDED);

        return costs.sharesHeld() > 0 ? (double) dividendCalculation.get().sum().getAmount() / (double) cost.getAmount()
                        : 0;
    }
```

Second, I understand that using for the cost basis the reporting period and therefore the valuation at start can be confusing. That is also true for purchase price and purchase value information in the security performance view. The idea was always to make instrument performance comparable even if the holding period is quite different.

Personally, I am not a fan of a personal dividend return which is taking the historic
purchase prices into account. As an investor, I always have to make the investment decision to keep an instrument or purchase a new instrument. For this decision, it is hard if I'm comparing numbers between an instrument which is held for 10 years and one year.
If I'm not mistaken, this change now proposes to use the historical purchase price. I haven't made up my mind - please argue here

And last but not least, I have to make up my mind how to communicate this incompatible change to the many users.

Issue #5547, opened 2026-03-04
Improvement: AI/CLI access to PP data


georgemac-labs:
I would like to give AI models a way to interrogate my Portfolio Performance data. Here are my thoughts on the subject.

I considered the REST idea (https://github.com/portfolio-performance/portfolio/pull/2478) but to me it seems unnecessarily complicated. I don’t need network access.

IMO the simplest option would be just adding a CLI mode, by intercepting command-line arguments early in the app lifecycle. If CLI flags are detected, the app loads the portfolio file, runs the requested operation using existing core APIs (ClientFactory.load, ClientSnapshot, performance calculations, etc.), prints results to stdout, and exits — never opening the GUI. If no flags are present, it starts normally as today.

This would give an AI agent (or a script) access to PP's full calculation engine — IRR, TWR, snapshots, currency conversion, taxonomy grouping — via simple shell commands like `PortfolioPerformance export securities file.portfolio`. The change is modest: arg parsing, a handful of command implementations that call existing APIs, and an early exit path.

Startup would not be particularly fast due to initialisation overhead, but I can live with that.

My MVP would be something like this:

- `export accounts --format csv <file.portfolio>` → CSVExporter.exportAccountTransactions()
- `export portfolios --format csv  <file.portfolio>` → CSVExporter.exportPortfolioTransactions()
- `export securities --format csv  <file.portfolio>` → CSVExporter.exportSecurityMasterData()
- `export prices --format csv  <file.portfolio>` → CSVExporter.exportMergedSecurityPrices()
- `export taxonomies <file.portfolio>` → TaxonomyJSONExporter.AllTaxonomies

Thinking ahead, my idea is to define JSON as the default format, but the current code can’t do that for the first four, so in the MVP you have to specify `--format csv`, otherwise you get an error. Alternatively we could make CSV the default, or make the `--format` switch mandatory.

Happy to do the implementation, but I would like to have maintainer feedback before I invest time in this.

buchen:
> I would like to give AI models a way to interrogate my Portfolio Performance data.

Haha, I was thinking about this also. With a co-worker, we were discussing what Claude Code could do with access to the portfolio. I was thinking one could build an MCP server and give access this way.

[...]

PR #5700, opened 2026-05-18
Headless Import Tool

matthid:
This adds a headless PDF import utility that can import documents into a Portfolio Performance XML file from the command line.                                                                                                                                  
                                                                                                                                                                                                                                                                  
Motivation: Portfolio Performance already has strong PDF importers, but reconstructing a long historical portfolio from a large archive of broker documents is difficult to do reproducibly through the UI alone. A headless importer makes it possible to run  
the existing import framework from scripts, keep intermediate XML snapshots, compare results after each batch, and repeat the process after importer fixes.
                                                                                                                                                                                                                                                                  
This is also useful for assisted workflows: in my case, Codex helped operate the import loop, inspect failed imports, compare generated transactions against the PDFs, and suggest or apply explicit manual corrections where the documents did not contain     
enough information. The tool itself is not AI-specific; it exposes the existing importer pipeline in a command-line form so humans, scripts, or AI-assisted agents can drive the same repeatable process.
                                                                                                                                                                                                                                                                  
For transparency: the real-life validation import processed roughly 2,000 real bank/broker documents and also involved manual post-processing of the generated XML, outside these patches. In particular, I added ticker symbols resolved from ISINs, inferred  
cash deposits/transfers to avoid negative cash balances on buys, and a few transfer/account corrections because the reference cash account was also used as a normal bank account. 
                                                                                                                                                                                                                                                                  
Full disclosure: the implementation and validation workflow were prepared with help from Codex. The combined branch passed `mvn -f portfolio-app/pom.xml clean verify`.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/HeadlessPDFImportTool.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/HeadlessPDFImportTool.java
@@ -1,0 +1,1212 @@
+package name.abuchen.portfolio.datatransfer.pdf;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.eclipse.core.runtime.NullProgressMonitor;
+
+import name.abuchen.portfolio.Messages;
+import name.abuchen.portfolio.datatransfer.Extractor;
+import name.abuchen.portfolio.datatransfer.ImportAction;
+import name.abuchen.portfolio.datatransfer.ImportAction.Status.Code;
+import name.abuchen.portfolio.datatransfer.actions.CheckCurrenciesAction;
+import name.abuchen.portfolio.datatransfer.actions.CheckForexGrossValueAction;
+import name.abuchen.portfolio.datatransfer.actions.CheckSecurityRelatedValuesAction;
+import name.abuchen.portfolio.datatransfer.actions.CheckTransactionDateAction;
+import name.abuchen.portfolio.datatransfer.actions.CheckValidTypesAction;
+import name.abuchen.portfolio.datatransfer.actions.DetectDuplicatesAction;
+import name.abuchen.portfolio.datatransfer.actions.InsertAction;
+import name.abuchen.portfolio.model.Account;
+import name.abuchen.portfolio.model.AccountTransaction;
+import name.abuchen.portfolio.model.BuySellEntry;
+import name.abuchen.portfolio.model.Client;
+import name.abuchen.portfolio.model.ClientFactory;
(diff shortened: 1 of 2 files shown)


Morpheus1w3:
#5547

buchen:
How would one start this tool? It probably runs in the IDE, but as a user, how to I construct the class path?

I understand the need for a CLI (and hopefully I have soon more time to invest in building one), but I would prefer to use the Eclispe mechanism to avoid maintaining, support, and troubleshoot multiple class path setups.

One more question: I'm wondering, do you want to work against the file and thereby you cannot have the desktop application running? Or would you rather want to have a command line / API against the desktop application which has the file open and you can inspect/review?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/HeadlessPDFImportTool.java
```diff
@@ -0,0 +1,1212 @@
...
+        return ClientFactory.load(options.clientFile, null, new NullProgressMonitor());
+    }
+
+    public static final class Options
+    {
+        private final List<File> inputFiles = new ArrayList<>();
+        private File clientFile;
+        private File outputFile;
+        private File reportFile;
+        private String baseCurrency = CurrencyUnit.EUR;
```
buchen: If we default the base currency, then to the base currency defined in the file. Everything else will be confusing for users that use other currencies than EUR.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/HeadlessPDFImportTool.java
```diff
@@ -0,0 +1,1212 @@
...
+    {
+        PortfolioTransaction transaction = entry.getPortfolioTransaction();
+        Portfolio portfolio = context.getPortfolio();
+
+        return portfolio.getTransactions().stream()
+                        .anyMatch(existing -> existing.getType() == transaction.getType()
+                                        && Objects.equals(existing.getDateTime(), transaction.getDateTime())
+                                        && Objects.equals(existing.getCurrencyCode(), transaction.getCurrencyCode())
+                                        && existing.getAmount() == transaction.getAmount()
+                                        && sameSecurity(existing.getSecurity(), transaction.getSecurity()));
```
buchen, line 257: Why not check for shares?

matthid:
I checked the import result again. The main issue I
  found was not a share-position problem, but value/      cost-basis handling for manually represented corporate
  actions and depot transfers. Some corporate-action
  deliveries were initially zero-valued, so later
  transfers could also carry zero values and distort
  performance/capital-gains views.

  I fixed my private generated XML in a separate cleanup
  snapshot by carrying cost basis through the affected    Shell, Kirkland/Agnico, Brookfield, and DouYu           corporate actions and then rerunning the zero-transfer
  cleanup. A few zero-value rows remain intentionally     zero, e.g. stock splits and cleanup-only rows, and
  those are documented.

 For the MR itself, I fixed the two code review points:  - default base currency now comes from the loaded file  unless explicitly overridden                            - exact duplicate detection now compares shares as well               
                                   
  I agree the current headless tool is more of an         importer development/QA harness than a finished user-   facing CLI. My use case was offline batch processing    against a closed XML file. Long-term, an Eclipse/       product-based command or an API against the running
  desktop app would be better for interactive review and
  manual corrections.


On a personal note: If you feel like this is not the right direction that is fine. Like written initially "it has done its job"". I sent the pr only in case someone finds it useful. Just close it, no hard feeling :). The same is practically true for the other prs we are no longer customers on those banks...

PR #5702, opened 2026-05-18
ING-DiBa Delivery Transfers

matthid:
This extends the ING-DiBa PDF extractor with support for delivery transfer documents: `Übertrag Ausgang` and `Wertpapier Eingang`. 

It also adds sanitized regression fixtures based on real documents and fixes a name handling issue so the second security-name line is not duplicated for delivery transfers. 

The importer was validated through the headless import workflow from `codex/headless-pdf-import-tool`, which made it possible to repeatedly import, inspect, fix, and re-run batches against a large historical document archive. Transfers were especially     
important because the portfolio history includes multiple depot transfers. In the broader validation workflow, this was part of reconstructing a historical Portfolio Performance depot from roughly 2,000 real bank/broker documents.

For transparency: the real-life validation import also involved manual post-processing of the generated XML, outside these patches. In particular, I added ticker symbols resolved from ISINs, inferred cash deposits/transfers to avoid negative cash balances 
on buys, and a few transfer/account corrections because the reference cash account was also used as a normal bank account.                                                                                                                                        

Full disclosure: the implementation, fixture preparation, and validation workflow were prepared with help from Codex. Tests added/updated in this PR are in `INGDiBaPDFExtractorTest`. The combined branch passed `mvn -f portfolio-app/pom.xml clean verify`.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/ingdiba/WertpapierEingang01.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/ingdiba/WertpapierEingang01.txt
@@ -1,0 +1,13 @@
+ING-DiBa AG
+Depotinhaber: Max Mustermann
+Direkt-Depot Nr.: 1234567890
+Datum: 15.10.2025
+Wertpapier Eingang
+Sehr geehrte Kundin, sehr geehrter Kunde,
+nachstehende Wertpapiere haben wir Ihrem Depot gutgeschrieben:
+Nominale / Stück Wertpapier-Informationen Schlusstag Auftragsnummer
+0,50 Stück Brookfield Corp. 10.10.2025 0025830418
+Registered Shares Cl.A o.N.
+ISIN (WKN): CA11271J1075 (A3D3EV)
+Wertpapierrechnung Kanada
+Die Gesellschaft hat einen Aktiensplit durchgeführt.
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
@@ -975,7 +975,7 @@ public class INGDiBaPDFExtractor extends AbstractPDFExtractor
                         .match("^(?<nameContinued>.*)$") //
                         .match("^ISIN \\(WKN\\): (?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9]) \\((?<wkn>[A-Z0-9]{6})\\)$") //
                         .assign((t, v) -> {
-                            v.put("currency", guessCurrencyCode(v.get("isin")));
+                            v.put("currency", getClient().getBaseCurrency());
 
                             t.setDateTime(asDate(v.get("date")));
                             t.setShares(asShares(v.get("shares")));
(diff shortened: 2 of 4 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/ingdiba/WertpapierEingang01.txt
```diff
@@ -0,0 +1,13 @@
...
+Datum: 15.10.2025
+Wertpapier Eingang
+Sehr geehrte Kundin, sehr geehrter Kunde,
+nachstehende Wertpapiere haben wir Ihrem Depot gutgeschrieben:
+Nominale / Stück Wertpapier-Informationen Schlusstag Auftragsnummer
+0,50 Stück Brookfield Corp. 10.10.2025 0025830418
+Registered Shares Cl.A o.N.
+ISIN (WKN): CA11271J1075 (A3D3EV)
+Wertpapierrechnung Kanada
+Die Gesellschaft hat einen Aktiensplit durchgeführt.
```
buchen, line 13: It say explicitly that this is a stock split. But this cannot be imported as an inbound delivery. The performance calculation would not be correct. The stock split is not adding external capital to the portfolio.

Unfortunately, PP does not support the import of the corporate action "stock split" (yet). Therefore the correct way is to mark this as failure item with a hint to the user what to do (create stock split manually). See other importer.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
```diff
@@ -939,6 +940,68 @@ private void addNonImportableTransaction()
...
+            return "EUR";
+
+        return switch (isin.substring(0, 2))
+        {
+            case "CA" -> "CAD";
+            case "GB" -> "GBP";
+            case "US" -> "USD";
+            default -> "EUR";
+        };
+    }
```
buchen: That is a tough one. There is no currency in the document (what a pity). A US or CA security held at ING can be modeled in EUR, USD, CAD, etc.; GB ISINs can also be GBX/GBP/USD/EUR depending on listing.

Right now the language support is German and Spanish. Defaulting to EUR does not work for Switzerland.

I am open: one could also default to the baseCurrency. What is your opinion?

buchen (changes requested):
The stock split must be handled differently - see comment.

The UebertragAusgang: is it a performance neutral outbound delivery? Or is this part of another corporate action? It also looks like a reduced example. It would be super helpful to have the full document. In the future, when the document structure is changed by the bank, it enables us to maintain the importer.

matthid:
Thanks, agreed.               
                                                                                  
I changed the stock split handling: the “Wertpapier Eingang” document is no longer treated as a normal inbound delivery when it contains the stock split text. It now creates a failed import item with MsgErrorTransactionSplitUnsupported, so the user gets   the hint to create the stock split manually.


For the missing currency in these ING delivery documents, I removed the ISIN-country heuristic. I
  agree that CA/US/GB from the ISIN is not reliable       because the security can be modeled in different
  currencies. The importer now defaults to the client base currency for these documents.                                                                              For “Übertrag Ausgang”: in my source material this is   an outgoing depot transfer to another depot, not a      corporate action, so I modeled it as a performance-neutral outbound delivery. However, your point about    the reduced fixture is valid. The original document is  a larger multi-position transfer document, and the      current test case only captures the logical row needed  for this parser. I can either replace it with a fuller
  anonymized fixture that preserves the document          structure, or remove this part from the PR until it is
  covered by such a fixture.

PR #5710, opened 2026-05-23
 Improve PDF-Importer use convenience constructors, normalize wrap blocks and ExtractorMatchers

Nirus2000:
Replace the verbose three-line subject lambda with the AccountTransaction(Type) convenience constructor introduced in #5614 across all PDF importers.

   **Additionally:**
   - Add BuySellEntry(PortfolioTransaction.Type) and PortfolioTransaction(Type) convenience constructors and apply the same single-line subject lambda pattern to all BuySellEntry and PortfolioTransaction subject lambdas
   - Convert the remaining three verbose AccountTransferEntry subject lambdas to method references (AccountTransferEntry::new)
   - Replace setCurrencyCode(CurrencyUnit.X) with setCurrencyCode(asCurrencyCode("X"))
   - Add missing t.getCurrencyCode() != null guard to bare t.getAmount() == 0 checks
   - Normalize all simple wrap blocks to a consistent if/return SkippedItem format, removing intermediate item variables and unused ctx parameters
   - Add final modifier to all var type = new DocumentType(...) declarations

   **Improvement Arkéa Direct Bank and ExtractorMatchers**
   - Extend `hasExDate(String)` to accept `null`, allowing dividend tests to explicitly assert that no ex-date is present in a document
   - Add `hasExDate(null)` assertions to all three dividend tests to document the absence of an ex-date in Arkéa Direct Bank documents

   **Normalize currency handling in DocumentContext**
   - Extend `hasExDate(String)` to accept `null`, allowing dividend tests to explicitly assert that no ex-date is present in a document
   - Add `hasExDate(null)` assertions to all three dividend tests to document the absence of an ex-date in Arkéa Direct Bank documents
   
   **Normalize currency handling in DocumentContext**
   - Fix missing asCurrencyCode() in ctx.put("currency", ...) in BundesschatzPDFExtractor (two document types stored the raw captured string instead of the normalized currency code)
   - Remove redundant asCurrencyCode() wrapping in 15 extractors where currency is read back via .documentContext("currency") — the value stored in the context is already normalized, so the second call was unnecessary
   (affected: ABNAMROGroup, AkfBank, AudiBank, AustrianAnadi, BarclaysBankIreland, BawagAG, BoursesDirect, Bundesschatz, Crowdestor, DKB, INGDiBa, RaiffeisenBankgruppe, SantanderConsumerBank, Tradegate, VolkswagenBank)
   
   **Normalize currency handling — fixed-currency extractors and raw section reads**
   - Introduce `private static final String XXX = "XXX"` constants and replace all `CurrencyUnit.XXX` object references and unwrapped bare string literals with `asCurrencyCode(XXX)` in fixed-currency extractors (affected:
   LimeTradingCorp, FidelityInternational, Computershare, AvivaPLC, ScorePriorityInc, Simpel, Sunrise, SutorBankGmbH, KFintech)
   - Wrap `v.put("currency", stripBlanks(...))` with `asCurrencyCode()` where currency is captured from raw PDF text (affected: Comdirect 4×, Commerzbank 1×, LibertyVorsorgeAG 7×)
   - Wrap `v.put("currency", v.get("currency1/2"))` with `asCurrencyCode()` for currency1/currency2 relay patterns (affected: Consorsbank 12×, DZBankGruppe 6×, DKB 6×, INGDiBa 6×, Postbank 6×)
   - Wrap `Money.of(v.get("currency/feeCurrency"), ...)` with `asCurrencyCode()` in BSDEXPDFExtractor where both currency fields are extracted locally from raw PDF text (8× across buy/sell sections)
   
   **Improve asCurrencyCode() in AbstractPDFExtractor and fix TextUtil.trim()**
   - Replace `currency.trim()` (Java's String.trim, strips only ASCII ≤ 0x20) with `TextUtil.trim()` (already statically imported) which additionally handles Unicode spaces such as U+00A0 (NO-BREAK SPACE) and U+FEFF
   (Zero-Width BOM)
   - Fix asymmetric bug in `TextUtil.trim()`: the trailing whitespace loop used `Character.isWhitespace()` while the leading loop used the more comprehensive `TextUtil.isWhitespace()` — U+00A0, U+202F and U+FEFF were
   therefore trimmed leading but not trailing; both loops now use `TextUtil.isWhitespace()`
   - As a side effect, `MessagesTest` now detects a stray U+202F (NARROW NO-BREAK SPACE) trailing in the Spanish translation key `AttributeSettings_LimitPrice_ColorSettings` — removed
   - Add `TextUtilTest.testTrim()` covering regular whitespace, U+00A0 and U+FEFF trimming on both sides
   - Add `AbstractPDFExtractorTest` covering: standard ISO codes, U+00A0 trimming, regular whitespace trimming, null input, and unknown-code fallback to base currency

### DIFF
--- a/.claude/rules/pdfextractors.md
+++ b/.claude/rules/pdfextractors.md
@@ -20,6 +20,84 @@ PDF importers extract transactions from bank and broker PDF statements.
   - Consistency in coding style between all extractors is more important than nice code.
   - Add comment-blocks with `@formatter:off` and `@formatter:on` before each section-block showing the format that is handled there.
 
+## Conventions
+
+**DocumentType declaration** — always `final`:
+```java
+final var type = new DocumentType("...");
+```
+
+**subject() lambda** — always use the single-argument convenience constructor or method reference, never the verbose 3-line form:
+```java
+// correct
+.subject(() -> new AccountTransaction(AccountTransaction.Type.DIVIDENDS))
+.subject(() -> new BuySellEntry(PortfolioTransaction.Type.BUY))
+.subject(() -> new PortfolioTransaction(PortfolioTransaction.Type.DELIVERY_INBOUND))
+.subject(AccountTransferEntry::new)   // no Type parameter — use method reference
+
+// wrong
+.subject(() -> {
+    var t = new AccountTransaction();
+    t.setType(AccountTransaction.Type.DIVIDENDS);
+    return t;
+})
+```
+
+**setCurrencyCode** — always use `asCurrencyCode()`, never `CurrencyUnit` constants:
+```java
+t.setCurrencyCode(asCurrencyCode("EUR"));   // correct
+t.setCurrencyCode(CurrencyUnit.EUR);         // wrong
+```
+
+**Fixed-currency extractors** — when an extractor always deals with one currency (e.g. a USD-only broker), declare a private String constant and use it with `asCurrencyCode()`. Keep `asCurrencyCode()` — it has normalization side-effects even for known codes:
+```java
+private static final String USD = "USD";
+
+v.put("currency", asCurrencyCode(USD));          // correct
+v.put("currency", CurrencyUnit.USD);              // wrong — CurrencyUnit object, not String
+v.put("currency", asCurrencyCode("USD"));         // works but prefer named constant
+```
+
+**DocumentContext currency** — when storing currency into the document context, always normalize with `asCurrencyCode()`. When the value is later read back via `.documentContext("currency")`, it is already normalized — do NOT wrap it with `asCurrencyCode()` again:
+```java
+// storing — always normalize
+.assign((ctx, v) -> ctx.put("currency", asCurrencyCode(v.get("currency"))))  // correct
+.assign((ctx, v) -> ctx.put("currency", v.get("currency")))                   // wrong
+
+// reading via documentContext — already normalized, no wrapping needed
+.section("tax").optional()
+.documentContext("currency")
+.assign((t, v) -> {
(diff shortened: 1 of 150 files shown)


buchen:
Thanks @Nirus2000. The change looks good to me.

Let me ask you about your opinion on one thing I noticed:

The DekaBank importer changes all failure items to skipped items. In the draft pull request #5562 in commit https://github.com/portfolio-performance/portfolio/pull/5562/changes/a77c56e8c51a7e634f98f52d9267e393d20240b6 I was proposing to keep the some as failure items. There is a merger of companies (German "Fusion"). PP does not support that as corporate action. I was thinking to mark it as failure because the user has to do something here manually. It applied only to items where we were able to parse shares (otherwise the items can really be skipped and the user does not need to bother about them).

The code was something like
```java
            .wrap((t, ctx) -> {
                if (t.getPortfolioTransaction().getCurrencyCode() != null && t.getPortfolioTransaction().getAmount() == 0)
                {
                    if (t.getPortfolioTransaction().getShares() != 0)
                        ctx.markAsFailure(Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
                    else
                        ctx.skipTransaction(Messages.MsgErrorTransactionTypeNotSupportedOrRequired);
                }
```

What do you think?

buchen:
From the update, I see that you have updated the DekaBank importer. I think that makes sense.

There are a couple more importer (I think Kauf28 of Consorsbank could also be a candidate) but I propose to do this in separate and smaller pull requests. This is already a borderline-sized pull request.

Issue #5714, opened 2026-05-25
Setup blocked at "Run tests": missing org.eclipse.pde.junit.runtime


Konilo:
Hello!

I'm trying to set myself up for contributing to Portfolio Performance by following [CONTRIBUTING.md](https://github.com/portfolio-performance/portfolio/blob/master/CONTRIBUTING.md). I got blocked at step 6 ("Run tests") and want to check with you what I might be missing before opening a doc PR.

## Setup steps I followed

* Installed Eclipse IDE for RCP and RAP Developers (2026-03)
* Installed Zulu JDK 21 and added it to the Installed JREs in Eclipse
* Installed the marketplace plugins (Infinitest, ResourceBundle Editor, Checkstyle, SonarQube, LcDsl)
* Installed Eclipse e4 Tools via Install New Software (M2E PDE Integration was already shipped with the RCP/RAP package, so I skipped it)
* Configured save actions, content assist, and favorites
* Forked, cloned, imported all projects
* Opened `portfolio-target-definition.target`, clicked "Set as Active Target Platform", and let it fully resolve
* Tried to run `JUnit Plug-in Tests` -> `PortfolioPerformance_Tests`

## What I got
```
Required plug-in 'org.eclipse.pde.junit.runtime' could not be found.
```

Running "Validate Plug-ins" on the launch config reports "No problems were detected": the missing bundle is implicit to the PDE JUnit launcher, not part of the bundles selected by LcDsl.

In `portfolio-target-definition.target`, I don't see PDE bundles included. The `download.eclipse.org/eclipse/updates/4.38/` location pulls in `org.eclipse.platform.feature.group` and `org.eclipse.e4.rcp.feature.group`, but no `org.eclipse.pde.*` feature.

I also noticed CI uses `mvn verify` (Tycho), which has its own runtime. So the Eclipse-PDE-launcher path may not be the one maintainers exercise day-to-day.

## What worked for me

In the Target Editor, I added an `Installation` location pointing at the running Eclipse IDE directory (`C:\Program Files\eclipse\rcp-2026-03\eclipse`). After re-activating the target, the tests run cleanly (4359/4359 passing).

## Question

Did I miss a step, or is this an actual gap in the setup? If it's a gap, I'd be happy to open a PR with whichever fix you'd prefer (doc update or target file update).

## Environment

* OS: Windows 11
* Eclipse: IDE for RCP and RAP Developers 2026-03
* JDK: Zulu 21
[...]

PR #5717, opened 2026-05-25
Move plugin versions into properties section

### LINKED ISSUE #3244: refactoring: moving versions into properties section
**current**
pom.xml of portfolio-app contains a lot of plugins in the pluginManagement section, but only one plugin has a version placeholder.
This makes it harder to find out if there is something outdated. e.g. there is an update for jacoco (0.8.7 -> 0.8.8)

**goal**
put the versions with a placeholder into the properties section

Konilo:
Hello!

This refactors `portfolio-app/pom.xml` so the remaining hardcoded plugin versions are declared in `<properties>` as `${<artifactId>.version}` placeholders, matching the existing `${tycho-version}` pattern. The goal, as discussed in the issue, is to make outdated plugins easier to spot via `mvn versions:display-plugin-updates`.

## Changes

* Lifted 11 plugin versions into the `<properties>` block: `maven-enforcer-plugin`, `maven-resources-plugin`, `maven-dependency-plugin`, `maven-jarsigner-plugin`, `maven-gpg-plugin`, `maven-checkstyle-plugin`, `sonar-maven-plugin`, `fix-info-plist-maven-plugin`, `exec-maven-plugin`, `jacoco-maven-plugin`, `protoc-jar-maven-plugin`.
* Used the dotted Maven convention `${<artifactId>.version}`, e.g. `${maven-resources-plugin.version}`.
* `${tycho-version}` is left as-is — already a property, shared by 8 Tycho plugins, so a short name made sense there. The new properties are 1:1 with their plugins, hence the more explicit `<artifactId>.version` form.

## Not in scope

* The checkstyle library dependency version inside `maven-checkstyle-plugin` (`13.4.2`) is intentionally left hardcoded — it is a plugin dependency, not a plugin version, and `display-plugin-updates` does not check it.

## Verification

`mvn -f portfolio-app/pom.xml clean verify -Plocal-dev` → BUILD SUCCESS, all 13 modules, 0 test failures.

Closes #3244

### DIFF
--- a/portfolio-app/pom.xml
+++ b/portfolio-app/pom.xml
@@ -1,6 +1,5 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 
 	<groupId>name.abuchen.portfolio</groupId>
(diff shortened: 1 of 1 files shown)


Konilo:
@Nirus2000 

This PR is unrelated to https://github.com/portfolio-performance/portfolio/issues/5714! :) I'll make a doc PR for #5714 later, though!

buchen:
@Konilo I believe so far dependabot was updating the dependencies. Will that work with the extracted versions (thinking about it: I guess because it used to work for Tycho)? And: what does the ```display-plugin-updates``` require this?

Konilo:
@buchen

**Dependabot:** confirmed, and it already works in this repo. Dependabot bumps both styles:
- property-based `${tycho-version}` e.g., #5382, where it edits the `<tycho-version>` declaration directly
- hardcoded plugin versions e.g., #5727 (sonar), #5703 (enforcer), where it edits the inline `<version>`

One minor difference: the Dependabot PR titles for these plugins will read "Bump maven-resources-plugin.version from X to Y" instead of the full artifact coordinates, like the tycho PRs do now.

**`display-plugin-updates`:** my PR description overclaimed a bit! `display-plugin-updates` effectively reports outdated plugins regardless of whether versions are hardcoded or stored as properties. So this refactor doesn't change anything on that front. The actual benefit is that, instead of browsing through `pluginManagement`, you could find all `<artifactId>.version`s at the top of the file. It also brings these plugins under the property-only goals (`versions:display-property-updates`, `versions:set-property`), which would be interesting if you ever want to script a bump (separate from the `display-plugin-updates` you run today).

To be transparent, I took on this [issue](https://github.com/portfolio-performance/portfolio/issues/3244) because it felt approachable for me as a new comer and was marked as `help_wanted`. I'll close my PR if this is not interesting in your eyes anymore, no problem :)

buchen:
Ah, thanks. Now I understand where you're getting from. Yes, I've merged the change. Nowadays, I'm relying on the dependabot updates (could be that the issue is from before that time). But the change improves readability and I have merged it.

PR #5753, opened 2026-05-30
Fix wrong excess return values in performance heatmap

### LINKED ISSUE #5752: Wrong Heatmap Values for Monthly Excess Returns (Baseline)
**Describe the bug**
Observation:
On monthly return heatmap, excess returns are calculated incorrectly against a benchmark when:

- The first transaction in the selected “Data Series“ for the widget is not the oldest transaction in the complete portfolio

**To Reproduce**

1. Create an empty portfolio
1. Create a securities account S1 referencing a deposit account D1
1. Create a cash account C1
1. Add the following transactions to C1:
    - 12 deposits on: 2024-01-01, 2024-02-01, …, 2024-12-01
1. Add the following transactions to D1:
    - 6 deposits on: 2024-06-15, 2024-07-15, …, 2024-12-15
1. Add the following transactions to S1:
    - 3 buy any security XYZ on: 2024-07-16, 2024-08-16, 2024-09-16,
1. Open a performance dashboard
1. Set the time interval for the dashboard to “3 years“
1. Add a monthly return heatmap widget
    - Select S1 account as “Data Series“ for the widget
    - Select security XYZ (or any other security) as “Excess Return → Benchmark“
1. Observe the wrong behavior

**Expected behavior**

- All month before the first buy on 2024-07-16 shall show ‘0’ as S1 has no securities
- Any transaction in the cash account C1 shall not have any influence on the monthly heatmap widget as:
    - the selected data series for the widget is S1
    - There are no connections between S1 and C1 accounts

**Observed behavior**

- The month before the first buy have non-zero values
- Testing several scenarios led to the following conclusion:
    - The number of month with non-zero values is the smallest of the following:
        - The time interval of the dashboard or
        - The first available historical quota for the benchmark or
        - The first available transaction in the complete portfolio
- In this scenario “the first available transaction in the complete portfolio” is the observed
- Note: The values on the heatmap are wrong for all month before the first buy AND for the month in which the first buy occurs!

**Simple test**

- If you remove the account C1 and all its transactions, the values in the heatmap widget change i.e. they turn to expected values.
- This shall not happen as C1 has no connection to data series selected for the widget

**Example**
Uploaded to the forum (link below)

**Reference**
https://forum.portfolio-performance.info/t/wrong-heatmap-values-for-monthly-excess-returns-baseline/38228

 - OS: Windows 11
 - PP Version  0.83.2 (originally reported on 0.80.3)

gitpie69:
When a baseline is configured, months before the primary data series
    has any holdings showed a spurious non-zero excess return.
    This happened because the display interval spanned the full reporting
    period, so months with zero primary return were still compared against
    a non-zero benchmark return (0 - benchmarkReturn != 0).

    Fix by restricting the display interval to start from the first date
    the primary series holds actual assets, so pre-activity months are
    shown as empty cells instead of misleading excess return values.

Closes #5752
Issue: https://forum.portfolio-performance.info/t/wrong-heatmap-values-for-monthly-excess-returns-baseline/38228

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
@@ -66,7 +66,13 @@ public class PerformanceHeatmapWidget extends AbstractHeatmapWidget<Double>
 
         // build heat map model for actual interval
 
-        Interval actualInterval = performanceIndex.getActualInterval();
+        // When an excess return baseline is configured, restrict the display
+        // interval to start from the first date the primary series has actual
+        // holdings. Without this, months before the first transaction show a
+        // spurious excess return (0 - benchmarkReturn != 0).
+        Interval actualInterval = benchmark != null
+                        ? getFirstActiveInterval(performanceIndex)
+                        : performanceIndex.getActualInterval();
 
         boolean showSum = get(HeatmapOrnamentConfig.class).getValues().contains(HeatmapOrnament.SUM);
         boolean showStandardDeviation = get(HeatmapOrnamentConfig.class).getValues()
@@ -119,6 +125,20 @@ public class PerformanceHeatmapWidget extends AbstractHeatmapWidget<Double>
         }
 
         return model;
+    }
+
+    private Interval getFirstActiveInterval(PerformanceIndex index)
+    {
+        long[] totals = index.getTotals();
+        LocalDate[] dates = index.getDates();
+
+        for (int i = 0; i < totals.length; i++)
+        {
+            if (totals[i] > 0)
+                return Interval.of(i > 0 ? dates[i - 1] : dates[i], dates[dates.length - 1]);
+        }
+
+        return index.getActualInterval();
     }
 
     private double getPerformanceFor(PerformanceIndex index, YearMonth month)


gitpie69:
@Nirus2000 I rebased my branch to the latest master. Not sure if I need to do anything else.

Nirus2000:
> @Nirus2000 I rebased my branch to the latest master. Not sure if I need to do anything else.

No, you don't have to do anything. Just wait. Time is our biggest issue right now. But we'll do our best to keep working through the pull requests.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
```diff
@@ -121,6 +127,20 @@ protected HeatmapModel<Double> build()
...
+        long[] totals = index.getTotals();
+        LocalDate[] dates = index.getDates();
+
+        for (int i = 0; i < totals.length; i++)
+        {
+            if (totals[i] > 0)
+                return Interval.of(i > 0 ? dates[i - 1] : dates[i], dates[dates.length - 1]);
+        }
+
+        return index.getActualInterval();
```
buchen, line 141: It there are no holdings whatsoever, then the fallback is an actual interval of the last day only. Otherwise, the spurious returns show up again

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/heatmap/PerformanceHeatmapWidget.java
buchen, line 64: I believe the SUM row would also respect the limited actual interval - otherwise the SUM and the monthly values do not fit.

buchen (commented):
Thanks for the contribution. I have merged the changes. I did some minor refactoring related to the comments. Basically I believe the sum row also then needs to respect the limited holding interval and also the partial months should then respect the holding interval because otherwise some and partial months don't fit together. But anyway thank you very much for fixing this long-standing bug.

PR #5602, opened 2026-03-28
Adjust TradeRepublic Importer to support new format

gamue:
Closes #5552 

- Also added test-cases from forum: 
  - https://forum.portfolio-performance.info/t/pdf-import-von-trade-republic/5107/819
  - https://forum.portfolio-performance.info/t/pdf-import-from-trade-republic/24425/137

Tests are working fine locally and build is failing due some eclipse-dependency download issue. Might resolve due retriggering the build later. I tried retriggering by adding a new test-case, however still failing :(

PR #5766, opened 2026-06-02
TradeRepublic PDF import parse fix

kimmerin:
added test for new type of PDF containing a trailing page with tax-informations leading to parse error. Fixed parse error

The dump has been provided at https://forum.portfolio-performance.info/t/pdf-import-von-trade-republic/5107/837?u=kimmerin

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
@@ -76,7 +76,7 @@ import name.abuchen.portfolio.online.impl.CoinGeckoQuoteFeed;
 @SuppressWarnings("nls")
 public class TradeRepublicPDFExtractorTest
 {
-    TradeRepublicPDFExtractor extractor = new TradeRepublicPDFExtractor(new Client())
+    TradeRepublicPDFExtractor cryptExtractor = new TradeRepublicPDFExtractor(new Client())
     {
         @Override
         protected List<SecuritySearchProvider> lookupCryptoProvider()
@@ -1136,7 +1136,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf01.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf01.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1170,7 +1170,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf02.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf02.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1204,7 +1204,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf03.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf03.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1238,7 +1238,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf04.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf04.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1272,7 +1272,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf05.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf05.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1306,7 +1306,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf06.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf06.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1340,7 +1340,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf07.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf07.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1374,7 +1374,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf08.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf08.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1408,7 +1408,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf09.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoKauf09.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
@@ -1442,7 +1442,7 @@ public class TradeRepublicPDFExtractorTest
     {
         List<Exception> errors = new ArrayList<>();
 
-        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoVerkauf01.txt"), errors);
+        var results = cryptExtractor.extract(PDFInputFile.loadTestCase(getClass(), "CryptoVerkauf01.txt"), errors);
 
         assertThat(errors, empty());
         assertThat(countSecurities(results), is(1L));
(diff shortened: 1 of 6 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
```diff
@@ -1499,7 +1513,7 @@ private void addDividendTransaction()
                                         // Dividendo con l'ex-tag 12.05.2023.
                                         // Distribuzione con l'ex-tag 17.08.2023. 
                                         // Ausschüttung mit dem Ex-Tag 12.09.2019.
-                                        // Dividende mit Ex-Datum22.05.2024.
+                                        // Dividende mit Ex-Datum 22.05.2024.
```
Nirus2000, line 1508: Um, have you looked at the files “Dividende15.txt” and “Dividende24.txt”?
This change is incorrect, and so is the regex.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
```diff
@@ -76,7 +76,7 @@
 @SuppressWarnings("nls")
 public class TradeRepublicPDFExtractorTest
 {
-    TradeRepublicPDFExtractor extractor = new TradeRepublicPDFExtractor(new Client())
+    TradeRepublicPDFExtractor cryptExtractor = new TradeRepublicPDFExtractor(new Client())
```
Nirus2000, line 79: We won't start distinguishing between the variables “extractor” and “cryptExtractor” in our tests. It just doesn't make sense.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
```diff
@@ -1256,7 +1261,16 @@ private void addDividendTransaction()
...
+        var firstRelevantLine = new Block("(?i)(AUSSCH.TTUNG" //
+                        + "|(STORNIERUNG DER )?DIVIDENDE( EN ESP.CES)?" //
+                        + "|REINVESTIERUNG" //
+                        + "|STORNO DIVIDENDE" //
+                        + "|(CASH )?DIVIDEND" //
+                        + "|DIVIDENDO" //
+                        + "|DISTRIBUZIONE" //
+                        + "|Distribution" //
+                        + "|KAPITALREDUKTION)", //
+                        BLOCK_ENDSWITH); //
```
Nirus2000: Hmm... I see the changes were included in PR #5602 and now they're being reverted, but in a different way.
I'd prefer it if we re-included the changes from #5602 (those weren't very good).

Nirus2000 (dismissed):

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
kimmerin: The problem is that the new format contains a last page that starts with the "TRADE REPUBLIC BANK GMBH"-text but contains no dividend- or other fitting identifiers to process and ends with an parse error because of this, even after it has successfully processed the dividend on the first page.

There is no mechanism that allows to skip the parsing of a single block when encountering some match (something like `markAsFailure` being used to reject dividend-cancellations), so the solution to this is reintroducing the specific identifiers for the different types plus an ends-with-block containing the text that is missing in the additional page.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
kimmerin: This change was done due to a SONAR-warning that all the extractors in the different test methods are hiding this member.

Nirus2000: Really? Sonar... you want to go through over 4,000 tests just because of one warning? Well then... make a separate pull request for that.

kimmerin: It's the only issue that showed up in this particular file (except the TODO) and it doesn't hurt to address them while at that class.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
kimmerin: The space was present in the Dividend-files I've checked (e.g. Dividende15.txt) so I've adjusted the comment because I've assumed that there was a copy/paste error. I haven't changed any regexes, so I'm not sure what you're referring to.

Edit: Dividende30.txt that is, not 15.txt.

kimmerin:
I've added another fix to make an educated guess about the decimal separator being used with shares. Currently - with less than 1000 shares but with decimals separated by a dot, the number is parsed as german number leading to vastly hight numbers.

A final solution would be to change all regexes that currently handle the currency to have another group containing the monetary amount that is always shown with two decimals and the decimal separator.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
Nirus2000: https://github.com/portfolio-performance/portfolio/blob/e505f903aeb1ac93efc54133689fd97df1807e40/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/Dividende15.txt#L10

Nirus2000: https://github.com/portfolio-performance/portfolio/blob/e505f903aeb1ac93efc54133689fd97df1807e40/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/Dividende24.txt#L10

kimmerin: https://github.com/portfolio-performance/portfolio/pull/5766/changes#diff-a106dab3a853a2f9aa93ea9fbd9ed9bd08e0241cb488858a9040a1e1404868eaR11

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
```diff
@@ -5083,6 +5088,16 @@ public Optional<AccountAmountTransactionItem> findItem(Integer line, Money money
...
+    @Override
+    protected long asShares(String value)
+    {
+        if (value.matches("^(0\\.\\d+|\\d+\\.\\d{1,2}|\\d+\\.\\d{4,})$"))
+        {
+            return asShares(value, "en", "US"); //
+        }
+        return super.asShares(value);
+    }
+
```
Nirus2000, line 5101: If you take a look at the Erste Bank PDF import feature, you might be able to implement a suitable solution.
https://github.com/portfolio-performance/portfolio/blob/e505f903aeb1ac93efc54133689fd97df1807e40/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ErsteBankPDFExtractor.java#L1292-L1308

😄

kimmerin: Not sure what's so funny. I've tried that first (didn't need to look into other classes, the same solution is copy/pasted in `asAmount` and `asExchangeRate`) and got a failed test with `testWertpapierVerkauf10` because of

https://github.com/portfolio-performance/portfolio/blob/e505f903aeb1ac93efc54133689fd97df1807e40/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/Verkauf10.txt#L15

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
buchen: It is one of those changes where there are two sides to the argument. The tests are hiding the member. The member is some very special thing because it is overriding a method specifically for crypto. Therefore, I see the argument for changing it. Therefore I propose to keep this change in.

I understand the comment from Nirus of course. But the same could be made for converting explicit variable declarations to "var". We do not go through all and change them, but change them as we go. And we do not do so if it "pollutes" the diff too much.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
buchen: I think we all agree that we have a heuristic here and that this is not optimal. It would be better to know the number format from the document instead of guessing. But, alas, this is not part of this change.

With respect to what heuristic to use, I think the heuristic suggested here is actually better. We see this in test case Verkauf10 which would be detected as English, but is actually German.

```
1. 0\.\d+ — 0. then any digits → 0.5, 0.123456. Leading zero ⇒ dot must be decimal.
2. \d+\.\d{1,2} — digits, dot, 1–2 trailing digits → 12.5, 100.25. Too few trailing digits to be a thousands group ⇒ decimal.
3. \d+\.\d{4,} — digits, dot, 4+ trailing digits → 12.3456. Too many for a thousands group ⇒ decimal
```

The open gap of the heuristic are numbers with three digits "1.000". It could be German (1000 shares) or English (1 share). Falling back to German is sensible (but only because most documents are in German). I think we should also consider English numbers which have the comma as thousands separator.

@Nirus2000 my proposals is to extract that into a reusable method - something along the lines of ```guessLocale``. What do you think?

Then I can make a proposal and push (I cannot re-open this pull request because the remote branch does not exist anymore)

kimmerin: > I cannot re-open this pull request because the remote branch does not exist anymore

I've restored the branch and reopened it.

Before you start doing things, I can give some feedback on your comments but got something else to do now, so I'll do that when I've got the time.

kimmerin: > my proposals is to extract that into a reusable method - something along the lines of ```guessLocale``. What do you think?

While I support the idea to avoid copy/paste-solutions and introduce global helper methods I'd prefer to have a solution that doesn't rely on guessing. For this particular problem the solution would be to look for a number we know for sure that it contains a decimal separator. For TR-PDFs that would be the monetary amount of the purchase price (per unit or the sum doesn't matter), e.g. 

````
Long @49.99 $ Monster Beverage Open End Turbo 1.000 Stk. 0,15 EUR 150,00 EUR
````

Currently the regexes only use them to get the currency but because monetary amounts (at least here but it should be true for most of the other banks) always include cents there has to be a decimal separator (in above example, we see, it's the comma, therefor the dot must be the thousand separator).

I've played with the idea to implement that in this fix but stepped back because it would require quite some effort and a lot of changes in the "background":

 - Change all regexes that currently handle currencies to catch the monetary amount by a new named group, e.g. `moa`, e.g. `.match("^(?<name>.*) [\\.,\\d]+ (Stk\\.|titre\\(s\\)|Pz\\.|Pcs\\.) (?<moa>[\\.,\\d]+) (?<currency>[A-Z]{3}) [\\.,\\d]+ [A-Z]{3}$")`
 - Add a member to `Transaction` to contain the formatting style (not necessarily the Locale) that can be used for the NumberParser to parse a number and set it accordingly if we encounter a line containing a monetary amount
 - Recalculate all previously processed numbers if we encounter the moa-entry at a later point in time that shows that we've used the wrong assumptions. Alternatively first only store the values to be parsed and do the parsing when wrapping up the transaction (using a default style if no moa-entry has been found).

Especially the last step kept me from going through with this because it's quite a deviation from the current process.

But before considering adding a proposed guesstimate-method and use that everywhere it might be worth considering to invest that time into thinking about a solution that isn't a lot of band aid that we already know can't work reliably with values like the mentioned `1.000`.

Comment (kimmerin) [reply name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java:5101]: I've played around with the idea, I've described above and created a test balloon for `addBuySellTransaction` (the one where there is one test with a number of shares formatted as `1.000`:

````
                                        // @formatter:off
                                        // Clinuvel Pharmaceuticals Ltd. 80 Stk. 22,82 EUR 1.825,60 EUR
                                        // Tencent Holdings Ltd. 0,3773 titre(s) 53,00 EUR 20,00 EUR
                                        // zjBAM Corp. 125 Pz. 29,75 EUR 3.718,75 EUR
                                        // Vonovia SE 0,781379 tít. 24,06 EUR 18,80 EUR
                                        // @formatter:on
                                        section -> section //
                                                        .attributes("shares", "moasample") //
                                                        .match("^.* (?<shares>[\\.,\\d]+) (Stk\\.|titre\\(s\\)|Pz\\.|t.t\\.) (?<moasample>[\\.,\\d]+) .*$") //
                                                        .assign((t, v) -> t.setShares(
                                                                        asShares(v.get("shares"), v.get("moasample")))),
                                        // @formatter:off
                                        // Berkshire Hathaway Inc. 0.3367 Pcs. 297.00 EUR 100.00 EUR
                                        // @formatter:on
                                        section -> section //
                                                        .attributes("shares", "moasample") //
                                                        .match("^.* (?<shares>[\\.,\\d]+) Pcs\\. (?<moasample>[\\.,\\d]+) .*$") //
                                                        .assign((t, v) -> t.setShares(
                                                                        asShares(v.get("shares"), v.get("moasample")))),
                                        // @formatter:off
                                        // Factor MSCI USA Quality ESG EUR Hedged (Acc) 0,100058 42,775 EUR 4,28 EUR
                                        // @formatter:on
                                        section -> section //
                                                        .attributes("shares", "moasample") //
                                                        .match("^.* (?<shares>[\\.,\\d]+) (?<moasample>[\\.,\\d]+) [A-Z]{3} [\\.,\\d]+ [A-Z]{3}$") //
                                                        .assign((t, v) -> t.setShares(
                                                                        asShares(v.get("shares"), v.get("moasample")))),
````

This "exctracts" an example of a monetary amount that is garantueed to contain a decimal separator.

````
    long asShares(String value, String moaSample)
    {
        // check for the occurrence of the last non-digit-character -> must be
        // the decimal separator
        Matcher m = decSepFinder.matcher(moaSample);

        // MOA is a number without fractions, so we apply our "magic" logic to
        // the value
        if (!m.matches())
            return asShares(value);

        // search for occurrence of decimal separator in value
        int decPos = value.indexOf(m.group(1));

        // no fraction in the actual value, remove all non-digit-characters and
        // simply convert - no magic needed
        if (decPos == -1)
            return super.asShares(value.replaceAll("\\D", ""));

        // remove all non-digit-characters in the number part, add a dot and the fractions
        // so we end up with a "regular" number for the en_US-Locale
        String toParse = value.substring(0, decPos).replaceAll("\\D", "") + //
                        "." + //
                        value.substring(decPos + 1);

        return super.asShares(toParse, "en", "US");
    }
````

Works but there is a catch if the line with the number of shares doesn't contain a monetary amount. This would require to save a monetary amount sample to the underlying transaction so that it is available when handling numbers like the number of shares here.

While this would solve the number parsing issue once and for all and makes it unnecessary to know all possible separators out there, it's quite a big change, so I suggest to stick with my solution in the PR that solves most of the formats TRs use with the known limitation of english formatted four digit long numbers.

Comment (buchen) [reply name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java:5101]: I appreciate the idea. But is also reads super complicated.

And there are cases where Trade Republic seems to mix formats. For example, the ExAnte03.txt contains both in one line:
```
EX-ANTE KOSTENINFORMATION ZUM WERTPAPIERKAUF
WERTPAPIER BESTELLUNG / BETRAG WERT
FTSE All-World USD (Acc)
Kauf 0.851192 Stk. 124,239984 €
```

I was wondering if the information "Stück" vs. "Pcs" could be used. But Sparplan12.txt mixes that as well:

```
Sparplanausführung am 02.06.2026 an der Lang und Schwarz Exchange.
Der Kontrahent der Transaktion ist Lang & Schwarz TradeCenter AG & Co. KG.
POSITION ANZAHL DURCHSCHNITTSKURS BETRAG
Core MSCI World USD (Acc) 3.226355 Stk. 123.9789 EUR 400.00 EUR
```

I tried a more generic ```guessNumberLocale```. It would work for Trade Republic, but not for other importers. Then I would start adding boolean flags because sometimes 1.000 is one thousand and sometimes 77.638 is actually seventy seven and a little bit more.

```
    public static Locale guessNumberLocale(String value, Locale fallback)
    {
        var v = trim(value).replaceAll("\\s", "");

        if (v.indexOf('\'') >= 0)
            return Locale.of("de", "CH");

        var hasDot = v.indexOf('.') >= 0;
        var hasComma = v.indexOf(',') >= 0;

        if (hasDot && hasComma)
            return v.lastIndexOf('.') > v.lastIndexOf(',') ? Locale.US : Locale.GERMANY;

        if (hasDot)
        {
            if (v.indexOf('.') != v.lastIndexOf('.')) // repeated dot = grouping
                return Locale.GERMANY;
            if (v.matches("[1-9]\\d{0,2}\\.\\d{3}")) // ambiguous thousands group
                return fallback;
            return Locale.US; // single dot = decimal point
        }

        if (hasComma)
        {
            if (v.indexOf(',') != v.lastIndexOf(',')) // repeated comma = grouping
                return Locale.US;
            if (v.matches("[1-9]\\d{0,2},\\d{3}")) // ambiguous thousands group
                return fallback;
            return Locale.GERMANY; // single comma = decimal comma
        }

        return fallback;
    }
```

Therefore I am thinking to leave the simplified regard expression here (and possibly extract the repeated search for dot and comma into a helper method)

buchen:
@kimmerin merged. thanks. I did some changes to the formatting to keep it consistent. And I uncommented the ex-date and changed it to the date found in the document. I was not quite sure if there was more reason to commenting out the test case. If so, please let me know.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/TradeRepublicPDFExtractor.java
kimmerin: > And there are cases where Trade Republic seems to mix formats. For example, the ExAnte03.txt contains both in one line:
>
>> EX-ANTE KOSTENINFORMATION ZUM WERTPAPIERKAUF
>> WERTPAPIER BESTELLUNG / BETRAG WERT
>> FTSE All-World USD (Acc)
>> Kauf 0.851192 Stk. 124,239984 €

Bummer. For a local test I've only changed the dividend-logic  and all tests were running through but my optimism seems to be premature.

> But Sparplan12.txt mixes that as well:
>
>> Core MSCI World USD (Acc) 3.226355 Stk. 123.9789 EUR 400.00 EUR

That's actually a consistent formatting and uses the dot as decimal separator that would be processed correctly using the moasample-approach.

> And I uncommented the ex-date and changed it to the date found in the document. I was not quite sure if there was more reason to commenting out the test case. If so, please let me know.

To be honest, I don't remember why I've commented it out, so I don't see a problem to reactivate the tests.

PR #5780, opened 2026-06-04
added test for a supplied dump containing an entry that leads to a parse error. Fixed parse error

kimmerin:
Dump provded by the PP forum: https://forum.portfolio-performance.info/t/pdf-import-von-bondora-go-grow/10760/141?u=kimmerin

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/BondoraCapitalPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/BondoraCapitalPDFExtractorTest.java
@@ -729,4 +729,69 @@ public class BondoraCapitalPDFExtractorTest
         assertThat(results, hasItem(removal(hasDate("2025-02-04"), hasAmount("EUR", 1111.00), //
                         hasSource("Kontoauszug17.txt"), hasNote("SEPA payment"))));
     }
+
+    @Test
+    public void testKontoauszug18()
+    {
+        var extractor = new BondoraCapitalPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Kontoauszug18.txt"), errors);
+
+        errors.forEach(Exception::printStackTrace);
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/BondoraCapitalPDFExtractorTest.java
```diff
@@ -729,4 +729,69 @@ public void testKontoauszug17()
...
+        {
+            assertThat(results, hasItem(interest( //
+                            hasDate((String) expLine[0]), //
+                            hasSource("Kontoauszug18.txt"), //
+                            hasNote("Go & Grow Zinsen"), //
+                            hasAmount("EUR", (double) expLine[1]), hasGrossValue("EUR", (double) expLine[2]), //
+                            hasTaxes("EUR", (double) expLine[3]), hasFees("EUR", (double) expLine[4]))));
+
+        }
+    }
```
Nirus2000, line 796: The good thing is that you apparently don't have to mainternance for the source, so new variants are constantly being created to implement tests or other features. Thankfully, there are contributing guidelines that you can read.
Would it be possible to stick to them?

Well, and if you can’t manage it yourself, then start claude code in the terminal and type in:

> Check the PDFImporter BondoraCapitalPDFExtractor and BondoraCapitalPDFExtractorTest against the new pdfextractor rules and follow the rules explicitly!

Maybe look in the .claude/rules folder....

Run this shit!

Nirus2000 (changes requested):

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/BondoraCapitalPDFExtractorTest.java
buchen: First, let's use polite language. Please turn down the level of sarcasm. Everybody is contributing their personal free time to the project.

Second, give the repetitive nature of the test case, duplicating the assertThat block does not seem like a good option. Using a loop seems like a good option to me. I do not see a reason why AI tools should have problem with such code. I am not sure what you are proposing as an alternative.

PR #5784, opened 2026-06-05
Bondora Capital: Fixed parse error in updated document format

buchen:
Added test for a supplied dump containing an entry that leads to a parse error.

Issue: https://forum.portfolio-performance.info/t/pdf-import-von-bondora-go-grow/10760/136

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/Kontoauszug18.txt
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/Kontoauszug18.txt
@@ -1,0 +1,46 @@
+PDFBox Version: 3.0.6
+Portfolio Performance Version: 0.84.0
+System: win32 | x86_64 | 21.0.5+11-LTS | Azul Systems, Inc.
+-----------------------------------------
+Zusammenfassung
+Zu Beginn 6.367,72 €
+kYNPH EYTrZD pRhlV
+Einzahlungen 31,59 €
+Datum 01.05.2026 - 31.05.2026 Auszahlungen 0 €
+Ergebnis 6.399,31 €
+Alle Ziele
+Datum Zahlungsart Eingänge Ausgänge Guthaben
+01.05.2026 Go & Grow Zinsen 1,01 € 6.368,73 €
+01.05.2026 Nicht  zugeordneter  Saldo  -
+Reserve 6.368,73 € 6.368,73 €
+02.05.2026 Go & Grow Zinsen 1,02 € 6.369,75 €
+03.05.2026 Go & Grow Zinsen 1,02 € 6.370,77 €
+04.05.2026 Go & Grow Zinsen 1,02 € 6.371,79 €
+05.05.2026 Go & Grow Zinsen 1,01 € 6.372,8 €
+06.05.2026 Go & Grow Zinsen 1,02 € 6.373,82 €
+07.05.2026 Go & Grow Zinsen 1,02 € 6.374,84 €
+08.05.2026 Go & Grow Zinsen 1,02 € 6.375,86 €
+09.05.2026 Go & Grow Zinsen 1,01 € 6.376,87 €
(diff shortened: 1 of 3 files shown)


buchen:
@kimmerin @Nirus2000
I was not able to reopen #5780 and therefore had to push a new pull request. But I think the Bondora importer deserves a fix.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/Kontoauszug18.txt
```diff
@@ -0,0 +1,46 @@
...
+-----------------------------------------
+Zusammenfassung
+Zu Beginn 6.367,72 €
+kYNPH EYTrZD pRhlV
+Einzahlungen 31,59 €
+Datum 01.05.2026 - 31.05.2026 Auszahlungen 0 €
+Ergebnis 6.399,31 €
+Alle Ziele
+Datum Zahlungsart Eingänge Ausgänge Guthaben
+01.05.2026 Go & Grow Zinsen 1,01 € 6.368,73 €
```
buchen, line 13: @kimmerin This interest payment is not part of the test case - but shouldn't it be imported as well?

kimmerin: I've created the expValue-array by hand and it seems that I've missed that transaction. The number of entries in the array is 30 rather than the 31, so there doesn't seem to be anything wrong with the importer, just with the test I've created. Simply adding

````
  { "2026-05-01", 1.01d, 1.01d, 0d, 0d }, //
````

at the beginning of the array should solve the issue. If this way of testing is done more than this time, another check

````
  assertThat(result.size(), is(expVals.length));
````

before the loop would make sure that an error like this doesn't happen again.

kimmerin:
> @kimmerin @Nirus2000 I was not able to reopen #5780 and therefore had to push a new pull request. But I think the Bondora importer deserves a fix.

I think more people are desperatley waiting for #5766 to be fixed. It's currently preventing dividend PDFs being parsed (at a time where you get many each day) and comes with magnitudes higher share numbers in the range of thousands and millions when parsing purchase-PDFs with fractional shares ;-)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/Kontoauszug18.txt
buchen: You are right. I noticed that the parsing includes this line, but the testing is missing it.

For now, I have asked Claude to convert the tests to the usual pattern for maintainability.

kimmerin:
Hm, I'm listed as reviewer but this is essentially my code so I think an approval should come from somebody else.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/bondoracapital/Kontoauszug18.txt
kimmerin: > For now, I have asked Claude to convert the tests to the usual pattern for maintainability.

Fine by me. Maybe there is time and motivation to discuss a more general approach to get the size of these tests down and therefore make them better maintainable and - this is something I try to always get - readability for humans. As an example just look at "my" array with expected values and try to see the differences between the transactions (date, monetary amounts, etc). Try to see the same differences when looking at the 200 lines of `assertThat`s Claude created, doing the same thing and I think you see where I'm going with this.

These type of tests are essentially crying to be dynamically created and I'm not talking about using an AI to do copy/paste-stuff but the use of `@RunWith(Parameterized.class)` to automatically create a test for each dump file it finds in a given directory. The expected result is then read from a second file (to be created), e.g. containing a JSON representing the data structure to check the current parse-result against.

If that is done, new tests are then created by simply adding two files into a directory. And I know this works, because I've done these kinds of things already multiple times for parsers like this ;-)

But as you've already said: Totally out of scope for this issue ;-)

PR #5785, opened 2026-06-06
Keep selection when changing view settings in SecuritiesPerformanceView

wahlbrink:
Currently, the selection in the SecuritiesPerformanceView is lost when the user changes, for example, the reporting period.

The change adds the methods hashCode and equals to the class RowElement (class of the items in the table), so that TableViewer automatically retains the selection.


buchen:
merged. I did a very small change because the Equus check at the moment assumes that the CKD performance record of the other object is always also non-nile. That is currently the case but might not be in the future so I slightly change it to be a little bit more defensive. Thanks!

PR #5786, opened 2026-06-07
ING PDF import: add additional entries

daha96:
add additional entries "Abbuchung", "Retoure" and "Bezuege"


kimmerin:
Just my two cents: Don't change existing dump-files but add new dumps containing the new types and create new test-methods that check the fixed importer. This way it's ensured that everything that worked before still works the same. By changing old files you might accidentially break things without realizing.

daha96:
I created an own test case for the additional entries. I hope that's okay.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/ingdiba/INGDiBaPDFExtractorTest.java
```diff
@@ -4323,6 +4323,67 @@ public void testGiroKontoauszug02()
...
+        var item = iter.next();
+
+        // assert transaction
+        var transaction = (AccountTransaction) item.getSubject();
+        assertThat(transaction.getType(), is(AccountTransaction.Type.REMOVAL));
+        assertThat(transaction.getCurrencyCode(), is("EUR"));
+        assertThat(transaction.getDateTime(), is(LocalDateTime.parse("2018-11-16T00:00")));
+        assertThat(transaction.getAmount(), is(Values.Amount.factorize(4.5)));
+        assertThat(transaction.getSource(), is("GiroKontoauszug07.txt"));
+        assertThat(transaction.getNote(), is("Abbuchung"));
```
buchen, line 4360: I converted these test to the new style

buchen:
rebased, squashed, and merged.

thanks for the contribution.jm

PR #5787, opened 2026-06-07
ING PDF import: import dividend fron Giro- and Extra-Kontauszug

daha96:


Morpheus1w3:
@daha96 
I don't think this is the right way to go.

It was regularly discussed and decided, not to extract security transactional data from the account statements due to missing information (e.g. tax, ex-date).

kimmerin:
@Morpheus1w3 was a bit faster than me with his comment ;-) The account statements are missing important informations of a dividend that are required for PP to do its job properly. E.g. in this particular case taxes and exchange rates. Lately the support of the dividends' ex-dates has been added, etc. so I think adding this feature shouldn't be done here and importing the single dividend reporting PDFs should stay the way to go (you can multi-select them in the file explorer of your choice and drag/drop them into the app, so it's not more effort than importing an account statement)

buchen:
I agree that importing dividend statements with limited information is not useful. In particular, if then the user is importing the original statements, we will end up with duplicates which cannot be replaced detected.

The only thing I'm wondering if it makes sense to include them as "skipped items". That indicates to the user to import the other dividend statements. What do you think?

daha96:
Thank you for the feedback. I agree that the method is missing important information. For me, it would be beneficial to be able to import old dividend income for which I no longer have access to the complete dividend reports. While this wouldn't be entirely accurate, it would still be better than not considering it at all.

Would it be possible to make the import optional in some way? Unfortunately, I'm not very familiar with the project and don't know what the best approach would be.

buchen:
> For me, it would be beneficial to be able to import old dividend income for which I no longer have access to the complete dividend reports.

That is fair, but I believe your problem is soft now, right? I mean, you can use this temporary code to import it once and then these all dividend statements are included in your file. But going forward, I am skeptical that we help users by importing these dividend statements in creating potentially a lot of duplicates with missing information.

daha96:
It's totally fine for me to use this as a temporary solution. I just wanted to make the code available in case there's any further need for it.

buchen:
Thanks. Closing. Dividend statements are better imported from the original documents with all information.

PR #5788, opened 2026-06-07
Ing pdf import: add more information to note

daha96:


kimmerin:
Esesentially the same comment here as I've said in #5786: Add new dump files and create new tests around it rather than amending existing ones. Also: The tests were in the process to be changed to use the hamcrest-way of testing. You can see that how it should be done now e.g. at https://github.com/portfolio-performance/portfolio/pull/5777/changes

daha96:
Should I adapt the modified tests to the new style, or is it planned to convert all tests to the new style at once?

buchen:
> Should I adapt the modified tests to the new style, or is it planned to convert all tests to the new style at once?

There's no need to convert existing tests, but new tests should use the more readable testing format.

buchen:
I was in the process of picking up your change and merging it. I also moved the additional comment about fixing something (your last commit). I had the same thought. :-)

buchen:
squashed and merged.

PR #5801, opened 2026-06-13
added support for PDFs containing the execution of an IPO order

kimmerin:
Dump provided at the PP forum: https://forum.portfolio-performance.info/t/pdf-import-von-trade-republic/5107/871?u=kimmerin


buchen:
merged

PR #5804, opened 2026-06-15
Improve detection of number format in bank documents

buchen:
- Introduce shared number-format guessing
    Adds central locale detection for formatted numbers, including a reliable detection variant that returns no locale for ambiguous cases. Also adds Locale-based amount/share conversion overloads so importers can
    pass detected formats directly.

  - Apply number-format guessing to Trade Republic
    Updates Trade Republic to derive the share-number format from a monetary amount on the same line. This removes importer-specific share heuristics and makes mixed German/US formats more robust.

  - Apply reliable number-format detection to Bondora
    Detects Bondora statement number formats from unambiguous summary/header amounts and, for slash-date transactions, also considers the balance as a local format sample. Ambiguous values are ignored so the importer
    falls back to its existing structural defaults.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ExtractorUtilsConvertToNumberLongTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ExtractorUtilsConvertToNumberLongTest.java
@@ -43,6 +43,13 @@ public class ExtractorUtilsConvertToNumberLongTest
     @Test
     public void testConvertToNumberLong()
     {
+        long actualOutput = ExtractorUtils.convertToNumberLong(input, Values.Amount, locale);
+        assertEquals(expectedOutput, actualOutput);
+    }
+
+    @Test
+    public void testConvertToNumberLongWithLanguageAndCountry()
+    {
         long actualOutput = ExtractorUtils.convertToNumberLong(input, Values.Amount, locale.getLanguage(),
                         locale.getCountry());
         assertEquals(expectedOutput, actualOutput);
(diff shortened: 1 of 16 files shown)


buchen:
@kimmerin I have picked up your idea of using another monetary amount in the same line to more reliably detect the number format. Take a look at the change in the trade republic importer.

The Bondora importer was more complicated because you cannot reliably rely on monetary amounts in the same line. So the logic is now using the monetary amount itself, other monetary amounts in the same line, and then falls back to the document header with the summary line.

For the existing test cases this works, but I'm still thinking if there are side effects to this. Do you have any opinion?

kimmerin:
@buchen I'm a bit surprised that the tests for the TR-importer go through because in one of your [comments](https://github.com/portfolio-performance/portfolio/pull/5766#discussion_r3403942459) you've shown `ExAnte03.txt` where the amount used a dot as decimal separator and the monetary amount used a comma. But apart this I think that adding this way of finding the correct format-settings should reduce the number of situations where this needs to be fiddled around with.

I don't see it necessary to change all the importers in bulk, this should be done on a need-to-change-basis because as long as they work as intended, there is no need to change them. Also I don't think that it's necessary to use Locales to hold the information what formatting is used. The only necessary information is the decimal separator, what the thousand separator is, is alrady irrelevant because you can simply remove all non-numeric characters that are not the evaluated decimal sepator and replace that to a dot afterwards. After that "normalization" you can parse the number with the fixed `Locale.US`. When using Locales you always need to know what Locale is to be used in dependence of the evaluated separator chars and if we already have them, we can simply create an instance of `DecimalFormatSymbols`, set the evaluated separators and use that as parameter for the new instance of `DecimalFormat` instead of mapping dec/thousand-sep combinations to a Locale, that is given to NumberFormat, that is using the Locale's underlying DecimalFormatSymbols to create an instance of DecimalFormat ;-)

buchen:
> I'm a bit surprised that the tests for the TR-importer go through because in one of your https://github.com/portfolio-performance/portfolio/pull/5766#discussion_r3403942459 you've shown ExAnte03.txt where the amount used a dot as decimal separator and the monetary amount used a comma.

That is a good point. I applied the logic only to those cases where there are "Stücke" and "Pieces" in the regular expressions which indicates it is not clear whether we have a comma or dot as decimal point. It is still a case by case application of this logic.

PR #5808, opened 2026-06-21
Fix FIL Fond Bank PDF Importer failing to parse data if the lines ends with spaces

kimmerin:
The regexes assume the line not ending with whitespaces leading to failing imports with some PDFs. This PR fixes that. Debug-data provided by the PP-forum at https://forum.portfolio-performance.info/t/pdf-import-von-frankfurter-fondsbank-ffb/4751/89?u=kimmerin


kimmerin:
Some mixup in the commits. Will create a new PR

PR #5809, opened 2026-06-21
fixed failing import if some of the lines end with trailing spaces

kimmerin:
The regexes assume the line not ending with whitespaces leading to failing imports with some PDFs. This PR fixes that. Debug-data provided by the PP-forum at https://forum.portfolio-performance.info/t/pdf-import-von-frankfurter-fondsbank-ffb/4751/89?u=kimmerin

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/filfondbank/FILFondbankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/filfondbank/FILFondbankPDFExtractorTest.java
@@ -169,6 +169,129 @@ public class FILFondbankPDFExtractorTest
     }
 
     @Test
+    public void testWertpapierKauf03()
+    {
+        var extractor = new FILFondbankPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Kauf03.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(3L));
+        assertThat(countBuySell(results), is(3L));
+        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
+        assertThat(results.size(), is(6));
+        new AssertImportActions().check(results, "EUR");
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("IE00B4X9L533"), hasWkn("A1C9KK"), hasTicker(null), //
+                        hasName("HSBC MSCI WORLD ETF"), //
+                        hasCurrencyCode("USD"))));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/filfondbank/FILFondbankPDFExtractorTest.java
```diff
@@ -168,6 +168,129 @@ public void testWertpapierKauf02()
...
+                        hasNote("Auftrags-Nr. 2667756179"), //
+                        hasAmount("EUR", 400.00), hasGrossValue("EUR", 400.00), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00), //
+                        check(tx -> {
+                            var c = new CheckCurrenciesAction();
+                            var s = c.process((PortfolioTransaction) tx, new Portfolio());
+                            assertThat(s, is(Status.OK_STATUS));
+                        }))));
+        assertThat(results, hasItem(purchase( //
+                        hasDate("2026-06-15T11:14:49"), hasShares(7.711), //
```
buchen, line 274: In the document, I do not see a time ("Handelsuhrzeit") for this transaction. I think the "time" must be cleared per block, not per document. Obviously, this is not introduced by this change.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FILFondbankPDFExtractor.java
```diff
@@ -52,7 +53,7 @@ private void addBuySellTransaction()
 
                         // Is type --> "Verkauf" change from BUY to SELL
                         .section("type").optional() //
-                        .match("^(?<type>(Splittkauf|Splitkauf|Wiederanlage|Kauf|Verkauf|Steuerliche Informationen \\(Einzeltransaktion\\))).*$") //
+                        .match("^(?<type>(Splittkauf|Splitkauf|Wiederanlage|Kauf|Verkauf|Steuerliche Informationen \\(Einzeltransaktion\\))).*\\s*$") //
```
buchen, line 56: The ```.*``` is already matching a space. Adding ```\\s*``` is redundant in those cases.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/FILFondbankPDFExtractor.java
buchen, line 253: I wonder if it makes sense to add this to fees and taxes as well. The sample Kauf03.txt does not include them, but other new documents could. I added those now because I think there is no particular harm to it. If I think about it, we could also add it to the sales, dividend, and fee transaction. Will do this as a separate commit.

buchen:
@kimmerin thanks for the contribution. rebased and merged. I added two commits on top regarding my comments. I want to include it into the release. If I break something, please let me know

kimmerin:
@buchen 

> If I break something, please let me know

I'm not a customer of FIL or TradeRepublic so I won't be able to tell if something is broken there because I only have got the PDF-dumps, you've got as well. And the imports that are relevant for me IRL haven't been changed for some time. ;-)

PR #5815, opened 2026-06-26
Allow to disable coloring of values of indicator and ratio widgets

Lukas0907:
It can be distracting when every value > 0 is green and every value < 0 is red in a dashboard. This PR adds an entry to the context menu of the ratio and the indicator widget so that coloring of values can be disabled:

image


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -271,13 +271,13 @@ public enum WidgetFactory
                                         return index.getVolatility().getStandardDeviation();
                                     }) //
                                     .withTooltip((ds, period) -> Messages.TooltipVolatility) //
-                                    .withColoredValues(false) //
+                                    .withColoredValueDefault(false) //
```
buchen: I believe volatility can never be negative. And if we color volatility, then by magnitude - say green for low volatility and red for high volatility.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -459,7 +459,7 @@ public enum WidgetFactory
                                     }) //
                                     .withBenchmarkDataSeries(false) //
                                     .with(ds -> ds.getInstance() instanceof Security) //
-                                    .withColoredValues(false) //
```
buchen, line 462: The latest security price cannot be negative

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/RatioWidget.java
```diff
@@ -67,5 +69,17 @@ public void update(Object rateOrValue)
 
         Double rate = (Double) rateOrValue;
         this.indicator.setText(!rate.isInfinite() ? Values.Percent2.format(rate) : "-"); //$NON-NLS-1$
+
+        boolean isNegative = rate < 0;
+        boolean isValueColored = get(ColoredValueConfig.class).isValueColored();
+
+        if (isValueColored)
+        {
+            indicator.setTextColor(isNegative ? Colors.theme().redForeground() : Colors.theme().greenForeground());
```
buchen: if we color it, then use ```ValueColorScheme.current().negativeForeground()/positiveForeground()```

The user can configure how to color negative or positive values.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/ColoredValueConfig.java
```diff
@@ -0,0 +1,49 @@
...
+        });
+
+        action.setChecked(this.isValueColored);
+        manager.add(action);
+    }
+
+    @Override
+    public String getLabel()
+    {
+        return Messages.LabelColorValue;
```
buchen: The label is used to construct the tooltip and should reflect the current state (say on/off)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
```diff
@@ -648,6 +648,7 @@ public class Messages extends NLS
     public static String LabelCollapsibleSection;
     public static String LabelColonSeparated;
     public static String LabelColorBy;
+    public static String LabelColorValue;
```
buchen: Just a minor thing, but I try to keep them sorted alphabetically.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/RatioWidget.java
```diff
@@ -67,5 +69,17 @@ public void update(Object rateOrValue)
 
         Double rate = (Double) rateOrValue;
         this.indicator.setText(!rate.isInfinite() ? Values.Percent2.format(rate) : "-"); //$NON-NLS-1$
+
+        boolean isNegative = rate < 0;
+        boolean isValueColored = get(ColoredValueConfig.class).isValueColored();
+
+        if (isValueColored)
```
buchen, line 77: what happens if the rate is infinite? Does it color it? Should it color it?

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/WidgetFactory.java
```diff
@@ -320,7 +320,7 @@ public enum WidgetFactory
                                                         Values.Percent5.format(vola.getStandardDeviation()),
                                                         Values.Percent5.format(vola.getSemiDeviation()));
                                     }) //
-                                    .withColoredValues(false) //
+                                    .withColoredValueDefault(false) //
```
buchen: I don't think semi deviation can be negative

buchen (commented):
I like reducing noise and unnecessary color is noise.

On the other hand, this change introduces noise because the user can configure the color for values where it is questionable. Volatility, the security price, semideviation -  they don't have negative values.

I am interested in your opinion: Couldn't we say if by default an indicator widget is colored, then the color can also be switched off. If by default it is not colored, then the coloring option doesn't exist. And the indicator widget could use this information whether to inject the configuration. Any thoughts?

Lukas0907:
Thanks for your input. I changed the PR so that it only allows to configure coloring if it makes sense (i.e. not when it is disabled already).

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/RatioWidget.java
Lukas0907: - 0/0 is shown as `NaN%` (green if enabled)
- x/0 is shown as `-` (green if x is positive, or red if x is negative)
- 0/x is shown as `0` (green)

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/IndicatorWidget.java
```diff
@@ -105,9 +105,15 @@ IndicatorWidget<N> build()
     private boolean isValueColored = true;
 
     public IndicatorWidget(Widget widget, DashboardData dashboardData, boolean supportsBenchmarks,
-                    Predicate<DataSeries> predicate)
+                    Predicate<DataSeries> predicate, boolean isValueColored)
     {
         super(widget, dashboardData, supportsBenchmarks, predicate);
+
+        this.isValueColored = isValueColored;
```
buchen, line 112: The member variable is not really needed anymore because the determining factor is now whether a config exists or not. I have removed it for maintainability.

name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/views/dashboard/IndicatorWidget.java
```diff
@@ -105,9 +105,15 @@ IndicatorWidget<N> build()
...
+        if (isValueColored)
+        {
+            addConfigAfter(LabelConfig.class, new ColoredValueConfig(this, isValueColored));
```
buchen, line 115: The config is added pretty much on the top. But I think it is not _that_ important. I moved it to the bottom.

buchen (approved):
merged. Thanks for your contribution.

PR #5820, opened 2026-06-30
Render retired asset icons and InvestmentVehicle with  alpha treatment

Morpheus1w3:
https://forum.portfolio-performance.info/t/inaktive-wertpapiere-logo/39471/

image

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Images.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Images.java
@@ -171,6 +171,18 @@ public enum Images
         return image;
     }
 
+    public static Image resolveWithAlpha(String file, int alpha)
+    {
+        String key = file + ":alpha:" + alpha; //$NON-NLS-1$
+        Image image = imageRegistry.get(key);
+        if (image != null)
+            return image;
+
+        image = ImageManager.getAlphaVersion(resolve(file, false), alpha);
+        imageRegistry.put(key, image);
+        return image;
+    }
+
     private static ImageDescriptor descriptor(String file)
     {
         ImageDescriptor descriptor = imageRegistry.getDescriptor(file);
@@ -198,6 +210,11 @@ public enum Images
     public Image image(boolean disabled)
     {
         return resolve(file, disabled);
+    }
+
+    public Image imageWithAlpha(int alpha)
+    {
+        return resolveWithAlpha(file, alpha);
     }
 
     public String getImageURI()
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/LogoManager.java
@@ -11,6 +11,7 @@ import name.abuchen.portfolio.model.Classification;
 import name.abuchen.portfolio.model.ClientSettings;
 import name.abuchen.portfolio.model.ImageManager;
 import name.abuchen.portfolio.model.InvestmentPlan;
+import name.abuchen.portfolio.model.InvestmentVehicle;
 import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.Security;
 import name.abuchen.portfolio.ui.Images;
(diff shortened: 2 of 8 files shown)


kimmerin:
There is already something in the app that renders logos differently if they are considered disabled (e.g. if you no longer hold any shares, etc.) Maybe you can have a look at

https://github.com/portfolio-performance/portfolio/commit/6e22fef8ed2f2db873c447e042105572cf7f7054
and
https://github.com/portfolio-performance/portfolio/commit/dee673278fef9520c9fbc8c5e270a36364816d09

how I've done it two years ago ;-)

buchen:
@kimmerin It looks to me somehow your capability broke - at least on macOS I do not see disabled icons for instruments anymore. Is this different maybe on Windows?

kimmerin:
> It looks to me somehow your capability broke - at least on macOS

@buchen Are you sure you've addressed the correct person? I haven't created this PR or made any recent changes to the disabling-feature ;-)

Or are we talking about the feature I've mentioned in my answer to @Morpheus1w3 and that shouldn't have changed for the last two years (at least I haven't seen anything in the git history indicating that). Here[TM] it works:

grafik

Securities without holdings are shown in a disabled state. You can see the difference with "Siemens Energy AG" and the corresponding options-entry one row further. I have to check the PR of back then but this feature was only added to the performance->calculation-view. Maybe you're looking at a different one?

buchen:
> Securities without holdings are shown in a disabled state. You can see the difference with "Siemens Energy AG" and the corresponding options-entry one row further.

I know you implemented the existing code. When running PP with the code of today, I am not seeing a difference between an active and disabled icon on macOS. That is the reason I am asking. (And I guess that cloud motivate the pull request). Then it seems to be operation-system specific.

I tested it with the custom logo for the price index:

Bildschirmfoto 2026-07-02 um 16 31 06

Same here:

Bildschirmfoto 2026-07-02 um 16 32 48

Bildschirmfoto 2026-07-02 um 16 36 26


Looking at the screenshots by @Morpheus1w3 - that is Windows.
@kimmerin on what platform was your screenshot?

kimmerin:
@buchen 

> on what platform was your screenshot?

Windows 10

kimmerin:
@buchen for funsies I've switched my working dir to use this PR and the calculation-view behaves as before:
grafik

The dimming of the logos work e.g. for accounts as well:

grafik

But as already mentioned: This kind of functionality already exists in performance->calculation where inactive elements are shown with the logo being grayscaled (by Eclipse itself) which shows the different status more clearly in my eyes than the logo with an alpha setting. The presentation should at least be consistent over the complete application, so this change should be changed to simply call the LogoManager with the parameter `inactive=true` if an entry should be shown in the greyed out state or my performance-calculation-view-change should be reverted so that the new presentation is used there as well.

buchen:
First, I agree with you @kimmerin that we want it consistent. That is why I am having this conversation.

Second, what confused me was this: There are two types of image manipulation:
* If the security is retired and has a standard icon, it gets a gray one (but: no change for instruments with icons)
* If the security has no holdings anymore but still shows up in Performance -> Calculation, it is disabled

BTW, the "default retired image" is additionally disabled as well.
```java
return security.isRetired() ? Images.SECURITY_RETIRED.image(disabled) : Images.SECURITY.image(disabled);
```

So the pull request has a valid argument: there is a missing "retired" version of a custom logo. And using the alpha for this sounds like a good compromise - and differentiates from the "disabled" (= no holdings state).

In the performance view it would look like this:


|                     | **Held** | **Not held anymore** |
|---------------------|----------|----------------------|
| **Active** | Bildschirmfoto 2026-07-03 um 09 52 41 | Bildschirmfoto 2026-07-03 um 09 58 19 |
| **Inactive** | Bildschirmfoto 2026-07-03 um 09 53 00 | Bildschirmfoto 2026-07-03 um 09 53 50 |

I like the alpha treatment. I do not think we should make the text color different. That gives to much weight and impairs readability. Plus it would need to use CSS to have theme-colored text.

Let me see if I can cherry-pick the alpha treatment from this pull request

buchen:
> I do not think we should make the text color different. That gives to much weight and impairs readability. Plus it would need to use CSS to have theme-colored text.

I realized that the AccountListView and the PortfolioListView actually does this. I have forgotten this, because I almost never display the retired accounts. I'll remove this for consistency. No other view, dialog, table was doing this.

PR #5821, opened 2026-06-30
Added support for PDF format containing a refund with a variation in the PDF

kimmerin:
added support for a variation of a PDF containing a refund as provided at the [PP-forum](https://forum.portfolio-performance.info/t/pdf-import-for-viac-wir-bank/21814/33?u=kimmerin).

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
@@ -2692,6 +2692,41 @@ public class WirBankPDFExtractorTest
     }
 
     @Test
+    public void testSteuerrueckerstattung09()
+    {
+        var extractor = new WirBankPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Steuerrueckerstattung09_English.txt"),
+                        errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(1L));
+        assertThat(countBuySell(results), is(0L));
+        assertThat(countAccountTransactions(results), is(1L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/wirbank/WirBankPDFExtractorTest.java
```diff
@@ -1935,6 +1935,40 @@ public void testDividende06()
...
+        new AssertImportActions().check(results, "CHF");
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("CH0030849613"), hasWkn(null), hasTicker(null), //
+                        hasName("UBS Canada"), //
+                        hasCurrencyCode("CHF"))));
+
+        // check dividend transaction
+        assertThat(results, hasItem(dividend( //
```
buchen: Is this a dividend or is this a tax refund?

The text in the document reads "Refund Swiss withholding tax"

kimmerin: I've assumed a dividend payment because that's what is mentioned in the dump:

````
Dividend Payment
We have booked the following payment for you:
Type of distribution: Refund Swiss withholding tax
````

I've asked in the forum for clarification.

buchen: Mmmh, I read this as. A refund of a withholding tax previously taxed on the dividend. If we record this as dividend, then we are not "cancelling" the taxes recorded earlier.

I wanted to ask in the forum, but I see you already did that. Thank you. Let's see

kimmerin: I've checked the other tax refund samples in the test and they are essentially the same (minus the "Swiss" in the text for the return). I've removed the if-block checking the type against the text in the PDF. The assign-block is only executed if the match-block matches and the regex already contains everything that is checked later on.

That duplicate checks was the reason why the importer treated this as dividend. The regex said tax-return (and made me happy that the parser stopped throwing IllegalArgumentExceptions) but the subsequent duplicate check wasn't amended by me so it never changed the type accordingly.

PR #5822, opened 2026-06-30
skip transaction if processing a second page only containing summary informations

kimmerin:
Extended PDF import to skip processing of a page if it only contains summy information and no actual transaction. Amended the parse-logic in PDFParser to not throw an IllegalArgumentException in case a skip took place.

PDF-Export provided at https://forum.portfolio-performance.info/t/pdf-import-von-raiffeisenbank-bankgruppe/12535/161?u=kimmerin

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
@@ -468,9 +468,13 @@ import name.abuchen.portfolio.model.TypedMap;
                     {
                         try
                         {
+                            var prevSkipReason = ctx.getSkipReason();
                             section.parse(filename, documentContext, lines, lineNo, lineNoEnd, ctx, target);
 
                             // if parsing was successful, then return
+
+                            if (prevSkipReason != null)
+                                ctx.skipTransaction(null); // reset skip marker
                             return;
                         }
                         catch (DuplicateSecurityException e)
@@ -484,7 +488,7 @@ import name.abuchen.portfolio.model.TypedMap;
                         }
                     }
 
-                    if (!isOptional)
+                    if (!isOptional && ctx.getSkipReason() == null)
                         throw new IllegalArgumentException(MessageFormat.format(
                                         Messages.MsgErrorNoneOfSubSectionsMatched, String.valueOf(subSections.size()),
                                         String.join(";\n", errors), lineNo + 1, //$NON-NLS-1$
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
@@ -211,7 +211,7 @@ import name.abuchen.portfolio.model.TypedMap;
         List<LineSpan> split(String[] lines);
     }
 
-    private static class PatternSplittingStrategy implements SplittingStrategy
+    public static class PatternSplittingStrategy implements SplittingStrategy
     {
         private Pattern startsWith;
         private Pattern endsWith;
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/PDFParser.java
```diff
@@ -468,9 +468,13 @@ public void parse(String filename, DocumentContext documentContext, String[] lin
                     {
                         try
                         {
+                            var prevSkipReason = ctx.getSkipReason();
                             section.parse(filename, documentContext, lines, lineNo, lineNoEnd, ctx, target);
 
                             // if parsing was successful, then return
+
+                            if (prevSkipReason != null)
```
buchen: Claude says: PDFParser skip-reset affects all extractors — latent regression risk. The new if (prevSkipReason != null) ctx.skipTransaction(null) clears a skip set by an earlier section whenever a later oneOf/optionalOneOf sub-section matches. Onvista sets a skip inside a required oneOf (Ertragsthesaurierung, ~L742) immediately followed by an optionalOneOf (~L768) — a real-world doc that skips and matches an fx alternative would be silently un-skipped and imported. Tests pass only because no such document exists in the suite. Notably, Raiffeisen's own happy path never reaches this reset branch (page 2 never matches a later oneOf), so the reset appears unnecessary for the feature.

I think that is a point. The sections can be nested. Do we want to reset here? What do you think?

kimmerin: Ah, there is a `parsedData.skipTransaction`. I've missed that and checked the references of `transactionContext.skipTransaction` and found no use in any of the parser (now I know why). That particular block was intended as some kind of insurance in case that line appears together with the actual dividend information. Since the decision to skip is done at the beginning (otherwise an exception will be thrown before we can decide to skip) that reset would have unskipped the transaction.

I've done changes that now follow a different approach without the use of the skipping mechanism (it would then show up in PP anyway, which might confuse users). This required me to "open" `PatternSplittingStrategy` in order to let the search for the parts to be parsed skip that particular block if it encouters the phrase that indicates that we're on a summary page without relevant data.

I've checked if I can do a similar approach like I've done with TradeRepublic, but this importer already uses the "startsWith/endsWith" specification and since the regex is done on the single lines instead of the whole block I can't use `(?!Steuerrelevante Daten)` to exclude text-blocks containing this text.

buchen: Before opening the PatternSplitStrategy, I wonder can't we change the pattern itself?

If we remove the "Private Banking" from the start pattern, the new test case is successful. I will comment in the relevant line. But that would be the only change needed (besides the sample file and the test case).

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/RaiffeisenBankgruppePDFExtractor.java
```diff
@@ -267,7 +271,25 @@ private void addDividendeTransaction()
 
         var pdfTransaction = new Transaction<AccountTransaction>();
 
-        var firstRelevantLine = new Block("^.*(Private Banking|Abrechnungsnr|Gesch.ftsart|Anlageverm.gen).*$", "^(Den Betrag buchen|Der Betrag wird|Dieser Beleg (tr.gt|wurde|wird)).*$");
```
buchen, line 270: Alternatively:

```java
var firstRelevantLine = new Block("^.*(Abrechnungsnr|Gesch.ftsart|Anlageverm.gen).*$", "^(Den Betrag buchen|Der Betrag wird|Dieser Beleg (tr.gt|wurde|wird)).*$");
```

would make the test successful. Do we really have generic documents that only have "Private Banking"? There is no sample.

buchen: Pushed it so you can have a look and comment: #5841

buchen:
implemented via #5841

PR #5816, opened 2026-06-26
Ledger Implementation - Corporate Actions

Morpheus1w3:
## Feature blocks in the pull request
- Adds a new Ledger subsystem as the client-owned transaction truth.
- Adds Ledger configuration, entry/posting definitions, parameter domains, validation, and structural guardrails.
- Materializes Ledger projections into the existing legacy transaction model.
- Adds XML and Protobuf persistence for Ledger truth while keeping compatibility projections derived.
- Adds compatibility creators and editor APIs for existing transaction families.
- Adds mutation, converter, deleter, repair, and owner-patch infrastructure.
- Integrates Ledger access into Client, AccountTransaction, PortfolioTransaction, and CrossEntry compatibility paths.
- Wires Ledger-backed transactions into dialogs, viewers, tables, and inline editing paths.
- Centralizes inline-edit routing and blocks unsupported structural edits on Ledger-backed projections.
- Adds importer, InvestmentPlan, and stock-split guardrails so write paths do not silently bypass Ledger truth.
- Adds a reporting/snapshot bridge so existing reporting consumers can continue to work through compatibility projections.
- Adds diagnostic codes, debug output, and NLS entries for Ledger validation and troubleshooting.
- Prepares native corporate-action modeling for SPIN_OFF, STOCK_DIVIDEND, BONUS_ISSUE, RIGHTS_DISTRIBUTION, and BOND_CONVERSION.

## Summary of the Ledger

The Ledger introduces a transaction model where one business operation is stored as one LedgerEntry with one or more LedgerPosting records. A LedgerProjectionRef connects that Ledger truth back to the existing AccountTransaction and PortfolioTransaction views.

The reason for this change is that many Portfolio Performance operations are not naturally single legacy transactions. BUY/SELL operations, transfers, deliveries, fees, taxes, investment-plan executions, repairs, and corporate actions can span account rows, portfolio rows, units, and CrossEntry links. In the old model, this creates multiple places that can appear to own the same business fact. That makes validation, persistence, migration, debugging, and future extension harder.

The Ledger does not remove the existing public transaction API. AccountTransaction, PortfolioTransaction, BuySellEntry, AccountTransferEntry, PortfolioTransferEntry, and CrossEntry remain visible because existing UI, reporting, import/export, checks, tests, and external callers still use them. For Ledger-backed transactions, those objects are compatibility projections derived from Ledger entries, not a second persisted truth.

[...]

PR #5823, opened 2026-07-01
Add local MCP server

AndreevDm:
Introduce a local MCP (Model Context Protocol) server that exposes the portfolio files currently open in the application to MCP clients. It provides tools to list open files, accounts, portfolios, securities and transactions, and to add, update and delete transactions.

Supported transaction operations cover simple account bookings (deposit, removal, interest, dividends, fees, taxes, refunds), security buy/sell, account-to-account transfers (including FX with a GROSS_VALUE forex unit), portfolio-to-portfolio security transfers and in/out deliveries. Inbound transfer legs are visible in list_transactions with counterparty and direction fields, and edits to a transfer leg (date, note, security, shares, amounts) propagate to the linked leg.

Input is validated (distinct from/to, positive amounts, required targetAmount for FX, and rejecting type changes on transfers) and the forex unit is preserved when editing fees/taxes. A preference page enables the server, configures the port and autosave, and shows the MCP client configuration.


Morpheus1w3:
Side effect to PR #5816

buchen:
Thanks @AndreevDm for the contribution. It absolutely makes sense to enable agents. I think your pull request already addresses some things: running the changes in the UI thread (to avoid concurrent updates) and enabling the server via the preferences.

I have decided to go a similar way with #5870. For now it is a REST API (although I would hope that with skills the agents can work with this or we add a MCP layer on top of that).

Therefore I decided to close this pull request.

PR #5824, opened 2026-07-01
Add Move left/right actions to reorder Dashboard views

Mikerion:
obrazek

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ConfigurationStore.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ConfigurationStore.java
@@ -131,17 +131,19 @@ public class ConfigurationStore
                                 a -> delete(config)));
 
                 int index = configSet.indexOf(config);
-                if (index > 0)
-                {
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/ConfigurationStore.java
```diff
@@ -155,6 +157,34 @@ private void createToolBarItems(ToolBarManager toolBar)
...
 
+    private void move(Configuration config, int direction, ToolBarManager toolBar)
+    {
+        int newIndex = configSet.indexOf(config) + direction;
+
+        if (newIndex < 0 || newIndex >= configSet.getConfigurations().count())
+            return;
+
+        configSet.remove(config);
+        configSet.add(newIndex, config);
```
buchen, line 168: I added a ```touch()``` to make sure the client is dirty and the change will be saved. Other than that, I merged the change.

I updated the other wording. Move left/right works on the x-axis. The previous "move to front" works on the z-axis (with does not fit here and now became apparent because the left/right was added). Thanks for the contribution.

buchen:
merged.

PR #5826, opened 2026-07-01
Make the regex to recognize more strict so that it doesn't accidentially match if the security name contains one of the phrases

kimmerin:
With the fix to handle multipage PDFs the regex for recognizing dividend-PDFs accidentially triggered if the name of the share contained one of the phrases ("dividend" in this case. I've changed the regex to be a multi-line-pattern to make sure that it only matches if the phrase is at the beginning of the line.


buchen:
Next time, please do not merge the master branch into the pull request. It makes it much more effort for me to rebase and adapt. Either rebase locally and resolve the conflicts, or I can also rebase. Thanks.

PR #5828, opened 2026-07-02
support the use of "Abwicklungspauschale" instead of "Abwicklungskostenpauschale" in the PDF to be recogized as fee.

kimmerin:
Traderepublic did it again and uses a variation of a word that the importer used to recognize fees. This change added that variation to the regex so that the fees of a buy/sell-transaction is correctly recognized again.

Dumps were provided at https://forum.portfolio-performance.info/t/pdf-import-von-trade-republic/5107/884?u=kimmerin

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
@@ -928,6 +928,40 @@ public class TradeRepublicPDFExtractorTest
     }
 
     @Test
+    public void testWertpapierKauf22()
+    {
+        var extractor = new TradeRepublicPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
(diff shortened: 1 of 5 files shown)


kimmerin:
I've also fixed the parsing of the provided PDF-dump containing a sale of a security that triggered the wrong parsing part of the importer (handling Vorabpauschalen in account statements)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/traderepublic/TradeRepublicPDFExtractorTest.java
```diff
@@ -6409,6 +6443,77 @@ public void testWertpapierVerkauf14()
...
+    @Test
+    public void testWertpapierVerkauf16()
+    {
+        var extractor = new TradeRepublicPDFExtractor(new Client());
+
+        List<Exception> errors = new ArrayList<>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "Verkauf16.txt"), errors);
+
+        errors.forEach(e -> e.printStackTrace(System.out));
```
buchen: I removed this for the merge.

if we need this often, then maybe we have an alternative for the 

```java
assertThat(errors, empty());
```

statement that prints all stack traces as part of the assertion failure.

PR #5832, opened 2026-07-04
Add IRR after tax column to securities performance view

Tanino70:
This PR adds a new "IRR (after tax)" column next to the existing IRR column
in the Securities Performance view.

Motivation:
The existing IRR (IRRCalculation) is intentionally computed on gross cash
flows: taxes withheld on dividends and paid on sales are added back to
reconstruct the value before taxes, while fees are always kept as real cash
flows. This is useful, but there is currently no way to see the return an
investor actually realized after taxes.

Implementation:
- New IRRCalculationAfterTax (package-private, extends IRRCalculation)
  that overrides the three visit() methods to use the actual net cash flows
  instead of reconstructing the gross value:
  - dividends: use the net amount as recorded, without adding tax back
  - buys/sells: use the amount actually paid/received, without adding tax back
  - standalone TAXES / TAX_REFUND transactions on the security (not already
    part of a buy/sell/dividend) are now included as real cash flows, since
    they represent an actual tax charge/refund independent of a specific
    transaction
  Fees continue to be treated exactly as in the gross calculation.
- IRRCalculation's internal dates/values lists are changed from private to
  protected so the subclass can reuse the ValuationAtStart/ValuationAtEnd
  handling without duplicating it.
- LazySecurityPerformanceRecord gets a new lazy irrAfterTax value/getter,
  following the exact same pattern as the existing irr field.
- SecuritiesPerformanceView gets a new (hidden by default) column
  "IRR (after tax)" right next to the existing IRR column.
- New unit test IRRCalculationAfterTaxTest, mirroring the existing
  IRRCalculationTest, plus a second test covering the standalone tax
  transaction case.

No existing behavior changes; the new column is opt-in (hidden by default).


Tanino70:
Hi! Just checking in on this PR — it's been a few weeks and the Maven Build workflow is still waiting for approval. 
Happy to make any changes if needed, just let me know 🙂

PR #5833, opened 2026-07-04
Add standalone IRR chart view to Performance

jmmnsantos1:
Extracts the IRR calculation from the generic configurable chart into its own dedicated and clean view in the Performance sidebar. Handles maximum and minimum bounds and preserves horizontal quartile markers for IRR independently.


cairesantos:
Thanks for working on bringing IRR into the performance views. I’ve been exploring the same gap and took a slightly different approach that may be useful to compare.

Instead of adding a separate IRR chart view, I extended the existing performance infrastructure so users can choose TTWROR or IRR in the current Performance Chart and related dashboard widgets. The implementation reuses Portfolio Performance’s existing PerformanceIndex, data-series/filtering, benchmark handling, and scalar getPerformanceIRR() logic as much as possible.

The only new calculation bridge is an IRRSeries helper for the chart, which produces the start-of-reporting-period → date IRR series needed for a timeline. Its endpoint is tested against the existing scalar IRR result.

I’ve also applied the same metric selection to:

- Performance Chart / Dashboard Performance Chart
- Top Performers
- Monthly Performance Heatmap
- Yearly Performance Heatmap

The main reason I preferred this route is to avoid introducing a second performance-chart implementation and to keep benchmarks, saved data-series configurations, portfolio/account/classification filters, aggregation, etc. within the existing chart architecture.

I noticed your PR already references IRRSeries, so there may be some overlap in the underlying idea. I’m happy to share the implementation/tests I’ve been working on and would be interested in maintainer feedback on whether IRR is better represented as a separate view or as a selectable performance metric in the existing views.

I don’t want to duplicate work without coordination, so I wanted to comment here before opening a separate draft PR.

jmmnsantos1:
I have no experience here on GitHub and with this type of information organization; I only know basic programming (like Matlab). So, could you explain to me:
- Who you are (a regular user like me or an administrator/implementer)?
- Will your solution be available to everyone?
- How can I see your solution in use?
Thank you in advance. Sorry for the basic questions, but I really don't understand how this works.
I proposed a solution and never received feedback or an implementation to use. GitHub has thousands of folders and files, where I got lost and don't understand anything. If you could clarify this for me, I would appreciate it.
The most important, your solution or me solution will be availabe on software?
Do I need to do something? If yes, please detail.

cairesantos:
@jmmnsantos1 ,

Thanks, and no problem at all.

I am also just a regular Portfolio Performance user/contributor, not an administrator or maintainer of the project.

I have now opened a separate draft pull request with the alternative implementation here: #5975.

If the Portfolio Performance maintainers review and merge that pull request, the functionality will become part of the normal Portfolio Performance application and will therefore be available to everyone in a future release.

At the moment it is still only a proposed change, so neither my implementation nor yours is part of the released software yet. The maintainers ultimately decide which approach, if any, is merged.

You do not need to do anything. Your original PR helped identify the same missing functionality, and I referenced it so the maintainers can see both approaches and decide how they want IRR to be integrated.

My main goal is the same as yours: to make IRR available properly in Portfolio Performance.

PR #5839, opened 2026-07-04
support for dividend PDF of UniCredit Bank Austria AG

kimmerin:
support for the PDF format provided by UniCredit Austria AG provied by the [PP-forum](https://forum.portfolio-performance.info/t/pdf-import-abrechnungsbeleg-unicredit-bank-austria/39482/8?u=kimmerin)


kimmerin:
@buchen something I've just recognized: `Dividende01.txt` and `Dividende02.txt" are identical, shall we remove one?

buchen:
> @buchen something I've just recognized: `Dividende01.txt` and `Dividende02.txt" are identical, shall we remove one?

Yes, let's remove this. I will add an additional commit. Thanks for the change.

Issue #5842, opened 2026-07-06
Add an annualized volatility dashboard widget


Konilo:
**Is your feature request related to a problem? Please describe.**
The "Volatility" widget scales the standard deviation to the length of the selected reporting period. It shows `sigma_daily × sqrt(N)`, where `N` is the number of trading days in the period (see `Risk.Volatility` in `Risk.java`: `sqrt(variance × count)`). So the value only matches the conventionally quoted *annualized* volatility when the selected period happens to be ~1 year; a shorter period shows a smaller number, a longer one a larger number, for the same underlying risk.

That's awkward for the dashboard specifically, because it's a monitoring view of the portfolio you actually hold, watched across whatever period you pick (e.g., YTD, since inception). The volatility figure therefore isn't stable across those periods, and isn't directly comparable to the annualized volatilities that fund fact sheets, indices, and brokers all quote. To benchmark your own portfolio against the outside world you have to re-annualize by hand.

Example from my own file: period May 8, 2024 to today (~2.16 years) shows 21.69%. The annualized equivalent is ~14.8% (`21.69% / sqrt(2.16)`). At a glance, 21.69% is easily misread as an annual figure (speaking for myself, at least!).

I saw that this came up in the forum (https://forum.portfolio-performance.info/t/rendite-volatilitaet-grafik/412), where annualizing volatility for comparability was requested and left open rather than declined.

**Describe the solution you'd like**
A new, distinct dashboard widget (e.g., `ANNUALIZED_VOLATILITY`) that always normalizes to one year: `sigma_daily × sqrt(252)`, independent of the selected period. Weekends and holidays are already excluded (`filterReturnsForVolatilityCalculation` in `PerformanceIndex`), so trading-day annualization with `sqrt(252)` looks like the right convention.

The current "Volatility" widget stays untouched; this just adds a sibling (mirroring how `TTWROR_ANNUALIZED` sits alongside the cumulative TTWROR widget). The annualization logic would live in a new method on `Risk.Volatility` so it can be unit-tested, with a matching case added to `RiskTest`.

Question for maintainers before I open a PR:

Would you prefer
1. a **separate widget** (as proposed above), or
2. an **optional toggle** on the existing widget (closer to the compromise floated in the forum thread, and to the opt-in `p.a.` preference from #3036)?

Happy to open a PR once we get aligned on the approach.

**Describe alternatives you've considered**

[...]

PR #5846, opened 2026-07-08
Add annualized volatility dashboard widget

Konilo:
Adds a "Volatility (annualized)" dashboard widget alongside the existing "Volatility" widget.

## Why

The existing "Volatility" widget scales the standard deviation to the length of the selected reporting period (`sigma_daily * sqrt(N)`, where `N` is the number of trading days in the period). The displayed value therefore depends on the period length and only equals the conventionally quoted *annualized* volatility when the period happens to be about one year. Since the dashboard is a monitoring view watched across arbitrary periods (e.g., YTD, since inception), the figure isn't stable across periods and isn't directly comparable to the annualized volatilities that fund fact sheets, indices, and brokers quote.

## What

A new additive `ANNUALIZED_VOLATILITY` widget that always normalizes to one year (`sigma_daily * sqrt(252)`), independent of the selected period. Weekends and holidays are already excluded from the returns (`filterReturnsForVolatilityCalculation` in `PerformanceIndex`), so trading-day annualization with `sqrt(252)` is the appropriate convention.

The existing "Volatility" widget is left unchanged. This mirrors how "True Time-Weighted Rate of Return (annualized)" sits alongside its cumulative counterpart.

- `Risk.Volatility` gains `getAnnualizedStandardDeviation()`, covered by a new `RiskTest` case.
- New `ANNUALIZED_VOLATILITY` entry in `WidgetFactory` mirroring `VOLATILITY`.
- New labels `LabelVolatilityAnnualized` / `TooltipVolatilityAnnualized`.

This is the first of two PRs discussed in #5842 (dashboard widget); a follow-up will add the matching annualized options to the Return/Volatility chart.

Part of #5842

Local run:
image

Hovering over the new widget
image

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/math/RiskTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/math/RiskTest.java
@@ -122,6 +122,22 @@ public class RiskTest
     }
 
     /**
+     * Tests the annualized volatility calculation with a set of returns.
+     * Scenario: The same return values as {@link #testVolatility()} are used.
+     * Expects: The annualized standard deviation equals the daily standard
+     * deviation scaled by the square root of 252 trading days, i.e. it is
+     * independent of the number of observations in the period.
+     */
+    @Test
+    public void testAnnualizedVolatility()
+    {
+        double[] delta = { 0.005, -1 / 300d, -0.005, 0.01, 0.01, -0.005 };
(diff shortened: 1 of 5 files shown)


Konilo:
Gentle bump 🙂

This implements PR1 (the dashboard widget) from #5842, as discussed with @Sn1kk3r5. CI is green and it's ready whenever you have a chance to review it. Happy to adjust anything, of course.

PR #5849, opened 2026-07-09
Resolve PortfolioPerformance.ini for scaling preferences

Morpheus1w3:
Resolve scaling preference INI access through the fixed PortfolioPerformance.ini file name next to the launcher, and make preference-page loading tolerate a missing INI in development launches.

This avoids resolving eclipse.ini from PDE bundle pools and keeps both SWT auto-scale settings on the shared generic VM-property manipulation path.

Unrelated preference behavior, product branding, and other INI handling were intentionally left unchanged.

https://forum.portfolio-performance.info/t/update-auf-version-0-85-0-vom-7-juli-2026/39491/

image


buchen:
Was the resolution of the PortfolioPerformance.ini not working before?

I also wonder: are the reports in the forum really scaling related - or are they CSS related? Bigger spaces in rows while having the same size of the font seems to sound more like a styling question.

Morpheus1w3:
It's a side effect while recently the Eclipse framework was updated. SWT enabled UI scalling now as default.  And if the UI scalling factor is 150% in general, forum user Missing app rescalling to previous PP scalling.

buchen:
Reading https://eclipse.dev/eclipse/markdown/?f=news/4.39/platform_isv.md, I thinking: 

Setting the property ```swt.autoScale.updateOnRuntime=true``` seems to be obsolete. It defaults to true with the update to Eclipse 4.4. I do not think we should offer an option to set it to false in the app. PP will not be able support a separate branch and I would assume the flag will also go away in the on the next updates.

Then about offering options for scaling: Let's see what the user reports in the forum when changing the parameters. Then we can decide to offer this as experimental option.

> Was the resolution of the PortfolioPerformance.ini not working before?

What about this question. It is still not clear to me why the reading of the file was refactored. Was it Codex also doing this unrelated change or was there a functional purpose to this?

buchen:
@stoeggich because you reacted to this - do you have a need for this feature? And if so, for what?

I am hesitant to add this as an official - albeit experimental - option. Why? First, I expect SWT to remove this over time. And then we easily create a configuration in the INI that prevents the start of PP. That is fine if somebody changed this manually, but not so much if we change it. Second, while there was some feedback in the forum, it wasn't that much after all. It sounds more like a specialized problem. And third, PP looks better now - or rather how it looks on other platforms as well.

The counter argument is: with the font size, one cannot get the icons smaller. And font size is an option, so why not add one here as well.

stoeggich:
I missed the notification, sorry. Since the last update, or the one before that, the font size has decreased; it was better before. I'm using 4K on Windows, so I thought this would be a good fix and a way to adjust the scaling individually, like some programs do, instead of using the system's default setting. @buchen @Morpheus1w3

PR #5830, opened 2026-07-03
Spin-off (corporate action) support — proof of concept

buchen:
Adds first-class spin-off support: one event decomposed into up to four legs in a single portfolio (source basis reduction, target receipt, fractional sale, cash-in-lieu), on a new N-ary CorporateActionEntry.

Model & calc
- New DISTRIBUTION_OUTBOUND/INBOUND transaction types. Cost basis is derived from a persisted basis ratio (never a stored amount) and is conserved by construction; spinco basis = ratio × parent basis at ex-date (cross-security derivation).
- CorporateActionEntry (implements CrossEntry + a Client registry) holds the legs and event properties; group-aware N-ary delete/edit.

Persistence
- Protobuf (new PCorporateAction message + additive wire types; legs serialised bare) and XStream. CURRENT_VERSION 70→71 so an older app cleanly refuses spin-off files.

UI
- CorporateActionDialog + SpinOffModel (create/edit/delete). Distribution ratio seeds an editable entitlement from holdings at ex-date; inline FMV→basis-ratio helper; multi-currency forex; live conservation summary; non-gating entry-time warning.

Integrity
- SpinOffConsistencyCheck (checks/impl) catches externally-broken groups on load with an offer-to-fix; entry-time validation blocks currency / ratio / exchange-rate errors.

Key decisions
- Shares are a persisted fact; basis is derived. Fractional remainder is always floored to cash. Events are per-portfolio. A single stored event rate drives both the market-value forex and the derived-basis conversion.

Deferred follow-ups: chained/cyclic spin-offs, third-currency cash accounts (currently blocked), plus assorted polish/coverage.

Issue #5466, opened 2026-02-15
Support for Corporate Actions (Spin-Off, Stock Dividend, Bonus Issue, …)


buchen:
PP currently lacks dedicated transaction types for corporate actions. Users work around this by (mis)using dividends or manual deliveries, which leads to incorrect performance calculations and confusing reporting.

### Background

Discussion in #3610 surfaced the need for a broader approach. Rather than adding one-off types, we should consider the full landscape of corporate actions:

```
Corporate Action
 ├─ Cash Distribution
 │   └─ Dividend (already exists)
 └─ Securities Distribution
     ├─ Stock Dividend
     ├─ Spin-Off
     ├─ Bonus Issue
     └─ Rights Distribution
```
Related: #5453 discusses bond/percentage-quoted instruments, which bring their own set of corporate actions (coupon payments, redemptions, calls, conversions, etc.).

```
Fixed Income Corporate Action
 ├─ Income
 │   ├─ Coupon Payment
 │   ├─ PIK Interest
 │   └─ Defaulted Interest
 ├─ Redemption
 │   ├─ Maturity
 │   ├─ Partial Redemption
 │   ├─ Call
 │   └─ Put
 └─ Structural / Credit
     ├─ Conversion
     ├─ Exchange
     ├─ Restructuring
     └─ Default
```

### Key Design Considerations

* Single type vs. multiple types: One generic "Corporate Action" type with subtype-specific data, or dedicated types per action?
* Two-instrument transactions: A spin-off touches both a source and a target security. This is similar to how purchases/sales/transfers use an Entry that manages two transactions. This pattern could apply here to keep FIFO and other per-instrument lookups efficient.
* Cash flow: Some corporate actions (e.g. rights issues) involve a cash component and therefore also touch an account.
* Reporting columns: Rather than a column per action type, the proposal is:
  * Keep "Sum of Dividends" for cash dividends only
  * Add "Securities Distribution" for non-cash corporate actions
  * Add "Total Distribution" combining both
* Performance calculation: ClientIndex, ClientFilter, and related classes will need updates.

### Related

• #3610 — Original spin-off PR and discussion
• #5453 — Bond support / percentage-quoted instruments

Sn1kk3r5:
@buchen 

Probably one mitigation step I discovered while playing around with ppxml2db.
[...]

PR #5856, opened 2026-07-11
Add tax-neutral transfers between mutual funds

PabloReyes:
## Summary

Add support for tax-neutral transfers between mutual funds, including Spanish fund transfers (`traspasos`).

The new **Fund transfer** workflow records linked outbound and inbound fund transactions while carrying the original acquisition lots into the target fund. This preserves acquisition dates and cost basis so that gains remain deferred until the target holding is eventually sold.

## Motivation

Spanish investors can transfer eligible mutual-fund holdings from one fund to another without realizing taxable gains at the transfer date.

Previously, these operations required delivery transactions or sell-and-buy pairs. Those workarounds either lose acquisition-lot continuity or incorrectly represent the transfer as a taxable disposal and new investment.

Related discussions:

- https://forum.portfolio-performance.info/t/transfer-fund-from-another-entity-and-subscribe-a-new-one/17640/8
- https://forum.portfolio-performance.info/t/transfer-between-funds-without-taxes/13455
- https://forum.portfolio-performance.info/t/mutual-fund-transfer/19799

Closes #5612

### Domain model and persistence

- Add `FundTransferEntry` with linked source and target portfolio transactions.
- Add carried acquisition lots containing:
  - original acquisition date
  - source shares
  - target shares
  - carried purchase value
  - optional source transaction reference
- Validate currencies, positive values, share totals, available source holdings, and distinct source and target securities.
- Persist fund transfers and carried lots in XML/protobuf.
- Include fund-transfer data in JSON exports.

### Calculations and reporting

- Remove transferred shares from the source fund and add the converted shares to the target fund.
- Preserve FIFO and moving-average purchase values across the transfer.
- Defer capital gains until shares of the target fund are sold.
- Carry acquisition lots into trade collection and security-performance calculations.
- Treat complete fund transfers as internal operations for client-level TTWROR and IRR:
  - no deposit or removal is reported
  - the transfer does not change full-client performance
- When a filtered view contains only one side, retain the fund-transfer transaction and its acquisition lots while treating the visible side as an external inbound or outbound flow for scoped performance reporting.
- Preserve fund transfers across portfolio, security, classification, and tax-filtered clients, including weighted classifications.

### User interface

- Add a dedicated **Fund transfer** action and dialog.
- Allow selection of source and target funds and securities accounts.
- Calculate source shares/value and target shares/value.
- Display the acquisition lots carried into the target fund.
- Allow source shares, target shares, and purchase values to be adjusted for broker-specific allocations.
- Explain that generated transactions use the transfer execution date while tax calculations retain the original acquisition dates.
- Add the new labels and validation messages to all managed translations.

## Scope

This PR introduces a new transaction type, so the implementation necessarily spans the domain model, persistence, calculations, filters, performance reporting, trade collection, and UI.

Covering these layers together ensures that a fund transfer behaves consistently after saving and reopening a portfolio, exporting data, applying filters, and calculating performance or capital gains.

## Testing

The standard Java 21 Maven/Tycho build passes:

```text
mvn -f portfolio-app/pom.xml clean verify

Core tests: 4,495 passed, 1 skipped
UI tests:     300 passed
Checkstyle:   0 violations
Build:        SUCCESS
```

Regression coverage includes:

- fund-transfer model invariants and cross-entry behavior
- XML/protobuf round trips and JSON export
- acquisition-lot construction, including multiple and same-day lots
- FIFO and moving-average cost basis
- deferred capital gains after a later target-fund sale
- trade collection
- full-client and scoped TTWROR/IRR behavior
- portfolio, security, classification, and tax filters
- dialog validation and editable carried lots
- managed translation bundles

image


buchen:
Hi @PabloReyes,

thanks for the contribution. I believe this discussion is very similar to the discussion as part of the corporate actions #5466 where there are two proposals at the moment - either introducing a more general ledger with #5816 or another proposals for spin-off which is extending CrossEntry with #5830.

To be honest, I have not decided which way to go. But if we are adding more CrossEntries, then I do not think it makes sense to add all the attributes specifically to the transaction (looking here at the decision of the PR where to extend the protobuf serialization).

Couple of questions to better understand:

How are the lots determined? Is it really a transfer of a specific purchase only and not the whole position held in the account? And is the user expected to run the dialog then per lot? (Haven't run it locally yet)

Do you happen to still have the plan markdown document against which this was developed?

Do you have a link to some official documentation how the fund transfers work? Maybe an ISO document or something similar?

PabloReyes:
Hi @buchen ,

Thanks for your review. I see how this overlaps with the other discussions you mentioned. 

For now, I used a dedicated `FundTransferEntry` because the operation connects two different funds and needs to keep the original purchase information. For protobuf, I followed the existing transfer approach: save one side and rebuild the other when loading.

I understand the complexity of adding fund-specific fields directly to the transaction... I am happy to change this once with your guidelines. 

Regarding the lots: the user runs the dialog only once for each transfer.

- The transfer can include the whole position or only part of it.
- The application automatically selects the oldest available purchases first using FIFO.
- If several purchases are involved, they appear as rows in the same dialog.
- Their original purchase dates and values are moved to the new fund.
- The user can adjust the rows if the broker provides a different breakdown, but the totals must still match the transfer.

About the plan, I started with a plan but ended up making some modifications after it. I'm attaching the design document and implementation plan. 
[fund-transfer-implementation-plan.md](https://github.com/user-attachments/files/30001356/fund-transfer-implementation-plan.md)
[fund-transfer-design.md](https://github.com/user-attachments/files/30001358/fund-transfer-design.md)

The implementation is based on Spanish regulation. I admit I did not look at other cases, since most of the questions I've seen on this matter are related to Spain. 

Regulation sources:

- [Spanish Personal Income Tax Law 35/2006, Article 94](https://www.boe.es/buscar/act.php?id=BOE-A-2006-20764#a94) states that qualifying transfers do not realize the gain or loss and that the new participations retain the acquisition value and date of the transferred participations.
- The [CNMV guide on transfers between investment funds](https://internet.cnmv.es/DocPortal/Publicaciones/Fichas/GR15_Traspasos_entre_fondos.pdf) describes both total and partial transfers.
- The [AEAT 2025 income-tax manual](https://sede.agenciatributaria.gob.es/Sede/ayuda/manuales-videos-folletos/manuales-practicos/irpf-2025/c11-ganancias-perdidas-patrimoniales/determinacion-importe-ganancias-perdidas-patrimon-especificas/transmisiones-reembolsos-acciones-iic/regimen-especial-diferimiento-tributacion.html) includes an example where one transfer carries several original acquisition lots with different dates and values.

This PR is still a draft and I need to make some changes still (hardening mostly), but I appreciate all feedback. 

Thanks

Rich28:
What you've outlined about 'Traspasos' @PabloReyes seems similar to a fund share class conversion which occasionally can occur in the UK market.  A fund share class conversion is when your investment stays in the same underlying fund, but the original share class that you acquired - which has a unique ISIN and NAV value -  is changed by its asset manager to another share class within the same fund - that has a different ISIN and NAV value .  

I had felt that PP could cater for such for a situation with the user creating an outbound delivery transaction relating to the discounted fund share class; and an inbound transaction, relating to its replacement class - and importantly featuring the original acquisition date in the underlying fund (i.e. the purchase date in the discounted share class holding).

Also,  I would like to suggest that thought is given to Middlefield Canadian Income Trust, which was an investment trust (so not a mutual fund) that has recently undergone a transition to the actively managed Middlefield Canadian Enhanced Income UCITS ETF.  Hence, I'm somewhat unsure of the appropriateness of specifically referring to this as 'Fund transfer'

Having said all this many thanks for bringing this forward!

PabloReyes:
Hi @buchen and team,

I came back to this and saw #5830 getting some traction. 

I do not think a fund transfer is a corporate action. A corporate action is initiated by the issuer, while a fund transfer is initiated by the investor. They share some technical requirements (linked transactions involving different securities) but they represent different things.

I would keep `FundTransferEntry` as a separate concept. The persistence approach in #5830 may still be relevant though: the cross-entry data could be stored in a dedicated top-level protobuf message instead of adding the carried-lot fields directly to `PTransaction`. What do you think? Can we keep the current model and UI but introduce something like `PFundTransfer` for persistence? Or do you think a shared persistence mechanism for different CrossEntry implementations would be useful for other cases?

Happy to get back to it and check against the current master once you provide some feedback. 

Thanks!

PR #5862, opened 2026-07-11
Stock split wizard: show a live example to clarify ratio direction

digitalerdude:
## Summary
The stock split wizard's "Split [new] for [old]" input fields are only explained by the static description text at the top of the wizard page (small, easy to skim past). The fields themselves are just labeled "Stücke (neu)" / "Stücke (alt)" with no context at the point of input, which has led to users entering the ratio backwards.

This adds a label below the ratio fields that updates live as the user types, spelling out both numbers in a full example (e.g. "Example: 1 old share(s) become 2 new share(s)"), so the direction is unambiguous at the moment of input rather than only in a description that can be missed.

## Test plan
- [x] Full `mvn verify -Plocal-dev` build: `name.abuchen.portfolio.ui` compiles, `name.abuchen.portfolio.ui.tests` passes
- [ ] Manual visual check of the wizard dialog — I don't have a way to launch/screenshot the Eclipse RCP desktop app in my environment, so this is compile + existing-test verified only, not eyeballed. Flagging that explicitly per the project's own verification guidance.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1443,6 +1443,7 @@ public class Messages extends NLS
     public static String SplitWizardDefinitionDescription;
     public static String SplitWizardDefinitionTitle;
     public static String SplitWizardErrorNewAndOldMustNotBeEqual;
+    public static String SplitWizardHintRatio;
     public static String SplitWizardLabelNewForOld;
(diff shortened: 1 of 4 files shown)


mierin12:
>  Manual visual check of the wizard dialog — I don't have a way to launch/screenshot the Eclipse RCP desktop app in my environment, so this is compile + existing-test verified only, not eyeballed. Flagging that explicitly per the project's own verification guidance.

It looks like this
Capture d’écran 2026-07-12 à 11 38 17

PR #5869, opened 2026-07-13
Time Machine: allow typing the date in the date selection dialog

huntingthesun:
The Time Machine's "Pick other date" dialog only offered a calendar grid, so selecting a date a few years back required many clicks through the months. This change adds a typeable date field above the calendar: the date — including the year — can now be entered directly, and the field and the calendar stay in sync in both directions.

The field reuses the existing cross-platform `DatePicker` widget (native `DateTime` on macOS/Windows, `CDateTime` on Linux), created without its drop-down calendar button since the dialog already shows a calendar. `DatePicker` gains a field-only constructor variant and an `addSelectionListener` method; the one-argument constructor is unchanged, so all existing `DatePicker` and `DateSelectionDialog` callers keep their current behavior.

PP datepicker with manual override

Issue: https://forum.portfolio-performance.info/t/direkte-datumseingabe-in-portfolio-time-machine/38803


buchen:
Thanks for the contribution. I have rebased and merged the change. I did some minor editing for (my perceived) future readability.

PR #5870, opened 2026-07-13
Local REST API for scripts and agents (first increment)

buchen:
Adds a loopback HTTP/JSON API to the running application so scripts and AI agents can inspect and modify the data of open portfolio files. This first increment establishes the plumbing — server, authentication, per-file authorization, request pipeline — and exposes a deliberately narrow set of endpoints on top of it.

What it does

- New plugin name.abuchen.portfolio.rest hosts the server. It depends only on the model plugin; the UI plugin contributes the open-file registry and the UI-thread executor via an SPI (HostApplication) and starts/stops the server.
- Server binds 127.0.0.1 on a fixed port (default 5712), off by default, enabled in the preferences.
- Endpoints (/v1): list API-enabled files; read instruments, accounts, and portfolios; update (PATCH) and delete an instrument. Every request is executed on the UI thread. Writes change the in-memory model and mark the file dirty — exactly like an edit in the UI. The API cannot save; only the user does.
- Errors follow RFC 9457 problem+json; validation failures report all violations at once.

Security model

- Every request needs a bearer token (shown in the preferences, revocable by regenerating). Requests must address the API as loopback (Host header) and must not carry a browser Origin header, so a website cannot reach the API through the user's browser — not even by pointing its own domain at 127.0.0.1 (DNS rebinding).
- Access is granted per file: the user enables individual files in the preferences. Files that are not enabled are indistinguishable from files that do not exist.
- Writes are rejected with 423 Locked while a modal dialog is open or the user is editing a table cell, so an API call cannot be silently overwritten by stale UI state.
- File identity is stored in the workspace (keyed by path), never inside the portfolio file — copies of a file therefore never inherit API access.


georgemac-labs:
@buchen thanks for this.

I actually prefer your https://github.com/portfolio-performance/portfolio/pull/5550 approach, because it is headless and security is handled by the normal OS mechanisms. Switching to REST means new mechanisms for authentication and authorisation. Furthermore, you have to ensure a GUI app is running and the right file opened, which I find to be an impractical paradigm for either automation or a network server. I appreciate that adding remote access could be a plus for some use-cases, but I would have thought most users would use such an interface on a single-user basis, with everything running locally?

However, this is just my appraisal – apologies if I have overlooked some benefits or misjudged the use-cases.

And I would welcome getting _any_ solution into master that enables AI access to my PP data  :)   Please let me know if I can assist in any way.

I have added a few links to try to consolidate info on this topic in https://github.com/portfolio-performance/portfolio/issues/5547#issuecomment-5083635131

For me, access to performance values like IRR is essential, which is why none of the attempts so far scratch the itch. And obviously we don't want to start reimplementing the calculation engine, thus I think building the solution within PP is the right route.

Would you consider merging https://github.com/portfolio-performance/portfolio/pull/5550 independently of this effort? I'm happy to do rebase, any improvements you want, and add access to performance data.

buchen:
@georgemac-labs thanks for your thoughts

> Furthermore, you have to ensure a GUI app is running and the right file opened, which I find to be an impractical paradigm for either automation or a network server.

Fair point. My thinking (and the use case I have in mind) is that I am locally running PP and *additionally* want to interact with the agent. For this, I need an API. The CLI has the problem, that I cannot have the file open at the same time - the risk for dual edits is too high and there is no conflict resolution (PP does not detect if the file was edited via some other app nor does it have a way to resolve conflicts if two parties modified the file).

Having said this, I think one could expose the same API - the one that is exposed via REST - also via CLI. The current API surface is still very limited because I am considering every interface carefully to avoid exposing to many idiosyncrasies that have been added to PP over the time.

buchen:
Here some hints if somebody wants to give the branch a test drive.

Run to install the [skills](https://github.com/portfolio-performance/skills) - preferable into a new local project
```
npx skills@latest add portfolio-performance/skills
```

Then ask some questions. The agents can resolve against the [openapi.yaml](https://github.com/buchen/portfolio-work/blob/5f786455c146cc40a55c6bf32dd2f3a8bd9dd8e1/name.abuchen.portfolio.rest/openapi.yaml) definition. Available are
* instruments (securities) with attributes - one can update properties and attributes
* cash accounts
* investment accounts
* holdings (basic information about the currently held instruments)
* performance (IRR, TTWROR for a given period).

The agent will attempt to get permission. You have to enable a file via the preferences first.

Bildschirmfoto 2026-07-23 um 10 21 12

Then:
Bildschirmfoto 2026-07-23 um 08 25 24

Bildschirmfoto 2026-07-23 um 08 25 37

Bildschirmfoto 2026-07-23 um 08 25 57

georgemac-labs:
@buchen yep – all depends on how you envisage using this  :)

Regarding parallelism – and this is again just my personal requirements – I only need read access, so that AI can review my positions and discuss with me. If I think ahead, down the line I'm sure I could make use of a write interface, but a read-only CLI would already be great.

But ultimately, I would use REST, CLI, MCP... ANY solution that provides access to calculated metrics! That's the part that's missing from the CLI prototype and the various side-projects. I am pondering building something myself using the Java classes, but obviously there are some downsides to that, so I wanted to try one more time to advocate (and assist if possible) to get a solution into master.

I prefer not to spend time going down dead ends, but if you think you'll ultimately merge this in some form, I'd be happy to test it.

auchri:
@buchen @georgemac-labs Wouldn't it be possible to implement a parameter to start Portfolio Performance with a given portfolio file, without the gui, but with REST-Api enabled? The cli-tool then could just call the Rest-API methods without a need to reimplemented the whole locking mechanism

buchen:
> Wouldn't it be possible to implement a parameter to start Portfolio Performance with a given portfolio file, without the gui, but with REST-Api enabled?

I know that Eclipse is locking on the workspace, i.e., one can only start one instance of PP. I have to check if the CLI mechanism is using the same lock - the there is a low risk of editing.

I think @georgemac-labs would point out that the CLI could be read-only - that avoids the whole problem domain of parallel edits. That could be a reasonable starting point.

buchen:
Bildschirmfoto 2026-07-27 um 20 09 15

Bildschirmfoto 2026-07-27 um 21 00 59

buchen:
Bildschirmfoto 2026-07-29 um 09 20 44


Bildschirmfoto 2026-07-29 um 09 57 36

manueldeprada:
I heavily extended this API and made it general for a personal side project that builds on it. I needed the API to transmit transactions, heavily extended attributes for holdings, and most of the data actually displayed in Portfolio Performance's UI.

The branch is https://github.com/manueldeprada/portfolio/tree/feature/rest-api-transactions-and-holdings-metrics

You can see the individual commits here: https://github.com/portfolio-performance/portfolio/compare/feature/rest-api...manueldeprada:portfolio:feature/rest-api-transactions-and-holdings-metrics

Just letting you know in case you want a subset of the commits in a PR :) It is mostly vibecoded and quite large, so all of them might not be useful for your agentic purposes.

EDIT: other than expanding the API, I have also added [caching and pagination](https://github.com/manueldeprada/portfolio/compare/feature/rest-api-transactions-and-holdings-metrics...manueldeprada:portfolio:feature/rest-api-caching-and-pagination?expand=1) and [headless mode](https://github.com/manueldeprada/portfolio/compare/feature/rest-api-caching-and-pagination...manueldeprada:portfolio:feature/rest-api-headless?expand=1) on top of each other

Issue #5873, opened 2026-07-14
Ledger representation for the local REST API


buchen:
# Ledger for the local REST API — spec

Lifts the ledger model, the legacy→ledger migration, and the ledger→legacy projection out of
the proof of concept (#5816) on top of the local REST API for scripts and agents (#5870), and
serves them under `/v1/files/{file}/ledger`.

The endpoint is the proof point for the model. It is deliberately not the beginning of a
migration: nothing about the ledger is persisted, and the legacy transactions stay the truth.

## Goal

* A ledger model that covers exactly today's transaction types.
* A migration that turns an open file's transactions into a ledger graph, in memory, on demand.
* A projection that turns a ledger entry into real legacy transactions and inserts them into the file.
* A REST surface that exercises both directions against messy real-world files, so the model
  meets data nobody thought of before it is cast into a persistence format.

## Non-goals

* **No persistence.** No `Ledger` field on `Client`, no protobuf, no XML, no `ClientFactory`
  change. The graph is built per request and thrown away.
* **No corporate actions.** No entry type, posting type, parameter, leg, or rule that exists only
  to describe them. They come back as an additive change when the model has been validated.
* **No UI.** No editors, converters, inline-editing policy, or inspector model.
* **No investment-plan integration.** Plan-generated transactions migrate like any other
  transaction; the plan keeps its legacy back-references.
* **No i18n.** Diagnostics are stable codes plus English detail text, read by scripts and agents.

## Decisions

| # | Decision | Rationale |
|---|---|---|
| D1 | The ledger is transient: derived per request, discarded after. | `Client` is untouched, so there is no cache to invalidate and no persistence risk. |
| D2 | The write path reuses `DerivedProjectionDescriptorService` and adds a materializer that **copies** descriptor values into plain legacy objects. | Read and write share one semantic vocabulary, and round-trip becomes a testable property. The PoC's `LedgerBacked*` classes delegate to a live ledger and cannot survive save/load — they are not lifted. |
| D3 | Migration is strict: any unmappable transaction fails the request — but the migrator collects **all** failures first and reports them together. | The target is total coverage. A single round-trip must tell the user (and us) everything in their file that surprised the model. Matches the API's existing "report all violations at once" convention. |
[...]

PR #5874, opened 2026-07-14
Ledger model for the local REST API (step 1)

buchen:
Lifts the ledger core model onto the REST API branch: entries, postings, typed parameters, the type vocabulary, and the structural validator that checks an entry against the semantic shape its type requires. Persistence, migration, and projection are out of scope and follow in steps 2 to 4.

The semantic shape is data in LedgerRoleRules rather than a hard-coded switch in the validator, and a finding carries a stable LedgerIssueCode that the REST API will expose as its error code.

Issue: #5873


name.abuchen.portfolio/src/name/abuchen/portfolio/model/ledger/configuration/LedgerRoleRules.java
```diff
@@ -0,0 +1,193 @@
...
+
+    private static final Map<LedgerPostingType, UnitRule> UNIT_RULES = new EnumMap<>(LedgerPostingType.class);
+
+    static
+    {
+        cashOnly(LedgerEntryType.DEPOSIT);
+        cashOnly(LedgerEntryType.REMOVAL);
+        cashOnly(LedgerEntryType.INTEREST);
+        cashOnly(LedgerEntryType.INTEREST_CHARGE);
+        cashOnly(LedgerEntryType.DIVIDENDS);
```
buchen, line 79: A dividend has no legal carrier for its instrument and shares.

cashOnly(DIVIDENDS) gives a dividend a single primary posting of type CASH, and LedgerPostingType.CASH declares itself neither SECURITY_BEARING nor SHARES_MEANINGFUL. Yet the validator demands an instrument somewhere on the entry (DIVIDEND_SECURITY_REQUIRED), and the EX_DATE parameter rule presupposes that the instrument sits on a posting (posting.getSecurity() != null). Today the tests satisfy this by hanging a Security on the CASH posting — a shape that contradicts that type's own characteristics. The alternative, an extra SECURITY-typed posting, is no better: shares on a SECURITY posting denote a movement of shares, and a dividend moves none. So it isn't that only one shape works — no shape is right.

Nothing breaks today, because the Characteristic vocabulary is almost entirely decorative: only CURRENCY_REQUIRED and SECURITY_REQUIRED are ever read (requiresCurrency(), requiresSecurity()). isSecurityBearing(), isSharesMeaningful(), isAccountReferenceMeaningful(), isPortfolioReferenceMeaningful() and isForexMeaningful() have no callers at all. A posting type can declare that it does not bear an instrument and then bear one, unchallenged.

It bites in step 2, which has to migrate PP's dividends: those carry an instrument and a share count as context on the transaction, with nothing changing hands. The ledger has no entry-level home for either, and no posting type whose characteristics permit an instrument-with-shares that is not a movement.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ledger/LedgerStructuralValidator.java
```diff
@@ -0,0 +1,517 @@
...
+                            "Posting exchange rate must be positive for posting " + posting.getUUID(), entry, posting));
+    }
+
+    private static void validateSemanticShape(LedgerEntry entry, List<ValidationIssue> issues)
+    {
+        if (entry.getType() == null)
+            return;
+
+        for (var rule : LedgerRoleRules.primaryRules(entry.getType()))
+            requirePrimary(entry, rule, issues);
```
buchen, line 118: The validator does not enforce UNIT_RULES.

LedgerRoleRules carries two tables. PRIMARY_RULES says which primary postings an entry type requires, and the structural validator checks entries against it. UNIT_RULES says how a fee, tax, or gross-value posting derives its semantic role and unit role from its posting type — and the validator never consults it. LedgerRoleRules.unitRule() is implemented and unit-tested, but no caller exists.

The consequence is the non-primary counterpart of the matchesPrimary bug fixed earlier: a posting typed FEE may claim unitRole=TAX, semanticRole=TAX and pass validation, because only primaries are matched against the table. Nothing cross-checks a non-primary posting's declared roles against the roles its posting type implies.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ledger/configuration/LedgerPostingType.java
```diff
@@ -0,0 +1,109 @@
...
+                    Characteristic.SECURITY_REQUIRED, Characteristic.SHARES_MEANINGFUL,
+                    Characteristic.PORTFOLIO_REFERENCE_MEANINGFUL, Characteristic.FOREX_MEANINGFUL), //
+    FEE(Characteristic.MONEY_BEARING, Characteristic.CURRENCY_REQUIRED, Characteristic.FOREX_MEANINGFUL), //
+    TAX(Characteristic.MONEY_BEARING, Characteristic.CURRENCY_REQUIRED, Characteristic.FOREX_MEANINGFUL), //
+    GROSS_VALUE(Characteristic.MONEY_BEARING, Characteristic.CURRENCY_REQUIRED, Characteristic.FOREX_MEANINGFUL);
+
+    public enum Characteristic
+    {
+        MONEY_BEARING, //
+        CURRENCY_REQUIRED, //
```
buchen, line 29: At least today, there is always a currency needed. What is the use case of not requiring a currency?

name.abuchen.portfolio/src/name/abuchen/portfolio/model/ledger/configuration/LedgerEntryType.java
```diff
@@ -0,0 +1,44 @@
...
+ * The type of a ledger entry, for example a buy, a dividend, or a cash
+ * transfer.
+ * <p>
+ * Every type has a stable code. Codes are part of the REST API: they must never
+ * be changed or reused for a different meaning.
+ */
+public enum LedgerEntryType
+{
+    DEPOSIT, //
+    REMOVAL, //
```
buchen, line 13: Shall we rename to "withdrawal". The name was wrong all along but could not be changed anymore. Now there is a chance.

buchen:
@Morpheus1w3 I lifted out the model as step 1 of the #5873. I started to drill down, but I am still in the process of understanding the semantics and the guarantees that the semantics make (or do not make). I plan to add more comments here - we can use the comments threads to discuss.

buchen:
@Morpheus1w3 here a list of bigger changes I applied on the model compared to the "full" pull request. Please comment if you see this differently.

---

On removing ```LedgerDiagnosticCode```: The PoC's numeric codes (LEDGER-STRUCT-007) were a hand-maintained registry: 286 entries and a lookup table needed to make sense of any one of them. Lifting them onto this branch would have meant carrying that bookkeeping forever for the 22 codes we actually use, and it left us with two stable identifiers for every single condition — LEDGER_STRUCT_007 and IssueCode.ENTRY_TYPE_REQUIRED — with nothing saying which one is the contract a client should rely on. We collapsed them into one. The surviving identifier is the symbolic one (entry-type-required), because it is self-describing, needs no registry to allocate, and matches the error shape the REST layer already publishes: RFC 9457 problem+json, where the code sits in its own field of a {field, code, message} error rather than being prefixed into the message text. The four LEDGER-CORE-* codes are gone too: they marked IllegalArgumentExceptions from fromCode, which are programming or parse errors, and what an unknown type code means on the wire is a decision for the REST handler that sees the request, not for an enum in the model. Nothing is lost that a consumer could have depended on — no client speaks these codes yet.

---
The three configuration enums (LedgerEntryType, LedgerPostingType, LedgerParameterType) each carried an explicit code string that duplicated the constant's own name verbatim — DEPOSIT("DEPOSIT"), and so on for all 21 constants — while LedgerIssueCode in the same feature derives its wire code from name(). That's two conventions doing one job in one package, and the redundant argument is a silent invitation for the code and the constant to drift apart. I've dropped the code field, its constructor parameter, and the Objects.requireNonNull guard in all three, so getCode() now simply returns name(). fromCode is unchanged in signature and behaviour; it just compares against getCode() rather than the field, which keeps getCode() as the single definition of the wire representation. This also lets the static initializer in LedgerEntryType go: it checked that no code was blank and that none collided, and neither can happen once the code is the constant name. No caller was touching the field directly, so nothing outside the enums changed. Core build is green and all 51 ledger tests pass, including the fromCode(getCode()) round-trip assertions.

---

matchesPrimary ignored the posting type of the rule it matched. LedgerRoleRules keys each primary rule on (entryType, postingType, direction), but the validator compared only unitRole, semanticRole, and direction — never rule.postingType(). The consequence: a BUY whose cash leg was carried by a FEE-typed posting claiming semanticRole=CASH, OUTBOUND, PRIMARY validated completely clean, no issues at all. The hole was one-directional, which is why the tests missed it — a wrong role on a correctly typed posting is caught (no rule matches, so SEMANTIC_PRIMARY_REQUIRED fires), while a wrong type under correct-looking roles sails through, and every test helper set type and role together. It matters because roles are client-supplied input rather than derived — that's precisely why D7 makes direction an explicit field — so the structural validator is the only guard between a nonsense posting and the model, and step 3 will project entries by mapping postings by role: it would have taken the FEE posting as the purchase's cash leg, turning the fee amount into the purchase amount and losing the fee as a unit. Fixed by comparing the posting type as well; added a regression test that retypes the cash primary of a valid BUY to FEE and expects SEMANTIC_PRIMARY_REQUIRED.

auchri:
@coderabbitai walkthrough

PR #5889, opened 2026-07-19
Show column menu on table header right-click

huntingthesun:
Today the show/hide-columns menu is only reachable via the small gear icon in each view's toolbar, and only some views additionally offer a per-column rename/hide menu when right-clicking a column header. Right-click is the more common convention for this in most desktop applications.

This combines both into a single header right-click menu: for views that already support column renaming, the rename/reset-name/hide-this-column items appear first, followed by the full show/hide-columns list; for views that don't support renaming, right-click now shows the show/hide-columns list directly (previously nothing useful happened there).

PP_feature_right-click-columns

No behavior change for the toolbar gear icon or for column renaming itself — this only adds a second, more discoverable way to reach the same functionality.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ShowHideColumnHelper.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/util/viewers/ShowHideColumnHelper.java
@@ -858,31 +858,37 @@ public class ShowHideColumnHelper implements IMenuListener, ConfigurationStoreOw
                             .forEach(c -> policy.create(c, null, c.getDefaultSortDirection(), c.getDefaultWidth()));
         }
 
-        if (isEditable)
-        {
-            ContextMenu headerMenu = new ContextMenu(policy.getViewer().getControl(), this::headerMenuAboutToShow);
+        var headerMenu = new ContextMenu(policy.getViewer().getControl(), manager -> {
+            if (isEditable)
+            {
+                headerMenuAboutToShow(manager);
+                if (manager.getItems().length > 0)
+                    manager.add(new Separator());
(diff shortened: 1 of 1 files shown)


buchen:
Thanks for the contribution. I think exposing the full menu on the header makes sense. (As a side remark: mid-term I am not sure that adding all columns to the menu makes sense - there are too many columns - and considering further configuration, it might deserve its own dialog).

However, I do not want to add the configuration menu for tables that do not have the configuration menu (gear item) at all. Let's say the table of the historical prices with "date" and "price". The table has these columns - removing them does not make sense. The other problem is: I assume I can add new columns to these tables and they become visible by default. That would not be the case anymore.

Therefore I did not change to the PR to enable the menu only if the table is editable in the first place.

To my understanding, the following tables to not have the header menu:

Dialogs
- EditClientFilterDialog — client-filter editor tree (createColumns())
- TroubleshootBalanceDiscrepancyDialog — balance discrepancy table

Views / tabs
- PerformanceView — 3 tables: the calculation tree, the "by securities" table, and the "by accounts" table. Its only gear (MenuConfigureView) holds pre-tax/cost-method options, not columns.
- SecurityQuoteQualityMetricsViewer — data-quality metrics table
- settings/AttributeListTab — attributes table
- settings/BookmarksListTab — bookmarks table
- payments/TransactionsTab — payments transactions table
- currency/CurrencyConverterTab — currency converter tree
- currency/ExchangeRatesListTab — exchange-rates table

Panes
- panes/HistoricalPricesPane — historical prices table
- panes/SecurityEventsPane — security events table
- panes/HistoricalPricesDataQualityPane — 2 tables (missing + unexpected quotes)

buchen:
merged.

PR #5892, opened 2026-07-20
Wrap column headers onto two lines when the column is too narrow

huntingthesun:
Long column headers (e.g. "Capital Gains (FIFO) 1 year") are currently truncated when the column is narrower than the label, leaving the header unreadable even when a two-line header would fit comfortably.

With this change, a column header wraps onto a second line when the full label does not fit the current column width, and snaps back to a single line as soon as the column is wide enough. The break point is chosen dynamically by measuring the actual rendered text width (greedy fill of the first line), not hardcoded into the label strings — so translations and renamed columns work without any per-language changes.

Headers wrapped when columns narrow
PP_feature_header-wrap

Header unwraps again when a column gets wider
PP_feature_header-unwrap

Implementation notes:
- Lives in `ShowHideColumnHelper`, so it applies uniformly to every table and tree in the application; no per-view changes and no message-string changes.
- The logical (unwrapped) label is tracked in widget data: column rename, "reset column name", and saved column configurations all persist the clean label — a rendered line break never leaks into saved state.
- Tested on macOS. Windows (Header32) and Linux (GTK) rendering of multi-line native headers is not verified.

### DIFF
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/Messages.java
@@ -1304,6 +1304,7 @@ public class Messages extends NLS
     public static String PrefMsgConfigureUpdates;
     public static String PrefMsgLanguageConfig;
     public static String PrefMyDividends24APIKey;
+    public static String PrefNoteExperimentsRestartRequired;
     public static String PrefNoteIndirectQuotation;
     public static String PrefNoteStoreSettingsNextToFile;
     public static String PrefPortfolioReportAPIKey;
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/messages.properties
@@ -2602,6 +2602,8 @@ PrefMsgLanguageConfig = To change the language, a restart is required.
 
 PrefMyDividends24APIKey = myDividends24 API Key
 
+PrefNoteExperimentsRestartRequired = A change takes effect for views opened afterwards. Quit and relaunch the application to apply it everywhere.
+
 PrefNoteIndirectQuotation = Quotes using a country's home currency as the price currency (for example, EUR 0.9009 = USD 1.00 in the Eurozone) are known as direct quotation or price quotation (from that country's perspective) and are used by most countries.\n\nQuotes using a country's home currency as the unit currency (for example, USD 1.11 = EUR 1.00 in the Eurozone) are known as indirect quotation or quantity quotation and are used in British newspapers and are also common in Australia, New Zealand and the Eurozone.\n\nSource: https://en.wikipedia.org/wiki/Exchange_rate\n\n\nExamples:\n\nIndirect (Quantity) Quotation: EUR/USD 1.1232\n"For one Euro one gets 1.1232 USD."\n\nDirect (Price) Quotation: USD/EUR 0.89\n"One USD costs 0.89 EUR."
 
 PrefNoteStoreSettingsNextToFile = The GUI settings contain configurations such as the start page, the width of some columns and window areas, the expanded nodes of the classification view, and more.\n\nBy default, the GUI settings are stored hidden in the 'workspace' folder. It is also possible to store them in a '.settings' file next to the Portfolio Performance file itself. This facilitates synchronizing the settings across multiple PCs.\n\nAttention: the GUI settings are not migrated. To keep the current settings, do the following: open a Portfolio Performance file -> then change the checkbox here -> then save the file under a new name.
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/Experiments.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/Experiments.java
@@ -12,7 +12,7 @@ public class Experiments
 {
     public enum Feature
     {
-        JULY26_PREVENT_UPDATE_WHILE_EDITING_CELLS
+        JULY26_PREVENT_UPDATE_WHILE_EDITING_CELLS, JULY26_WRAP_COLUMN_HEADERS
     }
 
     public boolean isEnabled(Feature feature)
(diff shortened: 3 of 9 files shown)


buchen:
This is a high-impact change. I know that users are carefully optimizing the screen real estate to display the exact number of columns and rows. I assume that the column width is not changed, but only the row height.

Is the wrapping limited to two rows or is it wrapping to as many rows as it needs? I think there should be some kind of limit.

Is there way to test on other platforms?

In general I like the idea, but I think it could then require a preference to keep the existing behavior. Or an experimental flag to first enable this as an experiment.

huntingthesun:
Thanks for the thoughtful review — answering each point:

**Column width and header row height:** Before this change, a column header row is already double the height of a normal data row — that's the existing table layout, unchanged by this PR. Today, header text is always written on one line, vertically centered within that already-tall header row; if the text doesn't fit the column's width, the excess is simply cut off. This PR doesn't change that layout at all — column width and header row height are both exactly as they are today. What changes is only how the *existing, already-reserved* vertical space in the header row is used: the header text now fills it with one line (as today) or two lines (never more) when the label is longer than the column is wide, instead of truncating. Screen real estate is unaffected either way.

**Row limit:** hard-capped at two lines, always — matching the header row's existing height. The wrap logic fills the first line with as many words as fit, then puts the remainder on the second line; there's no further splitting. If a column is narrow enough that even the second line overflows, it's truncated, same as a single-line header would be today. It cannot grow beyond two lines, and the row never grows to accommodate it.

**Cross-platform testing:** I found via examining the SWT bytecode that on Windows, header text is rendered natively with a single-line-only draw flag (`DT_SINGLELINE`) — an embedded line break wouldn't produce a second line there the way it does on macOS, and could render oddly. Rather than risk that, I've restricted the dynamic wrap to macOS specifically (platform-gated in code); it does nothing on Windows or Linux, leaving today's single-line behavior unchanged there. I haven't been able to verify Linux/GTK rendering yet.

**Preference/experimental flag:** done — this is now behind an opt-in experimental feature flag (`JULY26_WRAP_COLUMN_HEADERS`, default off, in Settings → Experiments), the same mechanism already used for `JULY26_PREVENT_UPDATE_WHILE_EDITING_CELLS`. Existing behavior is completely unchanged unless a user explicitly enables it and restarts the app.

buchen:
@huntingthesun writes
> Column width and header row height

Thanks for clarifying. I did not run the change before commenting - but I see now that the layout does not change. By limiting it to macOS only and having no impact on the overall layout, I would also say this is not a high-impact change after all. That should allow to remove the experimental flag then fairly quickly.

I haven't looked in detail on the code, but is there a chance to left align the second line as well?

Bildschirmfoto 2026-07-25 um 07 00 01

---

And: if I change the font size (in the preferences), then the header row does not increase and the two line wrapping looks sub-optimal. Is there a way to detect this and only wrap the line if there is enough vertical space?

Bildschirmfoto 2026-07-25 um 07 09 08

---

And: at least if I open the view initially (=click in the sidebar navigation), the column does not seem to be wrap. 


---


Claude commented on this as well:

**(Functional bug) Wrapped header text leaks into CSV export**

attachDynamicHeaderWrap mutates the live column text to include \n. The commit patches every internal reader of the header (serialize(), rename/reset menu, ORIGINAL_LABEL_KEY) to prefer WRAP_LABEL_KEY, but the CSV exporters read the widget text directly:

- TableViewerCSVExporter.java:61 → printer.print(column.getText())
- TreeViewerCSVExporter.java:52,56 → viewer.getTree().getColumn(ii).getText()

So on macOS with the flag on and a wrapped header, "Export → CSV" writes header cells containing an embedded newline. Depending on the CSV strategy that either quotes the field (ugly) or splits the header row. This is a genuine regression in a shipping feature, only masked by the experimental gate. The clean label is available in WRAP_LABEL_KEY, but it's a private static string in ShowHideColumnHelper, so the exporters can't reach it. Options: promote the key to a shared constant and have the exporters prefer it, or simplest — have the exporters do column.getText().replace('\n', ' '). Worth checking whether any other column.getText() reader (clipboard copy, print, accessibility) has the same exposure.

huntingthesun:
Thanks — all points addressed below.

**Alignment of the second line.** You're right, and the cause turned out to be in SWT rather than in this PR. On macOS, `TableColumn.drawInteriorWithFrame_inView` positions the whole text block according to the column's alignment — which is why the first line still looks correctly aligned — but then builds the string with a hard-coded `SWT.LEFT` paragraph alignment, so every line is left-aligned *inside* that block. With a single-line header the two can never disagree, because the draw rect is exactly as wide as the string, and headers follow the column alignment as expected (right for numeric columns, left for text). With two lines of different length the block is as wide as the longer line, so the shorter line starts at the longer line's left edge instead of ending at the column's right edge.

`Control.createString(...)` already maps `SWT.RIGHT` / `SWT.CENTER` onto the correct `NSTextAlignment`; the header call site simply passes the literal `SWT.LEFT` rather than the column's alignment bits. Passing the real alignment there would be a no-op for any header that fits its column, since the draw rect then equals the string width. To be precise about the one case where it is not a no-op: the rect is `min(stringWidth, columnRight - x)` and the string is created with `NSLineBreakByClipping`, so for a header that is *already truncated today* on a right-aligned column, the alignment change would alter which end of the label gets clipped. I noticed you already ship a patched cocoa fragment through the SWT fork, which would be the natural place for such a change if you think it is worth making.

There is nothing I can set from the application side to influence it: the attributed string is rebuilt on every paint, so an alignment set on the header cell cannot override that hard-coded paragraph alignment. If you'd rather not touch the fork for this, I can pad the shorter line with spaces so both lines render at the same width — that gets the alignment visually right to within a couple of pixels. It is a workaround rather than a fix, so I've left it out, but I'm happy to add it as an interim if you'd prefer the feature to look right in the meantime.

**Font size / vertical space.** Good catch, and it turned out to be a hard limit rather than something to compute around. On my machine the header row measured 28px in every configuration I tried — it does not grow with the font size, nor after the two-line text is set. Two lines measured 26px while the header is drawn in the small system font, which is the case as long as no font size is configured. As soon as one is, the style sheet sets that font on the table and the header cell inherits it: two lines then measured 30px at 12px and 32px at 13px, which is the clipping you saw. Those are my measurements on one macOS version rather than a guarantee, so rather than compute against them the wrap is simply not applied unless the configured font size is the default or smaller; above that the label stays on one line and truncates as before.

**Not wrapping when the view is first opened.** The columns are created before the style sheet has been applied, so the initial measurement could still be based on the pre-theme font, and a font change fires no resize event to trigger a recalculation. There is now a deferred recompute once the UI has settled.

**CSV export.** Genuine regression, fixed. The exporters read the header text directly, so a wrapped header would have written a line break into the exported file. Added a `getColumnLabel()` accessor that returns the logical label without the inserted line break, and used it in both the table and tree exporters; I checked the rest of the UI bundle and those were the only two places reading column text as data.

**Experimental flag.** Left in place for now while the above settles; happy to drop it whenever you think it's ready.

PR #5900, opened 2026-07-23
added support for dividends payed in USD with a conversion to EUR

kimmerin:
This PR extends the regexes to catch exchange-rate-informations and use that to calculate the gross amount and holding taxes because it's not in the PDF itself (only the fxgross-amount).

I've also extended the logic to obtain the share's name because that particular PDF contains a name that spans over two lines so it could be done in a tested way. I've considered adding this to the buy/sell-block as well but in the end refrained from it because there is no test PDF data to test against I didn't want to create a synthetic one.

This is my first implementation with different currencies in place, I hope I've covered everything ;-)

The PDF was provided in the [PP-forum](https://forum.portfolio-performance.info/t/pdf-import-von-st-galler-kantonalbank/39539/6?u=kimmerin).

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/StGallerKantonalbankPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/StGallerKantonalbankPDFExtractor.java
@@ -1,6 +1,9 @@
 package name.abuchen.portfolio.datatransfer.pdf;
 
+import static name.abuchen.portfolio.datatransfer.ExtractorUtils.checkAndSetGrossUnit;
 import static name.abuchen.portfolio.util.TextUtil.trim;
+
+import java.math.BigDecimal;
 
 import name.abuchen.portfolio.datatransfer.ExtractorUtils;
 import name.abuchen.portfolio.datatransfer.pdf.PDFParser.Block;
@@ -10,6 +13,7 @@ import name.abuchen.portfolio.model.AccountTransaction;
 import name.abuchen.portfolio.model.BuySellEntry;
 import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.PortfolioTransaction;
+import name.abuchen.portfolio.money.Money;
 import name.abuchen.portfolio.money.Values;
 
 /**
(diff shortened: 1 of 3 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/StGallerKantonalbankPDFExtractor.java
```diff
@@ -181,6 +203,23 @@ private void addDividendeTransaction()
...
+                        .match("^Betrag\\s+[A-Z]{3}\\s+(?<fxGross>['\\.,\\d]+)*$")
+                        .match("^Umrechnungskurs\\s+(?<baseCurrency>[A-Z]{3})\\/(?<termCurrency>[A-Z]{3})\\s+(?<exchangeRate>[\\.'\\d]+)\\s*$")
+                        .assign((t, v) -> {
+                            var rate = asExchangeRate(v);
+                            type.getCurrentContext().putType(rate);
+
+                            var fxGross = Money.of(rate.getTermCurrency(), asAmount(v.get("fxGross")));
+                            var gross = rate.convert(rate.getBaseCurrency(), fxGross);
+
+                            checkAndSetGrossUnit(gross, fxGross, t, type.getCurrentContext());
```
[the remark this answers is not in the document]
kimmerin: Should have been `\s*`. Fixed

buchen:
LGTM

@aragar as you added the extractor, anything that jumps into your eyes?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/StGallerKantonalbankPDFExtractor.java
```diff
@@ -137,17 +141,35 @@ private void addDividendeTransaction()
...
-                        .assign((t, v) -> t.setSecurity(getOrCreateSecurity(v)))
+                        .oneOf( //
+                                        section -> section //
+                                        // @formatter:off
+                                            // 10'000 N-Akt Sibanye Stillwater Limited
+                                            // Sponsored ADR Repr 4 Shs ADR/ADS
+                                            // Übrige Aktien Valoren-Nr.: 52619625, ISIN: US82575P1075
+                                            // Ausschüttung: USD 0.310942
+                                            // @formatter:on
+                                                        .attributes("name", "nameContinued", "wkn", "isin", "currency") //
```
buchen: Just for readability: I would propose to first have the comments, then start with the section, and have consistent indentation.

The oneOf section here has a subtle difference - one line - and depends on matching order. If we revisit this in a year from now, we will thank ourselves if we have it readable... 🙃

kimmerin: I've changed the order. The different intendation comes from the auto-formatting when saving which is enabled in the PP-project. When changing the order the indentation became a tad better but the first `@formatter:off`-line always ends up more to the left and it more than once drove me crazy.

I've now solved it by deactivating the auto-formatter before saving.

buchen: I only activate formatting of changed lines. And then exclude the PDF files and tests. Historically, the formatting diverged at some point... it is an annoyance, but reformatting the whole code base seemed also disruptive. All pull requests can go wrong. And the AI tools are good enough to handle the formatting anyway. That is way I refrained from reformatting...

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/StGallerKantonalbankPDFExtractor.java
```diff
@@ -204,12 +244,32 @@ private <T extends Transaction<?>> void addTaxesSectionsTransaction(T transactio
...
+                                                        .attributes("withHoldingTax", "currency", "termCurrency",
+                                                                        "baseCurrency", "exchangeRate")
+                                                        .match("^Quellensteuer [\\.,'\\d]+% (?<currency>[A-Z]{3}) (\\-)?(?<withHoldingTax>[\\.'\\d]+)$") //
+                                                        .match("^Umrechnungskurs\\s+(?<baseCurrency>[A-Z]{3})\\/(?<termCurrency>[A-Z]{3})\\s+(?<exchangeRate>[\\.'\\d]+)\\s*$") //
+                                                        .assign((t, v) -> {
+                                                            var rate = asExchangeRate(v);
+                                                            type.getCurrentContext().putType(rate);
+                                                            processWithHoldingTaxEntries(t, v, "withHoldingTax", type);
+                                                        }) //
+                                        ,
```
aragar, line 262: Sorry, I've never received a notification about this PR

Is this part needed? Isn't the exchange rate already kept in context while processing the dividend amount? Obviously, I'm very new to all this, but I tried calling the tests without this code, and they all seem to have passed.

kimmerin: I've tried the same now and can confirm that the exchange rate parts aren't necessary. When doing the implementation there was a time where the tests failed unless I've added these but I've also had to adjust parts of the dividendTransaction and I assume that this "fixed" the problem in the taxes section and made the exchange-part obsolete.

PR #5910, opened 2026-07-29
Fix FIFO capital gains ignoring the transaction exchange rate

### LINKED ISSUE #5908: FIFO capital gains ignore the transaction exchange rate
# FIFO capital gains ignore the transaction exchange rate

## Context

`CapitalGainsCalculation` (FIFO) converts a trade into the reporting currency with the plain
historic day rate, while its sibling `CapitalGainsCalculationMovingAverage` and the adjacent
`CostCalculation` both prefer the exchange rate **recorded on the transaction**. For a foreign
security bought through an account in a third currency, the FIFO figures therefore disagree with
the moving-average figures and with the cost basis, by the difference between the rate the user
actually got and the rate the exchange-rate provider published that day.

Found while specifying the per-instrument REST performance resource
(`2026-07-28 Per-instrument performance indicators in the REST API.md`, fact M12), where it shows
up as a residue in the identity `closingValue − periodCostBasis = unrealizedCapitalGains`. That
spec documents the residue and deliberately does **not** depend on this fix. This is a core defect
worth fixing on its own merits, and it is scoped, reviewed and released independently — it changes
numbers in PP's own Performance view, and bundling it with an API change would make an API review
responsible for auditing calculation semantics.

## The defect

Three engines consume the same portfolio transactions and convert them three ways:

| engine | acquisition basis | rate |
|---|---|---|
| `CostCalculation:98-99` | `t.getMonetaryAmount(converter)` / `t.getGrossValue(converter)` | prefers the transaction's rate |
| `CapitalGainsCalculationMovingAverage:44-45` | `t.getGrossValue(converter.with(securityCurrency))` / `t.getGrossValue(converter)` | prefers the transaction's rate |
| `CapitalGainsCalculation:77-78` | `t.getGrossValue().with(converter.at(t.getDateTime()))` | **always the historic day rate** |

`PortfolioTransaction.getGrossValue(CurrencyConverter)` (lines 177–192) prefers the embedded rate in
exactly one case: the term currency equals the **security's** currency, the transaction is booked in
another currency, and a `Unit.Type.GROSS_VALUE` unit is present — then it returns that unit's forex
amount. Otherwise it falls through to `converter.convert(getDateTime(), getGrossValue())`, which is
what `CapitalGainsCalculation` already does.

**So the divergence is confined to reporting in the security's own currency.** With base-currency
reporting the two paths agree exactly, which is why this has gone unnoticed.

Worked example, using the data already in `CapitalGainsCalculationTest.testFifoBuySellTransactions3`
— security in USD, account in EUR, inbound delivery of 100 shares recorded as EUR 4,500 / USD 5,000
at rate 0.90 — but valued with `?currency=USD` rather than EUR:

```
CostCalculation / MovingAverage:  getGrossValue(converter)  →  USD 5000.00   (the GROSS_VALUE unit)
CapitalGainsCalculation:          EUR 4500 × historic(2024-01-01)  →  USD 4500/r

r = 0.90  (the rate the user actually got)      →  USD 5000.00   agree
r = 0.909 (a plausible published day rate)      →  USD 4950.55   1 % low
```

The lot basis is wrong by the entire rate discrepancy, and every gain measured against it inherits
the error.

### The forex decomposition is coupled to the historic-rate assumption

This is why the fix is not a one-line substitution. `CapitalGainsCalculation` derives the currency
component by round-tripping the lot basis through historic rates (SELL branch ~lines 128–135, and
again in the `ValuationAtEnd` block ~lines 300–310):

```java
Money forex = convert2forex.convert(item.date, Money.of(termCurrency, item.value));
Money back  = forex.with(converter.at(t.getDateTime()));
forexGain   = back.getAmount() - start;
```

That recovers the foreign-currency amount *only if* `item.value` was itself derived from the
historic rate at `item.date`. Feed it an embedded-rate basis and the round-trip no longer inverts:
the currency component becomes internally inconsistent with the total it is supposed to be a
component of. Basis and decomposition must change together.

## Why this is a defect and not a decision

- **The sibling class already does it the other way, deliberately.**
  `CapitalGainsCalculationMovingAverage` carries a parallel security-currency cost
  (`movingAverageNetCostForex`) and derives the rate from the two recorded amounts, with the intent
  written into the comment: *"we want to take the actual exchange rate of the transaction"* and
  *"the account currency might be different than the reporting currency"*. An inconsistency between
  two members of the same calculation family, where one carries an explicit rationale, is an
  oversight in the other.
- **`ClientPerformanceSnapshot.determineTransferAmount` states the same principle** for cash
  transfers: *"because the actual exchange rate of the transferal might differ from the historical
  rate given by the exchange rate provider (e.g. ECB), we would get rounding differences if we do
  not take the original amount."*
- **`PortfolioTransaction.getGrossValue(CurrencyConverter)` exists for precisely this purpose** and
  is used by every comparable caller except this one.

## Fix

Bring `CapitalGainsCalculation` onto the pattern its moving-average sibling already establishes —
carry the basis in **both** currencies and derive the rate from the transaction, rather than
round-tripping through the rate provider.

- `LineItem` gains a second amount: the lot basis in the **security's** currency, obtained with
  `t.getGrossValue(converter.with(securityCurrency))`, alongside the term-currency basis from
  `t.getGrossValue(converter)`. Both are relieved proportionally on disposal, exactly as `value` is
  today, and exactly as `movingAverageNetCost` / `movingAverageNetCostForex` are in the sibling.
- The realized currency component becomes the moving-average formula: relieved forex basis at the
  disposal's own rate, less the relieved term-currency basis. Derive the rate from
  `netAmount / netAmountForex` as the sibling does, so the account currency need not be either the
  term or the security currency.
- The unrealized currency component in the `ValuationAtEnd` block follows the same shape, using the
  accumulated forex basis of the still-open lots against the closing valuation.
- Trails follow the selected amounts.

**Valuations need no change.** `ValuationAtStart` / `ValuationAtEnd` carry no transaction rate, so
the historic conversion at those dates is already correct — in the FIFO engine (line 61) and in the
moving-average one (line 30) alike. Only `visit(…, PortfolioTransaction)` and the forex
decomposition are in scope.

Deliberately **out of scope**: `squashForexValuationsAtStart`, the `TRANSFER_IN` / `TRANSFER_OUT`
branches, and `CostCalculation` (already correct). Resist widening this into a rewrite of the
capital-gains engine.

## Blast radius

The change is user-visible. Everything downstream of `CapitalGainsRecord` moves for portfolios that
hold foreign securities bought through accounts in a third currency, **when reported in the
security's currency**:

- `ClientPerformanceSnapshot` categories `CAPITAL_GAINS` and `REALIZED_CAPITAL_GAINS`
  (`addCapitalGainsFifo`, lines 375–383) — hence the Performance view and the REST
  `/v1/files/{file}/performance` breakdown
- `SecuritiesPerformanceView` columns `cg`, `cgforex`, `ucg`, `ucgforex` (FIFO variants only)
- anything reading `LazySecurityPerformanceRecord.getRealizedCapitalGains(CostMethod.FIFO)` /
  `getUnrealizedCapitalGains(CostMethod.FIFO)`

**There is no residual term to absorb the shift.** `addCapitalGains` sets the category valuation
directly from `getCapitalGains()`, and `addCurrencyGains()` derives `CURRENCY_GAINS` from cash
accounts only — foreign-currency account transactions plus closing funds. Every category is computed
independently, so the breakdown reconciles by construction rather than by plugging. Moving
`CAPITAL_GAINS` against fixed `INITIAL_VALUE`/`FINAL_VALUE` can therefore break reconciliation, and
that must be proven not to happen rather than assumed.

Requires a release-notes entry: users will see FIFO capital gains change on multi-currency
positions. Frame it as a correction — the figures now use the rate recorded on the transaction, as
the moving-average method already did.

## Existing test coverage, and what it does not pin

`CapitalGainsCalculationTest.testFifoBuySellTransactions3` is the only test combining
`GROSS_VALUE` units with capital gains. Two things about it matter:

- It reports in **EUR**, the transaction currency, so `getGrossValue(converter)` takes its
  pass-through branch and both implementations agree. **The test passes either way** — it does not
  pin the defect.
- It asserts `getForexCaptialGains()` for the moving-average variant only (lines 225, 242), never
  for FIFO. The FIFO currency component under `GROSS_VALUE` is therefore **entirely unpinned**.

Its moving-average assertions carry the intended arithmetic in comments and are the reference for
what the FIFO figures should become.

Reconciliation guards to run unchanged: `ClientPerformanceSnapshotTest`,
`scenarios/ClientPerformanceSnapshotTestCase`, `scenarios/CurrencyTestCase`,
`issues/IssueCurrencyGainsRoundingErrorTest`, `CostCalculationTest`, `WithoutTaxesFilterTest`.
Any movement in these is a finding to explain, not a baseline to update.

## Open questions for the implementer

These could not be settled from reading alone and should be resolved with a test first:

1. **Does the FIFO breakdown currently reconcile for the affected scenario at all?** If reporting in
   the security's currency already fails to reconcile today, this fix may *repair* the identity
   rather than threaten it — which changes the framing from "risky" to "overdue". Establish the
   before-state before changing anything.
2. **Do the two engines agree after the fix?** The per-instrument REST spec wants
   `closingValue − periodCostBasis = unrealizedCapitalGains` exact. Independent FIFO stacks with
   independent rounding (M11 of that spec) may still leave minor-unit differences. Measure the
   residual and report it, so that spec's `openapi.yaml` wording can be tightened or left as
   "agrees up to rounding".
3. **Should `getForexCaptialGains` for FIFO now equal the moving-average figure** on a single-lot
   portfolio? If the two methods are given one lot there is no averaging to differ over, so the
   currency components ought to coincide. That makes a sharp equivalence test — and if they still
   differ, there is a second defect underneath this one.

## Verification

- a new `CapitalGainsCalculationTest` case mirroring `testFifoBuySellTransactions3` but reporting in
  **USD** (the security's currency), asserting the FIFO lot basis is the `GROSS_VALUE` unit's forex
  amount rather than the historic conversion — this test must **fail before** the fix
- FIFO `getForexCaptialGains()` asserted explicitly, realized and unrealized, closing the gap noted
  above; ideally as an equivalence against moving-average on a single-lot portfolio (open question 3)
- reconciliation asserted for the affected multi-currency scenario, before and after
- the guard suite above, unchanged
- the existing `testFifoBuySellTransactions3` must keep passing verbatim: it reports in the
  transaction currency and its expectations are correct under either implementation

Build and test per CLAUDE.md — the core reactor list plus `:name.abuchen.portfolio.tests`.

buchen:
CapitalGainsCalculation (FIFO) converted a trade into the reporting currency with the historic day rate of the exchange-rate provider, while its moving-average sibling and CostCalculation both prefer the exchange rate recorded on the transaction. For a foreign security bought through an account in a third currency, the FIFO figures therefore disagreed with the moving-average figures and the cost basis when reported in the security's own currency.

Carry the lot cost basis in both the term and the security currency and derive the currency component from the transaction rate (via getGrossValue(converter)) rather than round-tripping through the rate provider, matching CapitalGainsCalculationMovingAverage. Valuations, squashForexValuationsAtStart, the TRANSFER branches and CostCalculation are left unchanged.

The realized/unrealized currency components now equal the moving-average figures on a single-lot portfolio. Two guard tests pinned the previous FIFO forex sub-component and are updated accordingly; the reconciling totals are unchanged.

Closes #5908

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/ClientPerformanceSnapshotTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/ClientPerformanceSnapshotTest.java
@@ -6,6 +6,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.Month;
@@ -623,12 +624,43 @@ public class ClientPerformanceSnapshotTest
                                                         // start of period
                                                         - (90.0 * 5)))));
 
+        // exchange rate purchase
+        var ratePurchase = client.getPortfolios().stream().flatMap(p -> p.getTransactions().stream())
+                        .filter(p -> p.getDateTime().equals(LocalDateTime.parse("2015-01-02T00:00"))).findFirst().get()
+                        .getUnit(Transaction.Unit.Type.GROSS_VALUE).get().getExchangeRate();
+
+        var rateSale = client.getPortfolios().stream().flatMap(p -> p.getTransactions().stream())
+                        .filter(p -> p.getDateTime().equals(LocalDateTime.parse("2015-01-09T00:00"))).findFirst().get()
+                        .getUnit(Transaction.Unit.Type.GROSS_VALUE).get().getExchangeRate();
+
+        // relieved USD cost basis in minor units, matching
+        // CapitalGainsCalculation#startForex (Math.round == HALF_UP)
+        var relievedForexBasis = Math.round(90.0 * 5 / ratePurchase.doubleValue() * 100);
+
+        // that basis valued at the sale's exchange rate, rounded like the
+        // production code (CapitalGainsCalculation#relievedForexInTerm), which
+        // uses RoundingMode.HALF_DOWN rather than Math.round's HALF_UP
+        var relievedForexBasisInTerm = BigDecimal.valueOf(relievedForexBasis)
+                        .multiply(BigDecimal.valueOf(rateSale.doubleValue()))
+                        .setScale(0, RoundingMode.HALF_DOWN).longValue();
+
         assertThat(snapshot.getCategoryByType(CategoryType.REALIZED_CAPITAL_GAINS).getPositions().get(0).getForexGain(),
                         is(Money.of(CurrencyUnit.EUR,
-                                        // value at start -> to USD with rate at
-                                        // start -> back to EUR with rate at end
-                                        // -> minus start value
-                                        Values.Amount.factorize((90.0 * 5 * 1.2043 / 1.1813) - (90.0 * 5)))));
+                                        // currency gains now use the exchange
+                                        // rate recorded on the transaction
+                                        // (matching the moving-average method):
+                                        // the relieved USD cost basis valued at
+                                        // the sale's exchange rate, less the
+                                        // relieved EUR cost basis. Because the
+                                        // USD basis is carried in minor units
+                                        // it differs from the historic
+                                        // round-trip (EUR 8.76) by one cent of
+                                        // rounding.
+                                        relievedForexBasisInTerm - Values.Amount.factorize(90.0 * 5))));
+
+        // with the explicit monetary amount
+        assertThat(snapshot.getCategoryByType(CategoryType.REALIZED_CAPITAL_GAINS).getPositions().get(0).getForexGain(),
+                        is(Money.of(CurrencyUnit.EUR, 877L)));
 
         assertThat(snapshot.getAbsoluteDelta(),
                         is(snapshot.getValue(CategoryType.FINAL_VALUE)
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
@@ -243,4 +243,75 @@ public class CapitalGainsCalculationTest
                         is(Money.of(CurrencyUnit.EUR, Values.Amount.factorize(-97.54))));
 
     }
+
+    /**
+     * Same portfolio as {@link #testFifoBuySellTransactions3()} but reported in
+     * the security's currency (USD) rather than the transaction/account currency
+     * (EUR). In that case the FIFO lot basis must be the {@code GROSS_VALUE}
+     * unit's forex amount (the rate recorded on the transaction) rather than the
+     * historic day rate published by the exchange-rate provider, matching the
+     * moving-average sibling and {@code CostCalculation}. See
+     * https://github.com/portfolio-performance/portfolio/issues/5908
+     */
+    @Test
+    public void testFifoBuySellTransactions3ReportedInSecurityCurrency()
+    {
+        var client = new Client();
+        client.setBaseCurrency(CurrencyUnit.EUR);
+
+        var security = new SecurityBuilder(CurrencyUnit.USD) //
+                        .addPrice("2025-01-01", Values.Quote.factorize(80)) //
+                        .addTo(client);
+
+        var account = new AccountBuilder(CurrencyUnit.EUR) //
+                        .addTo(client);
+
+        new PortfolioBuilder(account) //
+                        .inbound_delivery(security, "2024-01-01", Values.Share.factorize(100),
+                                        new Transaction.Unit(Transaction.Unit.Type.GROSS_VALUE,
+                                                        Money.of(CurrencyUnit.EUR, Values.Amount.factorize(4500)),
+                                                        Money.of(CurrencyUnit.USD, Values.Amount.factorize(5000)),
+                                                        BigDecimal.valueOf(0.90)) //
+                        )
+                        .inbound_delivery(security, "2024-02-01", Values.Share.factorize(50),
+                                        new Transaction.Unit(Transaction.Unit.Type.GROSS_VALUE,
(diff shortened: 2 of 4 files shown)


mierin12 (approved):
Looks good !
I was a bit confused at the end because the name of the variables "valueForex" are in the security currency, and at the very end when everyting has to be in the term currency we still use "forexGain" name, so "forex" in the variable name can be a bit misleading, sometimes in security currency sometimes in term currency. 

I like looking at the trail:
2 buys no sale
"thereof foreign currency gain":
Before :

Image

After :
Image
I think it makes sense, we now see the values of the Delivery transactions, both the ones in EUR and USD.

2 buys one partial sale (case of testFifoBuySellTransactions3ReportedInSecurityCurrency)
"capital gain when term currency is in security currency":
Before :

Image

After :

Image
I think it makes sense, we now see the values of the Delivery transactions, both the ones in EUR and USD.

Our examples are a bit extreme with transaction exchange rate quite different from the historical one (especially on 01 Feb 2024), I expect in reality the transaction rates from the brokers should not be that far from the historical one.

mierin12:
I wonder if this topic is not appearing also in Trades
```
this.entryValueWithoutTaxesAndFees = transactions.stream() //
                        .filter(t -> t.getTransaction().getType().isPurchase() == isLong)
                        .map(t -> t.getTransaction().getGrossValue()
                                        .with(converter.at(t.getTransaction().getDateTime())))
                        .collect(MoneyCollectors.sum(converter.getTermCurrency()));
```

-> getGrossValue(converter) ?


In EUR (currency of the file)
Capture d’écran 2026-07-29 à 23 41 39

In USD (currency of the security)
Capture d’écran 2026-07-29 à 23 42 03
The entry value are different now. So the profit loss also are different between the two cost methods.

buchen:
@mierin12 writes:

> I wonder if this topic is not appearing also in Trades

Yes, I agree, the entry value with USD reporting currency should be the USD value from the transaction. That is what has been bought. Let me see how to fix that. I think the getGrossValue(converter) method should be the fix here.

PR #5912, opened 2026-07-30
Let capital gains be measured against a fee-inclusive cost basis

buchen:
CostCalculation reports a cost basis both with and without the taxes and fees embedded in a trade, but the two capital gains engines only ever measured against the basis excluding them. Anything reporting a cost basis next to the gains on it therefore published two numbers that do not subtract: they differ by the charges embedded in the still-open lots.

The divergence arises only from purchases made inside the interval. A lot carried in is seeded from the same opening valuation by both engines, so a position held across the whole period reconciles either way - which is what makes the mismatch easy to miss.

CapitalGainsCalculation and CapitalGainsCalculationMovingAverage now take a TaxesAndFees, selecting the basis on both legs of a trade: a sale's gross value is the proceeds *before* the charges are deducted, so relieving a fee-inclusive lot against it would book the fees as a gain. The trails follow the selected basis on their own, as asGrossValue only inserts its "without taxes and fees" step when the basis differs from the transaction's own amount.

The change is additive. Each engine now names its basis explicitly at construction; the ex-fee basis - TaxesAndFees.NOT_INCLUDED - is the one ClientPerformanceSnapshot needs: its breakdown carries fees and taxes as separate terms and would subtract them twice if the gains absorbed them into the lot basis. The record's existing accessors default to that basis, so only a caller passing the new argument sees the other, and the existing snapshot and performance tests pass unchanged.

Calculation gains a perform overload taking an instance, since the existing one reflects a no-argument constructor.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
@@ -19,10 +19,14 @@ import name.abuchen.portfolio.model.CostMethod;
 import name.abuchen.portfolio.model.Portfolio;
 import name.abuchen.portfolio.model.PortfolioTransferEntry;
 import name.abuchen.portfolio.model.Security;
+import name.abuchen.portfolio.model.TaxesAndFees;
 import name.abuchen.portfolio.model.Transaction;
 import name.abuchen.portfolio.money.CurrencyUnit;
 import name.abuchen.portfolio.money.Money;
 import name.abuchen.portfolio.money.Values;
+import name.abuchen.portfolio.snapshot.trades.Trade;
+import name.abuchen.portfolio.snapshot.trades.TradeCollector;
+import name.abuchen.portfolio.snapshot.trades.TradeCollectorException;
 import name.abuchen.portfolio.util.Interval;
 
 @SuppressWarnings("nls")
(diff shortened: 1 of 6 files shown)


buchen:
@mierin12 I was also thinking about the naming. The ```TaxesAndFees``` is cumbersome. And the ```NOT_INCLUDED``` maybe reads better as "excluded". What do you think about the term "Charges". And then ```Charges.INCLUDED``` and ```Charges.EXCLUDED```. "charges" is also a shorter postfix to any variables we come across. What do you think? Would this improve readability?

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
```diff
@@ -314,4 +315,84 @@ public void testFifoBuySellTransactions3ReportedInSecurityCurrency()
...
+            var unrealizedInclusive = Money.of(CurrencyUnit.EUR, Values.Amount.factorize(1140));
+            assertThat(record.getUnrealizedCapitalGains(method, TaxesAndFees.INCLUDED).getCapitalGains(),
+                            is(unrealizedInclusive));
+
+            // the reconciliation this change is about: the fee-inclusive
+            // unrealized gains subtract cleanly from the fee-inclusive cost
+            // reported next to them (market value - cost with charges)
+            assertThat(record.getUnrealizedCapitalGains(method, TaxesAndFees.INCLUDED).getCapitalGains(),
+                            is(record.getCapitalGainsOnHoldings(method)));
+            assertThat(record.getCapitalGainsOnHoldings(method), is(unrealizedInclusive));
```
mierin12, line 398: It is a good check. Maybe another good check would be to have a open/closed Trade (Gross)Profit/Loss check vs a unrealized/realized capital gains

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculation.java
```diff
@@ -107,22 +122,33 @@ public void visit(CurrencyConverter converter, CalculationLineItem.TransactionIt
...
+        // currency
+        var termBasis = taxesAndFees.isIncluded() ? t.getMonetaryAmount(converter) : t.getGrossValue(converter);
+        var forexBasis = taxesAndFees.isIncluded() ? t.getMonetaryAmount(converter.with(securityCurrency))
+                        : t.getGrossValue(converter.with(securityCurrency));
+
+        // asGrossValue adds the "without taxes and fees" step only when the
+        // basis differs from the transaction's own amount, hence it collapses
+        // to nothing under TaxesAndFees.INCLUDED
+
+        TrailRecord txTrail = TrailRecord.ofTransaction(t).asGrossValue(basis);
```
mierin12, line 144: For Trail, I do not exactly understand the comment, for the moment anyway I understand a Trail with TaxesAndFees included is not available for user. It will be easier to check when we can activate this option in the Calculation view.
From kommer file with a 6 years reporting period:

Image
How would the trail in the mode "Capital Gain after taxes and fees" would look like ? One less row and one less column to compare 359.4 with 262.4 directly ?

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationMovingAverage.java
```diff
@@ -41,8 +56,14 @@ public void visit(CurrencyConverter converter, CalculationLineItem.TransactionIt
...
-        long netAmountForex = t.getGrossValue(converter.with(securityCurrency)).getAmount();
-        long netAmount = t.getGrossValue(converter).getAmount();
+        // the basis: the gross value leaves out the taxes and fees embedded in
+        // the trade, the monetary amount takes them in - on both legs, as a
+        // sale's gross value is the proceeds before the charges are deducted
+        var netAmountForex = taxesAndFees.isIncluded()
+                        ? t.getMonetaryAmount(converter.with(securityCurrency)).getAmount()
+                        : t.getGrossValue(converter.with(securityCurrency)).getAmount();
+        var netAmount = taxesAndFees.isIncluded() ? t.getMonetaryAmount(converter).getAmount()
+                        : t.getGrossValue(converter).getAmount();
```
mierin12: For clarity, I think the variable name can now be changed, no need for "net" in the name anymore, same for movingAverageNetCost, movingAverageNetCostForex.

mierin12 (commented):
Hello, this is great to have this option. I did not realized the "gains on holding" was after taxes and fees.

I put some comments For Information.

In a next step, I see 4 locations where those data could be available from user:

- As columns in the Statement of Assets
- As columns in the Performance/Securities View
- In the Performance/Calculation view,
- In the Performance Calculation widget.

For those last two, I guess, it means the view shall also be able to show Earnings after taxes and fees ?

mierin12:
> @mierin12 I was also thinking about the naming. The `TaxesAndFees` is cumbersome. And the `NOT_INCLUDED` maybe reads better as "excluded". What do you think about the term "Charges". And then `Charges.INCLUDED` and `Charges.EXCLUDED`. "charges" is also a shorter postfix to any variables we come across. What do you think? Would this improve readability?

I guess yes, in particular for EXCLUDED. Charges works, but taxAndFees has the benefit to be directly clear, self sufficient.

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculation.java
buchen: Agree. It is not testable in the UI at the moment. The point is simply that the "asGrossValue" does nothing if it is already a gross value, i.e., is not adding an additional line in the calculation trail that will be visible to the user if there is nothing to deduct for the gross value anyway.

name.abuchen.portfolio/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationMovingAverage.java
buchen: I have updated the names - agree.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/snapshot/security/CapitalGainsCalculationTest.java
buchen: Agree to extend the tests. Added one.

PR #5914, opened 2026-07-30
Comdirect PDF import fix for tax refunds

kimmerin:
fixed regex handling tax refunds by removing all spaces and underscores to support Comdirect's new format without actual spaces for spacing.

Added tests with PDF-dumps [provided by the forum](https://forum.portfolio-performance.info/t/pdf-import-von-comdirect/1647/648?u=kimmerin).

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
@@ -1374,56 +1374,6 @@ public class ComdirectPDFExtractor extends AbstractPDFExtractor
                                                             t.setAmount(asAmount(v.get("refundedTaxes")));
                                                         }))
 
-                        .optionalOneOf( //
-                                        // @formatter:off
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ComdirectPDFExtractor.java
```diff
@@ -1376,6 +1378,11 @@ private void addTaxesTreatmentTransaction()
                                                             t.setAmount(asAmount(v.get("refundedTaxes")));
                                                         }))
 
+                        // TODO fun fact: With the preparing of the lines this whole optionalOneOf-block
+                        //      can be commented out and all tests continue to run through. But it's not
+                        //      my decision to remove it from the source but I keep the original regex
+                        //      becasue I can't check if I break something due to no tests failing from the start
+
```
buchen: Fair comment. Apparently we have no sample for this: VerkaufMitSteuerbehandlung08.txt and VerkaufMitSteuerbehandlung09.txt match it, but because the instrument currency and the transaction currency to not differ, the value is removed again.

I propose deleting the block. If we have samples, it is super hard to keep it alive. And the gross amount parsed for test case 08 also looks off.

Generally, I think such a comment is better part of the PR and not part of the code.

PR #5929, opened 2026-08-05
Statement of Assets column: Quote Date & Update Status

JGKle:
Implement a new optional column for Statement of Assets that shows the quote date when not updating, and the status of the update when updating.

Avoids needing to switch to View->Price Update Status & makes it easier to see inline what's being waited on, especially with rate-limited providers like Coingecko.

Peek 2026-08-04 20-09


JGKle:
Is anything missing here to get this merged @buchen ?

PR #5937, opened 2026-08-06
Fix Donut chart sort inconsistency

Mikerion:
Sort was inconsistent with other part and also with Pie chart


buchen:
pushed to master. merged.

PR #5939, opened 2026-08-07
added (more) support for bonds for Consors and DAB

kimmerin:
Being hinted in #5896 that better support for bonds is appreciated I've gone through my archive of PDFs (and bad business decisions ;-) and tried the import of PDFs that contained transactions of bonds. The problems that showed up I've fixed, also I've done a bit of preparation for #5453 by extending the PDF importer's data structure to handle bonds and nominals. The former in the form of `asBond` currently only calls `asSecurity` but might do other things in the future, e.g. set the correct type, etc. The latter in the form of `asBondNominal` does the division by 100 so that importer don't need to do that all the time by themselves.

I haven't searched through the code base where this division currently takes place in a copy/paste manner, that's something that can be done if my idea is accepted and merged.

PR #5896, opened 2026-07-21
Modify DADAT PDF-Importer to support bond transactions

aragar:
Closes #5316

Account statements (KONTOAUSZUG) containing bonds quoted as a percentage of nominal were not imported. Adds support for:

- **Bond purchases** with nominal notation (e.g. `16.000,00000 EUR` instead of `… STK`). Uses the established `shares = nominal / 100` workaround for percentage quotation (16.000 nominal → 160 shares), so the percent quote becomes the implied per-share price. Accrued interest (Stückzinsen) folds into the gross value, matching the document's "steuerlicher Anschaffungswert".
- **Bond coupon payments** (`Ertrag`) whose booking line carries the interest-rate token (`Ertrag 2,8750 % Depot …`); nominal shares handled the same way.
- **Kapitalertragsteuer** withheld on coupon payments (line-anchored so it does not collide with the `Kurs 0,000000 Kapitalertragsteuer …` lines of the separate Steuern blocks).
- **Outgoing transfers** where the IBAN sits on the same line as the amount and date, imported as removals.

Test fixtures are the two anonymized statements from #5316. Verified with DADATBankenhausPDFExtractorTest (49 tests, all green).

PR #5941, opened 2026-08-07
added (more) support for bonds for Consors and DAB - #5939

kimmerin:
Being hinted in https://github.com/portfolio-performance/portfolio/pull/5896 that better support for bonds is appreciated I've gone through my archive of PDFs (and bad business decisions ;-) and tried the import of PDFs that contained transactions of bonds. The problems that showed up I've fixed, also I've done a bit of preparation for https://github.com/portfolio-performance/portfolio/issues/5453 by extending the PDF importer's data structure to handle bonds and nominals. The former in the form of asBond currently only calls asSecurity but might do other things in the future, e.g. set the correct type, etc. The latter in the form of asBondNominal does the division by 100 so that importer don't need to do that all the time by themselves.

I haven't searched through the code base where this division currently takes place in a copy/paste manner, that's something that can be done if my idea is accepted and merged.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
@@ -1332,6 +1332,74 @@ public class ConsorsbankPDFExtractorTest
     }
 
     @Test
+    public void testAnleiheKauf01()
+    {
+        var extractor = new ConsorsbankPDFExtractor(new Client());
+
+        var errors = new ArrayList<Exception>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "AnleiheKauf01.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(1L));
+        assertThat(countBuySell(results), is(1L));
+        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
+        assertThat(results.size(), is(2));
+        new AssertImportActions().check(results, "EUR");
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("GRR000000010"), hasWkn("A1G1UW"), hasTicker(null), //
+                        hasName("0,0 % GRIECHENLAND 12-42 IO GDP   15.OKT"), //
+                        hasCurrencyCode("EUR"))));
+
+        // check buy sell transaction
+        assertThat(results, hasItem(purchase( //
+                        hasDate("2017-12-19T17:00:34"), hasShares(4000), //
+                        hasSource("AnleiheKauf01.txt"), //
+                        hasNote("120321388.001 | Limitkurs  0,550000 %"), //
+                        hasAmount("EUR", 2210.45), hasGrossValue("EUR", 2200.00), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 5.50 + 4.95))));
+    }
+
+    @Test
+    public void testAnleiheKauf02()
+    {
+        var extractor = new ConsorsbankPDFExtractor(new Client());
+
+        var errors = new ArrayList<Exception>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "AnleiheKauf02.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(1L));
+        assertThat(countBuySell(results), is(1L));
+        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
+        assertThat(results.size(), is(2));
+        new AssertImportActions().check(results, "EUR");
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("GRR000000010"), hasWkn("A1G1UW"), hasTicker(null), //
+                        hasName("0,0 % GRIECHENLAND 12-42 IO GDP 15.OKT"), //
+                        hasCurrencyCode("EUR"))));
+
+        // check buy sell transaction
+        assertThat(results, hasItem(purchase( //
+                        hasDate("2025-04-09T12:03:33"), hasShares(10000), //
+                        hasSource("AnleiheKauf02.txt"), //
+                        hasNote("329893673.001 | Limitkurs 0,280000 % | Poolfaktor 0,904761904"), //
+                        hasAmount("EUR", 2546.56), hasGrossValue("EUR", 2533.33), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 1.95 + 6.33 + 4.95))));
+    }
+
+    @Test
     public void testWertpapierBezug01()
     {
         var extractor = new ConsorsbankPDFExtractor(new Client());
@@ -1873,6 +1941,40 @@ public class ConsorsbankPDFExtractorTest
                         hasNote("15681369.001"), //
                         hasAmount("EUR", 211.30), hasGrossValue("EUR", 222.20), //
                         hasTaxes("EUR", 0.00), hasFees("EUR", 0.95 + 5.00 + 4.95))));
+    }
+
+    @Test
+    public void testAnleiheEinloesung01()
+    {
+        var extractor = new ConsorsbankPDFExtractor(new Client());
+
+        var errors = new ArrayList<Exception>();
+
+        var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "AnleiheEinlösung01.txt"), errors);
+
+        assertThat(errors, empty());
+        assertThat(countSecurities(results), is(1L));
+        assertThat(countBuySell(results), is(1L));
+        assertThat(countAccountTransactions(results), is(0L));
+        assertThat(countAccountTransfers(results), is(0L));
+        assertThat(countItemsWithFailureMessage(results), is(0L));
+        assertThat(countSkippedItems(results), is(0L));
+        assertThat(results.size(), is(2));
+        new AssertImportActions().check(results, "EUR");
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("GRR000000010"), hasWkn("A1G1UW"), hasTicker(null), //
+                        hasName("Griechenland EO-FLR Secs 12(23-42) 1 IO GDP"), //
+                        hasCurrencyCode("EUR"))));
+
+        // check buy sell transaction
+        assertThat(results, hasItem(sale( //
+                        hasDate("2025-05-14T00:00"), hasShares(21000), //
+                        hasSource("AnleiheEinlösung01.txt"), //
+                        hasNote(null), //
+                        hasAmount("EUR", 5297.88), hasGrossValue("EUR", 5297.88), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00))));
     }
 
     @Test
@@ -3501,7 +3603,7 @@ public class ConsorsbankPDFExtractorTest
 
         // check dividends transaction
         assertThat(results, hasItem(dividend( //
-                        hasDate("2025-04-10T00:00"), hasExDate(null), //
+                        hasDate("2025-04-10T00:00"), hasExDate("2025-04-01T00:00"), //
                         hasShares(1000.00), //
                         hasSource("Dividende26.txt"), //
                         hasNote(null), //
@@ -3538,7 +3640,7 @@ public class ConsorsbankPDFExtractorTest
 
         // check dividends transaction
         assertThat(results, hasItem(dividend( //
-                        hasDate("2025-04-10T00:00"), hasExDate(null), //
+                        hasDate("2025-04-10T00:00"), hasExDate("2025-04-01T00:00"), //
                         hasShares(1000.00), //
                         hasSource("Dividende26.txt"), //
                         hasNote(null), //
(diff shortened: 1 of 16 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -57,12 +61,13 @@ private void addBuySellTransaction()
...
                         .section("type").optional() //
-                        .match("^(?<type>(Kauf|Verkauf|Gesamtf.lligkeit)) .*$") //
+                        .match("^(?<type>" + TRANSACTIONTYPE_REGEX + ") .*$") //
                         .assign((t, v) -> {
                             // @formatter:off
                             // Is type --> "Verkauf" change from BUY to SELL
                             // @formatter:on
-                            if ("Verkauf".equals(v.get("type")) || "Gesamtfälligkeit".equals(v.get("type")))
+                            var tType = Objects.toString(v.get("type"), "");
+                            if (tType.endsWith("Verkauf") || "Gesamtfälligkeit".equals(tType))
```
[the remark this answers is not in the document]
kimmerin: fixed by checking it using a regex (I haven't introduced that part btw ;-)

name.abuchen.portfolio/src/name/abuchen/portfolio/model/BuySellEntry.java
```diff
@@ -78,11 +78,21 @@ public void setSecurity(Security security)
         this.accountTransaction.setSecurity(security);
     }
 
+    public void setBond(Security security)
+    {
+        setSecurity(security);
+    }
+
```
buchen: I think having a helper ```asBondNominale``` in the extractor code is good, but we should not add this to the model API now - we do not even know how the bond structure will look like. And the extractor will also have to have some compatibility to import into (however it will look like) into bonds and into the existing workaround structure. I think it will be easier if we contain the "hack" to the importer code for now.

buchen:
> and bad business decisions ;-)

haha, I know the feeling...

name.abuchen.portfolio/src/name/abuchen/portfolio/model/BuySellEntry.java
kimmerin: My idea was that having this method being used by the PDF-importers make it easier to do whatever is decided over there at #5453 to do the corresponding adjustments. At the moment you can only hope to find everything by looking for number of shares being divided by 100. Having a distinct method would make things a lot easier (even if it might be removed at the end because a bond is implemented as a flag in Security, etc.)

> I think it will be easier if we contain the "hack" to the importer code for now.

You mean keep `asBond` and `asBondNominal` but remove `setBond` and `setBondNominal` in `Transaction` and `BuySellEntry`?

buchen: Not sure which ```asBond``` you mean, but yes, the ```AbstractPDFExtractor``` changes are a good idea.

I have split the commit now into three different commits: one for the change to ``AbstractPDFExtractor```, one for Consorsbank, and one for DAB. I will force push this into this pull request and then add some comments on some other things where I am not sure that I fully understand it.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
```diff
@@ -1331,6 +1331,74 @@ public void testWertpapierKauf31()
...
+                        hasName("0,0 % GRIECHENLAND 12-42 IO GDP 15.OKT"), //
+                        hasCurrencyCode("EUR"))));
+
+        // check buy sell transaction
+        assertThat(results, hasItem(purchase( //
+                        hasDate("2025-04-09T12:03:33"), hasShares(10000), //
+                        hasSource("AnleiheKauf02.txt"), //
+                        hasNote("329893673.001 | Limitkurs 0,280000 %"), //
+                        hasAmount("EUR", 2546.56), hasGrossValue("EUR", 2533.33), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 1.95 + 6.33 + 4.95))));
```
buchen, line 1399: What about the "Poolfaktor" in the document?

The document reads:
```
Einheit Umsatz Poolfaktor
EUR 1.000.000,00000 0,904761904
Preis pro Anteil 0,280000 % FRANCO COURTAGE
```

Shouldn't that be applied to the shares? And the
904 761,90 × 0,28 % = 2 533,33
is exactly the gross price we are matching here?

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
```diff
@@ -3834,6 +3936,76 @@ public void testDividende30WithSecurityInEUR()
...
+
+        // check security
+        assertThat(results, hasItem(security( //
+                        hasIsin("GR0128010676"), hasWkn("A1G1UA"), hasTicker(null), //
+                        hasName("Griechenland EO-Bonds 2012(23) Ser.1"), //
+                        hasCurrencyCode("EUR"))));
+
+        // check dividends transaction
+        assertThat(results, hasItem(dividend( //
+                        hasDate("2018-02-26T00:00"), hasExDate("2018-02-21T00:00"), //
```
buchen: About the ex-date: the document has Kupontag of 24.02. but it is asserted against "Schlusstag". I think we should Kupontag. Furthermore, both test document 1 and 2 are about the very same bond, but parse different ex-dates.

```
Bestand Kupontag
150 EUR 24.02.2018
```

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
```diff
@@ -462,7 +465,36 @@ private void addDividendTransaction()
...
+                                        //  3,00000 % Griechenland                    ZINSTERMIN:  24.FEBRUAR      
+                                        // EO-Bonds 2012(23) Ser.1                                      
+                                        // ZINS-/DIVIDENDENSATZ            3,000000  %   SCHLUSSTAG PER 23.02.2017 
+                                        // @formatter:on
+                                        section -> section //
+                                                        .attributes("wkn", "name", "nameContinued", "currency") //
+                                                        .match("^(?<currency>[A-Z]{3})\\s+[\\.,\\d]+\\s+WKN:\\s+(?<wkn>[A-Z0-9]{6}).*$") //
+                                                        .match("^(?<name>.*?)(ZINSTERMIN.*)$") //
+                                                        .match("^(?<nameContinued>.*)$") //
+                                                        .match("^(?i)(ZINS-\\/DIVIDENDENSATZ|(ERTRAGSAUSSCHUETTUNG|ERTRAGSTHESAURIERUNG) P\\. ST\\.) .* ([\\d,.]+\\s+%|(?<currency>[A-Z]{3}))\\s+SCHLUSSTAG PER [\\d]{2}\\.[\\d]{2}\\.[\\d]{4}.*$") //
```
buchen: Two times currency in the regular expressions? It will be overwritten into the map of values that is passed to the assign function.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
```diff
@@ -736,27 +796,60 @@ private void addEncashmentTransaction()
...
+                            .attributes("wkn", "isin", "name", "name1") //
+                            .find("Wertpapierbezeichnung Zinssatz WKN ISIN") //
+                            .match("^(?<name>.*?) [\\d,.]+ (?<wkn>[A-Z0-9]{6}) (?<isin>[A-Z]{2}[A-Z0-9]{9}[0-9])$") //
+                            .match("^(?<name1>.*)$") //
+                            .assign((t, v) -> {
+                                if (!v.get("name1").startsWith("Einheit"))
+                                    v.put("name", v.get("name") + " " + v.get("name1"));
+    
+                                t.setSecurity(getOrCreateSecurity(v));
+                            })
```
buchen: It looks almost identical to the block above - and the sample from the bank document is identical. Either we update the sample (there is ZINSEN and the [\\d,.]+). Alternatively, we could collapse it but I guess the duplication does improve readability.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dab/AnleiheKauf01.txt
```diff
@@ -0,0 +1,89 @@
+PDFBox Version: 3.0.7
```
buchen, line 1: Consorsbank tests say AnleiheKauf but the DAB tests say KaufAnleihe. I have no opinion what is better, but I think we should be consistent.

I propose to use AnleiheKauf. There are Kauf_*qualifier* test documents, but that is more about taxes and other special handling. And "Anleihe" does not feel like a qualifier, but like a new class of instruments. (Like we use CryptoKauf...)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dab/DABPDFExtractorTest.java
```diff
@@ -630,6 +630,48 @@ public void testWertpapierKauf14()
                         hasTaxes("EUR", 0.00), hasFees("EUR", 0.00))));
     }
 
+    @Test
+    public void testAnleiheKauf01()
```
buchen, line 634: ... and the test method is actually called AnleiheKauf - a mismatch to the name of the document.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -903,7 +941,7 @@ private void addBuySellTaxReturnBlock(DocumentType type)
     {
         var pdfTransaction = new Transaction<AccountTransaction>();
 
-        var firstRelevantLine = new Block("^(Kauf|Verkauf|Gesamtf.lligkeit) .*$", "Dieser Beleg wird .*$");
+        var firstRelevantLine = new Block("^" + TRANSACTIONTYPE_REGEX + ".*$", "Dieser Beleg wird .*$");
```
buchen: Here the space is lost (while all other places in the refactoring keep it). I guess is makes sense to match the term with a space here as well.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
```diff
@@ -14,11 +16,12 @@
 import name.abuchen.portfolio.model.Client;
 import name.abuchen.portfolio.model.PortfolioTransaction;
 import name.abuchen.portfolio.money.Money;
-import name.abuchen.portfolio.money.Values;
 
 @SuppressWarnings("nls")
 public class DABPDFExtractor extends AbstractPDFExtractor
 {
+    private static final String TRANSACTIONTYPE_REGEX = "(Kauf|(Spitze )?Verkauf|Gesamtf.lligkeit)";
```
buchen: Because it is wrapped into named group anyway, we could also do ```"(?:Kauf|(?:Spitze )?Verkauf|Gesamtf.lligkeit)"``` because the idea is not to capture them at all.

buchen (commented):
Besides the "asBond" discussion (I have split the commits into 3 separate ones), I am also have some feedback on the Consorsbank and DAB implementation (commit 2 and 3 respectively). Do they make sense? Feel free to push additional commits into this pull request - ideally keep them separate for Consorsbank and DAB so that we can squash them accordingly.

name.abuchen.portfolio/src/name/abuchen/portfolio/model/BuySellEntry.java
hporten: I support a central method as well. Not because of the little code sharing but as an enforcement of everyone using the same type of hack. Until a proper solution is found. A name like asBondHack() would ensure that every use is aware of the odd nature.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
kimmerin: at the end of the day (i.e. after the execution of the order) the nominal of 1.000.000 got added to my depot which is why I've used that instead of the number after applying the factor. I can add this information to the note if you want.

buchen: Interesting. I tried to read up. You are right, the shares are apparently fully booked. And the value is determined by ```Wert der Anleihe = Nominalwert * Kurs * Poolfaktor``` in the end. Probably the best is then to add the pool factor (unless it is 1) to the note. That is the best we can support at the moment.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
kimmerin: Changed it accordingly, now imports from Dividende26.txt come with an ex-date as well ;-)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
kimmerin: Copied the regex from line 454 and that named group was too far to the right so I've missed it.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
kimmerin: The new one is triggered when dealing with bonds and I've expected the assign-block to differ from the regular ones handling regular shares. But it also increases readability.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/consorsbank/ConsorsbankPDFExtractorTest.java
kimmerin: Added to note

kimmerin:
@buchen I've commited the changes to the consors-related classes but Eclipse insisted on specifying a branch and since this separation of commits is new to me I've created a new one to avoid breaking things in the old: https://github.com/kimmerin/portfolio/tree/pdfimport_bondsupport_consors. Hope that's OK that way.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dab/AnleiheKauf01.txt
kimmerin: The filename came to be because all buying files are called Kauf and I've intended to keep it "nearby". Then I've switched to the test class and tried to stick with the naming scheme there as well leading to the different names. Renamed the file now

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dab/DABPDFExtractorTest.java
kimmerin: see previous response

buchen:
> I've created a new one to avoid breaking things in the old

That is okay. I think I see the changes to the Consors one. I can pick that up (but not today ... too late for me)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
kimmerin: Added a \b to the identifiying regex, a space like this can easily be overlooked or can lead to problems we've recently had with Comdirect.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/dab/AnleiheKauf01.txt
kimmerin: And done the same with VerkaufAnleihe01.txt

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/DABPDFExtractor.java
kimmerin: I assume we're talking about the decision to change the type to sell? non-capturing group doesn't help here because everything is "wrapped" into the named group "type" and that includes e.g. the "Spitze" all the time (I've tried ;-)

kimmerin:
> That is okay. I think I see the changes to the Consors one. I can pick that up (but not today ... too late for me)

A relieve that I didn't break everything (wouldn't be the first with git ;-)

I've checked in the DAB-related changes to the original branch.

PR #5942, opened 2026-08-08
MOEX provider support

kiberdruzhinnik:
Hey there.

Kindly ask you to add this commit into upstream to support MOEX history and latest quotes provider.


buchen:
Can you link to the terms and conditions that explain if and how the API can be used? Please also paste the relevant paragraph here.

kiberdruzhinnik:
> Can you link to the terms and conditions that explain if and how the API can be used? Please also paste the relevant paragraph here.

Terms and conditions can be found here: https://www.moex.com/ru/datapolicy/.

If I interpret them correct, we could use free API (history and another data) for informational purposes with delays approx. 15 minutes to current trade time.

API docs can be found here: https://iss.moex.com

buchen:
Thanks @kiberdruzhinnik. On a high-level reading, it appears to allow programmatic access but not redistribution (which PP does not do because we not proxy, cache, aggregate, publish, or redistribute the data).

I used the contact form on the site to leave a brief inquiry. Let's see if we get an answer.

> I am the lead developer of Portfolio Performance, an open-source desktop application.
A contributor would like to implement support for the public MOEX ISS API, see:
https://github.com/portfolio-performance/portfolio/pull/5942
> 
> The application would make HTTPS requests directly from the end user's computer to iss.moex.com to retrieve historical/ delayed market prices for securities selected by that user. We do not proxy, cache, aggregate, publish, or redistribute MOEX market data through our own servers. Is this use of the ISS API permitted without a separate market-data agreement or license?

buchen:
I got a reply:

> The information obtained from the ISS is available for review only and cannot be used to generate profit or for any other purposes (except for review), including for providing services to third parties to supply this information or materials and data, products and services prepared based on it. Any use of the information (except for review) is permitted only under a contract with the Moscow Exchange.
The procedure for using Exchange information is available at https://www.moex.com/en/datapolicy/

I am not sure yet how to read the "for review only" part of the reply. There is no idea to generate a profit, that is for sure, but what means "for review" - is displaying the prices in the application and using it to calculate historic performance covered by the review scenario? Maybe the term got changed in the translation? I will try to find out.

PR #5945, opened 2026-08-09
Make the width of the chart lines configurable

### LINKED ISSUE #5944: Make chart line width configurable (1 / 2 / 3 px)
## Problem

Since 0.85.0 (Eclipse 2026-06, monitor-specific scaling on by default) SWT scales line widths with the monitor scaling factor. The hardcoded `setLineWidth(2)` in the price chart renders as ~3 physical pixels at 150 % scaling, and users report the chart looks too heavy:

> **WICHTIG** Die Linien bei der Kursdarstellung unter Wertpapiere sind seit dem letzten Update viel zu dick. Bitte wieder dünner darstellen. Oder die Option für eine dünne Linien­darstellung wieder implementieren!

— [forum thread, post #28](https://forum.portfolio-performance.info/t/update-auf-version-0-85-0-vom-7-juli-2026/39491) (also #34, #45, #46, #36)

The workaround circulating in the thread is editing `PortfolioPerformance.ini` (`-Dswt.autoScale=75`). That is not a foundation to build on — `swt.autoScale` is on its way out upstream, so anything depending on it breaks again later. It is also not a reasonable thing to ask of users; several in the thread said they would not touch the file.

There is also a genuine inconsistency independent of the scaling change: the Statement of Assets chart already has a per-series *Line width ▸ 1 / 2 / 3 px* menu (`DataSeriesChartLegend:319-329`), because its series are `DataSeries`-backed. The price chart builds its series programmatically and therefore has no equivalent.

This is worth doing independently of the scaling behaviour. The scaling change is what surfaced the demand, but the gap it exposed — that one chart lets you choose a line width and the other doesn't — stands on its own, and closing it does not depend on how SWT scales lines in any given release.

For reference, at the current scaling behaviour `1 px` reproduces the pre-0.85.0 appearance at the factors people are complaining about: at both 150 % and 200 %, a width of `1` renders at the 2 physical pixels that a width of `2` used to produce.

## Proposal

A global **chart line width** setting with values `1`, `2`, `3` (default `2`), offered as a *Line width ▸* submenu in the chart's right-click menu, replacing the hardcoded `2`s.

## Design decisions

**Stored per installation (preference store), not per file.** The cause is the monitor, so the value belongs to the machine. A per-file value follows a synced `.portfolio` from a 4K desktop to a notebook, where it is wrong. It would also mark the file dirty (`Client.setProperty` calls `touch()`) for a cosmetic display change.

**Only the `2`s change.** The code today is not uniform: most lines are `setLineWidth(2)`, a few are `setLineWidth(1)` (outlines of filled areas). The setting applies to the `2`s; the `1`s stay `1`. A derivation rule (`secondary = max(1, base - 1)`) was considered and rejected — it differs only at the `3 px` setting, by one pixel, and nobody has complained that area outlines are too thin. The simple rule also gives a mechanical scope: replace every literal `setLineWidth(2)`, touch nothing else.

**In the right-click menu, not a preference page.** Right-click is already this codebase's idiom for line width (the Statement of Assets menu). A preference page entry would not be found by a user staring at a chart they dislike.

**Opt-in per chart.** `TimelineChart` hosts both in-scope charts (account balance) and `DataSeries`-backed ones (Statement of Assets, Performance), so an unconditional menu item would appear on charts where it does nothing.

## Implementation

**1. Preference and holder**

- `UIConstants.Preferences.CHART_LINE_WIDTH`
- `PreferencesInitializer`: `store.setDefault(CHART_LINE_WIDTH, 2)`
- New `util/chart/ChartLineWidth` static holder, clamping to `1..3` (`IPreferenceStore.getInt` returns `0` for a corrupted value)

**2. Propagation** — follow `ValueColorScheme` exactly:

- `UIConstants.Event.Global.CHART_LINE_WIDTH_CHANGED = "global/chart-line-width-changed"`
- `Preference2EnvAddon`: new `@Inject @Optional` method setting the holder and posting the event (mirrors `setValueColorScheme`, `:38-45`)
- `AbstractFinanceView`: new `@UIEventTopic` method delegating to `onRecalculationNeeded()` (mirrors `:115-121`)

The menu action writes via `PortfolioPlugin.getDefault().getPreferenceStore().setValue(...)`. That store is `ScopedPreferenceStore(InstanceScope.INSTANCE, "name.abuchen.portfolio.ui")` (`PortfolioPlugin:48`), the same scope and qualifier E4 `@Preference` injection reads, so the addon fires and every open view updates.

**3. Menu** — in `ChartContextMenu.configMenuAboutToShow`, when enabled, a `Separator` plus a *Line width* submenu. Copy the shape of `DataSeriesChartLegend:319-329` so the two menus look identical: `IntStream.range(1, 4)`, labels `i + " px"`, `action.setChecked(...)`. Reuse `Messages.ChartSeriesPickerLineWidth` — **already translated in every locale, no new i18n needed**.

**4. Opt-in** — `setLineWidthConfigurable(boolean)` on `TimelineChart` and `StackedTimelineChart` (default `false`), forwarding to their `ChartContextMenu`. Set it in: `SecuritiesChart`, `PortfolioBalanceChart:86`, `AccountBalancePane:48`, `ExchangeRatesListTab:163`, `DrawdownChartWidget:59`, `ClientDataSeriesChartWidget:120`, `AbstractStackedChartViewer:128`.

## Call sites to change

Every literal `setLineWidth(2)` (line numbers at `d26a30f6a`):

| File | Lines |
|---|---|
| `views/SecuritiesChart.java` | `1166` (main quote series), `1996` (purchase price), `1368` (limit lines), `1393` (SMA), `1415` (EMA), `1788` / `1803` / `1819` (Bollinger), `1862` / `1881` (MACD) |
| `util/chart/TimelineChart.java` | `207` (`addDateSeries` — highest leverage; feeds account balance, exchange rates, drawdown, client data series), `269` (vertical marker lines) |
| `util/chart/StackedTimelineChart.java` | `108` |
| `views/payments/PaymentsAccumulatedChartBuilder.java` | `78` |
| `views/securitychart/SharesHeldChartSeries.java` | `116` |

`PortfolioBalanceChart` needs no call-site edit — its only `setLineWidth` calls are `1`s, and it picks up the change through `TimelineChart:207`.

The payments chart is a `PlainChart` (`PaymentsChartTab:68`) and has no context menu, so it honours the setting but does not offer it. That is acceptable: the value is global.

## Out of scope

- `views/dataseries/AbstractChartSeriesBuilder.java:44` — `DataSeries`-backed and already user-configurable per series. Overwriting it would discard the user's choice. `DataSeries.lineWidth` defaults to `2`, so the two systems agree out of the box.
- `SecuritiesChart` `:1146` / `:1155` and `PortfolioBalanceChart` `:399` / `:406` — width `1`, area outlines.
- `SecuritiesChart` `:1489, 1498, 1507, 1600, 1609, 1618, 1737` — these pass `0`, a local convention meaning "do not call `setLineWidth`" (see the guard at `SecuritiesChart:768`). These are marker series drawing plot symbols; setting a width would draw connecting lines through the markers.
- `TimelineChart:323` — `gc.setLineWidth(0)`, a real SWT hairline for the non-trading-day marker. Unrelated to the `0` convention above despite looking similar.
- `util/chart/MeasurementTool.java:100`, `util/chart/CrosshairTool.java:168` — width `1` transient tool overlays.
- `views/columns/QuoteRangeColumn.java:97` (`TICK_WIDTH`) — the *other* complaint in the same thread (post #4: the Kursspanne bars got wider). Same root cause, different fix, separate issue. Listed here so it is not changed by accident.

## Testing

Mostly manual. Automated coverage is limited to the holder's clamping behaviour (missing / `0` / `99` → `2`).

- Default with no preference set → `2` everywhere, rendering identical to today. This is the regression guard for the whole change.
- Changing the value updates **all** open charts immediately, not just the one whose menu was used.
- At `1`: quote line, SMA/EMA/Bollinger/MACD thin; area outlines still `1`; marker symbols unchanged with no connecting lines.
- At `3`: data lines thicken; markers still draw no connecting lines.
- Value survives a restart.
- Statement of Assets and Performance charts are unaffected, and their existing per-series *Line width* menu still works.
- Manual check at 100 % / 150 % / 200 % scaling on Windows.

buchen:
Charts that paint data series let the user pick a line width per series. The other charts painted their lines with a hard-coded width of 2 pixel, which since the monitor-specific scaling of Eclipse 2026-06 is scaled up to 3 pixel and more.

Add a "Line width" menu offering 1, 2 and 3 pixel. The value is stored per installation - not per file - because it compensates for how the monitor scales the lines: the same file opened on a notebook and on a 4K screen needs different values. Changing it updates all open charts.

The menu is offered by the chart context menu and, if the chart has one, by the configuration menu of the chart. Charts painting data series do not offer it as the user picks a line width per series there.

Closes #5944

### DIFF
--- a/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/util/chart/ChartLineWidthTest.java
+++ b/name.abuchen.portfolio.ui.tests/src/name/abuchen/portfolio/ui/util/chart/ChartLineWidthTest.java
@@ -1,0 +1,47 @@
+package name.abuchen.portfolio.ui.util.chart;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import org.junit.After;
+import org.junit.Test;
+
+@SuppressWarnings("nls")
+public class ChartLineWidthTest
+{
+    @After
+    public void resetToDefault()
+    {
+        ChartLineWidth.set(ChartLineWidth.DEFAULT_WIDTH);
+    }
+
+    @Test
+    public void testDefaultWidthIsUsedInitially()
+    {
+        assertThat(ChartLineWidth.get(), is(ChartLineWidth.DEFAULT_WIDTH));
+    }
+
+    @Test
+    public void testSupportedWidthsAreApplied()
+    {
+        for (int width = ChartLineWidth.MIN_WIDTH; width <= ChartLineWidth.MAX_WIDTH; width++)
+        {
+            ChartLineWidth.set(width);
+            assertThat(ChartLineWidth.get(), is(width));
+        }
+    }
+
+    @Test
+    public void testWidthsOutsideTheRangeFallBackToDefault()
+    {
+        // the preference can be edited manually and the injected value is 0 if
+        // it cannot be read as a number
+
+        for (int width : new int[] { 0, -1, ChartLineWidth.MAX_WIDTH + 1, 99 })
+        {
+            ChartLineWidth.set(width);
+            assertThat("width " + width + " must fall back to the default", ChartLineWidth.get(),
+                            is(ChartLineWidth.DEFAULT_WIDTH));
+        }
+    }
+}
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/UIConstants.java
@@ -44,6 +44,7 @@ public interface UIConstants
         {
             String DISCREET_MODE = "global/discreet-mode"; //$NON-NLS-1$
             String VALUE_COLOR_SCHEME_CHANGED = "global/value-color-scheme-changed"; //$NON-NLS-1$
+            String CHART_LINE_WIDTH_CHANGED = "global/chart-line-width-changed"; //$NON-NLS-1$
         }
     }
 
@@ -220,6 +221,13 @@ public interface UIConstants
         String ENABLE_SURVEY_REMINDER = "ENABLE_SURVEY_REMINDER"; //$NON-NLS-1$ //NOSONAR
 
         String ENABLE_SWTCHART_PIECHARTS = "ENABLE_SWTCHART_PIECHARTS"; //$NON-NLS-1$
+
+        /**
+         * Preference key for the width (in pixel) of the lines painted by those
+         * charts and dashboard widget series that do not offer a line width per
+         * data series.
+         */
+        String CHART_LINE_WIDTH = "CHART_LINE_WIDTH"; //$NON-NLS-1$
 
         /**
          * Preference key whether to activate in-place editing with a double
--- a/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/Preference2EnvAddon.java
+++ b/name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/addons/Preference2EnvAddon.java
@@ -25,6 +25,7 @@ import name.abuchen.portfolio.online.impl.TwelveDataSearchProvider;
 import name.abuchen.portfolio.ui.UIConstants;
 import name.abuchen.portfolio.ui.dialogs.transactions.PresetValues;
 import name.abuchen.portfolio.ui.util.ValueColorScheme;
+import name.abuchen.portfolio.ui.util.chart.ChartLineWidth;
 import name.abuchen.portfolio.util.FormatHelper;
 import name.abuchen.portfolio.util.TradeCalendarManager;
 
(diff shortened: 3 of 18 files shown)


buchen:
@OnkelDok here an proposal to configure the chart line width globally - to address the issue for folks that have the scaling issues.

OnkelDok:
> @OnkelDok here an proposal to configure the chart line width globally - to address the issue for folks that have the scaling issues.

Hi @buchen , did you mentioned me because of a specific reason?
Sorry for not having time to take part on coding to PP at the moment (since some time). Sometimes I have small feelings to just start (there are some interesting topics - configurable SMA, worst case calculation using limit as current price). But currently there is no freetime available for this. 😢

buchen:
@OnkelDok I know you did work on the chart, therefore I thought you might have feedback. But I know the situation about not having enough time... I struggle also with the many ideas, pull requests, chores, etc.

PR #5949, opened 2026-08-13
Support for older PDFs of Barclays Bank (Ireland) PLC

kimmerin:
While the Barclay's PDFs are quite stable some of the identifying elements have changed over time with the effect that older PDFs can't be parsed. This PR fixes that.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
@@ -4,10 +4,12 @@ import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.deposit;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasAmount;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasDate;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasFees;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasGrossValue;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasNote;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasSource;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasTaxes;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.interest;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.interestCharge;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.removal;
 import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countAccountTransactions;
 import static name.abuchen.portfolio.datatransfer.ExtractorTestUtilities.countAccountTransfers;
@@ -21,7 +23,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.collection.IsEmptyCollection.empty;
 
 import java.util.ArrayList;
-import java.util.List;
 
 import org.junit.Test;
 
@@ -38,7 +39,7 @@ public class BarclaysBankIrelandPLCPDFExtractorTest
     {
         var extractor = new BarclaysBankIrelandPLCPDFExtractor(new Client());
 
-        List<Exception> errors = new ArrayList<>();
+        var errors = new ArrayList<Exception>();
 
         var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "KreditKontoauszug01.txt"), errors);
 
@@ -86,7 +87,7 @@ public class BarclaysBankIrelandPLCPDFExtractorTest
     {
         var extractor = new BarclaysBankIrelandPLCPDFExtractor(new Client());
 
-        List<Exception> errors = new ArrayList<>();
+        var errors = new ArrayList<Exception>();
 
         var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "KreditKontoauszug02.txt"), errors);
 
@@ -146,7 +147,7 @@ public class BarclaysBankIrelandPLCPDFExtractorTest
     {
         var extractor = new BarclaysBankIrelandPLCPDFExtractor(new Client());
 
-        List<Exception> errors = new ArrayList<>();
+        var errors = new ArrayList<Exception>();
 
         var results = extractor.extract(PDFInputFile.loadTestCase(getClass(), "KreditKontoauszug03.txt"), errors);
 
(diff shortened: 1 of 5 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
```diff
@@ -184,4 +183,186 @@ public void testKreditKontoauszug03()
...
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00))));
+        assertThat(results, hasItem(deposit(hasDate("2020-08-12"), hasAmount("EUR", 19.98), //
+                        hasSource("KreditKontoauszug04.txt"), //
+                        hasNote("Onlineshop Umsatz 123-456-7890"), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00))));
+        assertThat(results, hasItem(removal(hasDate("2020-08-18"), hasAmount("EUR", 119.22), //
+                        hasSource("KreditKontoauszug04.txt"), //
+                        hasNote("Onlineshop2 Umsatz EXAMPLE.COM"), //
+                        hasTaxes("EUR", 0.00), hasFees("EUR", 0.00))));
+    }
```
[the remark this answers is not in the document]
kimmerin: added some checks and removed some others.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
```diff
@@ -38,7 +37,7 @@ public void testKreditKontoauszug01()
     {
         var extractor = new BarclaysBankIrelandPLCPDFExtractor(new Client());
 
-        List<Exception> errors = new ArrayList<>();
+        var errors = new ArrayList<Exception>();
```
buchen, line 42: Unrelated change: the three existing tests were rewritten from `List<Exception> errors = new ArrayList<>();` to `var errors = new ArrayList<Exception>();` (and the `List` import dropped). `CLAUDE.md` asks not to touch code unrelated to the change, and this is also the minority idiom — 130 of the PDF extractor test classes use `List<Exception>`, only 4 use `var`, and the documented test template in the project rules uses `List<Exception>`. I'd revert the three pre-existing methods and use `List<Exception>` in the two new ones as well.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
```diff
@@ -184,4 +183,158 @@ public void testKreditKontoauszug03()
                         hasNote(null), //
                         hasTaxes("EUR", 35.10 + 1.93), hasFees("EUR", 0.00))));
     }
+
+    @SuppressWarnings("java:S5961")
```
buchen: `@SuppressWarnings("java:S5961")` (Sonar "too many assertions per test") is new to the codebase — this file is the only one in the repo carrying it, and plenty of existing PDF extractor tests have far more assertions than these two without suppressing anything. Since the project rules explicitly require one assertion block per transaction and forbid consolidating values, the rule simply does not apply to this test style; the suppression is noise on both new methods. Suggest dropping it, or raising it centrally if Sonar is actually complaining in CI.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -19,6 +20,7 @@ public BarclaysBankIrelandPLCPDFExtractor(Client client)
         super(client);
 
         addBankIdentifier("Barclays Bank Ireland PLC");
+        addBankIdentifier("Barclays Bank PLC");
```
buchen, line 23: Adding the pre-Brexit UK entity as a second identifier is correct — `Barclays Bank Ireland PLC` is not a substring of it, so both entries are needed, and no other extractor in the repo claims a Barclays identifier, so there is no collision.

Worth noting that `getLabel()` still returns "Barclays Bank Ireland PLC", so a `Barclays Bank PLC` statement is shown under the Ireland label in the import dialog. That matches how the class is named and is probably the least disruptive choice, but if you want it visible to users, this is the place to decide.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -31,21 +33,26 @@ public String getLabel()
 
     private void addCreditcardStatementTransaction()
     {
-        final var type = new DocumentType("Kontoauszug", //
+        final var type = new DocumentType("(Kontoauszug|Konto.bersicht)", //
```
buchen: Nit: the group is only used for alternation, not for capturing — `(?:Kontoauszug|Konto.bersicht)` is the usual form in the other extractors. Using `.` for `ü` is right per the regex rules.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -31,21 +33,26 @@ public String getLabel()
...
                         documentContext -> documentContext //
                                         // @formatter:off
                                         // Beleg- Buchungs-/ Beschreibung Karte Betrag (EUR)
                                         // Gebucht am Wertstellung am Verwendungszweck Betrag (EUR)
+                                        // Abrechnungszeitraum: 19.06.2015 - 19.07.2015 Kreditrahmen: EUR
                                         // @formatter:on
                                         .section("currency") //
-                                        .match("^.* \\((?<currency>[A-Z]{3})\\)$") //
+                                        .match("^.* [\\(|:]\\s*(?<currency>[A-Z]{3})\\b.*$") //
```
buchen: The currency detection is broken in two ways; it only produces the right answer here by luck.

1. The stray `|` inside the character class makes the pipe-separated footer match and captures `BIC` as the currency.

`[\(|:]` is a character class, so `|` is a literal alternative, not alternation — the class means "`(` or `|` or `:`". The footer of the existing test documents is pipe-separated:

    Barclays Bank Ireland PLC Hamburg Branch | Gasstraße 4 c | 22761 Hamburg | barclays.de | BIC BARCDEHAXXX | Handelsregister HRB: 153530

` | BIC` matches `<space><|>\s*[A-Z]{3}\b` → `currency = "BIC"`. I verified this against the existing samples: KreditKontoauszug01, 02 and 03 each contain two such matching lines, all yielding `BIC`. They pass today only because an earlier line in each document happens to match with `EUR` first. Any statement where the footer appears before the first `(EUR)` / `: EUR` line silently imports every transaction as `asCurrencyCode("BIC")`. The class should be `[(:]`.

2. The line documented in the comment above does not actually match.

The pattern requires a space *before* the delimiter, and in `Abrechnungszeitraum: 19.06.2015 - 19.07.2015 Kreditrahmen: EUR` both colons are preceded by a letter. In KreditKontoauszug05.txt that line never matches — the only line in the whole document that does is on page 2:

    Umsätze Visa, Karten-Nr. 4912345678909876 : EUR      1.200,35

So the new format takes its currency from an unrelated summary line, and the verbatim comment documents something that isn't happening.

Suggestion: keep the pattern layout-specific rather than a catch-all — e.g. a `oneOf` over the two known headers, `^.* \((?<currency>[A-Z]{3})\)$` for the current format and an explicit `^Abrechnungszeitraum:.*Kreditrahmen: (?<currency>[A-Z]{3}).*$` for the old one — and fix the comment block so every documented line is really matched.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -54,24 +61,30 @@ private void addCreditcardStatementTransaction()
...
+                        .oneOf( //
+                                        section -> section //
+                                                        .attributes("date", "note", "amount", "type") //
+                                                        .documentContext("currency") //
+                                                        .match("^[\\d]{2}\\.[\\d]{2}\\.[\\d]{4} " //
+                                                                        + "(?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}) " //
+                                                                        + "(?<note>.*?)" //
+                                                                        + "\\/\\s+" //
+                                                                        + "(?<amount>[\\.,\\d]+)" //
+                                                                        + "(?<type>[\\-|\\+])$") //
```
buchen: The first alternative is looser than it needs to be and is tried first for *every* line the block matches, including all current-format lines.

`(?<note>.*?)` is unbounded, and the only thing pinning it is `\/\s+` immediately before the amount. In the current format the description column can legitimately contain a slash — KreditKontoauszug05.txt line 30 has `AUTOST BRENNERO/FROSIN TOLL WAY`, which survives only because `FROSIN` follows the slash rather than whitespace. A description ending in `... /` before the card column would be captured by this branch instead of the intended one and produce a wrong note.

Consider anchoring the branch to what actually distinguishes the "Weitere Kontotransaktionen" rows — they have no card column, so `(?<note>.{1,36}?)\s*\/\s+(?<amount>[\.,\d]+)` with the trailing `\s*` folded in, or bounding the note the same way the second branch does, would be tighter.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -54,24 +61,30 @@ private void addCreditcardStatementTransaction()
...
+                                                        .assign(this::assignDepositRemoval),
+
+
```
buchen: Stray double blank line between the two `oneOf` sections — single blank line elsewhere in this file and in the other extractors.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
```diff
@@ -125,4 +138,19 @@ private void addCreditcardStatementTransaction()
 
                         .wrap(TransactionItem::new));
     }
+
+    private void assignDepositRemoval(AccountTransaction t, ParsedData v)
+    {
+        // Is type --> "-"
+        // change from
+        // DEPOSIT to
+        // REMOVAL
```
buchen: Two issues with extracting the assign body into a helper.

First, the comment got mangled by the formatter. It was a single readable prose line before:

    // Is type --> "-" change from DEPOSIT to REMOVAL

and is now wrapped across four lines, one fragment each. `CLAUDE.md` says not to auto-format extractor files; please restore the one-liner.

Second, `assign(this::...)` with a private helper is unique in the codebase — a grep over all ~200 extractors in `datatransfer/pdf/` finds this method and this call site as the only occurrences. The house rule is explicit that "consistency in coding style between all extractors is more important than nice code", and the documented pattern for a runtime type switch is the inline `.assign((t, v) -> { ... })` body. Even with `oneOf` the convention elsewhere is to repeat the lambda in both sections. I'd inline it back into the two sections.

buchen:
@kimmerin just FYI - I used this PR to experiment how to publish the code review comments that I get locally also to GitHub. I will comment on each of them - it is not we want to apply them 1:1 (this is also what I have experiment with CodeRabbit more - but my gut feeling is that I get much better reviews from Claude Code).

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
buchen: Which warning is this fixing for you? Because I do not get warnings about too many asserts, but I now have a warning about unnecessary code - in the sections of warnings I try to keep empty.

Bildschirmfoto 2026-08-14 um 09 39 21

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
buchen: I think the change is okay - and I would leave it in here. But generally, I prefer unrelated changes to be as little as possible. Yes, the code base is then mixed, but with the var keyword I do not consider this too much of a problem.

So nothing to change here.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
buchen: I guess that comment is okay as well - we accept it - nothing to change.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
buchen: Fair nit. I would not do it manually, but probably asked Claude to fix it.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
buchen: Does this have merit?

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
buchen: About the second point: yes, we do not use this pattern yet, but I think it is okay. Anyway AI will do mass changes and should be able to take this pattern also into the refactoring.

Fixing the comment makes sense IMHO.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
buchen: If I asked Claude to fix it, it produces:

```java
                                                        .match("^[\\d]{2}\\.[\\d]{2}\\.[\\d]{4} " //
                                                                        + "(?<date>[\\d]{2}\\.[\\d]{2}\\.[\\d]{4}) " //
                                                                        + "(?<note>.*?)" //
                                                                        + "\\s\\/\\s+" //
                                                                        + "(?<amount>[\\.,\\d]+)" //
                                                                        + "(?<type>[\\-|\\+])$") //
```

where ```"\\s\\/\\s+"``` adds a matching space. Not sure about the merit of this but it sounds like it could not hurt. The other comments always end in "IT Visa" and similar statements therefore a trailing slash seems unlikely.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
kimmerin: Coderabbit complained and I've solved it by a text replacement leading to all the other occurrences to be replaced as well.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
kimmerin: It's Sonar's complaint about too many assertions in one method. If more is OK, the sonar-rules should be amended. In case you want to no longer see a warning for a non-supported SuppressWarnings:

grafik

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: Since I've already did changes somewhere else, I've changed it here already.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: Wasn't aware that `oneOf` is available in that context which is why I've tried to cram everything in one regex. Changed it to two distinct checks in a `oneOf` making the whole thing way more readable.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: I'll have to have a look into more statement-PDFs because it seems that there is a third variant for this:

grafik

No slash at the end but in the middle.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: "fixed"

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: Concerning the first comment: That's the autoformatter again...

About the second comment: copy/pasting the same logic is increasing the technical debt and makes changes error-prone if one of the many occurrences are missed (again using the bond-feature as an example. With `asBondNominal`you only need to do a change once if more than the division by 100 should happen in the future and not search for all occurrences of `/ 100` (76 at the moment). I.e. extracting logic into a method and reuse it isn't "nice code" or syntactic sugar but something that the coding rules should allow at least if not encourage.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
buchen: Apparently, the Sonar configuration is local. Because for whatever reason I do not get the warning about too many asserts. I think that is also okay to have that asserts in the PDF test cases. When I have time, I will check how to share the Sonar configuration with the project

kimmerin:
> I used this PR to experiment how to publish the code review comments that I get locally also to GitHub.

I thought that already, I recognized Claude's style so I was aware about the source of the "complaints" ;-)

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
kimmerin: > Apparently, the Sonar configuration is local.

The Eclipse instance I'm using for PP is set up according to your setup-document here on GitHub. So it's all your fault nontheless ;-)

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: added support for the overdraft interest block and amended the regex as suggested by Claude.

name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/BarclaysBankIrelandPLCPDFExtractor.java
kimmerin: reformatted the comment and this time the autoformatter stayed put

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/BarclaysBankIrelandPLCPDFExtractorTest.java
kimmerin: I've removed the SuppressWarnings from the method with less asserts than the threshold. I can change the remaining one to a NOSONAR, though.

name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/pdf/barclaysbankirelandplc/KreditKontoauszug06.txt
```diff
@@ -0,0 +1,124 @@
...
+         Seite 2 von 4 Bitte wenden
+Umsatzübersicht
+Beleg- Buchungs-/ Beschreibung Karte Betrag (EUR)
+datum Valutadatum
+17.03.2021 18.03.2021 Laden 4 Filiale        POECKING      DE Visa A             14,92-
+Sonstige Umsätze
+18.03.2021 18.03.2021 Monatl. Zinsen für Einkäufe/Überweisungen              1,68-
+Neuer Saldo am 18. März 2021            971,83-
+Hauptkarte/n, Vorname Nachname
+Umsätze Visa Karten-Nr. 491234XXXXXX9876  EUR           837,96
```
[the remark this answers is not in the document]
kimmerin: sums are adjusted

PR #5956, opened 2026-08-17
Fix IB Flex dividend reversal sign

### LINKED ISSUE #5930: IBKR Flex Query import handles dividend reversal as a positive dividend
**Describe the bug**

When importing an IBKR Flex Query XML containing a dividend reversal, Portfolio Performance imports the reversed dividend as a positive cash amount.

The number of shares is imported as negative, but the dividend amount remains positive. This causes an incorrect cash account balance.

Example from the IBKR Flex Query import:

Date: 2026-06-16

1. Dividend: +0.70 USD  
   Shares: -8

2. Dividend: +0.76 USD  
   Shares: +9

3. Transfer (Outbound): -0.06 USD

The first dividend is a reversal of the original dividend and should therefore reduce the cash account by 0.70 USD.

Actual calculation:

+0.70 USD + 0.76 USD - 0.06 USD = +1.40 USD

Expected calculation:

-0.70 USD + 0.76 USD - 0.06 USD = 0.00 USD

As a result, the USD cash account incorrectly shows a balance of 1.40 USD.

**To Reproduce**

1. Create an IBKR Flex Query containing a dividend and its subsequent reversal.
2. Export the Flex Query as XML.
3. Import the XML into Portfolio Performance.
4. Check the import preview.
5. Observe that the reversed dividend has a negative number of shares but a positive dividend amount.
6. Complete the import.
7. Check the related cash account balance.
8. The reversed dividend increases the balance instead of decreasing it.

**Expected behavior**

A dividend reversal should be imported with a negative cash amount or should reverse the original dividend transaction.

For this example, the resulting cash balance should be:

-0.70 USD + 0.76 USD - 0.06 USD = 0.00 USD

I have attached screenshots showing:

- the import preview with the reversed dividend imported as `+0.70 USD` and `-8` shares,
- the resulting USD cash account balance of `1.40 USD`.

**Screenshots**

Image 

Image

- OS: Windows
- Version: 0.86.1 (jul 2026)

sergio12S:
Fixes #5930.

IB Flex dividend reversals carry a negative `amount`, but the shared amount parser normalizes all amounts with `abs()`. The dividend share calculation retains the original sign, which explains the reported combination of negative shares and a positive cash amount.

This change restores the negative sign only for dividend and payment-in-lieu reversals. It also adds a regression fixture for the full reported reconciliation: `-0.70 + 0.76 - 0.06 = 0.00 USD`.

Verification:

```text
mvn -o -f portfolio-app/pom.xml verify \
  -pl :portfolio-target-definition,:name.abuchen.portfolio.pdfbox1,:name.abuchen.portfolio.pdfbox3,:name.abuchen.portfolio,:name.abuchen.portfolio.junit,:name.abuchen.portfolio.tests \
  -am -amd \
  -Dtest=name.abuchen.portfolio.datatransfer.ibflex.IBFlexStatementExtractorTest
```

Result: 31 tests passed; checkstyle clean.

### DIFF
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
@@ -3530,6 +3530,38 @@ public class IBFlexStatementExtractorTest
         assertThat(cashDelta, is(0L));
     }
 
+    /// A withholding tax lands on the dividend the user can actually import.
+    @Test
+    public void testDividendReversalDoesNotConsumeTheWithholdingTax() throws IOException
+    {
+        var extractor = new IBFlexStatementExtractor(new Client());
+        var activityStatement = getClass().getResourceAsStream("testIBFlexStatementFile31.xml");
+        var tempFile = createTempFile(activityStatement);
+        var errors = new ArrayList<Exception>();
+
+        var results = extractor.extract(Collections.singletonList(tempFile), errors);
+
+        assertThat(errors, empty());
+
+        // postProcessing pairs a tax with one dividend per date and security,
+        // and removes the tax item. The reversal and the replacement share both,
+        // so whichever is matched first takes the tax. It has to be the
+        // replacement: the reversal is flagged and never reaches the portfolio,
+        // which would drop the tax entirely.
+        assertThat(results, hasItem(dividend( //
+                        hasDate("2026-06-16"), hasShares(9), //
+                        hasSecurity(hasTicker("TEST")), //
+                        hasAmount("USD", 0.65), hasGrossValue("USD", 0.76), //
+                        hasTaxes("USD", 0.11), hasFees("USD", 0.00))));
+
+        // The reversal keeps the amount IB reported and stays flagged.
+        assertThat(results, hasItem(withFailureMessage( //
+                        Messages.MsgErrorCashDividendCorrectionUnmatched, //
+                        dividend( //
+                                        hasDate("2026-06-16"), hasShares(-8), //
+                                        hasAmount("USD", -0.70)))));
+    }
+
     @Test
     public void testIBFlexStatementFile27() throws IOException
     {
--- a/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
+++ b/name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
@@ -19,6 +19,7 @@ import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.hasTicker;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.inboundCash;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.outboundCash;
 import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.sale;
+import static name.abuchen.portfolio.datatransfer.ExtractorMatchers.withFailureMessage;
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio.tests/src/name/abuchen/portfolio/datatransfer/ibflex/IBFlexStatementExtractorTest.java
```diff
@@ -3490,6 +3490,33 @@ public void testIBFlexStatementFile26() throws IOException
...
+        var errors = new ArrayList<Exception>();
+
+        var results = extractor.extract(Collections.singletonList(tempFile), errors);
+
+        assertThat(errors, empty());
+        assertThat(countAccountTransactions(results), is(3L));
+        assertThat(results, hasItem(dividend( //
+                        hasDate("2026-06-16"), hasShares(-8), //
+                        hasSecurity(hasTicker("TEST")), //
+                        hasAmount("USD", -0.70))));
```
buchen: unfortunately, at the moment PP does not support corrections/reversal transactions. I think the best way for now would be to mark it as a "failed item" to give notice the user that something has to be fixed.

sergio12S:
Thanks — that is a better fit than what I had, and it turns out Portfolio Performance already has the exact convention for it, so I followed that rather than inventing anything.

`TradeRepublicCSVExtractor` emits a cash dividend correction as a failure `DIVIDENDS` item carrying `Messages.MsgErrorCashDividendCorrectionUnmatched`, and keeps the CSV's negative signs on the item. The IB Flex reversal is the same situation, so it now does the same thing:

- the negative `Dividends` / `Payment In Lieu Of Dividends` row is flagged with `MsgErrorCashDividendCorrectionUnmatched`, so the wizard tells the user the original has to be removed by hand;
- it keeps the amount IB reported (`USD -0.70`) instead of the `+0.70` that `asAmount` produced, because the item is shown in the preview and the figure the user has to reconcile should be the real one. That also matches how the Trade Republic corrections are stored.

No new message string, so no new translations. I did not add any matching between a reversal and its original — `CashDividendCorrectionMatcher` is package-private to `traderepublic`, and an IB reversal usually corrects a dividend from an earlier statement rather than one in the same import, so there would rarely be anything to pair with.

The regression now asserts the failure message, and adds `new AssertImportActions().check(results, CurrencyUnit.USD)`, which passes because the flagged item is skipped. Removing the `setFailureMessage` call makes it fail.

```
mvn -o -f portfolio-app/pom.xml verify \
  -pl :portfolio-target-definition,:name.abuchen.portfolio.pdfbox1,:name.abuchen.portfolio.pdfbox3,\
:name.abuchen.portfolio,:name.abuchen.portfolio.junit,:name.abuchen.portfolio.tests \
  -am -amd -Dtest=name.abuchen.portfolio.datatransfer.ibflex.IBFlexStatementExtractorTest
```

`IBFlexStatementExtractorTest`: 31 tests, 0 failures, 0 errors; checkstyle 0 violations.

sergio12S:
The CodeRabbit comment above is right, and it is a problem my change made worse rather than one it found lying around.

`postProcessing` keeps one dividend per `(date, security)` and then deletes the tax item it paired. A reversal and its replacement share both, so whichever comes first in document order takes the tax. Before, that was merely arbitrary — both items were importable. Now the reversal is flagged, so the tax attaches to an item that never reaches the portfolio and is removed from the results: the replacement dividend imports at its gross amount and the withholding tax is gone.

`dividendTransactionList` now skips failed items, so only a dividend the user can actually import is a candidate for the pairing.

New fixture `testIBFlexStatementFile31.xml` adds a `Withholding Tax` row for the same date and security. Without the filter the test fails on the replacement dividend, which arrives at `USD 0.76` with no tax unit instead of `USD 0.65` with `USD 0.11`.

`IBFlexStatementExtractorTest`: 32 tests, 0 failures. Full `name.abuchen.portfolio.tests` module: 4626 tests, 0 failures, 0 errors, 1 skipped; checkstyle 0 violations.

PR #5965, opened 2026-08-22
Fix Yahoo Finance exchange code for Saudi Stock Exchange (.SR)

### LINKED ISSUE #5951: Yahoo historical quotes feed generates wrong drop down value for Saudi stocks
Hi,

In the security historical quotes feed setup, when choosing Yahoo feed for a security list on on Saudi Exchange, the built-in drop down generate wrong symbol/ticker, for example:

Symbol 4060  Middle East Pharmaceutical Industries Company, the correct symbol to feed historical data from Yahoo is 4060.SR. The built-in drop down  generate symbol 4060.SAU which cause data feed error.

The built-in drop down should generate 4060.SR instead for yahoo feed.

Thanks,
Ali

MMarafi:
Correct the Yahoo Finance ticker suffix for the Saudi Stock Exchange (Tadawul) from `.SAU` to `.SR`.

Fixes #5951


buchen:
The build error is due to the sorting of the message keys. I will fix that and merge. Thanks!

buchen:
merged

PR #5972, opened 2026-08-26
Nitpick PR for Consors PDF importer

kimmerin:
Sometime between November and December 2020 Consors changed their PDF format leading to the info about interests paid/received when buying/selling bonds not being recognized if an older version is attempted to be imported. This PR fixes that.

At the moment this has only a cosmetic effect on the note but with an upcoming bond-support this might change.

While at it; I've changed the note-adding to remove multiple blanks, because sometimes the limit price appeared as `Limitpreis 12,34` and sometimes as `Limitpreis   12,34`. This had an effect on existing tests where this effect also occurred and was ignored.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
@@ -397,7 +397,7 @@ public class ConsorsbankPDFExtractor extends AbstractPDFExtractor
                         // Stückzins    Zinsvaluta    26.11.2020    277    Tage EUR 4,42
                         // @formatter:on
                         .section("note1", "note2").optional() //
-                        .match("^(Hallo)?(?<note1>St.ckzins) .* (?<note2>[\\d]+\\s+Tag(e)?\\s+([\\.,\\d]+ [A-Z]{3}|[A-Z]{3} [\\.,\\d]+))$") //
+                        .match("^(?<note1>St.ckzins) .* (?<note2>[\\d]+\\s+Tag(e)?\\s+([\\.,\\d]+ [A-Z]{3}|[A-Z]{3} [\\.,\\d]+))$") //
                         .assign((t, v) -> {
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/ConsorsbankPDFExtractor.java
```diff
@@ -392,12 +394,13 @@ private void addBuySellTransaction()
 
                         // @formatter:off
                         // Stückzins Zinsvaluta 10.07.2023 356 Tage 25,60 EUR
+                        // Stückzins    Zinsvaluta    26.11.2020    277    Tage EUR 4,42
                         // @formatter:on
                         .section("note1", "note2").optional() //
-                        .match("^(?<note1>St.ckzins) .* (?<note2>[\\d]+ Tag(e)? [\\.,\\d]+ [A-Z]{3})$") //
+                        .match("^(Hallo)?(?<note1>St.ckzins) .* (?<note2>[\\d]+\\s+Tag(e)?\\s+([\\.,\\d]+ [A-Z]{3}|[A-Z]{3} [\\.,\\d]+))$") //
```
buchen: Hallo? 🤔

kimmerin: First time, somebody answered ;-) It's a debug-regex I',m using to see what's happening when parsing a PDF (I've added System.out.println in PDFParser, showing me the current regex, the parsed line and if it matched. Have to remember to remove it when commiting, of course... ;-)

kimmerin: I've removed it now

buchen: if you have a good idea how to add debug information by default, feel free to propose something.

maybe if the -Dpdf.trace=true parameter and, if given, traces all matches or not

my fear is only that it creates then too much log statement that do not help in the end

PR #5975, opened 2026-08-27
Add IRR as selectable performance metric across charts and dashboard widgets

cairesantos:
## Why this matters

Currently only available as static metric, IRR is not only an alternative way to display performance. It is a standard annualized return measure used in portfolio management to compare investments with different cash-flow profiles, holding periods, asset classes, and strategies.

That makes IRR especially useful when comparing, for example:

- portfolios with different contribution/withdrawal patterns
- fixed-income instruments with different maturities
- pension or savings products with recurring contributions
- individual securities versus funds or portfolios
- strategies that deploy capital at different points in time

TTWROR answers a different question: how the underlying portfolio performed independently of the timing and size of external cash flows. That is useful for evaluating manager or strategy performance, but it is often not the return actually experienced by the investor and it is not always a suitable common annualized measure for comparing heterogeneous investments.

Exposing IRR consistently across the existing performance views therefore adds analytical capability rather than just an alternative visualization. It allows Portfolio Performance to use the same annualized, money-weighted return measure across securities, portfolios, benchmarks, and reporting periods.

## Summary

This PR adds **IRR as a selectable performance metric** alongside TTWROR in the existing performance views and dashboard widgets.

Rather than introducing a separate IRR-specific chart, it reuses Portfolio Performance's existing performance infrastructure and exposes IRR through the same data-series, filtering, benchmark, reporting-period, and dashboard configuration paths already used for TTWROR.

## What changes

IRR can now be selected in:

- Performance Chart
- Dashboard Performance Chart
- Top Performers
- Monthly Performance Heatmap
- Yearly Performance Heatmap

TTWROR remains the default, so existing behavior is preserved.

## Calculation approach

The implementation intentionally reuses existing Portfolio Performance calculation logic wherever possible:

- `PerformanceIndex`
- existing data-series/filtering infrastructure
- existing benchmark handling
- existing scalar `getPerformanceIRR()`
- existing security performance record IRR
- existing reporting-period logic

The only new calculation adapter is `IRRSeries`, needed because the existing code provides IRR for a reporting interval, but not a cumulative timeline of:

`IRR(reporting-period start -> day 1), IRR(start -> day 2), ...`

`IRRSeries` uses the existing IRR implementation and Portfolio Performance's existing external-cash-flow semantics rather than introducing a separate IRR calculation engine.

Tests verify, among other things, that:

- the final IRR series value matches the existing scalar `PerformanceIndex.getPerformanceIRR()`
- intermediate IRR series points match scalar IRR calculated over the equivalent truncated reporting interval
- benchmark IRR follows the existing benchmark semantics
- monthly and yearly IRR calculations use the existing reporting-interval IRR path

## Chart aggregation

For IRR, the full daily IRR series is calculated first and only then downsampled for chart display.

This avoids calculating IRR from an already aggregated `PerformanceIndex`, which would move valuation/cash-flow dates and change the result.

## Relation to #5833

This overlaps with the goal of #5833, but takes a different UI/architecture approach.

#5833 introduces a standalone IRR chart view. This PR instead treats IRR as another performance metric in the existing configurable Performance Chart and related dashboard widgets.

I commented on #5833 before opening this PR so the two approaches can be compared and discussed rather than developed independently without coordination.

## Scope

This PR intentionally does not change the existing IRR calculation itself. The goal is to expose the IRR functionality that already exists in Portfolio Performance more consistently across performance reporting.

CI is currently green on the feature branch.


cairesantos:
Hi @buchen 
just a heads-up that the required Maven workflow is still awaiting maintainer approval for this first contribution. I’m new to contributing on GitHub, which is why the account has very little public contribution history, but the same head has passed CI on my fork. 
Much appreciated!

cairesantos:
Hi @buchen following up on this since the PR has been open for a few weeks and the required CI is still waiting for maintainer approval. The branch remains mergeable and the same head has passed CI on my fork. Happy to adjust anything if needed. Thanks!

NL-TCH:
Nice work @cairesantos hopefully a maintainer sees value in our contributions!

PR #5976, opened 2026-08-28
ING: Added support for PDFs that contain tax adjustments

kimmerin:
Parsers like the one for Consors support this type of document so I've added support for the same documents by ING. The logic (and the fixed use of EUR as currency) is taken from the Consors parser. The ING document also lacks a specific mention of the currency being used and I'm not sure if it makes sense to parse the footer where `Euro` is mentioned unless I've got other doucments of the same type with a different (implicit) currency to see if the footer is relflecting that.

### DIFF
--- a/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
+++ b/name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
@@ -53,6 +53,7 @@ public class INGDiBaPDFExtractor extends AbstractPDFExtractor
         addAdvanceTaxTransaction();
         addAccountStatementTransaction();
         addNonImportableTransaction();
+        addTaxAdjustmentTransaction();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


name.abuchen.portfolio/src/name/abuchen/portfolio/datatransfer/pdf/INGDiBaPDFExtractor.java
```diff
@@ -924,6 +925,62 @@ private void addNonImportableTransaction()
                         .wrap(TransactionItem::new);
     }
 
+    private void addTaxAdjustmentTransaction()
+    {
+        final var type = new DocumentType("(?i)nachtr.gliche verlustverrechnung");
```
buchen, line 930: Gibt es einen speziellen Grund für das ```(?i)```?
Ich frage mich, weil es dafür kein Beispiel gibt - und wir bisher die regulären Ausdrücke so nicht formuliert haben.

kimmerin: I only have two PDFs of this type so I'm not sure if in an earlier or later version the text might be in a different case. That's why I often use "case insenstive" regexes in such a case so that "nachträglich" can occur at the beginning or within a sentence. I personally find this more readable than e.g. `[nN]achtr.gliche Verlustrechnung`

PR #5983, opened 2026-09-08
PDF: SGKB - support additional currencies and transaction types

RandomLambda:
### Summary
Extends the PDF extractor for **St.Galler Kantonalbank (SGKB)** to support additional currencies and transaction types.

### Key Changes
- **Currencies:** Added parsing for additional currencies in SGKB PDF documents.
- **Transaction Types:** Added extraction logic for additional transaction types.
- **Unit Tests:** Added unit test cases with anonymised PDF text samples for the new document variants.


RandomLambda:
Good catch. Flipping just the assertion wasn't enough though — the extractor itself had the bug: it took the security's currency from the "Ausschüttung:" (payment) line instead of the security's own nominal currency. Fixed the root cause in StGallerKantonalbankPDFExtractor.java (currency now derived from the "<CCY> <value> nom" suffix in the security name when present, falling back to the payment currency otherwise) and updated the assertion to EUR accordingly. Pushed in [ad063a7](https://github.com/portfolio-performance/portfolio/pull/5983/commits/ad063a7af468ff0e6726c5a9d4b78a2d3816fabe).

RandomLambda:
renamed to var. thanks for the catch

PR #6051, opened 2026-09-20
Attempt to rebook transaction in the UI during import

stoeggich:
I'm trying to get "Umbuchen" to work as "ABuchen". I need the @buchen ;)

## Concept: Transfers in the import dialog – automatic and manual

At its core, this is about **adapting the import UI for transfers**: between accounts (including foreign currency) and between portfolios. As a first step, they should be captured, checked and imported correctly both **automatically** from importers and **manually**.

The **flatex importer** (currency exchange statement) serves **only as an example** of an importer that produces foreign currency transfers.

I am aware that this is **a lot at once**. I am happy to split it into **smaller, separate PRs** (e.g. importer, manual entry, import dialog). For testing and feedback, however, I would like to show everything **together as a concept** first.

### New
- **Transfers in the header area:** separate offset account per currency pair (e.g. USD → EUR), portfolio with offset portfolio, new layout (accounts | portfolio | transfers); the selection is remembered.
- **Change the transaction type** via right-click or double-click: buy/sell ↔ delivery ↔ portfolio transfer, deposit/removal ↔ account transfer. Fees and taxes are kept as separate transactions, and every change can be undone.
- **Table:** account and portfolio columns show the values actually used, new column "Converted amount", empty columns are hidden, column widths are saved. The checkboxes "delivery" and "dividend + removal" are reflected in the table immediately.
- **Account and portfolio selection** via right-click or double-click: only suitable ones (active accounts in the matching currency), the current selection is checked.
- **Manual entry:** create new securities directly.
- **Core:** portfolio transfers with a direction (inbound/outbound), offset account of a transfer per currency pair.
- **flatex (example):** the currency exchange statement is imported as a foreign currency transfer.

### Fixes
- The offset account dropdown was invisible, and the offset account and portfolio columns stayed empty.
- The same account or portfolio as source and target, as well as a mismatching currency, are now marked as errors and not imported – neither are the related fees and taxes.
- The context menu offered accounts and portfolios that were not needed.
- Changed column widths were not applied when moving to the next page.

### Tests
New core tests (transfers per currency pair, direction of portfolio transfers, flatex) and UI tests (type conversion, account/portfolio and reference account checks, linking of fee entries).

https://github.com/user-attachments/assets/d56d5264-e386-410e-b7bf-aaef6c7aa270

https://github.com/user-attachments/assets/07bc783b-791f-4fa4-ac17-9ccc99397c9d

https://github.com/user-attachments/assets/a69df61c-e5ec-4d23-ace1-9a8b13001d60


stoeggich:
In addition, it would be highly beneficial to have an option to seamlessly display notes, as well as relevant fees and taxes, for optimal visibility.

stoeggich:
- New: type of a transaction can be changed by right-click or double-click (buy/sell ↔ delivery ↔ portfolio transfer, deposit/removal ↔ account transfer), fees and taxes are kept as separate transactions.
- New: import options are only shown when they apply, plus new options to show taxes/fees and the note as additional columns.
- New: manual entry – create new securities directly from the wizard.
- New: text view of the PDF – search field with next/previous and a toggle to highlight text passages.
- Fix: the check mark in the type menu now follows the option "convert buys/sells into deliveries".
- Fix: a new security can no longer be deleted while a transaction editor is open.
- Fix: the text view now scrolls properly.

https://github.com/user-attachments/assets/5e617872-271d-4252-8c02-07dee9ed0c09

PR #6053, opened 2026-09-21
Add "Performance per Year" dashboard widget

NL-TCH:
Adds a dashboard widget that draws the time-weighted rate of return of
several calendar years on top of each other: one line per year, each
starting at 0% on January 1st. This makes it possible to compare
visually how the years performed relative to each other.

The chart already exists in the mobile companion app — it is one of the
screenshots on its Google Play Store listing — but has never been
available on the desktop version. It has been requested in the forum:
https://forum.portfolio-performance.info/t/diagramm-performance-nach-jahren-auch-in-der-desktop-version-verfugbar-machen/31367

### Screenshots

Screenshot_20260921_163644
Screenshot_20260921_163734
Screenshot_20260921_163823

### How it works

The data per year is fetched with the existing `ReportingPeriod.YearX`
and `DashboardData#calculate`, the same way `YearlyPerformanceHeatmapWidget`
already does, so both widgets share the entries of the `DataSeriesCache`.

To draw the years on top of each other, the dates of every year are
projected onto one shared reference year. The reference year is a leap
year so that February 29th keeps a valid slot, and the reference date of
a reporting period — December 31st of the previous year, where the
accumulated performance is 0% — is projected onto the year before it and
therefore keeps sorting before January 1st.

Because the axis no longer shows real calendar dates, `TimelineChart`
gained an optional `DateTimeFormatter` for the labels of the time grid;
without it the axis would read "Q1 2000". The existing behaviour is
unchanged when no format is set, and `StackedTimelineChart` is not
affected.

The colors come from `PaymentsColors`, so a year has the same color here
as in the earnings charts.

### Configuration

Data series, the first year to show (`StartYearConfig`, reused from the
earnings widgets), whether to show the Y axis, and the chart height.
Years in which no assets are held are skipped — they would only add a
flat line at 0%.

### Tests

`PerformancePerYearWidgetTest` covers the date projection: the leap day,
the reference date sorting before January 1st, the same day of different
years landing on the same date, and that a full year stays unique and
ascending.


NL-TCH:
@buchen i saw your message on [this forumpost](https://forum.portfolio-performance.info/t/diagramm-performance-nach-jahren-auch-in-der-desktop-version-verfugbar-machen/31367/5) i like to have this as a permanent widget also available on desktop. Would you please consider merging this?

cairesantos:
Great initiative — I really like the idea of bringing **Performance per Year** into the dashboard as a permanent widget. It fits very naturally with the way I use Portfolio Performance for longer-term performance analysis.

I’m working on a related improvement on the performance-reporting side as well: [PR #5975 – Add IRR as selectable performance metric across charts and dashboard widgets](https://github.com/portfolio-performance/portfolio/pull/5975), which adds **IRR as a selectable performance metric alongside TTWROR** in the existing Performance Chart and several dashboard widgets.

The two changes are complementary: yours improves how performance is visualized across years, while mine expands which performance measure can be used in those views.

My PR has been open for a few weeks and is still waiting for the first-contributor CI approval, so I thought it was worth mentioning it here as well in case anyone interested in the performance/dashboard area wants to take a look.

Nice work on #6053 — I hope both improvements can eventually make it into PP.

